PR #1292, opened 2023-10-26
WIP: Aggregate events before executing them

JAORMX:
This adds yet another component to mediator called the EEA, which stands for the
Event Execution Aggregator. As the name suggests, it aggregates events so we
should process less of them in the end. We'll process an initial triggering event,
with successive events being batched into one until the executor is ready.

This was done by using a pessimistic soft-locking mechanism based on time which locks
execution for a single entity. If an entity execution is locked, we add the event to
a queue which effectively aggregates them into one. These events will be triggered
once a flush happens.

Note that I chose to soft-lock to avoid deadlocks. Failures would, at worst, let some
events through, which is not that problematic.


JAORMX:
superceded by https://github.com/stacklok/minder/pull/1654

PR #1528, opened 2023-11-06
Remove examples and use the official repo instead

rdimitrov:
The following PR removes the profiles and rule types from the examples folder and uses the official repo instead.


lukehinds (approved):

eleftherias (commented):
What are the files under cmd/dev/examples used for?

rdimitrov:
> What are the files under cmd/dev/examples used for?

They are used by the medev tool when you want to test/verify rule types and profiles locally without needing minder running.

PR #1538, opened 2023-11-06
Update activity.md

eryn-muetzel:
Updated title to reflect that Trusty score will be more than activity over time; updated copy to reflect user can set threshold.

Consider changing the option in the code from `pi_threshold` to "trusty_threshold' (and then update this doc) - but not urgent.


JAORMX (approved):

JAORMX:
I'm gonna rerun the failed job. It's not related but I was a see if it's a flake or if we actually broke CI elsewhere

PR #1569, opened 2023-11-07
Add sigstore signing for minder-server and helm/minder

rdimitrov:


.github/workflows/chart-publish.yml
```diff
@@ -20,6 +20,7 @@ jobs:
     permissions:
       contents: read
       packages: write
+      id-token: write # To sign the provenance.
```
jhrozek, line 23: we should test if this needs to be write, but the sigstore example workflow says it is.

jhrozek (approved):

PR #1575, opened 2023-11-07
docs: fixes broken link to CLI docs on the readme

andresgalante:
There is a broken link on the read me. This PR is fixing it :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ Now you can run `minder` commands against the public instance of Minder where yo
 and create custom profiles that would help ensure your repositories are configured consistently and securely.
 
 For more information about `minder`, see:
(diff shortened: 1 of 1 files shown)


JAORMX:
Thanks a lot for the contribution! Once CI is happy and the CLA is signed, we'll get this merged.

andresgalante:
Actually, should it be linked to here instead: https://minder-docs.stacklok.dev/ref/api ?

JAORMX (approved):

JAORMX:
> Actually, should it be linked to here instead: https://minder-docs.stacklok.dev/ref/api ?

I think the ref you linked it to is just fine. Though it would be handy to also have a reference to the API handy.

andresgalante:
I agree, since I have it open let me add a reference to the API too

PR #1578, opened 2023-11-07
Makes issue templates forms

andresgalante:
Feel free to close this PR if forms for issue templates are not your cup of tea. I enjoy them since they keep reports tidy.

You can see a [preview of it here](https://github.com/andresgalante/minder/issues/new/choose).

image


jhrozek:
This is something @dussab might have an opinion about

andresgalante:
The biggest drawback to this approach is that project maintainers don't need to follow the forms, and they become REALLY annoying. For that, there is a small link at the bottom of the template selection to open a blank issue:

image

dussab:
@andresgalante Thanks for these suggestions to keep things tidy! Some feedback and ideas on the proposals:

✅ **"Ask the community" issue report** - I love the idea of providing a link from the issue templates to the discussion forum

💡**Feedback on the bug report**
1. Could you make the Prerequisite section informational (not required checkboxes)?
2. On the "To Reproduce" section, make it optional in case they need assistance recreating the issues
3. For now, we can exclude the OS and Browser dropdowns
4. I like the freeform "version" textbox and the fact it is not required

💡**User Story template** 
1. If you could turn the "user story" template into a combined "Issue or Enhancement" template/form it would be ideal
2. Change the example text in the title so it's applicable to Issues and Enhancements
3. For the text area "Is your feature request related to a problem? Please describe.", provide an alternative label such as "Please describe the Issue or Enhancement". (required field)
4. For the "Describe the solution you'd like" section, I'd suggest calling it "Proposal" or "Solution Proposal" . 
5. I like all the additional fields you have in the form to provide additional context. I also like that they are optional so folks can just fill in what they know to get the conversation rolling

andresgalante:
@dussab The changes are made, you can [preview them here](https://github.com/andresgalante/minder/issues/new/choose). The original PR was based on the template Miner currently has.

Just a thought: "Issue or Enhancement" might be misleading since, as I understand it, an issue is a bug report, and an enhancement is a feature request. A better separation might be "Report a bug" and "Feature request"

dussab:
Yeah, I can see your point. While I have historically gone with a combined approach to have fewer options for folks to consider, I'd see value in separating them for clarity. Go for it!

andresgalante:
@dussab it's updated. Thanks for giving it a try, if it doesn't work for contributors or forms become a problem to maintainers feel free to revert it 😬

dussab (approved):
Thanks for making the updates. Looks great!

PR #1591, opened 2023-11-08
Enable skipping teardown when using the `run-docker` Makefile target

JAORMX:
Now you can use an environment variable to tell our `run-docker` target
to skip the teardown phase. It may be done as follows:

```bash
export RUN_DOCKER_NO_TEARDOWN=true
```

This will result in subsequent calls to `make teardown` to skip the
`docker-compose down` call at the beginning of the run.

this is handy for cases where you're doing rapid development and want
to keep the database changes.

### DIFF
--- a/.mk/develop.mk
+++ b/.mk/develop.mk
@@ -15,6 +15,8 @@
 
 DOCKERARCH := $(shell uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/')
 
+RUN_DOCKER_NO_TEARDOWN?=false
+
 .PHONY: run-cli
 run-cli: ## run the CLI, needs additional arguments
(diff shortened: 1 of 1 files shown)


.mk/develop.mk
```diff
@@ -23,15 +25,24 @@ run-cli: ## run the CLI, needs additional arguments
 run-server: ## run the app
 	@go run -ldflags "-X main.version=$(shell git describe --abbrev=0 --tags)" -tags '$(BUILDTAGS)' ./cmd/server serve
 
+.PHONY: run-docker-teardown
+run-docker-teardown:
```
rdimitrov, line 29: Nice, can you add a comment after the target so it shows when doing make help correctly 👍

rdimitrov (approved):

PR #1599, opened 2023-11-08
Fix database connections to actually rotate credentials

evankanderson:
Apparently, I over-optimized the caching on the credentials for IAM authorization... I was looking at fancy ways to fix this when I noticed the `sync.Once()` which isn't right here.

This does introduce new connection strings all the time, which may impact our ability to pool database connections.  My suggestion is to observe the behavior in staging and we can implement a cache in `getDBCreds` if needed.

### DIFF
--- a/internal/config/db_test.go
+++ b/internal/config/db_test.go
@@ -70,7 +70,7 @@ func TestDatabaseConfig_GetDBConnection(t *testing.T) {
 				urls = append(urls, MustParseURL(urlStr))
 			}
 			basePassword, _ := urls[0].User.Password()
-			// nolint: G101 // This one doesn't actually have a hard-coded password
+			// #nosec G101 // This one doesn't actually have a hard-coded password
 			expectedNoPasswdURL := "postgres://postgres:@localhost:5432/minder?sslmode=disable"
 			for i, u := range urls {
 				base, pw := URLExtractPass(u)
@@ -122,7 +122,8 @@ func TestDatabaseConfig_GetDBConnection(t *testing.T) {
 			connStrings := make([]string, 0, tt.numTries)
 			t.Logf("DB config is %v", c.Database)
 			for {
-				_, uri, _ := c.Database.GetDBConnection(context.Background())
+				_, closer, uri, _ := c.Database.GetDBConnection(context.Background())
+				defer closer()
 				connStrings = append(connStrings, uri)
 				if len(connStrings) >= tt.numTries {
 					break
--- a/internal/config/db_test.go
+++ b/internal/config/db_test.go
@@ -1,0 +1,178 @@
+//
+// Copyright 2023 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


internal/config/db_test.go
```diff
@@ -0,0 +1,178 @@
...
+	return nil
+}
+
+func (f *fakeDriver) Begin() (driver.Tx, error) {
+	return nil, errors.ErrUnsupported
+}
+
+func TestDatabaseConfig_GetDBConnection(t *testing.T) {
+	t.Parallel()
+	// sql.Register("postgres", &fakeDriver{})
```
JAORMX, line 67: Should this be commented like this?

JAORMX (commented):
I think this is a good change, we shouldn't have had that `sync.Once`. Good catch!  I'm not so sure it'll have a big impact on connections though. Where would this be valued?  Seems to me it only happens when we read the configuration on startup.

internal/config/db_test.go
evankanderson: This was from some earlier work, and isn't needed at alle

evankanderson:
> I think this is a good change, we shouldn't have had that `sync.Once`. Good catch! I'm not so sure it'll have a big impact on connections though. Where would this be valued? Seems to me it only happens when we read the configuration on startup.

Hmm, you might be right, I'll look further this evening.

evankanderson:
So, it looks like `sqlc` doesn't really want to re-open the DB after `New()` is called; all the generated coded assumes a `Queries` which contains a `DBTX` that assumes a `sql.DB` object.

I can see two ways forward:

1. Implement a shim in `internal/db` or `internal/config` that cycles the underlying DB when the DBTX methods are called.
2. Use the `PGPASSFILE` environment variable, and put the generated token into the `$PGPASSFILE` every 10 minutes (it's valid for 15).  My main problem here is that this would be obnoxious to test / manage the shutdown on, and we might need to add a tmpdir mount to avoid read-only root filesystem restrictions.

(2) is somewhat attractive, as the connection string would simply omit the database password (allowing us to print the connection string with impunity), but setting the env var and creating / overwriting the file feels a bit gross.  WDYT?

JAORMX:
I can see the gross nessus in 2, but it would also be the path of least resistance with sqlc. I'm leaning towards that.

JAORMX:
The other option is to use a password... not as fancy as RDS auth but it would be less extra code to maintain and dependencies.

JAORMX:
The more I think about it, the more I think using this aith mechanism is the right thing to do and it merits the extra machinery. I think we can try option 2.

evankanderson:
I've got two ideas for (2).  I'll update this PR to do the first (run a background task to fill in PGPASSFILE if `aws` is set, and use an empty volume for that).

The second option is to generate the password from a sidecar and use a shared mount between the two, like this example:

https://github.com/prometheus-community/postgres_exporter/issues/498#issuecomment-833579014

That sounds elegant in theory, but I fear the helm configuration needed to make that work optionally.

JAORMX:
@evankanderson I can help with the helm part. Also, this would be handy for other services using postgres, so the sidecar might be beneficial in other places too. I think it's a great idea.

evankanderson:
I've added code for approach (2) in the application.  I'll think a bit about where I'd want to land the sidecar code; it's definitely possible to add this to the helm chart, but making it all optional feels uglier than I'd like.

PR #1600, opened 2023-11-09
Add enhanced user agent informtion to CLI

JAORMX:
This replicates the user agent format that browsers use in order to
get more relevant debugging information without revealing any PII.


jhrozek (approved):
This works fine, do you also plan on using some of the more static strings as metrics?

rdimitrov (approved):

JAORMX:
@jhrozek what do you mean?

jhrozek:
> @jhrozek what do you mean?

I meant if we should have a counter per GOOS for example that we could expose in Grafana to see that we have this many connections from this OS

JAORMX:
@jhrozek oh yeah! That would be feasible

Issue #1589, opened 2023-11-08
read-only images can't update fulcio roots 


jhrozek:
**Describe the bug**
We're running our container images with read-only root. This means we get an error while attempting to update trust roots:
```
2023/11/08 07:20:06 error verifying image: error getting fulcio roots: initializing tuf: creating cached local store: mkdir /home/nonroot/.sigstore: read-only file system
```

**To Reproduce**
unsure

**Expected behavior**
we should check if we can point the code to use a different location that we would then mount an emptyDir over

rdimitrov:
Closing as fixed by #1611

PR #1606, opened 2023-11-09
Disable read-only FS until we fix issue #1589

jhrozek:
We track go-tuf writing into /home and not being allowed to in issue
artifact signatures.

### DIFF
--- a/deployment/helm/templates/deployment.yaml
+++ b/deployment/helm/templates/deployment.yaml
@@ -38,7 +38,6 @@ spec:
           securityContext:
             allowPrivilegeEscalation: false
             runAsNonRoot: true
-            readOnlyRootFilesystem: true
             seccompProfile:
               type: RuntimeDefault
             capabilities:


rdimitrov (commented):
We should update this setting on the docker compose file too 👍

JAORMX:
@rdimitrov so, there is no way to configure where go-tuf downloads these files?

rdimitrov:
> @rdimitrov so, there is no way to configure where go-tuf downloads these files?

@JAORMX - Alright, I'm silly 🤦‍♂️ 

I went back to the cosign code and found that in sigstore/sigstore there's a [`SIGSTORE_NO_CACHE`](https://github.com/sigstore/sigstore/blob/ee007c4780d59fec556ddf09607cfefcd63923e1/pkg/tuf/client.go#L57) env var that  apparently should allow for cosign to run in a readonly setup. Although it's not really documented anywhere so...there's that.

@jhrozek - did you found a way to reproduce this somewhat easily so we can try this?

jhrozek:
> > @rdimitrov so, there is no way to configure where go-tuf downloads these files?
> 
> @JAORMX - Alright, I'm silly 🤦‍♂️
> 
> I went back to the cosign code and found that in sigstore/sigstore there's a [`SIGSTORE_NO_CACHE`](https://github.com/sigstore/sigstore/blob/ee007c4780d59fec556ddf09607cfefcd63923e1/pkg/tuf/client.go#L57) env var that apparently should allow for tuf to run in a readonly setup which hopefully allows this for cosign too. Although it's not really documented anywhere so...there's that.
> 

Do we lose the ability to fetch the trust roots?

> @jhrozek - did you found a way to reproduce this somewhat easily so we can try this?

Not really, although it was visible several times a day in staging. We can push your patch and observe for 24h if you prefer.

rdimitrov:
> Do we lose the ability to fetch the trust roots?
> 

No, we shouldn't. It's just that if minder restarts cosign will rebuilt/update its metadata state. Otherwise it would've read it from disk.

> > @jhrozek - did you found a way to reproduce this somewhat easily so we can try this?
> 
> Not really, although it was visible several times a day in staging. We can push your patch and observe for 24h if you prefer.

Ah, so we haven't seen that locally, i.e. using `run-docker`? Alright, I'll open a PR adding this env var 👍

jhrozek:
> Ah, so we haven't seen that locally, i.e. using `run-docker`? Alright, I'll open a PR adding this env var 👍

I don't use run-docker, but run the server locally on foreground..

jhrozek:
Closing in favour of #1611

PR #1615, opened 2023-11-10
docs: adds prerequisites to the register a repo getting started page

andresgalante:
👋 As I was following the [Getting Started guide](https://docs.stacklok.com/minder/getting_started/) I got stuck on the [Register a repo step](https://docs.stacklok.com/minder/getting_started/register_repos) since even though I did auth, I missed enrolling.

For those of us who suck at reading docs, I am adding a Prerequisites section to the register a repo page, following the example of the other steps.


JAORMX (approved):
Thanks for the contributions! This is super useful!

Issue #1610, opened 2023-11-09
profile status is not updated on entity deletion


jhrozek:
I found this with PRs but this actually happens with any entity.

To reproduce:

- have two entities registered, one is passing all the checks, one is failing
- the profile is marked as failing
- delete the entity that is failing
- the profile is still marked as failing, should be passing now

PR #1618, opened 2023-11-10
Add a stored procedure to handle entity deletions

jhrozek:
In case an entity was deleted (typically a PR but a repo delete is also
a valid use-case), we would not update the profile status until the next
profile run.

For that, we need to add a new database trigger in a migration.

Sort-of-related is a bug in the original procedure that we had where if
all the current rules were skipped and we added a failure the overall
status would still be a failure. I added a new version of the old
procedure to cover that plus some tests.

Fixes: #1610

### DIFF
--- a/database/migrations/000007_status_delete_procedure.down.sql
+++ b/database/migrations/000007_status_delete_procedure.down.sql
@@ -1,0 +1,66 @@
+-- Copyright 2023 Stacklok, Inc
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+DROP TRIGGER IF EXISTS update_profile_status_after_delete ON rule_evaluations;
+
+DROP FUNCTION IF EXISTS update_profile_status_on_delete();
+
+-- This is the same version of the function from the 000001 migration. Just for completeness of down migration
+CREATE OR REPLACE FUNCTION update_profile_status() RETURNS TRIGGER AS $$
+DECLARE
+v_profile_id UUID;
+BEGIN
+    -- Fetch the profile_id for the current rule_eval_id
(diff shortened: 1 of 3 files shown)


database/migrations/000006_status_delete_procedure.up.sql
```diff
@@ -0,0 +1,116 @@
...
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- Start to make sure the function and trigger are either both added or none
+BEGIN;
+
+-- Define the trigger function (if it doesn't already exist)
+CREATE OR REPLACE FUNCTION update_profile_status_on_delete() RETURNS TRIGGER AS $$
```
JAORMX, line 19: this function makes sense.

database/migrations/000006_status_delete_procedure.up.sql
```diff
@@ -0,0 +1,116 @@
...
+
+-- Create the trigger
+CREATE TRIGGER update_profile_status_after_delete
+    AFTER DELETE ON rule_evaluations
+    FOR EACH ROW
+    EXECUTE FUNCTION update_profile_status_on_delete();
+
+-- Update overall profile status if a rule evaluation status is updated
+-- error takes precedence over failure, failure takes precedence over success
+CREATE OR REPLACE FUNCTION update_profile_status() RETURNS TRIGGER AS $$
```
JAORMX, line 69: Can you change the comment to mention what actually changed? It's tricky to figure out since we don't have a diff

database/migrations/000006_status_delete_procedure.up.sql
```diff
@@ -0,0 +1,116 @@
...
+DECLARE
+    v_profile_id UUID;
+BEGIN
+    -- Fetch the profile_id for the current rule_eval_id
+    SELECT profile_id INTO v_profile_id
+    FROM rule_evaluations
+    WHERE id = NEW.rule_eval_id;
+
+    raise warning 'profile_id: %', v_profile_id;
```
JAORMX, line 78: Do we want to keep this `raise warning`?

jhrozek: oops, no, sorry, this is a debug leftover

database/migrations/000006_status_delete_procedure.up.sql
jhrozek: sure. I just realised I should have included the old function in the down migration for completeness, too

jhrozek:
let me rebase atop the recent migration (I'll bump the file number)

JAORMX (approved):

Issue #1243, opened 2023-10-19
Reuse the inline comments from the OSV checker to provide more in-context suggestions in a GitHub Pull Request


jhrozek:
Currently we have the OSV evaluator which is able to provide inline comments for PRs with vulnerabilities. I tried to reuse it for the initial implementation of the Pi integration, but it turned out that some refactoring is needed and the inline comments are currently too coupled to vulnerabilities. We should enable the inline comments code to also comment on policies that use the Pi evaluator.

(Probably) Depends on: #1203

evankanderson:
We should do this in the following sprint

Issue #1203, opened 2023-10-15
Split the PR actions of the vulncheck evaluator into a proper remediate action


jhrozek:
The vulncheck evaluator does two things - checks the vulnerabilities and then performs an action based on the vulnerabilities found, like suggesting PR changes or commenting with a summary. This was all coded before we had remediations, but now that we do have a proper remediations engine we should split the actions part of the evaluator into a proper remediate module to avoid turning the vulncheck into spaghetti (esp visible once we started adding Pi support in addition to vulnerabilities)

jhrozek:
When addressing this, please take comments in https://github.com/stacklok/minder/issues/1203 into account

PR #1632, opened 2023-11-13
Suggest smallest delta in PR vulnerability check

### LINKED ISSUE #1607: OSV PR check: Don't suggest latest, but the smallest delta
Currently the OSV evaluator suggests the latest package. We should instead consult the fixedIn field and suggest the smallest possible delta. If there is no fixedIn and we'd assume that the CVE is not fixed at all, we should print and informational comment instead of suggesting the latest version blindly.

eleftherias:
Also, if there is not patch available, just add a comment.

Fix #1607

What it looks like in a Go update:
Screenshot 2023-11-13 at 13 54 11

What it looks like in a Python update:
Screenshot 2023-11-13 at 13 53 32

### DIFF
--- a/internal/engine/eval/vulncheck/pkgdb.go
+++ b/internal/engine/eval/vulncheck/pkgdb.go
@@ -51,11 +51,13 @@ func urlFromEndpointAndPaths(
 type patchLocatorFormatter interface {
 	LineHasDependency(line string) bool
 	IndentedString(indent int, oldDepLine string, oldDep *pb.Dependency) string
+	HasPatchedVersion() bool
 }
 
 // RepoQuerier is the interface for querying a repository
 type RepoQuerier interface {
-	SendRecvRequest(ctx context.Context, dep *pb.Dependency) (patchLocatorFormatter, error)
+	SendRecvRequest(ctx context.Context, dep *pb.Dependency, patched string, latest bool) (patchLocatorFormatter, error)
+	NoPatchAvailableFormatter(dep *pb.Dependency) patchLocatorFormatter
 }
 
 type repoCache struct {
@@ -116,6 +118,10 @@ func (pj *packageJson) LineHasDependency(line string) bool {
 	return strings.Contains(line, pkgLine)
 }
 
+func (pj *packageJson) HasPatchedVersion() bool {
+	return pj.Version != ""
+}
+
 // check that pypi repository implements RepoQuerier
 var _ RepoQuerier = (*pypiRepository)(nil)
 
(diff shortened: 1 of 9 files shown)


internal/engine/eval/vulncheck/vulncheck.go
```diff
@@ -134,6 +144,34 @@ func (e *Evaluator) Eval(ctx context.Context, pol map[string]any, res *engif.Res
 	return nil
 }
 
+// getPatchedVersion returns a version that patches all known vulnerabilities. If no such version exists, it returns
+// the version that patches the most vulnerabilities. If none of the vulnerabilities have patches, it returns the
+// empty string.
```
jhrozek, line 149: I thought this was interesting (not saying necessarily wrong) that there could be a lower package that fixes more CVEs than a newer version. Do you think that's an artifact of how fixes are tracked? Are there many cases where we'd suggest a version that's not the latest?

internal/engine/eval/vulncheck/pkgdb.go
```diff
@@ -51,11 +51,13 @@ func urlFromEndpointAndPaths(
...
 	LineHasDependency(line string) bool
 	IndentedString(indent int, oldDepLine string, oldDep *pb.Dependency) string
+	HasPatchedVersion() bool
 }
 
 // RepoQuerier is the interface for querying a repository
 type RepoQuerier interface {
-	SendRecvRequest(ctx context.Context, dep *pb.Dependency) (patchLocatorFormatter, error)
+	SendRecvRequest(ctx context.Context, dep *pb.Dependency, patched string, latest bool) (patchLocatorFormatter, error)
+	NoPatchAvailableFormatter(dep *pb.Dependency) patchLocatorFormatter
```
jhrozek, line 60: Do we actually use the `NoPatchAvailableFormatter` directly under any condition or do we display that we can't find a fix version and we use this formatter to fullfill a contract? (btw the formatter interface has outlived its lifespan I think and is ripe for refactoring (see also issues #1203 and #1243)

internal/engine/eval/vulncheck/vulncheck.go
```diff
@@ -134,6 +144,34 @@ func (e *Evaluator) Eval(ctx context.Context, pol map[string]any, res *engif.Res
...
+func getPatchedVersion(vulns []Vulnerability) (fixedVersion string, latest bool, noFix bool) {
```
jhrozek, line 150: Did you consider adding the three return parameters to the `Vulnerability` structure as unexported (lowercase) attributes and then calling `getPatchedVersion` directly from within the queryVulnDb function?

jhrozek (approved):
I think this is good to go. I left some comments, but none of them are blocking.

internal/engine/eval/vulncheck/pkgdb.go
eleftherias: We use the `LineHasDependency` method of the formatter to find which line to comment on, even if the comment says there's not fix available

internal/engine/eval/vulncheck/vulncheck.go
eleftherias: We suggest a version that's not the latest if an earlier version still fixes the vulnerability, to keep with the "smallest delta" update. The example that I was testing with is `github.com/cilium/cilium` where if we're already on version `1.13.0` we can update to `1.13.7`, instead of the latest, which is `1.14.2`

internal/engine/eval/vulncheck/vulncheck.go
eleftherias: True, it can probably go in `VulnerabilityResponse`. I'll create a followup issue.

stacklokbot (commented):

eleftherias:
@jhrozek I had to rebase

internal/engine/eval/vulncheck/vulncheck.go
jhrozek: OK, in the case there are still vulnerabilities in the later versions, we'll just detect them later right? That's fine!

jhrozek (approved):

PR #1648, opened 2023-11-14
Print project list as multiple rows in `whoami` subcommand

JAORMX:
This enables us to print several projects in an easy way in case it would
be enabled for a specific user.


eleftherias (approved):

rdimitrov (commented):
Let's add (or try reuse) this to the auth login table we show too 👍

PR #1651, opened 2023-11-14
Add a repository count gauge

jhrozek:
This seems like it might be a useful metric


internal/controlplane/metrics.go
```diff
@@ -69,6 +69,22 @@ func (m *metrics) initInstrumentsOnce(store db.Store) error {
...
+			c, err := store.CountRepositories(ctx)
+			if err != nil {
+				return err
+			}
+			observer.Observe(c)
+			return nil
+		}),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create profile count gauge: %w", err)
```
jhrozek, line 85: self-back: copy-pasta

evankanderson (commented):
We might want to export histograms of:

* Repos per user
* Profiles per user (tagged by remediate / alert)
* Rules per profile

(I realize that right now, all these will be very small, but it seems like they might be interesting over time)

jhrozek:
> We might want to export histograms of:
> 
> * Repos per user
> * Profiles per user (tagged by remediate / alert)
> * Rules per profile
> 
> (I realize that right now, all these will be very small, but it seems like they might be interesting over time)

thank you for the suggestions, I filed https://github.com/stacklok/minder/issues/1655

PR #1654, opened 2023-11-14
Process entity events in exclusively and in parallel

JAORMX:
This combines https://github.com/stacklok/minder/pull/1356 and https://github.com/stacklok/minder/pull/1292 to enable us to run the executor with more parallelism, whilst enabling us to lock an entity for execution.

To recap:

This adds yet another component to mediator called the EEA, which stands for the
Event Execution Aggregator. As the name suggests, it aggregates events so we
should process less of them in the end. We'll process an initial triggering event,
with successive events being batched into one until the executor is ready.

This was done by using a pessimistic soft-locking mechanism based on time which locks
execution for a single entity. If an entity execution is locked, we add the event to
a queue which effectively aggregates them into one. These events will be triggered
once a flush happens.

Note that I chose to soft-lock to avoid deadlocks. Failures would, at worst, let some
events through, which is not that problematic.

The aggregator is done as middleware for the executor, so boilerplate was added to
enable it.

Also, this limits the executor event handling code to just parse the message. It
subsequently spawns a goroutine to do the actual evaluation.

The intent is to not block the message handler when we're receiving events
so we'd have faster executions. We do this by spawning a goroutine when we
finish parsing the entity.


internal/eea/eea.go
```diff
@@ -0,0 +1,288 @@
...
+	caches, err := e.querier.ListFlushCache(ctx)
+	if err != nil {
+		// No rows to flush, this is fine.
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil
+		}
+		return fmt.Errorf("error listing flush cache: %w", err)
+	}
+
+	for _, cache := range caches {
```
jhrozek, line 200: Since this is happening at startup, I wonder about concurrency issues and error handling. Should we select each row with `FOR UPDATE` to lock it and just continue if the row can't be selected (was handled by another process)?

internal/eea/eea.go
```diff
@@ -0,0 +1,288 @@
...
+		cache := cache
+
+		eiw, err := e.buildEntityWrapper(ctx, cache.Entity,
+			cache.RepositoryID, cache.ArtifactID, cache.PullRequestID)
```
jhrozek, line 204: Similarly here, if the buildEntityWrapper (or the getters that retrieve the entity) return NoRows, should we just continue because the entity to update no longer exists?

internal/eea/eea.go
JAORMX: Here we'll simply get a state of all the flushes that exist on system startup. This will issue an event to the queue as opposed to doing anything with the flushed entry. This is why I didn't use a lock for this.

internal/eea/eea.go
JAORMX: This is a good point, let me fix this up.

internal/eea/eea.go
```diff
@@ -207,7 +207,9 @@ func (e *EEA) FlushAll(ctx context.Context) error {
 		}
 
 		msg, err := eiw.BuildMessage()
-		if err != nil {
+		if err != nil && errors.Is(err, sql.ErrNoRows) {
```
jhrozek, line 210: Shouldn't this special case rather be after the call to buildEntityWrapper ?

JAORMX: oops! done

jhrozek (approved):
I don't have anymore comments. I think we should just go with it in staging.

PR #1661, opened 2023-11-15
Remove trailing slash from rules-and-profiles entry in gitignore

JAORMX:
This allows us to create a symlink instead of relying on explicitly cloning the repo.


rdimitrov (approved):
Neat! 👍

Issue #1659, opened 2023-11-15
rule type creation should fail if the entity is not provided


JAORMX:
### Describe the issue

We should not allow the creation of a rule type that doesn't have the `in_entity` setting enabled. Today we do.

### To Reproduce

Create a rule type without the `in_entity` setting.

### What version are you using?

main

PR #1662, opened 2023-11-15
Re-introduce entity validation for rule types

JAORMX:
I don't know why I commented that out, but this re-introduces that check.

Closes: https://github.com/stacklok/minder/issues/1659


rdimitrov:
Should we add a simple test that ensures that for the future? 😃

PR #1666, opened 2023-11-15
Update package-lock.json parsing strategy

### LINKED ISSUE #1631: Unmarshalling package-lock.json can cause errors
### Describe the issue

With certain changes to the `package-lock.json` file, Minder isn't able to unmarshall the JSON correctly and check for vulnerabilities.

Example PR that reproduces the issue: https://github.com/eleftheria-test/public-repo/pull/21/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519

Error message:
```
{"level":"debug","profile":"stacklok-remediate-profile","ruleType":"pr_vulnerability_check","eval_status":"error","projectId":"6918633e-8b9f-493b-8147-b30c888db98a","repositoryId":"f596c281-8943-41fc-b4bb-f80c499f8f81","exception.message":"error ingesting data: error ingesting file package-lock.json: error parsing file package-lock.json: failed to unmarshal npm package: invalid character '}' after top-level value","Timestamp":1699877410335589000,"message":"result - evaluation"}
```

### To Reproduce

- Register an empty repository
- Create a profile with PR vulnerability scanning turned on
- Create an npm project in that repo, for example `npm init`
- Add a dependency to the npm project, for example `npm i --save lodash`
- Create a PR in the registered repo with the newly created `package.json` and `package-lock.json`
- Notice that Minder does not commend on the PR and Minder server has an error

(note: creating a brand new `package-lock.json` is only one scenario where the unmarshalling fails. There are other scenarios as well. See the example PR listed above.)

_No response_

eleftherias:
- Start by finding any version changes
- Then find the package name whose version was changed

Fix #1631

### DIFF
--- a/internal/engine/ingester/diff/parse.go
+++ b/internal/engine/ingester/diff/parse.go
@@ -17,13 +17,17 @@ package diff
 
 import (
 	"bufio"
-	"encoding/json"
-	"fmt"
+	"path/filepath"
 	"regexp"
 	"strings"
 
 	"github.com/stacklok/minder/internal/util"
 	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+)
+
+var (
+	versionRegex        = regexp.MustCompile(`^\+\s*"version"\s*:\s*"([^"\n]*)"\s*(?:,|$)`)
+	dependencyNameRegex = regexp.MustCompile(`\s*"([^"]+)"\s*:\s*{\s*`)
 )
 
 type ecosystemParser func(string) ([]*pb.Dependency, error)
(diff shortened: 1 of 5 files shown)


internal/engine/ingester/diff/parse.go
```diff
@@ -161,62 +160,69 @@ func goParse(patch string) ([]*pb.Dependency, error) {
...
+}
 
-	// Convert the string builder to a string and unmarshal it into the Package struct
-	root := &npmRoot{
-		Deps: make(map[string]*npmDependency),
+// findDependencyName iterates over all the previous lines to find the JSON key containing the parent dependency name.
+// If the parent key does not correspond to a dependency (i.e. it could be the root project), then an empty string is
+// returned.
+func findDependencyName(i int, lines []string) string {
+	dependencyNameRegex := regexp.MustCompile(`\s*"([^"]+)"\s*:\s*{\s*`)
```
jhrozek, line 194: do you think we could have these as variables global but not exported so we avoid compiling the regex on each call? same for the one in npmParse

eleftherias: Makes sense, updating now

jhrozek (approved):
This is a clear and nice improvement thank you.

While reading the PR I was wondering if the approach that I took with the original code where we parse the patch is just too fragile though and we'd find even more issues later. What if we did the following instead (later of course):
 - grab the whole file that was patched and parse all dependencies to get a slice of pb.Dependency
 - grab the original and parse it, too
 - see which dependencies changed
 - find them in the patch to get context
 - use that instead to add reviews at the right places

The reason I started thinking about this was really this test case:
```
+   },
+    "node_modules/undici-types": {
+      "version": "5.26.5",
+      "resolved": "https://registry.npmjs.org/undici-types/-/undici-types-5.26.5.tgz",
+      "integrity": "sha512-JlCMO+ehdEIKqlFxk6IfVoAUVmgz7cU7zD/h9XZ0qzeosSHmUJVOzSQvvYSYWXkFXC+IfLKSIffhv0sVZup6pA=="
```
and the fact that package-lock.json might be just so easy to parse otherwise and I bet there's also libraries we could use.

eleftherias:
> This is a clear and nice improvement thank you.
> 
> While reading the PR I was wondering if the approach that I took with the original code where we parse the patch is just too fragile though and we'd find even more issues later. What if we did the following instead (later of course):
> 
> * grab the whole file that was patched and parse all dependencies to get a slice of pb.Dependency
> * grab the original and parse it, too
> * see which dependencies changed
> * find them in the patch to get context
> * use that instead to add reviews at the right places
> 
> The reason I started thinking about this was really this test case:
> 
> ```
> +   },
> +    "node_modules/undici-types": {
> +      "version": "5.26.5",
> +      "resolved": "https://registry.npmjs.org/undici-types/-/undici-types-5.26.5.tgz",
> +      "integrity": "sha512-JlCMO+ehdEIKqlFxk6IfVoAUVmgz7cU7zD/h9XZ0qzeosSHmUJVOzSQvvYSYWXkFXC+IfLKSIffhv0sVZup6pA=="
> ```
> 
> and the fact that package-lock.json might be just so easy to parse otherwise and I bet there's also libraries we could use.

@jhrozek I agree that's a cleaner way to do it. We might need to rethink how we define the rule, because if we're getting the whole file instead of the patch we need something other than the diff ingester.

PR #1671, opened 2023-11-16
Add support for sidecars to migrate and minder, add tests for same

evankanderson:
Since many people probably care about their helm versions, I didn't make it part of the bootstrap command, and instead rely on helm already being installed for this test.

I started with trying to use https://github.com/mittwald/go-helm-client/, but it turns out that has a bunch of out-of-date dependencies, and all the APIs changed because that's how k8s API dependencies work.

### DIFF
--- a/deployment/helm/templates/deployment.yaml
+++ b/deployment/helm/templates/deployment.yaml
@@ -66,6 +66,9 @@ spec:
             value: "{{ .Values.trusty.endpoint }}"
           - name: "SIGSTORE_NO_CACHE"
             value: "true"
+          {{- if .Values.deploymentSettings.extraEnv }}
+          {{- toYaml .Values.deploymentSettings.extraEnv | nindent 10 }}
+          {{- end }}
           
           # ko will always specify a digest, so we don't need to worry about
           # CRI image caching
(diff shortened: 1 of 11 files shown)


evankanderson:
Note that this replaces https://github.com/stacklok/minder/pull/1599 and allows us to start retiring the `aws` cloud provider in favor of a sidecar for both GCP (sql proxy sidecar) and AWS (AWS using the pattern in https://github.com/prometheus-community/postgres_exporter/issues/498#issuecomment-833579014)

deployment/helm/templates/deployment.yaml
```diff
@@ -100,8 +103,11 @@ spec:
           - name: identity-secrets
             mountPath: /secrets/identity
           {{- if .Values.deploymentSettings.extraVolumeMounts }}
-          {{- toYaml .Values.deploymentSettings.extraVolumeMounts | nindent 12 }}
+          {{- toYaml .Values.deploymentSettings.extraVolumeMounts | nindent 10 }}
```
evankanderson, line 106: (At indent 12, this doesn't produce valid YAML, verified with test)

JAORMX (approved):

Issue #1636, opened 2023-11-13
Integrate PostgreSQL Watermill Pub/Sub for enhanced messaging in Minder


teodor-yanev:
## Description:
We aim to enhance our messaging capabilities within Minder by integrating a PostgreSQL Pub/Sub system. Currently, our application relies on the default Go channels provided by the Watermill library, which has served our initial needs well. However, as we scale and require more robust messaging patterns, the need for a more persistent and reliable Pub/Sub system has become evident.

The Watermill library offers PostgreSQL as a supported Pub/Sub option, which we believe could provide the reliability and persistence we need for complex workflows and asynchronous processing. This integration will also allow us to leverage PostgreSQL's native capabilities for transactional consistency and potentially improve overall system performance.

## The following tasks will be addressed:

- Evaluate the current messaging patterns and requirements.
- Set up the PostgreSQL Pub/Sub as per the Watermill documentation.
- Integrate the new Pub/Sub system with Minder, ensuring compatibility and minimal disruption.
- Conduct thorough testing to ensure message persistence and system reliability.
- Update any relevant documentation to reflect the changes in the messaging system.

Relevant docs: https://docs.google.com/document/d/1zd7CdwsRICjZHzoeSQGtHz1TiHUJAF0GkukxJqjv_S0/

PR #1680, opened 2023-11-16
Add PostgreSQL Messaging Pub/Sub Watermill

teodor-yanev:
Issue described in: https://github.com/stacklok/minder/issues/1636

### Testing

```
2023-11-16 10:20:18 {"level":"info","Timestamp":1700122818822984756,"message":"Initializing logger in level: debug"}
2023-11-16 10:20:18 {"level":"info","Timestamp":1700122818822995839,"message":"No cloud provider credentials specified, using password"}
2023-11-16 10:20:18 {"level":"info","Timestamp":1700122818827850756,"message":"Connected to DB"}
2023-11-16 10:20:18 {"level":"debug","component":"watermill","count":"1","Timestamp":1700122818827921006,"message":"Adding publisher decorators"}
2023-11-16 10:20:18 {"level":"debug","component":"watermill","count":"1","Timestamp":1700122818827937214,"message":"Adding subscriber decorators"}
2023-11-16 10:20:18 {"level":"debug","component":"watermill","count":"1","Timestamp":1700122818827971464,"message":"Adding middleware"}
2023-11-16 10:20:18 {"level":"debug","component":"watermill","count":"3","Timestamp":1700122818827978714,"message":"Adding middleware"}
2023-11-16 10:20:18 {"level":"info","component":"watermill","topic":"internal.entity.event","handler_name":"github.com/stacklok/minder/internal/engine.(*Executor).HandleEntityEvent-fm-internal.entity.event","Timestamp":1700122818837505464,"message":"Adding handler"}
2023-11-16 10:20:18 {"level":"info","component":"watermill","handler_name":"github.com/stacklok/minder/internal/reconcilers.(*Reconciler).handleRepoReconcilerEvent-fm-internal.repo.reconciler.event","topic":"internal.repo.reconciler.event","Timestamp":1700122818837799547,"message":"Adding handler"}
2023-11-16 10:20:18 {"level":"info","component":"watermill","topic":"internal.profile.init.event","handler_name":"github.com/stacklok/minder/internal/reconcilers.(*Reconciler).handleProfileInitEvent-fm-internal.profile.init.event","Timestamp":1700122818837813172,"message":"Adding handler"}
2023-11-16 10:20:18 {"level":"debug","component":"watermill","Timestamp":1700122818837889714,"message":"Loading plugins"}
2023-11-16 10:20:18 {"level":"info","component":"watermill","count":3,"Timestamp":1700122818837920672,"message":"Running router handlers"}
2023-11-16 10:20:18 {"level":"debug","component":"watermill","subscriber_name":"github.com/stacklok/minder/internal/engine.(*Executor).HandleEntityEvent-fm-internal.entity.event","topic":"internal.entity.event","Timestamp":1700122818838012047,"message":"Subscribing to topic"}
2023-11-16 10:20:18 {"Timestamp":1700122818838127922,"message":"2023/11/16 08:20:18 Starting gRPC server on 0.0.0.0:8090"}
2023-11-16 10:20:18 {"Timestamp":1700122818838998756,"message":"2023/11/16 08:20:18 Starting HTTP server on 0.0.0.0:8080"}
2023-11-16 10:20:18 {"Timestamp":1700122818839487422,"message":"2023/11/16 08:20:18 Starting metrics server on 0.0.0.0:9090"}
2023-11-16 10:20:18 [watermill] 2023/11/16 08:20:18.838062 schema.go:28: level=INFO  msg="Initializing subscriber schema" query="[ 
2023-11-16 10:20:18 CREATE TABLE IF NOT EXISTS "watermill_internal.entity.event" (
2023-11-16 10:20:18 "offset" SERIAL,
2023-11-16 10:20:18 "uuid" VARCHAR(36) NOT NULL,
2023-11-16 10:20:18 "created_at" TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
2023-11-16 10:20:18 "payload" JSON DEFAULT NULL,
2023-11-16 10:20:18 "metadata" JSON DEFAULT NULL,
2023-11-16 10:20:18 "transaction_id" xid8 NOT NULL,
2023-11-16 10:20:18 PRIMARY KEY ("transaction_id", "offset")
2023-11-16 10:20:18 );
2023-11-16 10:20:18  
2023-11-16 10:20:18 CREATE TABLE IF NOT EXISTS "watermill_offsets_internal.entity.event" (
2023-11-16 10:20:18 consumer_group VARCHAR(255) NOT NULL,
2023-11-16 10:20:18 offset_acked BIGINT,
2023-11-16 10:20:18 last_processed_transaction_id xid8 NOT NULL,
2023-11-16 10:20:18 PRIMARY KEY(consumer_group)
2023-11-16 10:20:18 )]" subscriber_id=01HFBKEY8BZBT23P4EEANZD004 
2023-11-16 10:20:18 [watermill] 2023/11/16 08:20:18.845592 schema.go:28: level=INFO  msg="Initializing subscriber schema" query="[ 
2023-11-16 10:20:18 CREATE TABLE IF NOT EXISTS "watermill_internal.repo.reconciler.event" (
2023-11-16 10:20:18 "offset" SERIAL,
2023-11-16 10:20:18 "uuid" VARCHAR(36) NOT NULL,
2023-11-16 10:20:18 {"level":"debug","component":"watermill","subscriber_name":"github.com/stacklok/minder/internal/reconcilers.(*Reconciler).handleRepoReconcilerEvent-fm-internal.repo.reconciler.event","topic":"internal.repo.reconciler.event","Timestamp":1700122818845564006,"message":"Subscribing to topic"}
2023-11-16 10:20:18 {"level":"info","component":"watermill","subscriber_name":"github.com/stacklok/minder/internal/engine.(*Executor).HandleEntityEvent-fm-internal.entity.event","topic":"internal.entity.event","Timestamp":1700122818845722756,"message":"Starting handler"}
2023-11-16 10:20:18 {"level":"debug","component":"watermill","subscriber_name":"github.com/stacklok/minder/internal/reconcilers.(*Reconciler).handleProfileInitEvent-fm-internal.profile.init.event","topic":"internal.profile.init.event","Timestamp":1700122818849883131,"message":"Subscribing to topic"}
2023-11-16 10:20:18 {"level":"info","component":"watermill","subscriber_name":"github.com/stacklok/minder/internal/reconcilers.(*Reconciler).handleRepoReconcilerEvent-fm-internal.repo.reconciler.event","topic":"internal.repo.reconciler.event","Timestamp":1700122818849898172,"message":"Starting handler"}
2023-11-16 10:20:18 {"level":"info","component":"watermill","topic":"internal.profile.init.event","subscriber_name":"github.com/stacklok/minder/internal/reconcilers.(*Reconciler).handleProfileInitEvent-fm-internal.profile.init.event","Timestamp":1700122818860212756,"message":"Starting handler"}
2023-11-16 10:20:33 {"level":"error","Resource":{"service":"minder.v1.UserService","method":"GetUser"},"Attributes":{"http.code":"NotFound","http.content-type":["application/grpc"],"http.duration":"2.34775ms","http.user_agent":["minder-cli/v0.0.15+ref.d8c9ae80 (darwin) arm64 (go1.21.3) grpc-go/1.59.0"],"exception.message":"rpc error: code = NotFound desc = user not found"},"Timestamp":1700122833197264929}
2023-11-16 10:20:33 {"level":"info","Resource":{"service":"minder.v1.UserService","method":"CreateUser"},"Attributes":{"http.code":"OK","http.content-type":["application/grpc"],"http.duration":"6.48ms","http.user_agent":["minder-cli/v0.0.15+ref.d8c9ae80 (darwin) arm64 (go1.21.3) grpc-go/1.59.0"]},"Timestamp":1700122833204679846}
2023-11-16 10:20:18 "created_at" TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
2023-11-16 10:20:18 "payload" JSON DEFAULT NULL,
2023-11-16 10:20:18 "metadata" JSON DEFAULT NULL,
2023-11-16 10:20:18 "transaction_id" xid8 NOT NULL,
2023-11-16 10:20:18 PRIMARY KEY ("transaction_id", "offset")
2023-11-16 10:20:18 );
2023-11-16 10:20:18  
2023-11-16 10:20:18 CREATE TABLE IF NOT EXISTS "watermill_offsets_internal.repo.reconciler.event" (
2023-11-16 10:20:18 consumer_group VARCHAR(255) NOT NULL,
2023-11-16 10:20:18 offset_acked BIGINT,
2023-11-16 10:20:18 last_processed_transaction_id xid8 NOT NULL,
2023-11-16 10:20:18 PRIMARY KEY(consumer_group)
2023-11-16 10:20:18 )]" subscriber_id=01HFBKEY8BZBT23P4EEANZD004 
2023-11-16 10:20:18 [watermill] 2023/11/16 08:20:18.849949 schema.go:28: level=INFO  msg="Initializing subscriber schema" query="[ 
2023-11-16 10:20:18 CREATE TABLE IF NOT EXISTS "watermill_internal.profile.init.event" (
2023-11-16 10:20:18 "offset" SERIAL,
2023-11-16 10:20:18 "uuid" VARCHAR(36) NOT NULL,
2023-11-16 10:20:18 "created_at" TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
2023-11-16 10:20:18 "payload" JSON DEFAULT NULL,
2023-11-16 10:20:18 "metadata" JSON DEFAULT NULL,
2023-11-16 10:20:18 "transaction_id" xid8 NOT NULL,
2023-11-16 10:20:18 PRIMARY KEY ("transaction_id", "offset")
2023-11-16 10:20:18 );
2023-11-16 10:20:18  
2023-11-16 10:20:18 CREATE TABLE IF NOT EXISTS "watermill_offsets_internal.profile.init.event" (
2023-11-16 10:20:18 consumer_group VARCHAR(255) NOT NULL,
2023-11-16 10:20:18 offset_acked BIGINT,
2023-11-16 10:20:18 last_processed_transaction_id xid8 NOT NULL,
2023-11-16 10:20:18 PRIMARY KEY(consumer_group)
2023-11-16 10:20:18 )]" subscriber_id=01HFBKEY8BZBT23P4EEANZD004 

```

Screenshot 2023-11-16 at 10 24 50

Update: after adding the second database "minder_queue", we have a nice separation of concerns:
image


stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

jhrozek:
Maybe a stupid question, but since there appears to be a single table, how does publishing and subscribing work with multiple minder processes? Does the subscriber implement some locking with e.g. SELECT FOR UPDATE?

JAORMX (commented):
I see that the postgres driver brings some new tables to the mix. Where are those created? The `minder-server` database user does not have permissions to create tables, only the migration user is able to do this. I'm guessing this would be an issue.

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

internal/config/config.go
```diff
@@ -36,6 +36,7 @@ type Config struct {
 	Tracing       TracingConfig      `mapstructure:"tracing"`
 	Metrics       MetricsConfig      `mapstructure:"metrics"`
 	Database      DatabaseConfig     `mapstructure:"database"`
+	DatabaseQueue DatabaseConfig     `mapstructure:"database_queue"`
```
JAORMX, line 39: This database configuration is explicitly part of the eventing system. Let's not make it a top-level configuration and instead put that inside of the `events` section, within the driver configuration.

config/config.yaml.example
```diff
@@ -100,6 +100,6 @@ github:
     events: ["*"]
 
 events:
-  driver: go-channel
+  driver: postgresql
```
JAORMX, line 103: If we're changing the defaults (I'm not sure if we should), we should also add defaults for the database configuration. I actually suggest not to change these defaults just yet.

config/config.yaml.example
```diff
@@ -100,6 +100,6 @@ github:
...
   router_close_timeout: 10
   go-channel: {}
```
JAORMX, line 105: It would also be good to add the queuing database configuration below here.

internal/events/eventer.go
```diff
@@ -98,7 +100,7 @@ var _ message.Publisher = (*Eventer)(nil)
 
 // Setup creates an Eventer object which isolates the watermill setup code
 // TODO: pass in logger
-func Setup(ctx context.Context, cfg *config.EventConfig) (*Eventer, error) {
+func Setup(ctx context.Context, cfg *config.EventConfig, db *sql.DB) (*Eventer, error) {
```
JAORMX, line 103: Instead of passing the `db` as a new argument, shouldn't we instead just rely on the configuration given within `cfg` and do the database connection instantiation as part of the builder?

JAORMX (changes requested):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

deployment/helm/templates/deployment.yaml
```diff
@@ -66,6 +66,11 @@ spec:
             value: "{{ .Values.trusty.endpoint }}"
           - name: "SIGSTORE_NO_CACHE"
             value: "true"
+          - name: "MINDER_DB_PASS"
+            valueFrom:
+              secretKeyRef:
+                name: minder-identity-secrets
+                key: password
```
JAORMX, line 73: Let's not pass this variable this way. Instead, let's set it in the helm chart invocation itself via `extraEnv` that's defined below.

internal/config/db.go
```diff
@@ -154,6 +154,42 @@ func RegisterDatabaseFlags(v *viper.Viper, flags *pflag.FlagSet) error {
...
+	}
+
+	err = util.BindConfigFlagWithShort(
+		v, flags, "events.sql.dbname", "db-events-name", "N", "minder-queue", "Database Events name", flags.StringP)
+	if err != nil {
+		return err
+	}
+
+	return util.BindConfigFlagWithShort(
+		v, flags, "events.sql.sslmode", "db-events-sslmode", "M", "disable", "Database Events sslmode", flags.StringP)
```
JAORMX, line 194: These are not needed. We'll only read this from the configuration or environment files.

JAORMX (changes requested):

deployment/helm/templates/deployment.yaml
```diff
@@ -66,6 +66,11 @@ spec:
             value: "{{ .Values.trusty.endpoint }}"
           - name: "SIGSTORE_NO_CACHE"
             value: "true"
+          - name: "MINDER_DB_PASSWORD"
```
jhrozek: I think there is a bit of confusion in how the variables are transformed into config values. I haven't tried this PR, but the way I read this hunk is that this would set the equivalent of what a `db.password` in config would set, wouldn't it? At least that's how I think `SetViperDefaults` works - you'd strip the MINDER prefix and then substitute underscores for dots. Note that we set `MINDER_IDENTITY_SERVER_CLIENT_SECRET_FILE` in this file as an example.

And as far as datebase goes, I don't see a `db` field in the `Config` struct at all, but a `database` field instead, so if you want to set the DB password, you'd set `MINDER_DATABASE_PASSWORD`.

internal/config/db.go
```diff
@@ -154,6 +154,42 @@ func RegisterDatabaseFlags(v *viper.Viper, flags *pflag.FlagSet) error {
...
 	}
 
-	return util.BindConfigFlagWithShort(
+	err = util.BindConfigFlagWithShort(
 		v, flags, "database.sslmode", "db-sslmode", "s", "disable", "Database sslmode", flags.StringP)
+	if err != nil {
+		return err
+	}
+
+	err = util.BindConfigFlagWithShort(
```
jhrozek, line 163: I'm confused why do we need these? Can't we just rely on the configuration?

jhrozek (changes requested):
I'll remove the block once we discuss the env variable, but on first glance it seems incorrect

stacklokbot (commented):

stacklokbot (commented):

JAORMX (approved):

internal/config/db.go
teodor-yanev: I agree..., removed.

deployment/helm/templates/deployment.yaml
teodor-yanev: Thanks! Remove env changes here, there will be a follow-up PR in infra for the rest of the configuration: https://github.com/stacklok/infra/pull/319

The name will be ```MINDER_EVENTS_SQL_CONNECTION_DBPASS``` to reflect the nested struct structure we'll have.

teodor-yanev:
> Maybe a stupid question, but since there appears to be a single table, how does publishing and subscribing work with multiple minder processes? Does the subscriber implement some locking with e.g. SELECT FOR UPDATE?

Good question, to answer this one, I did some digging in files like https://github.com/ThreeDotsLabs/watermill-sql/blob/master/pkg/sql/schema_adapter_postgresql.go and https://github.com/ThreeDotsLabs/watermill/blob/master/_examples/real-world-examples/exactly-once-delivery-counter/README.md, and the scarce documentation in https://watermill.io/pubsubs/sql/ so I can present an overall idea of how it functions:

There are essentially two types of tables involved for each topic and consumer group:

### Message Table:
- Contains the actual messages that are published.
- Each message typically has a unique identifier, the payload, and possibly other metadata.
- This table is used for both publishing (inserting messages) and subscribing (reading messages).

### Offsets Table:
- Used for tracking the progress of each consumer or consumer group.
- It records the last message offset (or ID) that has been acknowledged or processed by a consumer group.
- Ensures that each message is processed exactly once by each consumer group and helps maintain the order of message processing.
- Uses mechanisms like SELECT ... FOR UPDATE to lock rows and manage concurrent access by different consumers.

### Relationship Between the Tables
The message table and the offsets table are related in the sense that the offsets table keeps track of which messages from the message table have been processed. The offsets table stores an identifier (like an offset or a transaction ID) that corresponds to records in the message table.

jhrozek:
One more question, but I guess this can be a follow-up..how is the InitSchema attribute used? I can't seem to find the usage:
```
jakub@flatline-2 ~/devel/mediator (postresql-messaging-watermill●)$ git grep init_schema
internal/config/events.go:      InitSchema bool           `mapstructure:"init_schema" default:"true"`
jakub@flatline-2 ~/devel/mediator (postresql-messaging-watermill●)$ git grep InitSchema
internal/config/events.go:      // InitSchema is whether or not to initialize the schema
internal/config/events.go:      InitSchema bool           `mapstructure:"init_schema" default:"true"`
```

JAORMX:
It isn't as of now. I made it configurable because I wanna figure out if it's possible to move this logic out of the workload, so then we could just toggle that.

teodor-yanev:
> It isn't as of now. I made it configurable because I wanna figure out if it's possible to move this logic out of the workload, so then we could just toggle that.

Should we remove it for now and re-add it when it's present in the application logic?

JAORMX (approved):

jhrozek (approved):

Issue #1242, opened 2023-10-19
Revisit our use of text/template for hardened options


jhrozek:
**Describe the bug**
Since we are displaying potentially untrusted data with some of our templates (data that might come from external sources), we should take a look at our use of text/template and make sure we are using secure defaults. Things we might want to check out:

- make sure we are HTML-escaping things that we render. Consider using html/template for content like tables instead.
- consider using `missingkey=error` to make sure we rather don't render anything and throw an error rather than garbage

(Since we are "only" rendering to GH, I would hope that they do some sanitisation on their part, too, but better safe from the start than sorry later)

PR #1682, opened 2023-11-16
Use missingkey=error for text/template

jhrozek:
This should make errors more pronounced when someone has a bug in their rule and the remediation or evaluation template doesn't expand fully

Fixes: #1242


jhrozek:
btw trusty is giving me a 500 so I haven't been able to test that part. Please don't merge until trusty is fixed and I test the trusty reply rendering.

rdimitrov (approved):

PR #1690, opened 2023-11-17
Remove `combined.yml` from gitignore

JAORMX:
This was previously used by our helm generation but fell out of use.

We shouldn't be ignoring this file anymore as it's no longer generated.


teodor-yanev (approved):
👍

PR #1716, opened 2023-11-23
Allow regex matching of artifacts with artifact ingester

JAORMX:
This changes the ingester to also allow a `tag_regex` field as part of its
configuration. The idea is that folks would be able to write a profile with
a configuration that matches all release-related tags for their containers.

Such a configuration would look as follows:

```yaml
artifact:
  - type: artifact_signature
    params:
      tag_regex: "v*"
    def:
      is_signed: true
      is_verified: true
      is_bundle_verified: true
```

### DIFF
--- a/internal/engine/ingester/artifact/artifact.go
+++ b/internal/engine/ingester/artifact/artifact.go
@@ -109,6 +109,10 @@ func getApplicableArtifactVersions(
 
 	// get all versions of the artifact that are applicable to this rule
 	for _, artifactVersion := range artifact.Versions {
+		if !isProcessable(artifactVersion.Tags) {
+			continue
+		}
+
 		if tagMatcher.MatchTag(artifactVersion.Tags...) {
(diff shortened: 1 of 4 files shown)


eleftherias (approved):
Just a reminder that our docs for rule types are not autogenerated. So when we update the rule definition in our rules-and-profiles repo, we'll also need to update `artifact_signature.md` here.

JAORMX:
@eleftherias yeah. However, we can't change that rule until this code is shipped in both staging and prod. Else, we'll have folks trying that out and failing to use it.

JAORMX:
@rdimitrov https://github.com/stacklok/minder/pull/1716/commits/de1d57178a94c6c2ecd67875d185698b91ad368c addresses the feedback you gave via slack.

JAORMX:
@rdimitrov found another issue; I'll work on it.

JAORMX:
Fixed

rdimitrov (approved):

PR #1723, opened 2023-11-23
run go mod tidy

jhrozek:
I ran go.mod tidy in another patch that adds frizbee as a dependency and
got a lot of unrelated changes. It's better to submit them separately to
not confuse reviewers later on


stacklokbot (commented):

JAORMX (approved):

PR #1725, opened 2023-11-23
Use frizbee to add a new function ListGithubActions to the rego library

jhrozek:
Adds a new function called `ListGithubActions` to the rego library along
with tests. The function calls the frizbee function and formats the
output as a Rego set to allow for easy comparisons later.

Also adds frizbee as a dependency for minder.


stacklokbot (commented):

JAORMX (approved):

PR #1727, opened 2023-11-24
WIP: Pass ingest result to remediator

JAORMX:


rdimitrov:
Why do we want to pass the ingest result to actions? I like having the linear chain of events - ingesting -> eval -> actions, so actions continue to respond based on the eval result and the eval result is based on the ingest result.

JAORMX:
> Why do we want to pass the ingest result to actions? I like having the linear chain of events - ingesting -> eval -> actions, so actions continue to respond based on the eval result and the eval result is based on the ingest result.

this is a WIP PR. But the idea is to be able to do remediations based on the contents of what was ingested. e.g. being able to do changes to the git tree that was ingested to form PRs.

JAORMX:
This doesn't break the linearity at all

rdimitrov:
> > Why do we want to pass the ingest result to actions? I like having the linear chain of events - ingesting -> eval -> actions, so actions continue to respond based on the eval result and the eval result is based on the ingest result.
> 
> this is a WIP PR. But the idea is to be able to do remediations based on the contents of what was ingested. e.g. being able to do changes to the git tree that was ingested to form PRs.

Ah, I see 👍 I guess it fits better to the `ActionsParams` interface then instead of updating the Actions interface?

JAORMX:
I can try that

PR #1728, opened 2023-11-24
Add an optional violation_format to rego rules

jhrozek:
Adds an optional parameter for the rego evaluator that allows specifying
that if the constraints mode is used, then the constraints message
should be a valid JSON object with a key and a value so that it decodes
into `map[string]any`.

This is done by passing an outputFormat into the rule, the usage can be
seen in unit tests.

The default is still "text" to keep backwards compatibility.

If the evaluator asks for JSON, but back comes just a string, we can
assume that the policy doesn't support JSON output, so we marshall the
string ourvelves into `{ "msg": $response }`.

The main use-case is rules that print a list of items violating a
policy, those can then be summarized using jq like this:
```
./bin/minder profile_status list --provider=github -i actions-github-profile -d -ojson 2>/dev/null | jq '.ruleEvaluationStatus | map(select(.ruleName == "repo_acti
on_list" and .status == "failure")) | map({repo_name: .entityInfo.repo_name, details: .details | fromjson})'
[
  {
    "repo_name": "testrepo",
    "details": [
      {
        "actions_not_allowed": [
          "docker/build-push-action",
          "docker/login-action",
          "docker/metadata-action",
          "docker/setup-buildx-action"
        ]
      }
    ]
  },
  {
    "repo_name": "bad-go",
    "details": [
      {
        "actions_not_allowed": [
          "docker/build-push-action",
          "docker/login-action",
          "docker/metadata-action",
          "docker/setup-buildx-action"
        ]
      }
    ]
  }
]
```


jhrozek:
to test, feel free to use https://github.com/stacklok/minder-rules-and-profiles/pull/20

JAORMX (approved):

rdimitrov (approved):

Issue #1729, opened 2023-11-27
Migrate the pull_request remediator to use git-go instead of direct GH API


jhrozek:
This would enable us to expand the scope of the PR remediator to also modify the contents in place better.

PR #1773, opened 2023-11-29
Make the pull-request remediator use git-go instead of plan GH API calls

jhrozek:
This would allow us to modularize better the actual action that's done
(currently creating a file in git) and then keep on adding more actions like calling frizbee to change actions' tag to a SHA.

Fixes: #1729

### DIFF
--- a/internal/engine/actions/remediate/pull_request/pull_request.go
+++ b/internal/engine/actions/remediate/pull_request/pull_request.go
@@ -23,12 +23,21 @@ import (
 	"errors"
 	"fmt"
 	htmltemplate "html/template"
+	"io"
 	"log"
-	"net/http"
 	"os"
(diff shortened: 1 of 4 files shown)


JAORMX (approved):
I somehow am able to reason better about the code now that we're using `go-git`. Great work! It ended up being a lot more involved than I thought, and I really appreciate all the effort you put in this!

jhrozek:
> I somehow am able to reason better about the code now that we're using `go-git`. Great work! It ended up being a lot more involved than I thought, and I really appreciate all the effort you put in this!

Honestly it took most time to write unit tests because I spent way too much time trying to bend memfs based repos into accepting a remote for pushing before realising I can just use a fs-based repo and be happy..

PR #1784, opened 2023-11-30
Adds commit lint to the CI

### LINKED ISSUE #1310: Add CI check for commit messages
We can add a job to our CI that checks if commit messages are formatted according to our contributor guidelines.

andresgalante:
closes #1310

This is my first time doing any CI work, so even though I've tested it, and it is annoyingly good at detecting mistakes on commit messages, I'd suggest taking a close look before merging it.

It'd also be good to add the same check on the local build; that way, contributors can check errors before submitting a PR 😄


jhrozek:
Thanks for the PR!

do you know what config does the action use by default? For some reason I'm not able to configure `commitlint` locally - I wanted to pipe a couple of commit messages through the linter so see what would get flagged.

andresgalante:
@jhrozek, thanks for taking a look. I've tested several cases, and it looks like the default follows Minder's guidelines and should allow setting up semantic release. 

I am not 100% sure but I think that [this is the default config](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional).

To test it, please open as many PRs as you want on[ my fork](https://github.com/andresgalante/minder/pulls), and you'll see the errors: There are some example PRs open.

If there is any [rule](https://commitlint.js.org/#/reference-rules) that needs to change, we can create a [config](https://commitlint.js.org/#/reference-configuration) for it or base it on the other [config examples](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint) in the commitlint repo.

andresgalante:
Ignore my last comment, I don't know why I thought that Minders guidelines were asking for semantic release commit messages, when they don't 🤦

I'll update this PR with a config that follows Chris Beans recommendations tomorrow 🙏

andresgalante:
@jhrozek great idea! I've just added the rule to ignore WIP messages.

Thanks for taking the time to review this one.

Issue #1664, opened 2023-11-15
Fix why dirty is set to true for the released minder binary


rdimitrov:
### Describe the issue

The dirty flag is set to true for the related minder binary which shouldn't be like that. 

Perhaps this happens because of some artifact that goreleaser creates in the repo and is caught by the `git diff` since the releaser job is called with `dirty: clean`. We should investigate what causes this and fix it accordingly :thu

Helpful resources:
* https://github.com/stacklok/minder/blob/main/internal/constants/version.go
* https://github.com/stacklok/minder/blob/main/internal/constants/version.go#L69
* https://github.com/stacklok/minder/blob/main/internal/constants/version.go#L53
* https://github.com/stacklok/minder/actions/runs/6877089039/job/18704037344#step:6:15

Thanks to @JAORMX for noticing this.

### To Reproduce

Run minder version using the released minder cli.

```bash
$ minder version
No config file present, using default values.
Version: 0.0.16
Go Version: go1.21.3
Git Commit: d8c9ae802dbb52984890709b204655a7286f17fa
Commit Date: 2023-11-15T11:19:12Z
OS/Arch: linux/amd64
Dirty: true
User Agent: minder-cli/0.0.16 (linux) amd64 (go1.21.3)
```

### What version are you using?

0.0.16

PR #1785, opened 2023-11-30
Ignore /dist directory generated by goreleaser

dio:
Fix https://github.com/stacklok/minder/issues/1664.

This makes sure we have `Dirty: false` for `minder version` output.

I tested this locally, by simulating running `goreleaser release --clean` in a simulated clean directory with a dummy tag (while the publishing will fail, but the `/dist` directory is created). Before this change, running `minder version` gave me:

```console
...
Dirty: true
...
```

With this change:

```console
...
Dirty: false
...
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ docs/.docusaurus
 
 # Ignore examples cloned repo
 examples/rules-and-profiles
+
+# Generated goreleaser artifacts
+/dist


rdimitrov (approved):
@dio - Thanks for taking care of that! Much appreciated! 🚀

JAORMX (approved):
great!!

PR #1790, opened 2023-12-01
Add simplified FGA model and tests for same

evankanderson:
Fixes https://github.com/stacklok/epics/issues/131

Followup to auth model discussion on Nov 30 and #1781

Closes: #1843


evankanderson:
Passing tests look like:

```
(PASSING) check-inheritance: Checks (50/50 passing) | ListObjects (0/0 passing)
```

Failing tests look like:

```
(FAILING) check-inheritance: Checks (34/38 passing) | ListObjects (0/0 passing)
✓ Check(user=user:admin1,relation=creator,object=project:001, context=<nil>)
✓ Check(user=user:admin1,relation=viewer,object=project:001, context=<nil>)
✓ Check(user=user:admin1,relation=repo_updater,object=project:001, context=<nil>)
✓ Check(user=user:admin1,relation=provider_creator,object=project:001, context=<nil>)
✓ Check(user=user:admin1,relation=artifact_updater,object=project:001, context=<nil>)
✓ Check(user=user:admin1,relation=creator,object=project:002, context=<nil>)
✓ Check(user=user:admin1,relation=viewer,object=project:002, context=<nil>)
✓ Check(user=user:admin1,relation=repo_updater,object=project:002, context=<nil>)
✓ Check(user=user:admin1,relation=provider_creator,object=project:002, context=<nil>)
✓ Check(user=user:admin1,relation=artifact_updater,object=project:002, context=<nil>)
✓ Check(user=user:admin2,relation=repo_updater,object=project:001, context=<nil>)
✓ Check(user=user:admin2,relation=provider_creator,object=project:001, context=<nil>)
✓ Check(user=user:admin2,relation=artifact_updater,object=project:001, context=<nil>)
✓ Check(user=user:admin2,relation=creator,object=project:001, context=<nil>)
✓ Check(user=user:admin2,relation=viewer,object=project:001, context=<nil>)
✓ Check(user=user:admin2,relation=creator,object=project:003, context=<nil>)
✓ Check(user=user:admin2,relation=viewer,object=project:003, context=<nil>)
✓ Check(user=user:admin2,relation=repo_updater,object=project:003, context=<nil>)
✓ Check(user=user:admin2,relation=provider_creator,object=project:003, context=<nil>)
✓ Check(user=user:admin2,relation=artifact_updater,object=project:003, context=<nil>)
✓ Check(user=user:nonadmin1,relation=repo_updater,object=project:001, context=<nil>)
✓ Check(user=user:nonadmin1,relation=provider_creator,object=project:001, context=<nil>)
✓ Check(user=user:nonadmin1,relation=artifact_updater,object=project:001, context=<nil>)
✓ Check(user=user:nonadmin1,relation=provider_getter,object=project:001, context=<nil>)
✓ Check(user=user:nonadmin1,relation=creator,object=project:001, context=<nil>)
✓ Check(user=user:nonadmin1,relation=viewer,object=project:001, context=<nil>)
✓ Check(user=user:nonadmin1,relation=artifact_updater,object=project:002, context=<nil>)
✓ Check(user=user:nonadmin1,relation=provider_getter,object=project:002, context=<nil>)
✓ Check(user=user:nonadmin1,relation=creator,object=project:002, context=<nil>)
✓ Check(user=user:nonadmin1,relation=viewer,object=project:002, context=<nil>)
✓ Check(user=user:nonadmin1,relation=repo_updater,object=project:002, context=<nil>)
✓ Check(user=user:nonadmin1,relation=provider_creator,object=project:002, context=<nil>)
ⅹ Check(user=user:nonadmin1,relation=creator,object=project:001, context=<nil>): expected=true, got=false, error=<nil>
✓ Check(user=user:nonadmin1,relation=viewer,object=project:001, context=<nil>)
ⅹ Check(user=user:nonadmin1,relation=repo_updater,object=project:001, context=<nil>): expected=true, got=false, error=<nil>
ⅹ Check(user=user:nonadmin1,relation=provider_creator,object=project:001, context=<nil>): expected=true, got=false, error=<nil>
ⅹ Check(user=user:nonadmin1,relation=artifact_updater,object=project:001, context=<nil>): expected=true, got=false, error=<nil>
✓ Check(user=user:nonadmin1,relation=provider_getter,object=project:001, context=<nil>)
```

auth/model.fga
```diff
@@ -0,0 +1,42 @@
...
+
+    define provider_getter: viewer
+    define provider_creator: admin
+    define provider_updater: admin
+    define provider_deleter: admin
+
+    define rule_type_getter: viewer
+    define rule_type_creator: editor
+    define rule_type_updater: editor
+    define rule_type_deleter: editor
```
JAORMX, line 42: Love the minimal model. Makes sense. one usability question... would it make sense to rename the relationships 'getter' in favor of using verbs? Might be more digestible when doing the actual permissions check. What do you think? Same for updater and the rest.

auth/README.md
```diff
@@ -0,0 +1,6 @@
+# About the model
+
+We use per-resource-type permissions off of "project" because
+we don't allow granting permissions on individual resources, only
+on projects.  This allows us to minimize the amount of state we
```
jhrozek, line 5: (I know I was arguing in favour of only importing the projects yesterday, too)

If we ever decide to make the model more granular to make a distinction between "a user can delete a repo" and "a user can delete a build environment", apart from updating the rules, how disruptive would changing of the state we keep in the DB be? Does FGA allow "migrations" of its DB?

JAORMX: It is possible and there is a process for it https://openfga.dev/docs/modeling/migrating/migrating-relations

evankanderson: I'm imagining that we'd define the new relations in the model, possibly bootstrapping off the existing permissions at first as:

```
type rule_type
  relations
    define project: [project]

    define getter: [user] or rule_type_getter from project
    define creator: [user] or rule_type_creator from project
    define updater: [user] or rule_type_updater from project
    define deleter: [user] or rule_type_deleter from project
```

Next, we would update our code on create / update of resources to create the relationships as resources were updated.  Following that, we would run a big pass to sync all our existing objects into the relationship database. (Basically, this is a dual-write scheme -- we might end up syncing some relationships _twice_, but we wouldn't miss any if we start with the live updates.)

Once we think the database is synced, we could do a "dark launch" where we check both the existing project-level permission _and_ the new permission and log any differences, using the result from the legacy permission.  Assuming no differences during the launch window, we'd switch the code to checking only the new permission, then remove the `resource_actioner` permissions from project.

auth/model.fga
```diff
@@ -0,0 +1,42 @@
+model
+  schema 1.1
+
+type user
+
+type project
```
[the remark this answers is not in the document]
evankanderson: I added support for hierarchical groups, see `group.tests.yaml` for example of this working.

evankanderson:
FYI, I removed the `README.md` because I figured out that the FGA DSL *does* support some comments in some places, and was able to move the comment into the model.

I also filed https://github.com/openfga/language/issues/115, because the comment support was clearly half-baked.  😁

JAORMX (changes requested):
Let's rename this directory to `authz` to denote it doesn't have to do with authentication.

Can you add the license headers to the new files?

Otherwise I think this is good to go! Thanks for researching this!

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

auth/model.fga
evankanderson: Done.

authz/tests/fga_test.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	if len(files) == 0 {
+		t.Fatal("no test files found")
+	}
+	for _, f := range files {
+		file := f
+		t.Run(file, func(t *testing.T) {
+			t.Parallel()
+
+			output, err := exec.Command("fga", "model", "test", "--tests", file).CombinedOutput()
```
JAORMX, line 41: We don't do `make boostrap` before the unit tests. We'd need to start doing that, or run the FGA tests as a separate CI job or test unit.

tools/go.mod
```diff
@@ -287,3 +309,5 @@ require (
 	mvdan.cc/lint v0.0.0-20170908181259-adc824a0674b // indirect
 	mvdan.cc/unparam v0.0.0-20221223090309-7455f1af531d // indirect
 )
+
+replace github.com/stacklok/minder => ../.
```
JAORMX, line 313: oh! this is handy! Perhaps it could go in a separate PR?

tools/go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/stacklok/minder/tools
 
-go 1.21
-
-toolchain go1.21.3
+go 1.21.5
```
JAORMX, line 3: Do we need the patch number here? thought we could make due with just major and minor

JAORMX (commented):
I really like how the model ended up. Looks quite understandable.

tools/go.mod
evankanderson: I can do that, but it needs to be a pre-requisite, so that I can reference the `fga` binary in the same repo that I'm adding in this PR.

tools/go.mod
evankanderson: I ran `go mod tidy` and it did this.

authz/tests/fga_test.go
JAORMX: It is cleaner, but make bootstrap is also very time consuming. What about separating the FGA tests and have a dedicated target that does the docker invocation?

tools/go.mod
evankanderson: Oh!  This is pulling the go version from `./go.mod` because of the `replace` directive...

authz/tests/fga_test.go
evankanderson: There's no docker invocation here, just the CLI process.

Oh, but I could actually just import the command code... let me try that.

tools/go.mod
aaguiarz: Maybe this helps :) 

https://github.com/datumforge/datum/blob/07ebc89d89ac14a656f4a4787932cc4c1581b176/.buildkite/pipeline.yaml#L25

authz/tests/fga_test.go
evankanderson: Okay, I've fixed this by actually importing the FGA CLI bits and invoking them directly, so we don't even need `bootstrap` anymore.

The `fga_test.go` still doesn't realize that it depends on `minder.fga`, so it will use cached results if you make changes, but that seems like a minor thing that we can figure out later.

tools/go.mod
evankanderson: I removed the need to have FGA installed, which seems even better.  (I also wanted folks to be able to run the tests with just `go test ./...`, since they'll probably do that from their IDEs.

authz/tests/fga_test.go
```diff
@@ -0,0 +1,70 @@
...
+	if len(files) == 0 {
+		t.Fatal("no test files found")
+	}
+	for _, f := range files {
+		file := f
+		t.Run(file, func(t *testing.T) {
+			// We invoke cobra commands directly, which references some global state in FGA.
+			// Call .SetEnv here to hint to paralleltest that this can't be parallelized.
+			// See https://github.com/kunwardeep/paralleltest/pull/33
+			t.Setenv("PARALLELTEST_NO_PARALLEL", "true")
```
evankanderson, line 44: I would love to have a comment or something that can quiet `parelleltest`, but the code seems to be constructed to parse the AST for methods, rather than comment-driven.  https://github.com/kunwardeep/paralleltest/pull/33 provides a hook to do this, though it also forces us to make this test not run at all in parallel with our other tests.  It runs in 0.2s on my MacBook, so it's not too bad at the moment.

evankanderson: ... and `golangci-lint` hasn't been released to pick up that change yet, so I needed to edit the workflow, too.

evankanderson: I just figured out that I can just use a file-level mutex around calls to `.Execute()`, which is _much_ simpler.

JAORMX (dismissed):
I like your new solution a lot better. thanks for working on this!

JAORMX (dismissed):

JAORMX (approved):

PR #1793, opened 2023-12-01
Run go mod tidy to prune the old go-github version from go.mod and go.sum

jhrozek:
We recently upgraded our go-github version, but because I forgot to run
go mod tidy, the old version was still present in go.mod

Sorry about the PR spam!


JAORMX (approved):

JAORMX:
@jhrozek right, it's because nobody did it.

PR #1797, opened 2023-12-01
WIP: Add POC showing casbin authorization

JAORMX:
Fixes https://github.com/stacklok/epics/issues/129

### DIFF
--- a/cmd/server/app/migrate_up.go
+++ b/cmd/server/app/migrate_up.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 
+	sqladapter "github.com/Blank-Xu/sql-adapter"
 	"github.com/golang-migrate/migrate/v4"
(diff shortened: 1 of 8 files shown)


stacklokbot (commented):

JAORMX:
so, this POC doesn't work yet, but I just wanted to show the approach and idea behind it. The nice thing about using Casbin is that we can use the same database and it's just one more table that gets added to the schema. Also, the code for the model is quite minimal and supports hierarchies. The policy itself is still in progress.

JAORMX:
Still working on the model:
image

It's not as straight-forward to do roles that inherit from each other, and it starts to get somewhat verbose. It's worth noting that requests would read as such: Can subject W with domain X do action Y on object Z? So, we'd check the pair W and X to verify the action Y. However, if the user has multiple roles on the tree, we'd need to do multiple checks (one per role in a domain) to verify if they have permissions. This is not ideal.

Issue #1805, opened 2023-12-04
When ingestion fails with an error, don't attempt to remediate


jhrozek:
### Describe the issue

I ran into a condition when the ingestor failed hard with an error (which is shouldn't but that's another matter). But from what I can tell, the engine still tried to remediate. I think it shouldn't, an error is not defined state and we shouldn't try to remediate because we can't make a reasonable decision on the ingested entity.

### To Reproduce

There's probably other ways but here's how I stumbled upon the bug:
1. enroll a repo that has no main, only master
2. create a profile with a dependabot rule

The ingestion will fail, but remediation was still attempted according to the profile status:
```
profileStatus:
    lastUpdated: "2023-12-04T10:54:31.758415Z"
    profileId: df9a52ce-86c3-45c5-9d1e-31590ff6baaa
    profileName: black-hat-profile
    profileStatus: error
ruleEvaluationStatus:
    - details: 'error ingesting data: could not clone repo: could not clone repo: couldn''t find remote ref "refs/heads/main"'
      entity: repository
      entityInfo:
        provider: github
        repo_name: pam_hbac
        repo_owner: jhrozek
        repository_id: 66c5e152-6bd9-4a5d-b417-39d883e19892
      guidance: |
        Dependabot enables Automated dependency updates for repositories.
        It is recommended that repositories have some form of automated dependency updates enabled
        to ensure that vulnerabilities are not introduced into the codebase.

        For more information, see
        https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
      lastUpdated: "2023-12-04T10:54:31.758415Z"
      profileId: df9a52ce-86c3-45c5-9d1e-31590ff6baaa
      remediationDetails: 'cannot get ref from: error getting commit ref: GET https://api.github.com/repos/jhrozek/pam_hbac/git/ref/heads/main: 404 Not Found []'
      remediationLastUpdated: "2023-12-04T10:54:31.761939Z"
      remediationStatus: error
      ruleId: 59da5553-71d1-47a5-a174-e80ddcd4e90d
      ruleName: dependabot_configured
      status: error
```

PR #1814, opened 2023-12-05
Skip actions on error in the executor

JAORMX:
This detects if there was an error when evaluating and skips actions if it was
so.

The idea is that evaluation error is an unknown state, and actions shouldn't
be performed if we don't actually know the current state of the entity we're
evaluating.

Closes: #1805

### DIFF
--- a/internal/engine/actions/actions.go
+++ b/internal/engine/actions/actions.go
@@ -148,11 +148,6 @@ func (rae *RuleActionsEngine) processAction(
 func shouldRemediate(prevEvalFromDb *db.ListRuleEvaluationsByProfileIdRow, evalErr error) engif.ActionCmd {
 	_ = prevEvalFromDb
 	_ = evalErr
-	// To be used in the future in case we decide to implement specific condition handling for performing a remediation
-	// that is not solely based on the current evaluation error.
-	// Example: Skip remediation if remediate_type is PR, the evalErr is failing, but we already have the last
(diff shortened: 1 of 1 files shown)


internal/engine/executor.go
```diff
@@ -240,10 +241,21 @@ func (e *Executor) evalEntityEvent(
...
-			evalParams.SetEvalErr(rte.Eval(ctx, inf, evalParams))
-
-			// Perform actions, if any
-			evalParams.SetActionsErr(ctx, rte.Actions(ctx, inf, evalParams))
+			evalErr := rte.Eval(ctx, inf, evalParams)
+			evalParams.SetEvalErr(evalErr)
+
+			if evalErr == nil {
+				// Perform actions, if any
+				evalParams.SetActionsErr(ctx, rte.Actions(ctx, inf, evalParams))
```
jhrozek, line 249: Hmm, doesn't this mean that we perform the action if evaluation succeeded and the entity matches the profile? Don't we want to run the actions on `errors.Is(err, ErrEvaluationFailed)` ?

JAORMX: duh, thanks, I missed this.

rdimitrov (commented):
What's left is to update the `Do` method of each remediation type to take into account the command, because so far this was ignored.

rdimitrov (approved):

Issue #1756, opened 2023-11-28
Create an "apply" command for minder profile


rdimitrov:

JAORMX:
Note that I did not modify the `minder profile update` sub-command because of https://github.com/stacklok/minder/issues/1757

PR #1815, opened 2023-12-05
Add subcommand to apply profiles

JAORMX:
This adds the `minder profile apply` sub-command. It also adds
some handy utility functions to simplify writing other profile
sub-commands such as `profile create`.

Closes: #1756

### DIFF
--- a/cmd/cli/app/profile/common.go
+++ b/cmd/cli/app/profile/common.go
@@ -1,0 +1,72 @@
+//
+// Copyright 2023 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


rdimitrov (approved):
Awesome! 🚀 

Not affecting this PR, but have you considered making some of the flags persistent? I know there's a separate issue and this is not part of this scope, just asking so we know if we should visit this command/sub-command pair or it's already covered.

rdimitrov:
> @rdimitrov I think we could start making them persistent, but we could do that as part of addressing that PR.

You're right, let's finish all of this first so we have everything we have to update in the other issue afterwards 💯

PR #1833, opened 2023-12-06
Point Trusty helm chart default to public instance

evankanderson:
It turns out we were encoding some internal pre-launch configuration details in the public helm chart.  This switches the defaults so they should work outside Stacklok's EKS cluster.

### DIFF
--- a/deployment/helm_tests/basic.yaml
+++ b/deployment/helm_tests/basic.yaml
@@ -16,6 +16,9 @@ hostname: "minder-test.example.com"
 db:
   host: postgres.postgres.svc
 
+trusty:
+  endpoint: http://trusty.trusty:8080
+
 serviceAccounts:
   server: minder
   migrate: migrate-job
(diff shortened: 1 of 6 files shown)


rdimitrov (approved):

JAORMX (approved):

teodor-yanev (dismissed):

evankanderson:
> The helm tests need to be updated prior to merging 👍

It turns out I forgot to save `values.yaml`, which was the whole point of the change.  🤦

deployment/helm_tests/basic.yaml
```diff
@@ -16,6 +16,9 @@ hostname: "minder-test.example.com"
 db:
   host: postgres.postgres.svc
 
+trusty:
+  endpoint: http://trusty.trusty:8080
```
evankanderson, line 20: I added this just to have a test of the values override.

JAORMX (approved):

PR #1842, opened 2023-12-07
Remove eventer retry middleware and retrable error construct

JAORMX:
The middleware has caused more issues than anything it solves, moreover,
we're not really leveraging anything that would be retriable.

Let's revisit this once we encounter a use-case.


internal/events/eventer.go
```diff
@@ -132,18 +132,6 @@ func Setup(ctx context.Context, cfg *config.EventConfig) (*Eventer, error) {
 	router.AddMiddleware(
 		// CorrelationID will copy the correlation id from the incoming message's metadata to the produced messages
 		middleware.CorrelationID,
-
-		// The handler function is retried if it returns an error.
-		// After MaxRetries, the message is Nacked and it's up to the PubSub to resend it.
```
evankanderson, line 137: I think the problem here is that the SQL router will retry errors, so we need something to consume the errors before it makes it back to SQL (possibly putting it onto a dead-letter queue).

evankanderson: Sorry, just getting back to this now:

https://github.com/ThreeDotsLabs/watermill-sql/blob/master/pkg/sql/subscriber.go#L375 is most relevant.

PR #1859, opened 2023-12-07
Get default provider from those enrolled in project

### LINKED ISSUE #1745: Make "github" the default value for every use of the "--provider" flag

### LINKED ISSUE #1752: Make the default provider to be taken from the list of enrolled providers for that user

eleftherias:
Fix #1752
Fix #1745

Also, make GitHub the default provider in the CLI, since that's the only option we allow to get past the CLI (this was initially suggested in #1745, which I had mistakenly assumed would be fixed by the server change)

### DIFF
--- a/internal/controlplane/common.go
+++ b/internal/controlplane/common.go
@@ -20,12 +20,14 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/google/uuid"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
 	"github.com/stacklok/minder/internal/auth"
+	"github.com/stacklok/minder/internal/db"
 	"github.com/stacklok/minder/internal/util"
 )
 
@@ -33,6 +35,12 @@ import (
 type ProjectIDGetter interface {
 	// GetProjectId returns the project ID
 	GetProjectId() string
+}
+
+// ProviderNameGetter is an interface that can be implemented by a request
+type ProviderNameGetter interface {
+	// GetProvider returns the provider name
+	GetProvider() string
 }
 
 func providerError(err error) error {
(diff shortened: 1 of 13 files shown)


rdimitrov (dismissed):
@eleftherias - Looks great! 💯🚀

internal/controlplane/common.go
```diff
@@ -35,6 +37,12 @@ type ProjectIDGetter interface {
 	GetProjectId() string
 }
 
+// ProviderNameGetter is an interface that can be implemented by a request
+type ProviderNameGetter interface {
+	// GetProvider returns the provider name
+	GetProvider() string
+}
+
```
evankanderson, line 45: I ❤️  creating an interface like this to fetch the common field across many different protos.  Nice job!

internal/controlplane/common.go
```diff
@@ -59,3 +67,32 @@ func getProjectFromRequestOrDefault(ctx context.Context, in ProjectIDGetter) (uu
...
+	providers, err := store.ListProvidersByProjectID(ctx, projectId)
+	if err != nil {
+		return db.Provider{}, status.Errorf(codes.InvalidArgument, "cannot retrieve providers: %s", err)
+	}
+	// if we do not have a provider name, check if we can infer it
+	if in.GetProvider() == "" {
+		if len(providers) == 1 {
+			return providers[0], nil
+		}
+		return db.Provider{}, status.Errorf(codes.InvalidArgument, "cannot infer provider name: %s", err)
```
evankanderson, line 86: Do you want to include `len(providers)` in the error?  Also, I think `err` is `nil` here, so you probably don't want to return that.

Consider using `util.UservisibleError` so that the error gets back to the CLI.

```suggestion
		return db.Provider{}, util.UserVisibleError(codes.InvalidArgument, "cannot infer provider, there are %d providers available", len(providers))
```

internal/controlplane/common.go
```diff
@@ -59,3 +67,32 @@ func getProjectFromRequestOrDefault(ctx context.Context, in ProjectIDGetter) (uu
...
+	}
+
+	matchesName := func(provider db.Provider) bool {
+		return provider.Name == in.GetProvider()
+	}
+
+	i := slices.IndexFunc(providers, matchesName)
+	if i == -1 {
+		return db.Provider{}, status.Errorf(codes.InvalidArgument, "invalid provider name: %s", err)
```
evankanderson, line 95: (Again, `err` is `nil` here.  You probably also want to use `util.UserVisibleError`)

```suggestion
		return db.Provider{}, util.UserVisibleError(codes.InvalidArgument, "invalid provider name: %s", in.GetProvider())
```

evankanderson (approved):
This is great!  A couple small nits on the returned error messages, but this looks good and will remove some amount of my annoyance.

I was going to ask whether we needed to change the default provider on the client side given that we'll support `""` on the server, but this seems fine for now, and we can always change the client defaults again later.  (And it's good to remove the flag as required with an empty default, so +1 to that!)

JAORMX (approved):

PR #1873, opened 2023-12-08
Make status a sub-command of profile

### LINKED ISSUE #1760: Refactor the minder cli so "profile_status" becomes a "status" sub-command for "profile"

eleftherias:
Fix #1760


evankanderson (commented):
I just noticed that we forgot to update `docs/docs/getting_started/remediations.md` and the other docs when we moved these commands.

Issue #581, opened 2023-08-02
API calls should include Context in the request instead of org_id/group_id


yrobla:
**Describe the bug**
Now we have the concept of Context, where we should be able to pass either a group id or a context id. We should reflect that on the proto files.
We should also include the full context in the response of the calls

PR #1892, opened 2023-12-11
Refactor-add context to grpc reqests

teodor-yanev:
Closes: https://github.com/stacklok/minder/issues/581

This PR adds Context to most of our RPCs.
As part of this, it also reverts one of the changes in https://github.com/stacklok/minder/pull/1831 to make the "provider" field optional. This is due to the constraint of optional fields not being allowed to be used as path parameters.


proto/minder/v1/minder.proto
```diff
@@ -996,7 +1028,7 @@ message Provider {
 // Context defines the context in which a rule is evaluated.
 // this normally refers to a combination of the provider, organization and project.
 message Context {
-    optional string provider = 1;
+    string provider = 1;
     optional string project = 3;
```
JAORMX, line 1032: I think `project` shouldn't be optional.

evankanderson (commented):
Added a few comments -- I think we want to bridge compatibility between old clients and new server a bit more cleanly, but I put in a suggestion.

teodor-yanev:
Closing in favour of: https://github.com/stacklok/minder/pull/1911

PR #1898, opened 2023-12-11
Retire message Provider -- it is not used

evankanderson:


rdimitrov (approved):

evankanderson:
> I left this there because I was indeed planning on taking this into use in the future for the provider API. I guess I can revert this once that work gets started.

That would be my preference -- in particular, we might discover that we need slightly different fields, so writing it fresh seems better than having people patch it forward without consideration for the next 2 months.

JAORMX (approved):

Issue #1754, opened 2023-11-28
Converge on using one type of table output for minder, i.e. profile list is different from the rest


rdimitrov:

rdimitrov:
AI: Explore if lipgloss has a autoscaling enabled for presenting tables so we converge on it

PR #1899, opened 2023-12-11
Use terminal width for Minder output

evankanderson:
This was bothering me when runing `minder auth login` and other commands in a default 80x24 MacOS terminal window.

Before:

image
(note that the output is cropped top & bottom -- more than 24 lines of output, since all the fancy banner lines wrap)

After:
Screenshot 2023-12-11 at 11 48 24 AM

### DIFF
--- a/internal/util/cli/style.go
+++ b/internal/util/cli/style.go
@@ -22,8 +22,11 @@
 package cli
 
 import (
+	"os"
+
 	"github.com/charmbracelet/bubbles/table"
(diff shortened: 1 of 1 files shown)


evankanderson:
Mostly, I wanted to fix this because the output looks crummy with the default terminal settings.  It may be related to https://github.com/stacklok/minder/issues/1754

rdimitrov (approved):
This looks nice 👍

PR #1908, opened 2023-12-12
Add default branch as a repository field

JAORMX:
This will be handy to stop hard-coding the branch in the Git-related
evaluations, we could just rely on the default branch for the project.

In order to facilitate adding/removing fields from the repo protobuf, this
also introduces a helper function that centralizes setting the values
from a database entry.

Note that if no default branch is set, this just assumes an empty string on
the protobuf level.

Note that at the moment this will still only work with new repositories,
as we're missing periodic reconciliation.


eleftherias (commented):
Is the default branch used currently, or is the idea that we'll store it now and use it in the future?

JAORMX:
@eleftherias it's not used. The idea is that we'll store it now and use it in our profiles/rules later. We could use it easily in the git ingester to not have to specify the branch at all and rely on the default.

eleftherias (approved):

PR #1918, opened 2023-12-13
Split minder server and CLI config

### LINKED ISSUE #1740: Split and create a dedicated config files for minder and minder-server

eleftherias:
Fix #1740

The CLI continues to use the default name `config.yaml`, to keep backwards compatibility for SaaS users.
The server now picks up a file called `server-config.yaml` by default.


JAORMX (commented):
should we have a dedicated golang structure to set the client CLI? it would help us do validations without having to solely rely on the server config as we do right now for parsing the client config.

rdimitrov:
> should we have a dedicated golang structure to set the client CLI? it would help us do validations without having to solely rely on the server config as we do right now for parsing the client config.

I was thinking about it too. Currently we don't parse the CLI config using the server config structure. It would be good to have a dedicated CLI config struct too so we can state explicitly what is expected to have in the config file.

JAORMX:
@eleftherias the dedicated CLI config struct can be done in a separate PR if you prefer. But first I'd like to read your thoughts no this.

eleftherias:
> @eleftherias the dedicated CLI config struct can be done in a separate PR if you prefer. But first I'd like to read your thoughts no this.

I agree it would be nice to have a CLI config struct. Since we never used the struct for the CLI, I would prefer to do it in separate PR.

JAORMX (approved):

rdimitrov (approved):

Issue #1758, opened 2023-11-28
Revisit and rename all command flags accordingly so they are consistent


rdimitrov:

JAORMX:
@rdimitrov I don't remember what this one was about? Can you add more description?

rdimitrov:
@JAORMX - Of course 👍 It's about making sure all flags have the same long and short name across all commands, i.e. things like `project` and `project-id`.

JAORMX:
ah! got it! thanks

PR #1919, opened 2023-12-13
Minder CLI improvements - table interface, refactor use of flags, context, etc.

rdimitrov:
**The following PR:**
* Leverages the use of PersistentFlags and FlagGroups. This resulted in a lot of housekeeping on helper functions we don't really had to have since they are already provided by cobra/viper. Essentially there a no major code logic changes but because of this cleanup the PR looks way bigger than I expected.
* Refactors the use of tables within the cli. Adds a Table interface with two supported tables (Simple and Glossy). Currently everything is migrated to the Simple one for consistency, but the glossy one is left for eventual PoC since it looks much better, but lacks the appropriate features (or we don't know how to implement them yet).
* Adds provider and project values (I have started this earlier so I haven't rebased the latest context changes, will do though that's why it's WIP)
* Went through the CLI stack of commands and made the codebase consistent, refer to the same default values, etc.
* Fixed an issue where errors were showing twice
* Went through all cli commands and made sure they use the cli.MessageAndReturn function. Updated the function to wrap a custom error type we later unwrap in ExitNicely so we can handle message outputs better.
* Reduced the details input of an GRPC error to description only
* Renamed `--repo` to `name` when doing repo register for consistency with the rest of the repo commands
* Added a new context for each batched profile call
* Converged on using the basic table rendering framework, at least for the time being. We can revisit this but let's not stretch this PR more.

**What's left:**
- [x] Rebase the latest changes related to context and config files
- [x] Finalise the tables usage
- [x] Ensure a separate context is used for all CLI commands that batch GRPC calls 
- [x] Ensure and update docs to reflect the latest state
- [x] Context.Project is UUID, not name. Either implement or fix that in the docs. (Added it in the description, the switch from UUID to name would be a separate PR)

Fixes: #1758
Fixes: #1754 
Fixes: #1751 
Fixes: #1749 
Fixes: #1748


cmd/cli/app/artifact/artifact_get.go
```diff
@@ -17,60 +17,71 @@ package artifact
...
-		art, err := client.GetArtifactById(ctx, &pb.GetArtifactByIdRequest{
-			Id:             artifactID,
-			LatestVersions: latest_versions,
-			Tag:            tag,
-		})
-		if err != nil {
-			return cli.MessageAndError(cmd, "Error getting artifact by id", err)
-		}
+	// Ensure provider is supported
+	if !app.IsProviderSupported(provider) {
```
JAORMX, line 51: I would actually like to get rid of these checks in the CLI. We will support other providers in the future, and we should just output a relevant error message if the provider isn't registered in the db. So it should be a server-side check instead.

cmd/cli/app/artifact/artifact_list.go
```diff
@@ -18,92 +18,93 @@ package artifact
...
+// listCommand is the artifact list subcommand
+func listCommand(ctx context.Context, cmd *cobra.Command, conn *grpc.ClientConn) error {
+	client := minderv1.NewArtifactServiceClient(conn)
+
+	provider := viper.GetString("provider")
+	project := viper.GetString("project")
+	format := viper.GetString("output")
+
+	// Ensure provider is supported
+	if !app.IsProviderSupported(provider) {
```
JAORMX, line 51: Let's get rid of this.

JAORMX (changes requested):

JAORMX (changes requested):
This mostly looks good, the only thing I don't agree with is the `app.IsProviderSupported` function, which I think should be removed as it'll prevent us from adding other providers in the future. Let's not add code we have to remove later for little benefit.

JAORMX (approved):
Trying this locally, the tables look good! the only one that looks a little off is the `whoami` list which looks a bit too slim. Is there a way for us to enforce minimal width?

```
+---------------+--------------------------------------+
|      KEY      |                VALUE                 |
+---------------+--------------------------------------+
| Minder Server | api.stacklok.com:443                 |
+---------------+--------------------------------------+
| Project       | jaormx /                             |
|               | xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx |
+---------------+--------------------------------------+
```

JAORMX (approved):

JAORMX:
@rdimitrov what changed?

rdimitrov:
> @rdimitrov what changed?

@JAORMX - Jakub merged his PR so I had to rebase. Also there was a diff in the make gen output.

rdimitrov:
> Trying this locally, the tables look good! the only one that looks a little off is the `whoami` list which looks a bit too slim. Is there a way for us to enforce minimal width?
> 
> ```
> +---------------+--------------------------------------+
> |      KEY      |                VALUE                 |
> +---------------+--------------------------------------+
> | Minder Server | api.stacklok.com:443                 |
> +---------------+--------------------------------------+
> | Project       | jaormx /                             |
> |               | xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx |
> +---------------+--------------------------------------+
> ```

Added a minimum width for tables of type key/value 👍

JAORMX (approved):
It still looks a little too slim but we can fix that later

PR #1924, opened 2023-12-13
Create context between steps in quickstart

jhrozek:
I was using the quickstart command to record a demo and was frequently
hitting deadline exceeded if I wasn't quick enough.

Let's add a context around each step so that the user actually has time
to read the text we present.

### DIFF
--- a/cmd/cli/app/quickstart/quickstart.go
+++ b/cmd/cli/app/quickstart/quickstart.go
@@ -159,7 +159,10 @@ var cmd = &cobra.Command{
 			return err
 		}
 
-		// Step 2 - Confirm repository registration
+		// Step 2 - Confirm repository registration (new context, so we don't time out)
+		ctx, cancel := getQuickstartContext(cmd.Context(), viper.GetViper())
+		defer cancel()
+
 		yes = cli.PrintYesNoPrompt(cmd,
 			stepPromptMsgRegister,
 			"Proceed?",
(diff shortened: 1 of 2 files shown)


rdimitrov (dismissed):
I know I added it for some of the steps, but apparently it wasn't enough to cover up for the time spend explaining between steps 😃 Thanks!

JAORMX (dismissed):

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -143,7 +143,10 @@ var cmd = &cobra.Command{
 			return nil
 		}
 
-		// Step 1 - Confirm enrolling
+		// Step 1 - Confirm enrolling (new context, so we don't time out)
+		ctx, cancel := cli.GetAppContext(cmd.Context(), viper.GetViper())
```
AGMETEOR, line 147: Just a question. Does this actually change the context deadline?

jhrozek: yes and no. It doesn't actually change the timeout (the new patch I just pushed does), but what it does is instantiate a new context per step so that the old doesn't time out.

JAORMX (approved):

cmd/cli/app/quickstart/quickstart.go
AGMETEOR: I see. Since the output from the command requires some reading, would increasing the deadline slightly be a bad idea?

rdimitrov (approved):

cmd/cli/app/quickstart/quickstart.go
jhrozek: that's what the new version of the patch I pushed does (doubles the timeout from 10 to 20 seconds for each step)

PR #1926, opened 2023-12-14
Add protobuf for configuring the actions tag-to-sha replacement

JAORMX:
We need to somehow configure the tag-to-sha remediation. This adds the protobuf
that would allow us to do so.


proto/minder/v1/minder.proto
```diff
@@ -1272,6 +1272,11 @@ message RuleType {
                     optional string mode = 3;
                 }
 
+                message ActionsReplaceTagsWithSha {
+                    // List of actions to exclude from the replacement
```
jhrozek, line 1276: Not for this patch, but it would be nice to get good documentation for the rule_type when we get to amending it, e.g. can I exclude only `actions/checkout` from being expanded or can I also exclude `actions/checkout@v3` to still expand `v4`.

jhrozek (approved):

PR #1934, opened 2023-12-14
Parse GitHub Action exclusion from rule def

JAORMX:
In the PR remediation that replaces tags for checksums, we need a way to
pass in an exclusion list from the user input in the profile.

This implements that by parsing an `exclude` key from the `def` section
of the rule.


jhrozek (approved):
seems to work fine with a custom rule_type

Issue #1935, opened 2023-12-14
`minder profile get -i <fdfdfd>` results in an error.


JAORMX:
Per @AGMETEOR's comment:

Meanwhile `minder profile get -i <fdfdfd>` doesn't seem to be working. I keep getting `INVALID_ARGUMENT`
```
Error getting profile: Code: 3
Name: INVALID_ARGUMENT
Description: Invalid argument
Details: InvalidArgument indicates client specified an invalid argument.
```

_Originally posted by @AGMETEOR in https://github.com/stacklok/minder/issues/1932#issuecomment-1855784907_

JAORMX:
@AGMETEOR working on this now.

AGMETEOR:
Seems it has to be fixed from the server itself. Is the server code also open source?

rdimitrov:
> Seems it has to be fixed from the server itself. Is the server code also open source?

Yes, it's in `cmd/server` 👍

AGMETEOR:
> > Seems it has to be fixed from the server itself. Is the server code also open source?
> 
> Yes, it's in `cmd/server` 👍

Gotcha

JAORMX:
We have a hypothesis of what's causing this issue. A potential fix will be posted soon.

AGMETEOR:
Let me ask this here since it's related to the fix. Is there a make target to generate `minder.pb.go`?

rdimitrov:
> Let me ask this here since it's related to the fix. Is there a make target to generate `minder.pb.go`?
Answered in https://github.com/stacklok/minder/pull/1936#discussion_r1426730074

JAORMX:
@AGMETEOR the fix is in prod now.

JAORMX:
> Let me ask this here since it's related to the fix. Is there a make target to generate `minder.pb.go`?

you can run `make buf` or `make gen` (`gen` will run all code generators)

AGMETEOR:
> @AGMETEOR the fix is in prod now.

Confirmed it's now working. Thanks for the fix.

PR #1936, opened 2023-12-14
Make GetProfileById authorized on profile ownership

jhrozek:
We didn't set any permissions on this method which always denied access.

Fixes: #1935


JAORMX (approved):

pkg/api/protobuf/go/minder/v1/minder.pb.go
```diff
@@ -15,7 +15,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.31.0-devel
+// 	protoc-gen-go v1.31.0
```
jhrozek, line 18: there seems to be a lot of spam in this diff is it related to the different protoc version I have?

JAORMX: probably, yes

rdimitrov (approved):

JAORMX:
superseded by https://github.com/stacklok/minder/pull/1937 (to reduce the noise from `protoc`)

pkg/api/protobuf/go/minder/v1/minder.pb.go
AGMETEOR: Is there a make target I can use to generate this file?

rdimitrov: yeah, `make gen` takes care of all generation-related things 👍

PR #1939, opened 2023-12-15
Set documentation for context appropriately

JAORMX:
The dev docs were leaking into the OpenAPI that's generated from minder.
This fixes that to generate more relevant docs.


teodor-yanev (approved):
Thanks 👍

PR #1942, opened 2023-12-15

stacklokbot:

and not a repository fork.

For more information, see

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -25,10 +25,8 @@ on:
   workflow_dispatch:
   release:
     types: [created]
-
 permissions:
   contents: write
-
 jobs:
   ldflags_args:
     runs-on: ubuntu-latest
@@ -39,7 +37,7 @@ jobs:
       tree-state: ${{ steps.ldflags.outputs.tree-state }}
     steps:
       - id: checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
         with:
           fetch-depth: 0
       - id: ldflags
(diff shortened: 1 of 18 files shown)


.github/workflows/releaser.yml
```diff
@@ -108,7 +105,7 @@ jobs:
       actions: read # To read the workflow path.
       id-token: write # To sign the provenance.
       contents: write # To add assets to a release.
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.9.0
```
rdimitrov, line 111: @JAORMX @jhrozek - I recall we should exclude some of these (one had a problem with the digest reference)

JAORMX: @rdimitrov you are right, https://github.com/stacklok/frizbee/blob/main/.frizbee.yml#L5

JAORMX: This is configurable via minder though, it's just about setting up the right `excludes` in there.

rdimitrov: until it is configured we should hold off merging this one then 👍

jhrozek: This is our configuration:
```
  - type: actions_check_pinned_tags
    def:
      exclude:
        - slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@1.9.0
```
It also marks the frizbee repo as succeeded, so currently I'm unsure what's going on

rdimitrov: Also strange that the CLA got triggered, I remember we added the bot when we did the first enrolment. cc: @dussab

JAORMX: @jhrozek aha! the tag doesn't match. The configuration uses `1.9.0` while we have configured `v1.9.0` in the action.

JAORMX: @jhrozek the right tag seems to be `v1.9.0`

rdimitrov: hawk-eye-ozz in action! 🦅 😃

jhrozek: This was fixed by modifying the profile. Thanks for the sharp eyes!

jhrozek:
Going to close this PR and let minder create a new one to get a nice screen for a blog post :-)

PR #1943, opened 2023-12-15
Take into account frizbee configuration in repository

JAORMX:
This modifies the PR action `minder.actions.replace_tags_with_sha` to also
take into account the `frizbee.yml` configuration if it's available in the
repository it's currently evaluating. This is slightly better usbility, as
it allows for overwrites and custom configurations for the different repos.

A repo may still use the global configuration.


jhrozek (approved):
love it

PR #1944, opened 2023-12-15
Don't attempt to reopen a PR after updating a branch if it already exists

jhrozek:
The PR remediation was checking if a PR with the same contents already
exist as a fast and cheap check to avoid triggering the remediation
again. But in case the profile changes or the contents change for
another reason, we want to push to the existing branch to update it, but
don't try to open a PR again as that would fail with 422.

To do that, this patch adds a check if a PR is already open from the
branch we just pushed to and if it is, just returns before opening the
PR again.

### DIFF
--- a/internal/engine/actions/remediate/pull_request/pull_request.go
+++ b/internal/engine/actions/remediate/pull_request/pull_request.go
@@ -207,7 +207,7 @@ func (r *Remediator) Do(
 	var remErr error
 	switch remAction {
 	case interfaces.ActionOptOn:
-		alreadyExists, err := prAlreadyExists(ctx, r.ghCli, repo, magicComment)
+		alreadyExists, err := prWithContentAlreadyExists(ctx, r.ghCli, repo, magicComment)
 		if err != nil {
 			return nil, fmt.Errorf("cannot check if PR already exists: %w", err)
 		}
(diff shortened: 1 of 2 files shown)


JAORMX (approved):

rdimitrov (approved):

JAORMX:
Oops! You need to update the unit tests with the new expected GH API calls:
```
❌ TestPullRequestRemediate/open_a_PR (190ms)
      pull_request.go:481: Unexpected call to *mockgh.MockGitHub.ListPullRequests([context.Background stacklok minder 0xc0006a6000]) at /home/runner/work/minder/minder/internal/engine/actions/remediate/pull_request/pull_request.go:481 because: 
          expected call at /home/runner/work/minder/minder/internal/engine/actions/remediate/pull_request/pull_request_test.go:197 has already been called the max number of times
      controller.go:269: missing call(s) to *mockgh.MockGitHub.CreatePullRequest(is anything, is equal to stacklok (string), is equal to minder (string), is equal to Add Dependabot configuration for gomod (string), is equal to <!-- minder: pr-remediation-body: { "ContentSha": "1041e57c2fac284bdb7827ce55c6e3cb609e97b9" } -->
          
          Adds Dependabot configuration for gomod (string), is equal to refs/heads/minder_add_dependabot_configuration_for_gomod (string), is equal to main (string)) /home/runner/work/minder/minder/internal/engine/actions/remediate/pull_request/pull_request_test.go:398
      controller.go:269: aborting test due to missing call(s)
```

JAORMX (changes requested):
Unit test updates needed

PR #1945, opened 2023-12-15

stacklokbot:

and not a repository fork.

For more information, see


rdimitrov:
cc: @dussab

dussab:
The issue might be related to the fact the CLA Tool cannot determine the user account associated with the commit. The Stacklok bot commit seems to be missing the email address associated with the Stacklokbot commit. Here is the commit for reference: https://github.com/stacklok/minder/commit/a25953167b4774c52a930645714f97cf9c68be4c.patch @jhrozek is investigating this further.

PR #1948, opened 2023-12-18
Upgrade golangci-lint to v1.55.2 in CI

JAORMX:


.github/workflows/golangci-lint.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.54.2
+          version: v1.52.2
```
teodor-yanev, line 21: Did you want to make this `v1.55.2` instead?

JAORMX: duh, thanks

teodor-yanev (approved):

PR #1954, opened 2023-12-18
Default to v1 and profile if profile type or version are not set to fix old clients

jhrozek:
We added profile type and validation recently which is great, but since
those fields are not set by older clients who didn't have those fields
in their protobuf definitions, those fields would come to the server as
empty and the validation would have failed.

This would impact e.g. an older client running quickstart against a new
server.

We should remove this workaround after a couple of releases when we can
be reasonably sure everyone has migrated to a new client.


JAORMX (approved):

JAORMX:
oops! Thanks for adding it.

rdimitrov (approved):

PR #1956, opened 2023-12-18
Add metric for message processing latency

### LINKED ISSUE #1839: Add a metric to report on message publish -> process latency
While Watermill-SQL's Postgres implementation happens to include a `created_at` timestamp, it's not exposed anywhere (and there's no clear place in `message.Message` to store it.

We should add a `timestamp` field to our Metadata, and then use that when handling an event to report (`NOW() - timestamp`) as an OpenTelemetry Gauge metric, which would allow measuring our queue processing latency.

eleftherias:
Fix #1839

Opted for a histogram instead of gauge, because the library doesn't have a synchronous gauge.

### DIFF
--- a/internal/events/eventer.go
+++ b/internal/events/eventer.go
@@ -35,6 +35,8 @@ import (
 	promgo "github.com/prometheus/client_golang/prometheus"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/metric"
 
 	"github.com/stacklok/minder/internal/config"
 )
@@ -50,6 +52,7 @@ const (
 	SQLDriver       = "sql"
 
 	DeadLetterQueueTopic = "dead_letter_queue"
+	PublishedKey         = "published_at"
 )
 
 const (
@@ -100,8 +103,14 @@ type Eventer struct {
 	// webhookSubscriber will subscribe to the webhook topic and handle incoming events
 	webhookSubscriber message.Subscriber
 	// TODO: We'll have a Final publisher that will publish to the final topic
+	metrics *messageInstruments
 
 	closer driverCloser
+}
+
+type messageInstruments struct {
+	// message processing time duration histogram
+	messageProcessingTimeHistogram metric.Int64Histogram
 }
 
 var _ Registrar = (*Eventer)(nil)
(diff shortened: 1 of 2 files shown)


JAORMX (commented):
Note this Middleware is only added to the executor. If you want to capture general message processing, you should add this to the eventer itself

evankanderson (approved):
A Histogram may be more useful in the future, if we end up with an implementation which can handle multiple in-flight events at once (i.e. out-of-order execution).  I almost thought about asking for one, then decided it was overkill.  😁

JAORMX (changes requested):
oops, meant to request changes.

evankanderson:
I'd be comfortable with a Middleware.  I'm also wondering whether we should contribute to https://github.com/ThreeDotsLabs/watermill/blob/master/components/metrics/handler.go#L62 rather than adding our own additional Middleware.

evankanderson:
Actually, we might want to hook `DecoratePublisher` and `DecorateSubscriber` from here: https://github.com/ThreeDotsLabs/watermill/blob/v1.3.5/components/metrics/builder.go to set the Metadata fields we need.

evankanderson:
I filed https://github.com/ThreeDotsLabs/watermill/issues/417 to see if this would be accepted upstream.  In the meantime, it feels like it _would_ make sense to implement this as a Middleware / PublisherDecorator in `eventer.go`.

eleftherias:
@JAORMX @evankanderson I moved the code into the eventer. 
I see a few options for where the metric can be read, let me know if you think there's a better place for it than the handler function.

internal/events/eventer.go
```diff
@@ -280,6 +302,12 @@ func (e *Eventer) Register(
 		topic,
 		e.webhookSubscriber,
 		func(msg *message.Message) error {
+			if builtAt := msg.Metadata.Get(BuiltKey); builtAt != "" {
```
evankanderson, line 305: I think this is okay, but we could also add it at line 141 in the handler chain for the router.  (I think I have a slight preference for that -- it will tell us when we're taking the message off the queue, before we dead-letter, retry, etc.) On the other hand, you'll have to worry about conflicting with @teodor-yanev 's #1994 , so I can see a tactical argument for doing it here.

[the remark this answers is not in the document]
eleftherias: Yes, it will return an error failing to parse the time

eleftherias: I moved it to where you suggested. If there are merge conflicts they should be easy enough to fix.

internal/events/eventer.go
```diff
@@ -48,6 +50,7 @@ const (
 
 	GoChannelDriver = "go-channel"
 	SQLDriver       = "sql"
+	BuiltKey        = "built_at"
```
JAORMX, line 53: Can you add some docs on what `BuiltKey` is supposed to be? Is that the time that the message was built?

JAORMX: Ah! I see now, it's the time when the message was published.

JAORMX: @eleftherias should this be `published_at`?

JAORMX (approved):
shipit :ship:

JAORMX (changes requested):
oops, the unit test issue seems legit

eleftherias:
> oops, the unit test issue seems legit

@JAORMX  On it.

JAORMX (approved):

PR #1960, opened 2023-12-18
Pass through error messages from AuthFromMD for debugging

evankanderson:
See https://github.com/stacklok/minder/discussions/1957 -- requests though the GRPC gateway don't seem to have proper auth (returning `codes.Unauthenticated`, but we're not providing anything more useful in terms of error message to distinguish different cases.

I'd really love it if I'd written tests for `UnaryAuthorizationInterceptor` earlier, but I compounded the technical debt here to get the quick fix in.

### DIFF
--- a/deployment/helm/Chart.lock
+++ b/deployment/helm/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: oci://registry-1.docker.io/bitnamicharts
-  version: 2.13.3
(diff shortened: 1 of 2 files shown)


AGMETEOR (approved):

JAORMX (approved):

JAORMX:
Oh, fails the linter

JAORMX (approved):

PR #1967, opened 2023-12-19
Artifact list: filter per repo

jhrozek:
With more repositories enrolled, it's not trivial to get an overview of
what artifacts did minder download. This commit adds a `--from` switch
to artifact list that allows to simply filter the artifacts by a
repository.

I also realied that we don't really have any docs for the artifact support,
but that will come in a subsequent PR.


JAORMX (approved):
handy

PR #1983, opened 2023-12-20
Remove config parsing from `mindev`

JAORMX:
The dev tool does not use a configuration file. Let's remove the parsing from there.


rdimitrov (commented):
Lintin' needs fixin'

JAORMX:
I really need to configure pre-commit hooks in my setup

rdimitrov (approved):

Issue #1762, opened 2023-11-28
Revisit and update the "minder artifact" command, i.e. show more meaningful output


rdimitrov:

jhrozek:
@rdimitrov so far I'm thinking about these changes:

* for artifact list:
  - add a `from` switch that would currently only support `=repository=owner/name`, e.g. `artifact list --from=repository=jakubtestorg/testrepo`. This would be extensible for the future if we ever add support for artifacts from registries
* for artifact get:
 - get the artifact by name. I guess simply owner/repo/artifact_name, the point is to just pass a string in the protobuf and deal with it on the server side
 - add a table output (might be done after your PR) which would be the default. Also JSON and YAML output as for other commands
 - in the table output, add whether there is a verified signature as unicode characters like for profile status
 - maybe filter by tags?
 - maybe filter by signature OK/not OK?

Does that sound good to you?

rdimitrov:
@jhrozek - Sounds great! 👍 Perhaps one more thing - should we make the signed part a bit more informational. For example if the artifact is signed show the entity that signed it (the action) along with a date when that happened.

jhrozek:
Since there are already quite a few issues around CLI, let's just use a list here:
 - [ ] artifact list by repo
 - [ ] artifact get by name
 - [ ] artifact get table
 - [ ] artifact get verified signature and signing entity
 from the above list, filter by tags was already there.

PR #1990, opened 2023-12-20
Add get-by-name to artifact get

jhrozek:
- Add SQL search for artifact versions by artifact name
- Add gRPC call to get artifact by name
- artifact get: Get artifact by name

Fixes: #1762


rdimitrov (commented):
Let's use cobra for the checks about the flags (if they are present or mutually exclusive or where at least one should be present out of a pair). We already do that for some commands so you can use them as examples.

jhrozek:
> Let's use cobra for the checks about the flags (if they are present or mutually exclusive or where at least one should be present out of a pair). We already do that for some commands so you can use them as examples.

Good idea, thank you. I added a new patch into the PR for easier review (they'll all be squashed during push)

rdimitrov (dismissed):

proto/minder/v1/minder.proto
```diff
@@ -78,6 +78,16 @@ service ArtifactService {
             auth_scope: OBJECT_OWNER_PROJECT
         };
     }
+
+    rpc GetArtifactByName (GetArtifactByNameRequest) returns (GetArtifactByNameResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/artifact/{name}"
```
evankanderson, line 84: Do we have to worry about names and IDs colliding?  If so, we could add some additional text to this one to ensure it doesn't conflict.

rdimitrov: I think the expected format for `--name` is `repoOwner/repoName/artifactName` which should not allow for such conflicts, right?

jhrozek: yes but I think Evan has a point and I was just being sloppy. Let me change the path

jhrozek: @evankanderson @rdimitrov one more patch in this PR, WDYT?

rdimitrov: Ah, I was too fast to reply to this one. Got it now 👍

rdimitrov (approved):

proto/minder/v1/minder.proto
jhrozek: thans for the ack, I'm going to wait for Evan to see if this is what he meant

jhrozek: actually it seems like folks would like to do a release, do I'm going to go YOLO and just merge this.

PR #1993, opened 2023-12-20
Add `type` to diff ingester protobuf

JAORMX:
The diff ingester was initially just coded to support a diff in dependencies.

Howver, this ingester is very handy for other things such as full diffs or
diffs of github actions.

This introduces a `type` to the protobuf which allows us to set the diff
type as needed.

This also introduces validation which makes sure that a user can only set
an empty or `dep` diff type. Empty keeps backwards compatibility.


JAORMX:
@jhrozek @teodor-yanev from today's conversation.

jhrozek (approved):

proto/minder/v1/minder.proto
```diff
@@ -1139,7 +1139,15 @@ message DiffType {
...
     }
+
+    // ecosystems is the list of ecosystems to be used
+    // for the "dep" diff type.
     repeated Ecosystem ecosystems = 1;
+
+    // type is the type of diff ingestor to use.
+    // The default is "dep" which will leverage
+    // the ecosystems array.
+    string type = 2;
```
evankanderson, line 1150: This seems fine -- I might have suggested using a `oneof`, but I seem to recall that those can actually be a bit a bit tricky in proto (plus we'd have the issue of doing a compatibility bridge dance).

evankanderson (approved):

proto/minder/v1/minder.proto
JAORMX: The main thing preventing us from using `oneof` (which would have been handy for many cases) is stuff like these:
* https://github.com/stacklok/minder/blob/main/pkg/api/protobuf/go/minder/v1/rule_types.go#L30-L37
* https://github.com/stacklok/minder/blob/main/internal/util/jsonyaml/jsonyamlutils.go#L52-L60

changing those to use `protojson` instead of `json`, while it would do the trick to correctly parse protobuf as we'd need.. For some reason it also currently breaks parsing either rule types or profiles. So we haven't changed it because of that.

Issue #1804, opened 2023-12-04
Failing to clone a branch should not be an error but a failure


jhrozek:
### Describe the issue

I managed to register a repo that had no main (an old one with master only) branch and because the git ingestor couldn't ingest the main branch, the profile ended with an error. It should probably be a failure instead (just like branch protection for a missing branch is not a a failure)

### To Reproduce

1. register a repo that doesn't have the branch that the rules refer to
2. create a profile that ingests the repo (e.g. dependabot)

PR #2001, opened 2023-12-21
Surface git clone errors as failures if the branch doesn't exist

JAORMX:
This changes the logic for the git ingester to output an evaluation
failure if the branch didn't exist.

Closes: #1804


JAORMX:
@jhrozek is this what you had in mind for the bug?

jhrozek (approved):
yes, great, thank you

PR #2002, opened 2023-12-21
Deploy docs on merges to `main`

JAORMX:
The minder docs are mean to be bleeding edge. This changes the deploy
trigger to build on merges to `main` instead of on every release.


rdimitrov:
The reasoning is there's we've come to a difference between:
* https://docs.stacklok.com/minder/getting_started/first_profile
and 
* https://minder-docs.stacklok.dev/getting_started/first_profile

rdimitrov (approved):

PR #2008, opened 2023-12-21
Use variadic parameters for the table interface

JAORMX:
Why? It looks fancier, that's all


JAORMX:
> Nice! The AddRowWithColor method should also be updated in that case 👍

we can't really do that since it takes two sets of strings, the headers and the colors. You can't use or more two variadic sets of parameters in one function.

rdimitrov (approved):

PR #2009, opened 2023-12-21
Introduce struct to handle colored columns

JAORMX:
This makes adding colored columns a little more ergonomic


JAORMX:
uhm... there's a bug somewhere

teodor-yanev (dismissed):

JAORMX:
Found the bug

Issue #2018, opened 2023-12-22
Profile status colors are not set correctly.


JAORMX:
### Describe the issue

When printing a status of a profile that has both remediations and alerts, the colors are not printed correctly.

e.g.

![image](https://github.com/stacklok/minder/assets/145564/b9555f97-4d29-48c8-bad9-4559e06a16c0)

### To Reproduce

The colors should match the status. e.g. red for failures and green for success.

### What version are you using?

`main`

PR #2019, opened 2023-12-22
Remove extra empty string from color list of remediation status

JAORMX:
This was causing the colors to be rendered incorrectly.

Closes: https://github.com/stacklok/minder/issues/2018


JAORMX:
While this fixes the issue, https://github.com/stacklok/minder/pull/2022 would prevent such issues from happening.

jhrozek (approved):

PR #2021, opened 2023-12-22
Add newline to `No config file present, using default values.`

JAORMX:
The newline was missing which was messing with the output.


jhrozek (approved):
Thanks for catching this

Issue #1858, opened 2023-12-07
`mindev` segfaults when testing a rule type using a profile with alerts on


JAORMX:
### Describe the issue

When running `mindev` we get the following error:

```
Profile valid according to the JSON schema!
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x111d4aa]

goroutine 1 [running]:
github.com/stacklok/minder/internal/engine/actions.shouldAlert(0x0, {0x0?, 0x0?}, {0x19dc1a0, 0x243ce00?}, {0x149abd5, 0x4})
	github.com/stacklok/minder/internal/engine/actions/actions.go:171 +0x4a
github.com/stacklok/minder/internal/engine/actions.(*RuleActionsEngine).DoActions(0xc000571220, {0x19e9128, 0x24a3740}, {0x19dc060, 0xc00028a700}, {0x19ee0e8?, 0xc0003402d0?})
	github.com/stacklok/minder/internal/engine/actions/actions.go:119 +0x4de
github.com/stacklok/minder/internal/engine.(*RuleTypeEngine).Actions(...)
	github.com/stacklok/minder/internal/engine/rule_types.go:282
github.com/stacklok/minder/cmd/dev/app/rule_type.runEvaluationForRules(0xc000623500, 0xc000949c38, {0xc00013e4f0, 0x1, 0x0?})
	github.com/stacklok/minder/cmd/dev/app/rule_type/rttst.go:172 +0x225
github.com/stacklok/minder/cmd/dev/app/rule_type.testCmdRun(0xc00024eb00?, {0x1499a21?, 0x4?, 0x1499a25?})
	github.com/stacklok/minder/cmd/dev/app/rule_type/rttst.go:142 +0x72b
github.com/spf13/cobra.(*Command).execute(0xc0003d6300, {0xc0001117a0, 0x6, 0x6})
	github.com/spf13/cobra@v1.8.0/command.go:983 +0xabc
github.com/spf13/cobra.(*Command).ExecuteC(0x2458800)
	github.com/spf13/cobra@v1.8.0/command.go:1115 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.8.0/command.go:1039
github.com/stacklok/minder/cmd/dev/app.Execute()
	github.com/stacklok/minder/cmd/dev/app/root.go:43 +0x1a
main.main()
[...]

PR #2023, opened 2023-12-22
Don't execute actions when running `mindev ruletype test`

JAORMX:
This overwrites the remediation and alert to be `off` since we don't want to
execute actions when testing. This is not _yet_ supported and causes a segfault.

So, let's turn it off for now. If the need comes to test the actions themselves
with `mindev` we can work on turning it on later. But, for now, no segfault
is better.

Closes: https://github.com/stacklok/minder/issues/1858

### DIFF
--- a/cmd/dev/app/rule_type/rttst.go
+++ b/cmd/dev/app/rule_type/rttst.go
@@ -106,6 +106,11 @@ func testCmdRun(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("error reading fragment from file: %w", err)
 	}
 
+	// Disable actions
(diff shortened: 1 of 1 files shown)


jhrozek:
Do we need to disable remediation too? I use that for testing ...

JAORMX:
@jhrozek does it work? I thought it segfaulted too.

jhrozek:
Yes I use that all the time

JAORMX:
So it's only alerts. Got it.

JAORMX:
@jhrozek fixed.

jhrozek (approved):

PR #2024, opened 2023-12-22
Make webhook tests more resilient

JAORMX:
The webhook tests start an http server and spawn it in the background. This leads
to sporadic test failures, as the caller code might sometimes run before the
server is up.

This adds a retry with backoff in those tests, so if this happens, the caller will
simply retry.

If the retry fails, this will actually be a real issue with the code.

### DIFF
--- a/internal/controlplane/handlers_githubwebhooks_test.go
+++ b/internal/controlplane/handlers_githubwebhooks_test.go
@@ -31,6 +31,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jhrozek:
The test changes look good what are the helm chart changes about?

JAORMX:
@jhrozek good question, I've been running a bunch of stuff and don't know how those got in there.

JAORMX:
@jhrozek fixed.

teodor-yanev (approved):

Issue #1710, opened 2023-11-21
test_handleEntityEvent is flaky


jhrozek:
I've seen this a handful of times already so it's probably worth an issue
```
tExecutor_handleEntityEvent (0s)
      executor_test.go:272: Running eventer
      executor_test.go:310: waiting for eventer to start
      executor_test.go:314: waiting for flush
      executor_test.go:306: Running entity event handler
  panic: test timed out after 10m0s
  running tests:
  	TestExecutor_handleEntityEvent (10m0s)
```

PR #2026, opened 2023-12-22
entity event handler test fixes

JAORMX:
The idea is that the eventer should be ready to issue events before we
try to parse and process an event.

This also adds another log line to know if the test is blocked because of the
last `Wait` call.

Moreso, this adds logging to the mock event handler plugin.

Finally, this makes sure handler registration happens before the eventer runs, so we don't loose events.

Closes https://github.com/stacklok/minder/issues/1710

### DIFF
--- a/internal/util/testqueue/passthroughqueue.go
+++ b/internal/util/testqueue/passthroughqueue.go
@@ -16,18 +16,26 @@
 // Package testqueue contains queue utilities for testing
 package testqueue
 
-import "github.com/ThreeDotsLabs/watermill/message"
+import (
+	"testing"
+
+	"github.com/ThreeDotsLabs/watermill/message"
+)
 
 // PassthroughQueue is a queue that passes messages through.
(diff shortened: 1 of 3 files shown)


internal/util/testqueue/passthroughqueue.go
```diff
@@ -38,6 +46,7 @@ func (q *PassthroughQueue) GetQueue() <-chan *message.Message {
 
 // Pass passes a message through the queue
 func (q *PassthroughQueue) Pass(msg *message.Message) error {
+	q.t.Logf("Passing message through queue: %s", msg.UUID)
```
rdimitrov, line 49: Would it make sense to log the event topic too?

JAORMX: That's not available through the message structure. Maybe if we start having several test queues in our tests, but that's not the case right now.

rdimitrov (approved):

PR #2033, opened 2023-12-22
Add flag for identity server base URL

### LINKED ISSUE #1665: Add config flag for identity issuer URL
Similar to how we have `--db-host` and `--metric-host`, we should add a flag for the configuration option of `identity:server:issuer_url`.

Then, we should set this value in the `docker-compose.yaml` when starting minder.

eleftherias:
Fix #1665


JAORMX (approved):

JAORMX (commented):
Uhm... don't we only need this for the auth commands?

eleftherias:
@JAORMX  This is on the server side. It's used mainly for fetching the JWKS for validating the incoming token. It's also used for user deletion.

JAORMX:
Oops, got confused. Shipit :ship:

PR #2043, opened 2023-12-31
Create a config structure for the minder CLI

### LINKED ISSUE #1920: Create a dedicated config structure for the minder CLI
Motivated by the discussions in https://github.com/stacklok/minder/pull/1918

Vyom-Yadav:
Fixes #1920 

---

Major Changes:

- Migrated options in the config file to a dedicated struct
- Eliminated direct access to viper store for config file options
- Added unit tests for testing various CLI configurations
- Refactored server config code to be present in `config/server`

### DIFF
--- a/cmd/cli/app/auth/auth_delete.go
+++ b/cmd/cli/app/auth/auth_delete.go
@@ -24,6 +24,7 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/stacklok/minder/internal/auth"
+	clientconfig "github.com/stacklok/minder/internal/config/client"
 	"github.com/stacklok/minder/internal/util"
 	"github.com/stacklok/minder/internal/util/cli"
 	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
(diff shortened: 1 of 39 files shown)


evankanderson (commented):
I know that historically we've used the same config for both the minder CLI and the server, but if we're creating a `client_config.go`, I'd be tempted to put it in a separate module, e.g. `config/client`.  That would also avoid the collision on the `IdentityConfig` name.

Other than that, this looks great, thank you!

Vyom-Yadav:
> I know that historically we've used the same config for both the minder CLI and the server, but if we're creating a client_config.go, I'd be tempted to put it in a separate module, e.g. config/client. That would also avoid the collision on the IdentityConfig name.

Nice suggestion. I refactored to create two packages, `config/client` and `config/server`.

Also, do we want to include the following in cli-config file?

- provider
- output
- project

eleftherias (approved):
Thanks @Vyom-Yadav! This is great!

> Also, do we want to include the following in cli-config file?
> 
> * provider
> * output
> * project

Yes, it makes sense to me to include those as well, but that can be a separate PR. I'll go ahead and merge this one.

PR #2045, opened 2024-01-01
`repo register --name` should skip already registered repositories

### LINKED ISSUE #1826: `repo register --repo` should skip an already registered repository
### Describe the issue

When a repository is already registered and a user tries to run `minder repo register --repo owner/name --provider=github` again, the command says that the repo can't be found and presents the user with a list of not-yet-registered repositories to choose from. Instead, it should just exit gracefully.

### To Reproduce

1. enroll a provider
2. register a repository
3. try to register the same repository again using the `--repo` switch

_No response_

Vyom-Yadav:
Fixes #1826 

---

Major Changes:

-  `repo register --name` now skips registered repos, printing to the console that repo is already registered
- Added UT for  `getUnregisteredInputRepos` (new function)

### DIFF
--- a/cmd/cli/app/repo/repo_register.go
+++ b/cmd/cli/app/repo/repo_register.go
@@ -18,14 +18,13 @@ package repo
 import (
 	"context"
 	"fmt"
-	"os"
-	"slices"
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"google.golang.org/grpc"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/stacklok/minder/cmd/cli/app"
 	"github.com/stacklok/minder/internal/util/cli"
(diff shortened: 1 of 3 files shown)


cmd/cli/app/repo/repo_register.go
```diff
@@ -67,6 +66,25 @@ func RegisterCmd(ctx context.Context, cmd *cobra.Command, conn *grpc.ClientConn)
...
+		alreadyRegisteredReposSet[cli.GetRepositoryName(repo.Owner, repo.Name)] = true
+	}
+
+	inputRepositories := strings.Split(repoList, ",")
+	unregisteredInputRepos, cliOutput := getUnregisteredInputRepos(inputRepositories, alreadyRegisteredReposSet)
+
+	if len(cliOutput) > 0 {
+		for _, output := range cliOutput {
+			cmd.Println(output)
+		}
```
evankanderson, line 80: If we're printing these only as warnings (and we might want to call `cliOutput` "`warnings`"), then you don't need the `if len()`, check, as you'll be iterating over a zero-length slice.

cmd/cli/app/repo/repo_register.go
```diff
@@ -67,6 +66,25 @@ func RegisterCmd(ctx context.Context, cmd *cobra.Command, conn *grpc.ClientConn)
 		return cli.MessageAndError("Error getting list of repos", err)
 	}
 
+	alreadyRegisteredReposSet := make(map[string]bool)
```
evankanderson, line 69: We're already using [`k8s.io/apimachinery/pkg/util/sets.Set`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set) elsewhere in the code.  With that, this would be:

With that, you can use `alreadyRegisteredReposSet.HasAny(inputRepositories)` or `.Intersection(inputRepositories)` to find the overlapping items.

In fact, you may be able to use `Results` as a set item, though I suspect there's something in the proto definition to prevent the copying needed...

cmd/cli/app/repo/repo_register.go
```diff
@@ -171,10 +197,9 @@ func getSelectedRepositories(repoList []*minderv1.UpstreamRepositoryRef, flagRep
 	// Create a slice of strings to hold the selected repos
 	var allSelectedRepos []string
 
-	// If the --repo flag is set, use it to select repos
-	if flagRepos != "" {
-		repos := strings.Split(flagRepos, ",")
-		for _, repo := range repos {
+	// If the --name flag is set, use it to select repos
+	if len(inputRepositories) != 0 {
```
evankanderson, line 201: Again, we probably don't need `len()` here, just use the loop iteration.

Vyom-Yadav: Right 🤦🏽‍♂️ . Done.

cmd/cli/app/repo/repo_register.go
Vyom-Yadav: Done.

cmd/cli/app/repo/repo_register.go
Vyom-Yadav: I changed to [`k8s.io/apimachinery/pkg/util/sets.Set`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set).

> With that, you can use alreadyRegisteredReposSet.HasAny(inputRepositories) or .Intersection(inputRepositories) to find the overlapping items.

Functional programming style would be more precise, but here, the sets are quite different. To do set differences and stuff like that, we would have to convert set types to be the same and create redundant collections. This is fine IMO.

> though I suspect there's something in the proto definition to prevent the copying needed

No, Results is comparable, so it would work as a key, but it's not worth storing everything, we would still need a string set to compare with input.

Vyom-Yadav:
@evankanderson Please trigger the workflows.

cmd/cli/app/repo/repo_register.go
```diff
@@ -67,6 +66,23 @@ func RegisterCmd(ctx context.Context, cmd *cobra.Command, conn *grpc.ClientConn)
...
+	}
+
+	allInputReposRegistered := inputRepositories[0] != "" && len(unregisteredInputRepos) == 0
+	if allInputReposRegistered {
+		return nil
+	}
+
 	// Get a list of remote repos
 	remoteListResp, err := client.ListRemoteRepositoriesFromProvider(ctx, &minderv1.ListRemoteRepositoriesFromProviderRequest{
 		Context: &minderv1.Context{Provider: &provider, Project: &project},
```
evankanderson, line 88: Reading all this again, it feels like this would be better to split behavior based on `repoList`:

```golang
unregisteredRepos []string
if repoList != "" {
  // split repoList, remove already-registered repos
  // add remainder to unregisteredRepos
  if len(unregisteredRepos) == 0 {
    cmd.Println("All repos already registered, nothing to be done")
    return nil
  }
} else {
  // call ListRemoteRepositoriesFromProvider, select from menu, etc
}

reposToAdd []*minderv1.UpstreamRepositoryRef
for _, selectedRepo := range unregisteredRepos {
  // Parse repo name, store in reposToAdd
}
```

evankanderson: BTW - I think the factoring and testing you're doing here is great, and I'm okay leaving this PR as-is for the testing improvement.  But I'm also looking at the overall code flow, and it does feel like there's a bunch of weird edge cases here that it would be great if you wanted to clean up.

evankanderson (dismissed):

cmd/cli/app/repo/repo_register.go
Vyom-Yadav: > Reading all this again, it feels like this would be better to split behavior based on repoList:

I modified it to use `repoList`. Indeed, accessing array indices for boolean conditions is not the best solution. I have simplified the program flow by splitting it into functions. 

> bunch of weird edge cases here that it would be great if you wanted to clean up.

I refactored the code for visual simplification. I feel all cases are covered except the survey prompt, invoking RPCs, and a few `len` checks. Please let me know if you'd like me to add some mocking or any other test.

evankanderson (approved):
Beautiful, thank you!

PR #2046, opened 2024-01-01
Add instructions for updating webhook configuration for minder server

Vyom-Yadav:
This PR addresses the following documentation aspects:

- The docs don't mention updating the webhook configuration for running the minder server (AFAIK). 
- There was confusion about whether a single OAuth app or two of them are required. Configuring the repository provider for Minder should be present on the same page.
- The tutorial on the particular page mentions using `docker compose` for both Keycloak and `Postgres`; the same for `Minder` was added.

### DIFF
--- a/docs/docs/run_minder_server/run_the_server.md
+++ b/docs/docs/run_minder_server/run_the_server.md
@@ -176,10 +176,66 @@ auth:
    token_key: "./.ssh/token_key_passphrase"
 ```
 
+## Configure the Repository Provider
+
+At this point, you should have the following:
+
+- A running PostgreSQL database, with the `minder` database created
+- A running Keycloak instance
+- A GitHub OAuth application configured for social login using Keycloak
+
+**Prior to running the application**, you need to configure your repository provider. Currently, Minder only supports GitHub.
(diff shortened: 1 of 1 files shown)


docs/docs/run_minder_server/run_the_server.md
```diff
@@ -176,10 +176,66 @@ auth:
...
+
+```yaml
+database:
+  dbhost: "postgres" # Changed from localhost to postgres
+  dbport: 5432
+  dbuser: postgres
+  dbpass: postgres
+  dbname: minder
+  sslmode: disable
+```
```
evankanderson, line 233: IIRC, some of these like hostname can be overriden from the command-line, but I think this is fine.

evankanderson (approved):

docs/docs/run_minder_server/run_the_server.md
Vyom-Yadav: `minder-server` binary has the flag:
```
-H, --db-host string                     Database host (default "localhost")
```
But that would be overridden in the docker compose file. Alternatively, we could pass the flag earlier for `localhost` while doing migrate up, but that would have to be added to make too. So, this is fine IMO.

PR #2048, opened 2024-01-02
Refactor eventer middleware

teodor-yanev:
We have two places where middlewares are added, one of them being a duplicate. This commit fixes that.


JAORMX (approved):
I'm very surporised we had missed this. Good find! thanks!

Issue #1438, opened 2023-11-03
vulncheck's `Eval()` has too high cyclomatic complexity


jhrozek:
We pushed a hotfix in PR #1437 to silence PR spam, but the cyclomatic complexity of the function is now too high. We should refactor the Eval function instead of silencing the lint warning.

PR #2050, opened 2024-01-02
Refactor vulncheck's Eval() to address high cyclomatic complexity

gregfurman:
Fixes https://github.com/stacklok/minder/issues/1438

### DIFF
--- a/internal/engine/eval/vulncheck/vulncheck.go
+++ b/internal/engine/eval/vulncheck/vulncheck.go
@@ -57,30 +57,42 @@ func NewVulncheckEvaluator(_ *pb.RuleType_Definition_Eval_Vulncheck, pbuild *pro
 }
 
 // Eval implements the Evaluator interface.
-//
-//nolint:gocyclo
 func (e *Evaluator) Eval(ctx context.Context, pol map[string]any, res *engif.Result) error {
+	vulnerablePackages, err := e.getVulnerableDependencies(ctx, pol, res)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


internal/engine/eval/vulncheck/vulncheck.go
```diff
@@ -57,30 +57,42 @@ func NewVulncheckEvaluator(_ *pb.RuleType_Definition_Eval_Vulncheck, pbuild *pro
 }
 
 // Eval implements the Evaluator interface.
-//
-//nolint:gocyclo
 func (e *Evaluator) Eval(ctx context.Context, pol map[string]any, res *engif.Result) error {
```
evankanderson, line 60: Do we need to break out getVulnerableDependencies and Eval?  I think moving 93-133 into a separate method definitely makes sense, thank you!

evankanderson (approved):
I'm okay with the entire refactor, but it feels like splitting Eval and getVulnerableDependencies currently adds bookkeeping for `vulnerablePackages` which would be simpler otherwise.

internal/engine/eval/vulncheck/vulncheck.go
gregfurman: I suppose we don't actually need to break them up. Thought it would be more readable since it allows for one to easily see what `Eval` is doing at a higher level (i.e getting vulnerable dependencies and returning an err if any are found). 

Happy to re-consolidate `getVulnerableDependencies` back into the `Eval` though.

evankanderson: I'll buy that.

Issue #1879, opened 2023-12-08
Implement TelemetryStore logging as Watermill Middleware


evankanderson:
Implement [the Watermill `Middleware` interface](https://watermill.io/docs/middlewares/#introduction) in [`LoggingInterceptor`](https://github.com/stacklok/minder/blob/main/internal/logger/logging_interceptor.go#L84) so that we can collect telemetry (via structured JSON logs) for rule actions executed by Minder.  (See also #1878)

Logs should include the following:

* Details on the resource being operated on (repo / artifact / pull request)
* Project context, if available from the originating message

PR #2059, opened 2024-01-03
Add TelemetryStore interface

evankanderson:
Related-To: #1879

I documented a business record / telemetry store interface in [the design document](https://docs.google.com/document/d/132l2YCasne_p-ILyXG923v0qeLnELVW0qP1Gflza4Dw/edit#heading=h.tfrp6ktgz14f), but I don't seem to have created a separate issue for defining this interface.

### DIFF
--- a/internal/logger/logging_interceptor.go
+++ b/internal/logger/logging_interceptor.go
@@ -87,9 +87,10 @@ func Interceptor( /*logLevel string, logFormat string, logFile string*/ ) grpc.U
 		if info.FullMethod == "/minder.v1.HealthService/CheckHealth" {
 			return handler(ctx, req)
 		}
+		ts := TelemetryStore{}
 		// Attach the resource to all logging events in the context
 		logger := zerolog.Ctx(ctx).With().Dict("Resource", resource(info.FullMethod)).Logger()
-		ctx = logger.WithContext(ctx)
+		ctx = ts.WithTelemetry(logger.WithContext(ctx))
 		now := time.Now()
 
 		resp, err := handler(ctx, req)
@@ -101,11 +102,12 @@ func Interceptor( /*logLevel string, logFormat string, logFile string*/ ) grpc.U
 			logMsg = logger.Error()
 
 			attrs = attrs.Err(err)
-			// Log the body of errors (TODO: do need to scrub requests?)
+			// Log the body of errors (TODO: do we need to scrub requests?)
 			if jsonText, err := json.Marshal(req); err != nil {
 				logMsg = logMsg.RawJSON("Body", jsonText)
 			}
 		}
+		ts.Record(logMsg)
 
 		// Note: Zerolog makes it hard to add attributes in multiple calls.
 		logMsg.Dict("Attributes", attrs).Send()
--- a/internal/logger/telemetry_store.go
+++ b/internal/logger/telemetry_store.go
@@ -1,0 +1,75 @@
+// Copyright 2024 Stacklok, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logger
+
+import (
+	"context"
+
+	"github.com/rs/zerolog"
+)
+
+type key int
+
+const (
+	// telemetryContextKey is a key used to store the current telemetry record's context
+	telemetryContextKey key = iota
+)
+
+// TelemetryStore is a struct that can be used to store telemetry data in the context.
+type TelemetryStore struct {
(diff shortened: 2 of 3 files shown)


evankanderson:
I'd like a couple eyes/opinions on this.  I'm particularly not thrilled about relaxing from `json.Marshaller` to `any` in the interface type, but trying to use it with `json.Marshaller` was a pain.

I'm wondering if I should change the interface to support checking the marshalling of the data early, e.g. `BusinessRecord(ctx, key, data)`, rather than supporting assignment as described in the document (a more complex type system might be able to express what I want... i.e. `json.Marshaller` or `string` or `int`).

JAORMX:
> I'd like a couple eyes/opinions on this. I'm particularly not thrilled about relaxing from `json.Marshaller` to `any` in the interface type, but trying to use it with `json.Marshaller` was a pain.
> 
> I'm wondering if I should change the interface to support checking the marshalling of the data early, e.g. `BusinessRecord(ctx, key, data)`, rather than supporting assignment as described in the document (a more complex type system might be able to express what I want... i.e. `json.Marshaller` or `string` or `int`).

I agree with your concern. On one hand, I like the simplicity this brings. On the other hand, this also feels quite easy to misuse. Whether having a dedicated structure that's typed would allow us to have greater safety. I guess having a typed interface as opposed to `map[string]any` would be a little harder to do without limiting us a lot?

What about using this simple assignment-based interface for now and evolving it with time?

eleftherias:
> I'd like a couple eyes/opinions on this. I'm particularly not thrilled about relaxing from `json.Marshaller` to `any` in the interface type, but trying to use it with `json.Marshaller` was a pain.
> 
> I'm wondering if I should change the interface to support checking the marshalling of the data early, e.g. `BusinessRecord(ctx, key, data)`, rather than supporting assignment as described in the document (a more complex type system might be able to express what I want... i.e. `json.Marshaller` or `string` or `int`).

I'm not too worried about using `map[string]any` since it's only used internally and I have a hard time imagining us wanting to record something that can't be serialized, since most types in Go are JSON serializable.

Still, I think it's better to guard against it on the Minder side, as opposed to when we attempt to read the record. I like your proposal of using `BusinessRecord(ctx, key, data)`. In my opinion it's also cleaner than accessing the map directly, because whoever adds a new record doesn't need to know the internal implementation uses a map. This also lets us add a doc comment explaining that `data` must be serializable.

evankanderson:
> > I'd like a couple eyes/opinions on this. I'm particularly not thrilled about relaxing from `json.Marshaller` to `any` in the interface type, but trying to use it with `json.Marshaller` was a pain.
> > I'm wondering if I should change the interface to support checking the marshalling of the data early, e.g. `BusinessRecord(ctx, key, data)`, rather than supporting assignment as described in the document (a more complex type system might be able to express what I want... i.e. `json.Marshaller` or `string` or `int`).
> 
> I'm not too worried about using `map[string]any` since it's only used internally and I have a hard time imagining us wanting to record something that can't be serialized, since most types in Go are JSON serializable.
> 
> Still, I think it's better to guard against it on the Minder side, as opposed to when we attempt to read the record. I like your proposal of using `BusinessRecord(ctx, key, data)`. In my opinion it's also cleaner than accessing the map directly, because whoever adds a new record doesn't need to know the internal implementation uses a map. This also lets us add a doc comment explaining that `data` must be serializable.

I added a `BusinessRecord2` and `BusinessRecord3` to illustrate this second option, since when I tried to include `json.Marshaler` in the type for `BusinessRecord2`, I got the following error:

> cannot use encoding/json.Marshaler in union (encoding/json.Marshaler contains methods)

evankanderson:
(I'd rename `BusinessRecord2` to `BusinessRecord` and come up with another name for the `2` version if we prefer it...)

eleftherias:
> I added a BusinessRecord2 and BusinessRecord3 to illustrate this second option

Would we want `Recordable` to also include slices and maps?

When you initially mentioned checking the marshalling of the data early, I imagined that `BusinessRecord(ctx, key, data)` would attempt to Marshall `data` and log an error if data cannot be marshalled, instead of enforcing the parameter type.

evankanderson:
Rationale for preferring an explicit set of fields in `TelemetryStore`: this will make it easier for us to maintain some governance and schema for the data that we collect (and screen it for accidental inclusion of unwanted personal information), rather than needing to track it across the entire codebase.

internal/logger/logging_interceptor.go
```diff
@@ -101,11 +102,12 @@ func Interceptor( /*logLevel string, logFormat string, logFile string*/ ) grpc.U
 			logMsg = logger.Error()
 
 			attrs = attrs.Err(err)
-			// Log the body of errors (TODO: do need to scrub requests?)
+			// Log the body of errors (TODO: do we need to scrub requests?)
```
jhrozek, line 105: I know this code wasn't changed as part of your pR, but isn't the error condition below reversed? It seems to be logging the body only if marshalling the body failed?

evankanderson: I wonder if we should actually be logging this... what do you think of removing this code?

evankanderson: We're going to put this logging behind a flag, to avoid leaking PII into logs without consideration.  (And fix the condition)

jhrozek (approved):
This works for me, feel free to ping Ria or Ozz async if you want their approval, too since they contributed to the PR review earlier.

PR #2061, opened 2024-01-03
Use project ID in context and as CLI flag

### LINKED ISSUE #2051: CLI --project inconsistently refers to either the project ID or project name
Note: this only occurs in the current version of minder on main. None of the released versions have this bug.

In some cases, the `--project` flag expects a project ID. For example:
```
minder provider enroll --project ID
```

In some other cases, the `--project` flag expects a project name. For example:
```
minder profile list --project name
```
Although, in the `profile list` command, the documentation states that it expects a project ID, it will fail if the ID is provided.
```
  -j, --project string           ID of the project
```

At minimum we need to correct the documentation.
Ideally, we should use `--project` consistently so that it only refers to the name or ID.

eleftherias:
Fix #2051

### DIFF
--- a/internal/controlplane/handlers_profile.go
+++ b/internal/controlplane/handlers_profile.go
@@ -81,12 +81,13 @@ func (s *Server) ensureDefaultProjectForContext(ctx context.Context, inout *mind
 		return status.Errorf(codes.InvalidArgument, "cannot infer project id")
 	}
 
-	g, err := s.store.GetProjectByID(ctx, gid)
+	_, err = s.store.GetProjectByID(ctx, gid)
 	if err != nil {
 		return status.Errorf(codes.InvalidArgument, "cannot infer project id")
 	}
 
-	inout.Project = &g.Name
+	project := gid.String()
+	inout.Project = &project
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


JAORMX (commented):
So, the intent is to use project IDs everywhere. Right? I guess we'd also need to change the docs. Right?

eleftherias:
> So, the intent is to use project IDs everywhere. Right? I guess we'd also need to change the docs. Right?

Yes, both are right. The CLI docs were already saying that we use project ID everywhere. The swagger and proto docs still need to updated, but I assumed the docs GitHub action would create a PR for that after this one is merged.

JAORMX (approved):

evankanderson:
Do we want to use ID or name?  Name seems like it would be more friendly for humans.

eleftherias:
> Do we want to use ID or name? Name seems like it would be more friendly for humans.

Sharing what was said in a separate discussion. ID is used instead of name for 2 reasons:
1) For backwards compatibility, since we were already using project ID in commands like `minder provider enroll`
2) Our project naming is likely to evolve, we currently default the project name to the logged in user's username, but we may use a different approach in the future

PR #2063, opened 2024-01-03
Split token validation and authorization middleware

### LINKED ISSUE #2049: Separate authorization and authentication middleware
Currently in `AuthUnaryInterceptor` we check that a bearer token is present and valid, and also populate the users permissions and perform some authorization logic.

We should split this up into 2 interceptors instead.

eleftherias:
Also remove unused superadmin role lookup.

Fix #2049

### DIFF
--- a/internal/auth/jwtauth.go
+++ b/internal/auth/jwtauth.go
@@ -124,6 +124,7 @@ func NewJwtValidator(ctx context.Context, jwksUrl string) (JwtValidator, error) 
 }
 
 var tokenContextKey struct{}
+var userSubjectContextKey struct{}
 
 // GetPermissionsFromContext returns the claims from the context, or an empty default
 func GetPermissionsFromContext(ctx context.Context) UserPermissions {
@@ -137,6 +138,20 @@ func GetPermissionsFromContext(ctx context.Context) UserPermissions {
 // WithPermissionsContext stores the specified UserClaim in the context.
 func WithPermissionsContext(ctx context.Context, claims UserPermissions) context.Context {
 	return context.WithValue(ctx, tokenContextKey, claims)
+}
+
+// GetUserSubjectFromContext returns the user subject from the context, or nil
+func GetUserSubjectFromContext(ctx context.Context) string {
+	subject, ok := ctx.Value(userSubjectContextKey).(string)
+	if !ok {
+		return ""
+	}
+	return subject
+}
+
+// WithUserSubjectContext stores the specified user subject in the context.
+func WithUserSubjectContext(ctx context.Context, subject string) context.Context {
+	return context.WithValue(ctx, userSubjectContextKey, subject)
 }
 
 // GetDefaultProject returns the default project id for the user
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -16,20 +16,12 @@ package controlplane
 
 import (
 	"context"
-	"fmt"
-	"strings"
 
 	"github.com/google/uuid"
-	gauth "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/auth"
-	"github.com/lestrrat-go/jwx/v2/jwt/openid"
-	"github.com/rs/zerolog"
 	"golang.org/x/exp/slices"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	"google.golang.org/protobuf/proto"
-	"google.golang.org/protobuf/reflect/protoreflect"
-	"google.golang.org/protobuf/reflect/protoregistry"
 
 	"github.com/stacklok/minder/internal/auth"
 	"github.com/stacklok/minder/internal/db"
@@ -38,10 +30,6 @@ import (
 )
 
 type rpcOptionsKey struct{}
-
-func withRpcOptions(ctx context.Context, opts *minder.RpcOptions) context.Context {
-	return context.WithValue(ctx, rpcOptionsKey{}, opts)
-}
 
 func getRpcOptions(ctx context.Context) *minder.RpcOptions {
 	// nil value default is okay here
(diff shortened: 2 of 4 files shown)


internal/auth/jwtauth.go
```diff
@@ -139,6 +140,20 @@ func WithPermissionsContext(ctx context.Context, claims UserPermissions) context
 	return context.WithValue(ctx, tokenContextKey, claims)
 }
 
+// GetUserSubjectFromContext returns the user subject from the context, or nil
+func GetUserSubjectFromContext(ctx context.Context) *string {
+	subject, ok := ctx.Value(userSubjectContextKey).(string)
```
evankanderson, line 145: If we enforce that `subject` is always non-empty, we could just return an empty string, which might simplify other code.

internal/controlplane/handlers_authz.go
```diff
@@ -54,27 +47,12 @@ func isSuperadmin(claims auth.UserPermissions) bool {
...
-				}
-			}
-		}
-	}
-	return false
-}
-
 // lookupUserPermissions returns the user permissions from the database for the given user
-func lookupUserPermissions(ctx context.Context, store db.Store, tok openid.Token) (auth.UserPermissions, error) {
+func lookupUserPermissions(ctx context.Context, store db.Store, subject string) (auth.UserPermissions, error) {
```
evankanderson, line 51: If we have `GetUserSubjectFromContext`, do we need to pass in subject?

internal/controlplane/handlers_token.go
```diff
@@ -0,0 +1,87 @@
...
+func TokenValidationInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler) (any, error) {
+
+	opts, err := optionsForMethod(info)
+	if err != nil {
+		// Fail closed safely, rather than log and proceed.
+		return nil, status.Errorf(codes.Internal, "Error getting options for method: %v", err)
+	}
+
+	ctx = withRpcOptions(ctx, opts)
```
evankanderson, line 46: Move `withRpcOptions` into this file, too?

internal/controlplane/handlers_authz.go
```diff
@@ -148,62 +125,21 @@ func AuthorizedOnProject(ctx context.Context, projectID uuid.UUID) error {
...
-	}
-
 	server := info.Server.(*Server)
 
-	parsedToken, err := server.vldtr.ParseAndValidate(token)
-	if err != nil {
-		return nil, status.Errorf(codes.Unauthenticated, "invalid auth token: %v", err)
+	subject := auth.GetUserSubjectFromContext(ctx)
+	if subject == nil {
+		return nil, status.Errorf(codes.Internal, "Error getting user subject")
```
evankanderson, line 136: Won't this always fail for methods marked as `anonymous`, because they don't store a subject?

internal/controlplane/handlers_authz.go
```diff
@@ -114,7 +92,6 @@ func lookupUserPermissions(ctx context.Context, store db.Store, tok openid.Token
 		Roles:          roles,
 		ProjectIds:     projects,
 		OrganizationId: userInfo.OrganizationID,
-		IsStaff:        containsSuperadminRole(tok),
```
evankanderson, line 117: \o/

internal/controlplane/handlers_authz.go
```diff
@@ -54,27 +47,12 @@ func isSuperadmin(claims auth.UserPermissions) bool {
 	return claims.IsStaff
```
evankanderson, line 47: Since we no longer set `.IsStaff`, should we remove this method as well?

internal/controlplane/handlers_authz.go
```diff
@@ -148,62 +125,21 @@ func AuthorizedOnProject(ctx context.Context, projectID uuid.UUID) error {
...
 	}
 
 	// get user authorities from the database
 	// ignore any error because the user may not exist yet
-	authorities, _ := lookupUserPermissions(ctx, server.store, parsedToken)
-
-	if opts.GetRootAdminOnly() && !isSuperadmin(authorities) {
-		return nil, status.Errorf(codes.PermissionDenied, "user not authorized")
-	}
+	authorities, _ := lookupUserPermissions(ctx, server.store, *subject)
```
evankanderson, line 141: It looks like we're ignoring the `error` return value here?  Intentional?

eleftherias: It is intentional, since this interceptor is called on every request, including the one that creates a user. So if a user doesn't exist yet, then they don't have any permissions. Maybe it's better to remove the error from response?

internal/controlplane/handlers_authz.go
eleftherias: We are still using it in some tests to bypass authorization. I created a new issue to remove the superadmin role, since it needs some more refactoring and a better testing appoach. https://github.com/stacklok/minder/issues/2070

internal/controlplane/handlers_authz.go
eleftherias: We don't, I refactored

internal/controlplane/handlers_authz.go
eleftherias: I moved this call to `lookupUserPermissions`, so now anonymous methods won't error, but will simply have empty permissions.

JAORMX (dismissed):

internal/controlplane/handlers_authz.go
evankanderson: There are three different `return err` cases in `lookupPermissions`, and I believe it's only called from this one location.  I think it would be clearer to remove the error return if we believe we've handled all the errors within the function and don't need to do anything at the interceptor level.

evankanderson: Though, I'm now seeing that we've always been ignoring that error, so I'm not going to hold this up on it.

eleftherias: I removed the error from the method response

evankanderson (approved):

JAORMX (approved):

PR #2072, opened 2024-01-04
No longer print usage once the minder command has started doing real work

evankanderson:
I noticed this while validating the most recent production service push

Before this change:

```
minder repo get -n evankanderson/bad-python                             
Usage:
  minder repo get [flags]

Flags:
  -h, --help            help for get
  -i, --id string       ID of the repo to query
  -n, --name string     Name of the repository (owner/name format)
  -o, --output string   Output format (one of json,yaml) (default "json")

Global Flags:
      --config string            Config file (default is $PWD/config.yaml)
      --grpc-host string         Server host (default "api.stacklok.com")
      --grpc-insecure            Allow establishing insecure connections
      --grpc-port int            Server port (default 443)
      --identity-client string   Identity server client ID (default "minder-cli")
      --identity-url string      Identity server issuer URL (default "https://auth.stacklok.com")
  -j, --project string           ID of the project
  -p, --provider string          Name of the provider, i.e. github (default "github")

Message: Error getting repo by name
Details: Not found
```

After:

```
Message: Error getting repo by name
Details: NotFound means some requested entity (e.g., file or directory) was
not found.
```

I'm not sure why the output is slightly different, but it's a bit closer to friendly now.  In particular, it doesn't make you wonder if you entered the command wrong when you get an error back from the server.


JAORMX (approved):
Nice!

JAORMX (changes requested):
Oh, compilation error

JAORMX (approved):

rdimitrov:
@evankanderson - should we instead disable this for all? For example: 
```bash
...
// RootCmd represents the base command when called without any subcommands
	RootCmd = &cobra.Command{
		SilenceUsage:  true,
		Use:   "minder",
...
```

rdimitrov:
> No, I think we want to print usage when there's a problem with the arguments -- I added these assignments after any argument checking and parsing.

Ah, that makes sense 👍

PR #2075, opened 2024-01-07
Handle null values in config file to prevent incorrect data retrieval from viper store

### LINKED ISSUE #2067: config library doesn't respect precedence if empty configuration is provided + command line flags
### Describe the issue

In the case where the configuration looks as follows:

```yaml
---
grpc_server:
identity:
```

And some command line arguments are provided, e.g. `--grpc-host=192.168.1.7 --grpc-port=1234 --identity-url=http://localhost:1654`

The command line client does not respect the precedence order. That is, it'll randomly pick either the default for the config or the command line flag.

This is not an issue if there is no configuration file or if the configuration file actually contains the relevant values.

### To Reproduce

_No response_

`main`

Vyom-Yadav:
Fixes #2067

---

For rationale see https://github.com/stacklok/minder/issues/2067#issuecomment-1877102081

Major Changes:

- Check config file contents before adding it to the viper store
- UT for `GetKeysWithNullValueFromYAML`

### DIFF
--- a/cmd/server/app/root.go
+++ b/cmd/server/app/root.go
@@ -19,18 +19,19 @@ package app
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
 	"github.com/stacklok/minder/internal/auth"
+	"github.com/stacklok/minder/internal/config"
 	serverconfig "github.com/stacklok/minder/internal/config/server"
 	"github.com/stacklok/minder/internal/util/cli"
 )
 
 var (
-	cfgFile string // config file (default is $PWD/server-config.yaml)
 	// RootCmd represents the base command when called without any subcommands
 	RootCmd = &cobra.Command{
 		Use:   "minder-server",
@@ -50,7 +51,7 @@ func Execute() {
 
 func init() {
 	cobra.OnInitialize(initConfig)
-	RootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $PWD/server-config.yaml)")
+	RootCmd.PersistentFlags().String("config", "", "config file (default is $PWD/server-config.yaml)")
 	if err := serverconfig.RegisterDatabaseFlags(viper.GetViper(), RootCmd.PersistentFlags()); err != nil {
 		log.Fatal().Err(err).Msg("Error registering database flags")
 	}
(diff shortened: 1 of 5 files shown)


cmd/server/app/root.go
```diff
@@ -75,7 +96,7 @@ func initConfig() {
 	viper.SetConfigType("yaml")
 	viper.AutomaticEnv()
 
-	if err := viper.ReadInConfig(); err != nil {
+	if err = viper.ReadInConfig(); err != nil {
```
Vyom-Yadav, line 99: This is not foolproof checking. Ideally, every time, we update the viper store to get values from the config file, using `ReadInConfig()` or any other method, we should first check the config file contents to prevent polluting the viper store.

Viper keeps track of the config file, a rare case, but the user can:

- Pass a correct config file
- Pollute it will null keys after the sanity check has been performed

If we do sort of a "hot-reload" of config file anywhere in the code by invoking appropriate viper methods, the store could end up having incorrect data. To fix it from our side, we should always perform a sanity check before reading the config file, directly or indirectly.

internal/config/utils.go
```diff
@@ -91,3 +94,83 @@ func doViperBind[V any](
...
+		cfgFilePath, err = filepath.Abs(defaultCfgPath)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	cleanCfgFilePath := filepath.Clean(cfgFilePath)
+	if info, err := os.Stat(cleanCfgFilePath); err == nil && info.IsDir() || err != nil && os.IsNotExist(err) {
+		return nil, nil
+	}
```
Vyom-Yadav, line 117: While mounting, if no config file is present on local, docker will create an empty mount path in the container, but that would be a directory, so we skip that. 

We ignore if no config file exists, the system will revert to the defaults.

evankanderson: (I'm approving without requiring a change, but this would be a great comment in the code itself, since it's not immediately obvious what these two cases are.)

evankanderson (approved):
Very nice (and very nice tests)!

cmd/server/app/root.go
Vyom-Yadav: We unmarshal the values from the viper store. Having a strongly typed struct won't prevent viper from auto updating or choosing the wrong values.

Like I mentioned in https://github.com/stacklok/minder/issues/2067#issuecomment-1877102081, if the key `http_server` has `null` value in yaml, it would be unmarshalled properly, but the viper store still can be corrupted. 

Having both keys `http_server` and `http_server.host` in viper config will give unpredictable results, we could resort to default even if we specified `http_server.host` by a flag. User passes the initial sanity checks and later modifies the yaml can cause problems.

> I think the race condition you're mentioning here is short

If by short you mean only exists in a short time window, then this isn't true for server. Functions invoked on server could read the config file again internally which can corrupt the viper store.

internal/config/utils.go
Vyom-Yadav: > (I'm approving without requiring a change, but this would be a great comment in the code itself, since it's not immediately obvious what these two cases are.)

I'll cover this in next PR.

PR #2082, opened 2024-01-08
Remove context from user CRUD requests

eleftherias:
The Context currently contains the provider and the project. A user is not tied to either of those, so we don't need to include the context in user related operations like fetching and deleting.

### DIFF
--- a/proto/minder/v1/minder.proto
+++ b/proto/minder/v1/minder.proto
@@ -732,7 +732,7 @@ message CreateUserResponse {
 }
 
 message DeleteUserRequest {
-    Context context = 1;
+    reserved 1; // deprecated context
 }
 
 message DeleteUserResponse {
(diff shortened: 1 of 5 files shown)


proto/minder/v1/minder.proto
```diff
@@ -732,7 +732,7 @@ message CreateUserResponse {
 }
 
 message DeleteUserRequest {
-    Context context = 1;
+    reserved 1; // deprecated context
```
eleftherias, line 735: Although we previously prefixed a field with `retired_` when we deprecated it, I think this approach may be cleaner (where we reserve the field ID to avoid reusing it)

evankanderson: Works for me!

evankanderson (approved):

teodor-yanev (approved):

PR #2083, opened 2024-01-08
Add the ability to enable/disable payload logging (default disabled)

evankanderson:
Followup from https://github.com/stacklok/minder/pull/2059#discussion_r1444409522

### DIFF
--- a/internal/logger/logger_setup.go
+++ b/internal/logger/logger_setup.go
@@ -27,8 +27,7 @@ import (
 
 // FromFlags configures logging and returns a logger with settings matching
 // the supplied cfg.  It also performs some global initialization, because
-// that's how zerolog works, and we've piggybacked that pattern for payload
-// management.
+// that's how zerolog works.
 func FromFlags(cfg config.LoggingConfig) zerolog.Logger {
 	zlevel := viperLogLevelToZerologLevel(cfg.Level)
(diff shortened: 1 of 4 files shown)


internal/logger/logger_setup.go
```diff
@@ -22,13 +22,13 @@ import (
...
 
-	serverconfig "github.com/stacklok/minder/internal/config/server"
+	config "github.com/stacklok/minder/internal/config/server"
 )
 
 // FromFlags configures logging and returns a logger with settings matching
 // the supplied cfg.  It also performs some global initialization, because
 // that's how zerolog works.
-func FromFlags(cfg serverconfig.LoggingConfig) zerolog.Logger {
+func FromFlags(cfg config.LoggingConfig) zerolog.Logger {
```
evankanderson, line 31: I renamed this import because there was only one logic `config` it could be here.  I was also (earlier) using a global.  I can take out this change if it annoys anyone.

jhrozek:
the logic and intent is good, just please fix the lint errors and I'll ack right away

evankanderson:
> the logic and intent is good, just please fix the lint errors and I'll ack right away

Oops, that'll teach me to do one more fix and walk away without trying to run things...

jhrozek (approved):

PR #2091, opened 2024-01-09
Increase size of latency buckets, add tests for exported Eventer metrics.

evankanderson:
I noticed that our latency buckets for event processing time only went up to around 10s due to the otel default bucket sizes.  On increasing the bucket sizes, I decided to add some tests for the metrics, because I want them to actually work.  :grin:

I am sad about adding the `time.Sleep(2ms)`, but I couldn't find a more elegant way to get `--count=200` to pass consistently -- without the sleep, it was around a 1-2% flake rate with missing telemetry.


jhrozek (approved):
nice work on the tests

teodor-yanev (approved):

PR #2093, opened 2024-01-09
Add OpenFGA deployment to docker-compose

JAORMX:
This adds a simple OpenFGA deployment to the docker-compose in order for us
to test and work with this.

It uses an in-memory datastore and no authentication, so this is not suitable
for production.


jhrozek (approved):
```
KO_DOCKER_REPO=ko.local services="postgres keycloak migrate openfga" make run-docker
```
this works.

Issue #2078, opened 2024-01-08
Populate associated project in middleware


eleftherias:
In order to decide if a user is authorized, we first need to know which project they want to operate on.
We should populate the project associated with the requested action in a middleware, so that it's available before we make the authorization decision.

PR #2096, opened 2024-01-09
Use provider and project from context

eleftherias:
- Remove or deprecate provider and project proto fields
- Add additional HTTP mapping for endpoints containing provider
- Always fetch provider and project from context

Deprecated provider field and old endpoints remain for backwards compatibility. I will create a followup issue to remove them.

Prerequisite for #2078


proto/minder/v1/minder.proto
```diff
@@ -523,33 +551,35 @@ message CheckHealthResponse {
...
     string redirect_uri = 5;
     Context context = 6;
+
+    reserved 2; // deprecated project_id
 }
 
 message StoreProviderTokenRequest {
-    string provider = 1;
-    string project_id = 2;
+    string provider = 1 [deprecated=true];
```
jhrozek, line 577: This might be my lack of grpc knowledge, but why are some fields marked as deprecated=true and some moved as reserved? If you have a link to a document I could read, I'll be glad to learn more about how to deprecate grpc fields properly

jhrozek (commented):
Looks good and cursory testing of new server with old client seems to work. Left one comment inline.

proto/minder/v1/minder.proto
eleftherias: The fields that we want to remove but are still used somewhere in the codebase for backwards compatibility reasons are marked with `deprecated=true`. In our case, they are used in the HTTP endpoints, for example `get: "/api/v1/artifacts/{provider}"`. I've added alternative endpoints that don't use the deprecated field, but kept the old ones for backwards compatibility. Marking a field with `deprecated=true` will give a warning when you attempt to use the field, but will not remove the field. [link to docs](https://protobuf.dev/programming-guides/proto3/#options)

The fields that are no longer used in the codebase have their field number marked as reserved to prevent new code from reusing the old field number in the future. [link to docs](https://protobuf.dev/programming-guides/proto3/#fieldreserved)

internal/controlplane/handlers_repositories.go
```diff
@@ -364,16 +364,16 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
-
 	provider, err := getProviderFromRequestOrDefault(ctx, s.store, in, projectID)
 	if err != nil {
 		return nil, providerError(fmt.Errorf("provider error: %w", err))
 	}
 
+	zerolog.Ctx(ctx).Debug().
+		Str("provider", provider.Name).
+		Str("projectID", projectID.String()).
+		Msgf("listing repositories for provider: %s", provider.Name)
```
rdimitrov, line 375: Perhaps we can skip the provider name in `Msgf` since we already log it as a parameter to the same log message.

rdimitrov (approved):

rdimitrov (approved):

PR #2101, opened 2024-01-10
Initial implementation of telemetry store watermill middleware

JAORMX:
This is a simple initial implementation of the middleware for logging
telemetry store data.

closes: #1879

### DIFF
--- a/cmd/server/app/serve.go
+++ b/cmd/server/app/serve.go
@@ -59,7 +59,8 @@ var serveCmd = &cobra.Command{
 		}
 
 		ctx = logger.FromFlags(cfg.LoggingConfig).WithContext(ctx)
-		zerolog.Ctx(ctx).Info().Msgf("Initializing logger in level: %s", cfg.LoggingConfig.Level)
+		l := zerolog.Ctx(ctx)
+		l.Info().Msgf("Initializing logger in level: %s", cfg.LoggingConfig.Level)
 
 		// Database configuration
 		dbConn, _, err := cfg.Database.GetDBConnection(ctx)
@@ -120,9 +121,13 @@ var serveCmd = &cobra.Command{
 
 		s.ConsumeEvents(aggr)
 
+		tsmdw := logger.NewTelemetryStoreWMMiddleware(l)
+
 		exec, err := engine.NewExecutor(ctx, store, &cfg.Auth, evt,
 			engine.WithProviderMetrics(providerMetrics),
-			engine.WithAggregatorMiddleware(aggr))
+			engine.WithMiddleware(aggr.AggregateMiddleware),
+			engine.WithMiddleware(tsmdw.TelemetryStoreMiddleware),
+		)
 		if err != nil {
 			return fmt.Errorf("unable to create executor: %w", err)
 		}
--- a/internal/logger/telemetry_store_watermill.go
+++ b/internal/logger/telemetry_store_watermill.go
@@ -1,0 +1,101 @@
+// Copyright 2024 Stacklok, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logger
+
+import (
+	"fmt"
+
+	"github.com/ThreeDotsLabs/watermill/message"
+	"github.com/rs/zerolog"
+
+	"github.com/stacklok/minder/internal/engine"
+	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+)
+
+// TelemetryStoreWMMiddleware is a Watermill middleware that
+// logs the relevant telemetry data.
+type TelemetryStoreWMMiddleware struct {
+	l *zerolog.Logger
+}
+
+// NewTelemetryStoreWMMiddleware returns a new TelemetryStoreWMMiddleware.
+func NewTelemetryStoreWMMiddleware(l *zerolog.Logger) *TelemetryStoreWMMiddleware {
+	return &TelemetryStoreWMMiddleware{l: l}
+}
+
+// TelemetryStoreMiddleware is a Watermill middleware that
+// logs the relevant telemetry data.
+func (m *TelemetryStoreWMMiddleware) TelemetryStoreMiddleware(h message.HandlerFunc) message.HandlerFunc {
+	return func(msg *message.Message) ([]*message.Message, error) {
+		inf, err := engine.ParseEntityEvent(msg)
+		if err != nil {
+			return nil, fmt.Errorf("error unmarshalling payload: %w", err)
+		}
+
+		typ := inf.Type
+		ent, err := getEntityID(inf)
+		if err != nil {
+			return nil, fmt.Errorf("error getting entity ID: %w", err)
+		}
+
+		ts := &TelemetryStore{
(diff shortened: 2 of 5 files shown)


JAORMX:
One thing to note about this implementation is that yet again it parses the entity from the Watermill message. I think at this point we can have a separate middleware to do this parsing once and set it in the message's Context. But I rather leave this optimization for a separate PR.

internal/logger/telemetry_store_watermill.go
```diff
@@ -0,0 +1,88 @@
...
+	case minderv1.Entity_ENTITY_ARTIFACTS:
+		ts.Resource = artID.UUID.String()
+	case minderv1.Entity_ENTITY_PULL_REQUESTS:
+		ts.Resource = prID.UUID.String()
+	}
+
+	ctx := ts.WithTelemetry(msg.Context())
+	msg.SetContext(ctx)
+
+	ts.Record(m.l.Info()).Str("event", msg.UUID).
```
jhrozek, line 82: Looking at this record, I wonder if we should extend the struct TelemetryStore to include EntityType and EntityName instead of the current Resource.

internal/logger/telemetry_store_watermill.go
```diff
@@ -0,0 +1,88 @@
...
+		Str("entity-type", typ.String()).
+		Str("provider", inf.Provider).
```
jhrozek, line 84: Same for the provider, isn't this something that should be part of the struct instead of a one-off value?

internal/logger/telemetry_store_watermill.go
```diff
@@ -0,0 +1,88 @@
...
+	// In minder this would be the entity
+	switch typ {
+	case minderv1.Entity_ENTITY_UNSPECIFIED:
+		return fmt.Errorf("unspecified entity type")
+	case minderv1.Entity_ENTITY_BUILD_ENVIRONMENTS:
+		return fmt.Errorf("build environments not supported")
+	case minderv1.Entity_ENTITY_REPOSITORIES:
+		ts.Resource = repoID.String()
+	case minderv1.Entity_ENTITY_ARTIFACTS:
+		ts.Resource = artID.UUID.String()
```
jhrozek, line 74: Do you think the UUID would tell us much?

JAORMX: we have the resource name + this UUID

internal/logger/telemetry_store_watermill.go
evankanderson: I'm fine with including more fields in the TelemetryStore that make better sense.  For example, I don't have a strong feeling at the moment about type + name vs a semi-URL resource, or even a type + semi-URL resource.

evankanderson, line 82: I think we might not want to call `Send()` on the Zerolog.Event until _after_ `h` is run, because `h` might want to augment the event with additional details (for example, the names of rules that were run).

internal/logger/telemetry_store_watermill.go
```diff
@@ -0,0 +1,88 @@
...
+// TelemetryStoreMiddleware is a Watermill middleware that
+// logs the relevant telemetry data.
+func (m *TelemetryStoreWMMiddleware) TelemetryStoreMiddleware(h message.HandlerFunc) message.HandlerFunc {
+	return func(msg *message.Message) ([]*message.Message, error) {
+		if err := m.Log(msg); err != nil {
+			return nil, fmt.Errorf("error logging telemetry: %w", err)
+		}
+
+		return h(msg)
+	}
```
evankanderson, line 47: I think you want to pass the TelemetryStore along to be available for the rules engine, i.e.

```suggestion
	return func(msg *message.Message) ([]*message.Message, error) {
		ts := TelemetryStore{}
		logger := zerolog.Ctx(msg.Context())
		ctx := ts.WithTelemetry(logger.WithContext(msg.Context()))
		msg.SetContext(ctx)

		msgs, err := h(msg)
		
		// Record telemetry
		logMsg := logger.Info()
		if err != nil {
			logMsg = logger.Error()
		}
		ts.Record(logMsg)
		logMsg.Send()

		return msgs, err
	}
```

The "Record telemetry" part could partially go before the call to `h`, but I think we want to call `ts.Record` _after_ `h`, because the handler might want to store more telemetry that's part of the same record.

jhrozek:
Looks good! let me know when this is out of draft so I can ack

jhrozek (approved):

cmd/server/app/serve.go
```diff
@@ -120,9 +121,13 @@ var serveCmd = &cobra.Command{
 
 		s.ConsumeEvents(aggr)
 
+		tsmdw := logger.NewTelemetryStoreWMMiddleware(l)
+
 		exec, err := engine.NewExecutor(ctx, store, &cfg.Auth, evt,
 			engine.WithProviderMetrics(providerMetrics),
-			engine.WithAggregatorMiddleware(aggr))
+			engine.WithMiddleware(aggr.AggregateMiddleware),
+			engine.WithMiddleware(tsmdw.TelemetryStoreMiddleware),
```
evankanderson, line 129: Do we want to also collect telemetry from the reconciler?

internal/logger/telemetry_store_watermill.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/stacklok/minder/internal/engine"
+	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+)
+
+// TelemetryStoreWMMiddleware is a Watermill middleware that
+// logs the relevant telemetry data.
+type TelemetryStoreWMMiddleware struct {
+	l *zerolog.Logger
+}
```
evankanderson, line 31: If we store a logger here, does that override [our usage of `zerowater.NewZerologLoggerAdapter`](https://github.com/stacklok/minder/blob/main/internal/events/eventer.go#L126)?

internal/logger/telemetry_store_watermill.go
```diff
@@ -0,0 +1,101 @@
...
+			logMsg = m.l.Error()
+		}
+		ts.Record(logMsg).Send()
+
+		return msgs, err
+	}
+}
+
+func resourceFromEntity(typ minderv1.Entity, ent string) string {
+	return fmt.Sprintf("%s/%s", typ.ToString(), ent)
```
evankanderson, line 77: If it makes more sense to leave these as separate fields, feel free to add them both (or a sub-struct) to TelemetryStore

evankanderson: (We could even use proto here if it makes sense)

evankanderson (approved):
Thanks for the test!  This looks close enough to iterate on later in the codebase if you want, or you can address the "why not use the logger from context" now if you prefer.  (In fact, it may be that the logger adapter already adds some of the fields we need, I'm not sure...)

jhrozek:
> Thanks for the test! This looks close enough to iterate on later in the codebase if you want, or you can address the "why not use the logger from context" now if you prefer. (In fact, it may be that the logger adapter already adds some of the fields we need, I'm not sure...)

I think your comments are valid, but my 2c is to merge the PR quickly so I can propose mine over it (I did a local rebase this afternoon and there appeared a wild import loop, so I'd rather get all the pieces into main..)

JAORMX:
I'll get this merged and address the comments on separate PRs

cmd/server/app/serve.go
JAORMX: I'm not sure how much value do we get from that. But if we do, then we'd need dedicated middleware, as it does not use the entity info wrapper.

internal/logger/telemetry_store_watermill.go
JAORMX: that adapted is only usable if you have access to the watermill router. Which we don't here. I think a good thing to do would be to make the eventer also receive a zerolog Logger instance, so they both can use the same level at least.

internal/logger/telemetry_store_watermill.go
JAORMX: I'm not sure what makes sense since we haven't started using this telemetry data yet. What do you think?

Issue #1881, opened 2023-12-08
Add Rule evaluation results to Minder telemetry logs


evankanderson:
When evaluating rules, record the following data in the `TelemetryStore` associated with the message context (See #1878 and #1879):

* Resource evaluated on (`artifact_id`, `repository_id`, or `pr_id`)
* Rules evaluated
  * Rule outcomes / actions taken, including remediations or alerts created

PR #2104, opened 2024-01-10
Rule evaluator business telemetry

jhrozek:
Extends the telemetry store so that all the rules being evaluated are stored in the telemetry store for sending later.

Fixes: #1881

### DIFF
--- a/internal/engine/executor.go
+++ b/internal/engine/executor.go
@@ -32,6 +32,7 @@ import (
 	"github.com/stacklok/minder/internal/engine/ingestcache"
 	engif "github.com/stacklok/minder/internal/engine/interfaces"
 	"github.com/stacklok/minder/internal/events"
+	minderlogger "github.com/stacklok/minder/internal/logger"
 	"github.com/stacklok/minder/internal/providers"
 	providertelemetry "github.com/stacklok/minder/internal/providers/telemetry"
 	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
@@ -116,6 +117,8 @@ func (e *Executor) Wait() {
 // HandleEntityEvent handles events coming from webhooks/signals
 // as well as the init event.
 func (e *Executor) HandleEntityEvent(msg *message.Message) error {
+	// Grab the context before making a copy of the message
+	msgCtx := msg.Context()
 	// Let's not share memory with the caller
 	msg = msg.Copy()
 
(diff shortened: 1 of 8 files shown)


jhrozek:
Still a draft until we merge #2101 but rebased atop that PR, with some tests etc..

JAORMX:
@jhrozek if you're gonna move the entity info wrapper ( which I think is a good idea) I suggest doing so in a dedicated PR

jhrozek:
> @jhrozek if you're gonna move the entity info wrapper ( which I think is a good idea) I suggest doing so in a dedicated PR

Done. I think this PR is now ready for review.

internal/engine/executor.go
```diff
@@ -380,6 +399,7 @@ func (e *Executor) releaseLockAndFlush(
 func logEval(
 	ctx context.Context,
 	inf *entities.EntityInfoWrapper,
+	profileName string,
```
rdimitrov, line 402: This is already part of the evalParams, i.e. `params.Profile.Name`

internal/engine/executor.go
```diff
@@ -407,6 +427,10 @@ func logEval(
 		Str("action", "alert").
 		Str("action_status", string(evalerrors.ErrorAsAlertStatus(params.GetActionsErr().AlertErr))).
 		Msg("result - action")
+
+	// log business logic
+	minderlogger.BusinessRecord(ctx).
+		AddRuleEval(params.Rule.Type, profileName, params)
```
rdimitrov, line 433: It's not going to be that explicit, but I think you already have the rule and profile names through params (the ActionsParams interface)

JAORMX (commented):
telemetry store parts lgtm.

jhrozek:
> telemetry store parts lgtm.

Thanks for the review, I'll address @rdimitrov's concerns as well

internal/engine/executor.go
jhrozek: addressed in the new patch

internal/engine/executor.go
jhrozek: thanks, addressed in the new patch

internal/engine/interfaces/interface.go
```diff
@@ -76,6 +76,10 @@ const (
 	ActionOptUnknown
 )
 
+func (a ActionOpt) String() string {
+	return [...]string{"on", "off", "dry_run", "unknown"}[a]
```
JAORMX, line 80: what is this witchery

JAORMX (approved):

internal/engine/interfaces/interface.go
jhrozek: it just says that the size of the array is determined by the number of the elements

PR #2106, opened 2024-01-10
Bump sigstore-go version

### LINKED ISSUE #2105: Update sigstore-go version
Based on https://sigstore.slack.com/archives/C0440BFT43H/p1704315100650419

image

rdimitrov:
Fixes #2105


internal/verifier/sigstore/sigstore.go
```diff
@@ -49,7 +49,7 @@ func New(trustedRoot, accessToken, cacheDir string) (*Sigstore, error) {
 		return nil, err
 	}
 	sev, err := verify.NewSignedEntityVerifier(trustedMaterial, verify.WithSignedCertificateTimestamps(1),
-		verify.WithTransparencyLog(1), verify.WithOnlineVerification())
```
eleftherias, line 52: Is this because the API changed between versions, or is there a different reason for it?

rdimitrov: This one is an API change 👍

eleftherias (approved):

JAORMX (approved):

internal/verifier/sigstore/sigstore.go
```diff
@@ -49,7 +49,7 @@ func New(trustedRoot, accessToken, cacheDir string) (*Sigstore, error) {
...
+		verify.WithTransparencyLog(1), verify.WithOnlineVerification(), verify.WithObserverTimestamps(1))
```
haydentherapper, line 52: Just wondering, what is the need for online verification? Is it that there is no state and bundles aren't persisted?

internal/verifier/sigstore/sigstore.go
haydentherapper: Yea, we've separated requiring a number of transparency log entries from timestamps. Using the new WithObserverTimestamps API states that you require at least one timestamp from any source (the preferred API to use).

internal/verifier/sigstore/sigstore.go
rdimitrov: @haydentherapper - Currently Minder handles sigstore verification by processing artifact publish events from GitHub - when a signature comes, extract whatever is needed from it (build a v1 bundle, prev. it was using cosign verify) and verify it. 

My impression was that we can't avoid the online verification **yet** since the signature we get is missing the inclusion proof rekor details we should provide in order to use the v2 bundle version(cosign does not store it). 

I recall we touched this in the sigstore-go PR about the OCI example, so thus why I defaulted to online verification (ref. [comment](https://github.com/sigstore/sigstore-go/pull/30#issuecomment-1845096189)). At least this was my understanding for it back then, but I'll be happy to revisit and align this to sigstore-go's preferred approach if you have suggestions 💯

haydentherapper: Yep, you’re totally right, forgot this is because it’s using the converted v0.1 bundles.

rdimitrov: Thanks for the input though! 👍 I would be happy to keep minder on par with the latest sigtore-go, i.e. update to the latest bundle version once this is possible, etc.

PR #2107, opened 2024-01-10
Remove duplicate top level context

### LINKED ISSUE #2103: Fix the correct use of context, i.e. in profile commands like create, delete, etc.
Noticed that the profile CLI and backend implementation still use a context that is part of the Profile object and not the intended one that is part of the Request message.

For example:

https://github.com/stacklok/minder/blob/f6ed42c5918fbd2d5914de6d197793e6e16f7df3/internal/controlplane/handlers_profile.go#L125 where `in` is https://github.com/stacklok/minder/blob/f6ed42c5918fbd2d5914de6d197793e6e16f7df3/internal/controlplane/handlers_profile.go#L123

eleftherias:
- Remove the top level context from protos that already have a nested context (ruletype, profile)
- Add function to fetch nested context
- Allow CLI command flags to override YAML context

Note: this does not change the DB implementation, where ruletypes, profiles and repos are still tied to a provider and project.

Fix #2103

### DIFF
--- a/cmd/cli/app/profile/apply.go
+++ b/cmd/cli/app/profile/apply.go
@@ -61,7 +61,6 @@ func applyCommand(_ context.Context, cmd *cobra.Command, conn *grpc.ClientConn) 
 	applyFunc := func(ctx context.Context, f string, p *minderv1.Profile) (*minderv1.Profile, error) {
 		// create a profile
 		resp, err := client.CreateProfile(ctx, &minderv1.CreateProfileRequest{
-			Context: &minderv1.Context{Provider: &provider, Project: &project},
 			Profile: p,
 		})
 		if err == nil {
@@ -79,7 +78,6 @@ func applyCommand(_ context.Context, cmd *cobra.Command, conn *grpc.ClientConn) 
 		}
 		// The profile already exists, so update it
 		updateResp, err := client.UpdateProfile(ctx, &minderv1.UpdateProfileRequest{
-			Context: &minderv1.Context{Provider: &provider, Project: &project},
 			Profile: p,
 		})
 		if err != nil {
@@ -90,7 +88,7 @@ func applyCommand(_ context.Context, cmd *cobra.Command, conn *grpc.ClientConn) 
 	}
 	// cmd.Context() is the root context. We need to create a new context for each file
 	// so we can avoid the timeout.
-	if err := ExecOnOneProfile(cmd.Context(), table, f, cmd.InOrStdin(), project, applyFunc); err != nil {
+	if err := ExecOnOneProfile(cmd.Context(), table, f, cmd.InOrStdin(), project, provider, applyFunc); err != nil {
 		return cli.MessageAndError(fmt.Sprintf("error applying profile from %s", f), err)
 	}
 
--- a/cmd/cli/app/profile/common.go
+++ b/cmd/cli/app/profile/common.go
@@ -29,7 +29,7 @@ import (
 )
 
 // ExecOnOneProfile is a helper function to execute a function on a single profile
-func ExecOnOneProfile(ctx context.Context, t table.Table, f string, dashOpen io.Reader, project string,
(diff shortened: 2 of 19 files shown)


JAORMX (commented):
we currently parse the context from the YAML files; have you tested this thoroughly? I guess we'd need to remove this from our samples too.

JAORMX:
The main concern I have with removing the context from the Profile itself (and rules) is that it would make it harder for us to be able to manage this via a GitOps workflow. We should still cater for that use-case even if we don't support that yet.

rdimitrov:
I guess the provider is still relevant for rule types, not only because of its definition, but also their action's implementations are provider-specific - i.e. the API calls we use for remediation and alerts are GitHub's API.

JAORMX:
> Edit: that's for the YAML file at least. I think we can still remove it from the nested protos and use the generic context instead when passing the requests.

So, the flow to interact with minder in a git-based flow would be to first set up context and then apply the file you want? That doesn't sound very GitOps:y to me :/. The ideal scenario would be a k8s-style application approach, where the file determines the necessary context and the server verifies if sufficient permissions are available.

eleftherias:
I would remove the nested context from the protos and keep it in the YAML definition. Then in the CLI map the nested context to the generic context before sending to the backend. Also the generic context would take precedence over the nested context if they are different (i.e if you do `minder ruletype create -f some-ruletype.yaml --project A` then the ruletype will be created for project `A` even if the YAML defines a different project)
If we do implement a GitOps flow in the future, the service that reads the YAML definition would also map the nested context to the generic context.

rdimitrov:
> > Edit: that's for the YAML file at least. I think we can still remove it from the nested protos and use the generic context instead when passing the requests.
> 
> So, the flow to interact with minder in a git-based flow would be to first set up context and then apply the file you want? That doesn't sound very GitOps:y to me :/. The ideal scenario would be a k8s-style application approach, where the file determines the necessary context and the server verifies if sufficient permissions are available.

Ah, I see. Still this is valid only for the provider though, not for the project, right? My take is project is user specific and does not translate to making the rule type being reusable, the provider on the other hand is implementation-specific option and make sense to be added as part of the YAMLs? Also in that case the dedicated context being sent along the request should match and take into account that it cannot overwrite the overlapping values from the embedded context, i.e. the provider.

JAORMX:
> I would remove the nested context from the protos and keep it in the YAML definition. Then in the CLI map the nested context to the generic context before sending to the backend. Also the generic context would take precedence over the nested context if they are different (i.e if you do `minder ruletype create -f some-ruletype.yaml --project A` then the ruletype will be created for project `A` even if the YAML defines a different project)
> If we do implement a GitOps flow in the future, the service that reads the YAML definition would also map the nested context to the generic context.

Gotcha! This makes sense to me. thanks for the clarification

eleftherias:
I ended up removing the top level context when a nested context already existed, and adding a top level `GetContext` method that fetched the nested context. This prevented having multiple structs that were almost identical.

rdimitrov:
> I ended up removing the top level context when a nested context already existed, and adding a top level `GetContext` method that fetched the nested context. This prevented having multiple structs that were almost identical.

Oh, but I thought we decided to leave the top-level context in the request and in the YAML description and instead remove the overlapping values from the proto object?

JAORMX (approved):

eleftherias:
> Oh, but I thought we decided to leave the top-level context in the request and in the YAML description and instead remove the overlapping values from the proto object?

I started down that path, but ran into 2 issues:
- We would then have 2 very similar structs, one for the proto without the context and one for the YAML with the context, and we would have to remember to add any new fields to both.
- For the endpoints that list the profiles / ruletypes, we would still need a proto that has the context nested, assuming that the profiles in the response come from different projects (maybe a parent project and a sub-project in the future) so each item in the list would need its own context.

It's doable, but after comparing the options I thought that keeping the nested context was more maintainable. We can discuss it further if you think keeping the top-level context is better.

JAORMX:
@eleftherias that works for me.

rdimitrov (approved):

PR #2109, opened 2024-01-11
Repository deletion should remove the associated webhook

### LINKED ISSUE #1783: repo delete should also remove the webhook
### Describe the issue

when deleting a repo, the webhook is left intact. 

this means a repo has been removed from minder, yet minder will still recieve webhook payload events for the deleted repo.

the webhook should be removed as well. 

https://docs.github.com/en/rest/repos/webhooks?apiVersion=2022-11-28#delete-a-repository-webhook

### To Reproduce

1. register a repo `minder repo register --provider github`
2. delete some repo `minder repo delete  --provider github --name lukehinds/some-repo`

After deleting, check the webhook setttings on the GitHub UI:

https://github.com/lukehinds/some-repo/settings/hooks

0.0.19

Vyom-Yadav:
Fixes #1783

### DIFF
--- a/internal/controlplane/handlers_githubwebhooks.go
+++ b/internal/controlplane/handlers_githubwebhooks.go
@@ -35,6 +35,8 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/viper"
 	"golang.org/x/exp/slices"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
@@ -319,7 +321,7 @@ func (s *Server) registerWebhookForRepository(
 			}
 			if parsedURL.Host == parsedOriginalURL.Host {
 				// it is our hook, we can remove it
-				err = client.DeleteHook(ctx, repo.Owner, repo.Name, h.GetID())
+				_, err = client.DeleteHook(ctx, repo.Owner, repo.Name, h.GetID())
 				if err != nil {
 					errorStr := err.Error()
 					regResult.Status.Error = &errorStr
@@ -352,6 +354,40 @@ func (s *Server) registerWebhookForRepository(
 	regResult.Repository.DefaultBranch = repoGet.GetDefaultBranch()
 
 	return regResult, nil
+}
+
+func (s *Server) deleteWebhookFromRepository(
+	ctx context.Context,
+	provider db.Provider,
+	projectID uuid.UUID,
+	dbrepo db.Repository,
+) error {
+	pbOpts := []providers.ProviderBuilderOption{
+		providers.WithProviderMetrics(s.provMt),
+	}
+	providerBuilder, err := providers.GetProviderBuilder(ctx, provider, projectID, s.store, s.cryptoEngine, pbOpts...)
+	if err != nil {
+		return status.Errorf(codes.Internal, "cannot get provider builder: %v", err)
+	}
+
+	client, err := providerBuilder.GetGitHub(ctx)
+	if err != nil {
+		return status.Errorf(codes.Internal, "cannot create github client: %v", err)
+	}
+
+	webhookId := dbrepo.WebhookID
+	if webhookId.Valid {
+		resp, err := client.DeleteHook(ctx, dbrepo.RepoOwner, dbrepo.RepoName, int64(webhookId.Int32))
+		if err != nil {
+			if resp != nil && resp.StatusCode == http.StatusNotFound {
+				// if the hook is not found, we can ignore the error, user might have deleted it manually
+				return nil
+			}
+			return status.Errorf(codes.Internal, "cannot delete webhook: %v", err)
+		}
+	}
+
+	return nil
 }
 
 func (s *Server) parseGithubEventForProcessing(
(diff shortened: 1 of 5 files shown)


Vyom-Yadav:
>  Error: : # github.com/stacklok/minder/internal/engine/eval/vulncheck [github.com/stacklok/minder/internal/engine/eval/vulncheck.test]
  Error: internal/engine/eval/vulncheck/review_test.go:64:57: cannot use mockClient (variable of type *mockgh.MockGitHub) as "github.com/stacklok/minder/pkg/providers/v1".GitHub value in argument to newReviewPrHandler: *mockgh.MockGitHub does not implement "github.com/stacklok/minder/pkg/providers/v1".GitHub (wrong type for method DeleteHook)

Generating mocks....

internal/controlplane/handlers_repositories.go
```diff
@@ -445,3 +450,40 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
 
 	return out, nil
 }
+
+func deleteRepositoryAndWebhook(
+	s *Server,
```
gregfurman, line 455: Should this not rather be defined as a method of `Server`?

Vyom-Yadav: Yes, it should be. Done.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -354,6 +356,40 @@ func (s *Server) registerWebhookForRepository(
...
+
+	webhookId := dbrepo.WebhookID
+	if webhookId.Valid {
+		resp, err := client.DeleteHook(ctx, dbrepo.RepoOwner, dbrepo.RepoName, int64(webhookId.Int32))
+		if err != nil {
+			if resp != nil && resp.StatusCode == http.StatusNotFound {
+				// if the hook is not found, we can ignore the error, user might have deleted it manually
+				return nil
+			}
+			return status.Errorf(codes.Internal, "cannot delete webhook: %v", err)
```
evankanderson, line 386: I was going to ask if you use `%w` here and above to wrap the error, but looking at https://github.com/grpc/grpc-go/issues/3616, it looks like the GRPC team rejected supporting `Unwrap` as it might lead to expectations that it would work between client and server.

internal/controlplane/handlers_repositories.go
```diff
@@ -290,8 +290,13 @@ func (s *Server) DeleteRepositoryById(ctx context.Context,
 		return nil, err
 	}
 
-	// delete the repository
-	if err := s.store.DeleteRepository(ctx, repo.ID); err != nil {
+	provider, err := getProviderFromRequestOrDefault(ctx, s.store, in, repo.ProjectID)
+	if err != nil {
+		return nil, providerError(fmt.Errorf("provider error: %w", err))
```
evankanderson, line 295: `providerError` will already wrap with `provider error: %w` if the error is not a "no rows found" error, so you should just be able to pass in `err` here.

evankanderson: (Yeah, most of the other examples get this wrong, I'll send a PR to fix)

internal/controlplane/handlers_repositories.go
```diff
@@ -445,3 +450,31 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+	defer s.store.Rollback(tx)
+
+	qtx := s.store.GetQuerierWithTransaction(tx)
+	if err := qtx.DeleteRepository(ctx, repo.ID); err != nil {
+		return status.Errorf(codes.Internal, "error deleting repository: %v", err)
+	}
+
+	if err := s.deleteWebhookFromRepository(ctx, provider, projectID, repo); err != nil {
+		return err
+	}
```
evankanderson, line 473: While it doesn't particularly matter since you're using a transaction, I'd tend to put the call to remove the webhook _before_ the database delete, such that failing to clean up the webhook doesn't even need to touch the database.  (Since you're in a transaction, you won't generate a conflict between other updates to the repository and this delete until the `Commit`, at which point you will have either deleted the webhook or not.)

In general, we probably need a bit of anti-entropy here (e.g. periodically re-check that the webhook is installed, and on unexpected webhooks, try to clean them up), but I'm not going to ask you to solve that.

evankanderson (approved):
This looks good -- I had a few small comments, but I don't think they're blocking, so I'll merge later today unless you decide you want to address them.

internal/controlplane/handlers_repositories.go
Vyom-Yadav: >  I'd tend to put the call to remove the webhook before the database delete, such that failing to clean up the webhook doesn't even need to touch the database. 

1. User decides to delete the repository
2. User invokes the command, but it fails while deleting the repo record (DB Call), but the webhook was deleted (GH API Call)
3. User thinks this is a signal from god and decides to keep the repo
4. Webhook is now gone. Yikes!

> In general, we probably need a bit of anti-entropy here (e.g. periodically re-check that the webhook is installed, and on unexpected webhooks, try to clean them up),

Yeah, we need some sort of reconciliation mechanism to stay up to date with the state of the webhooks.

> but I'm not going to ask you to solve that.

I would like to solve this, can we discuss the exact scope in a new issue and some other details?

internal/controlplane/handlers_repositories.go
Vyom-Yadav: Yeah, I copied other examples, should have checked. Done.

evankanderson (approved):

PR #2116, opened 2024-01-12
Move EntityInfoWrapper to a separate package to avoid import loop

jhrozek:
With the addition of the business telemetry, the engine will depend on
the telemetry package, but since we need to also manage entity info
wrappers in the telemetry, we would have a dependency fro the telemetry
package on the engine as well.

To avoid this import loop, let's move the entity info wrapper into a
separate package.


JAORMX (approved):

internal/events/eventer.go
```diff
@@ -61,6 +61,13 @@ const (
 	metricsSubsystem = "eventer"
 )
 
+const (
+	// ExecuteEntityEventTopic is the topic for internal webhook events
+	ExecuteEntityEventTopic = "execute.entity.event"
```
jhrozek, line 66: I wasn't sure about moving these constants here, but I couldn't find another place that would be outside the top-level engine package and wouldn't cause yet another import loop

JAORMX: This works for now. We could move it to a `constants` package if we see the need for it.

Issue #1880, opened 2023-12-08
Use TelemetryStore to record additional information about RPCs


evankanderson:
For RPCs from end-users, add the following to the telemetry records (see #1878) once the request is authenticated:

- `login_sha`: SHA256 hash of the user's identity from Keycloak (`sub` field from the token)
- `project_id`: Project ID that the user is operating on
- `repository_id`, `artifact_id`, `rule_id`, `pr_id`, `profile_id` of the resource that the user is operating on.

PR #2117, opened 2024-01-12
Add BusinessTelemetry to authenticated RPC calls

rdimitrov:
The following PR adds telemetry data for authenticated RPC calls for each Minder service.

**Context related:**
* `login_sha` - SHA256 hash of the user's identity from Keycloak (sub field from the token)
* `provider` name and `project` UUID (project name was removed in https://github.com/stacklok/minder/pull/2124) 

**Resource related:**

* `repository` and `artifact` ID 
* Both IDs and Names of `profiles` and `rule types` in requests and in evaluation results

**Improvements:** (not part of this PR)

* I saw @eleftherias's comment at https://github.com/stacklok/minder/pull/2130#issuecomment-1894104431 and that would be nice once we have it, because then we should be able to move most of the telemetry code from this PR out of the handlers and replace it with one in the middleware.

**Questions:**

- [x]  Noticed that the EvalParams we now log use names (not IDs) for referring to the profile and rule type. Should we provide additional parameters for providing the human-readable format and not just the UUID for objects where this is possible? Or perhaps converge on one?
- [x]  Should we include the OrgID also (wasn't part of the issue) or wait for when it starts making more sense? A: No, plan is to remove it completely.

Fixes: https://github.com/stacklok/minder/issues/1880

### DIFF
--- a/internal/controlplane/handlers_artifacts.go
+++ b/internal/controlplane/handlers_artifacts.go
@@ -19,6 +19,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"github.com/stacklok/minder/internal/logger"
 	"slices"
 	"strings"
 
@@ -59,6 +60,10 @@ func (s *Server) ListArtifacts(ctx context.Context, in *pb.ListArtifactsRequest)
 	if err != nil {
 		return nil, fmt.Errorf("failed to list artifacts: %w", err)
 	}
+
+	// Telemetry logging
+	logger.BusinessRecord(ctx).Provider = provider.Name
+	logger.BusinessRecord(ctx).Project = projectID
 
 	return &pb.ListArtifactsResponse{Results: results}, nil
 }
@@ -120,6 +125,12 @@ func (s *Server) GetArtifactByName(ctx context.Context, in *pb.GetArtifactByName
 		return nil, status.Errorf(codes.Unknown, "failed to get artifact versions: %s", err)
 	}
 
+	// Telemetry logging
+	logger.BusinessRecord(ctx).Provider = artifact.Provider
+	logger.BusinessRecord(ctx).Project = artifact.ProjectID
+	logger.BusinessRecord(ctx).Artifact = artifact.ID
+	logger.BusinessRecord(ctx).Repository = artifact.RepositoryID
+
 	return &pb.GetArtifactByNameResponse{Artifact: &pb.Artifact{
 		ArtifactPk: artifact.ID.String(),
 		Owner:      artifact.RepoOwner,
@@ -173,6 +184,12 @@ func (s *Server) GetArtifactById(ctx context.Context, in *pb.GetArtifactByIdRequ
 	if err != nil {
 		return nil, status.Errorf(codes.Unknown, "failed to get artifact versions: %s", err)
 	}
+
+	// Telemetry logging
+	logger.BusinessRecord(ctx).Provider = artifact.Provider
+	logger.BusinessRecord(ctx).Project = artifact.ProjectID
+	logger.BusinessRecord(ctx).Artifact = artifact.ID
+	logger.BusinessRecord(ctx).Repository = artifact.RepositoryID
 
 	return &pb.GetArtifactByIdResponse{Artifact: &pb.Artifact{
 		ArtifactPk: artifact.ID.String(),
--- a/internal/controlplane/handlers_artifacts.go
+++ b/internal/controlplane/handlers_artifacts.go
@@ -19,7 +19,6 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
-	"github.com/stacklok/minder/internal/logger"
 	"slices"
 	"strings"
 
@@ -31,6 +30,7 @@ import (
 
 	"github.com/stacklok/minder/internal/db"
 	"github.com/stacklok/minder/internal/engine"
+	"github.com/stacklok/minder/internal/logger"
 	"github.com/stacklok/minder/internal/util"
 	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 )
--- a/internal/logger/telemetry_store.go
+++ b/internal/logger/telemetry_store.go
@@ -16,8 +16,8 @@ package logger
 
 import (
 	"context"
-	"github.com/google/uuid"
 
+	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 
 	"github.com/stacklok/minder/internal/engine/actions/alert"
--- a/internal/logger/telemetry_store.go
+++ b/internal/logger/telemetry_store.go
@@ -16,6 +16,7 @@ package logger
 
 import (
 	"context"
+	"github.com/google/uuid"
 
 	"github.com/rs/zerolog"
 
@@ -32,6 +33,18 @@ const (
 	telemetryContextKey key = iota
 )
 
+// RuleType is a struct describing a rule type for telemetry purposes
+type RuleType struct {
+	Name string    `json:"name"`
+	ID   uuid.UUID `json:"id"`
+}
+
+// Profile is a struct describing a Profile for telemetry purposes
+type Profile struct {
+	Name string    `json:"name"`
+	ID   uuid.UUID `json:"id"`
+}
+
 // ActionEvalData reports
 type ActionEvalData struct {
 	// how was the action configured - on, off, ...
@@ -42,8 +55,8 @@ type ActionEvalData struct {
 
 // RuleEvalData reports
 type RuleEvalData struct {
-	RuleName    string `json:"rule_name"`
-	ProfileName string `json:"profile_name"`
+	RuleType RuleType `json:"ruletype"`
+	Profile  Profile  `json:"profile"`
 
 	EvalResult string                                   `json:"eval_result"`
 	Actions    map[interfaces.ActionType]ActionEvalData `json:"actions"`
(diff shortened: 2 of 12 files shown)


eleftherias:
> Should we include the OrgID also (wasn't part of the issue) or wait for when it starts making more sense?

I would still like to remove the concept of org entirely and just use project.

internal/controlplane/handlers_artifacts.go
```diff
@@ -121,6 +126,12 @@ func (s *Server) GetArtifactByName(ctx context.Context, in *pb.GetArtifactByName
 		return nil, status.Errorf(codes.Unknown, "failed to get artifact versions: %s", err)
 	}
 
+	// Telemetry logging
+	logger.BusinessRecord(ctx).Provider = in.GetContext().GetProvider()
```
jhrozek, line 130: Are there any calls that don't operate with `EntityContext`? Could we have a utility function that populates the telemetry using values read from the EntityContext that's itself read from the ctx?

rdimitrov: I was talking about this with @evankanderson. Essentially I was wondering if there's a better approach to declutter and limit the repetitions we do in the RPC handlers. Should there be a wrapper similar to what @JAORMX did for the CLIs? Or would it make sense to move context-related values like project and provider to context.Context and process/store them from the logging interceptor. I don't have a clear view of what's the best approach but it feels like there should be one.

rdimitrov:
> > Should we include the OrgID also (wasn't part of the issue) or wait for when it starts making more sense?
> 
> I would still like to remove the concept of org entirely and just use project.

I think that answers the question! 😃 Thanks! 👍

internal/logger/telemetry_store.go
```diff
@@ -72,6 +72,21 @@ type TelemetryStore struct {
 
 	// Rules evaluated during processing
 	Evals []RuleEvalData `json:"rules"`
+
+	// Repository is the repository ID that the request was associated with.
+	Repository string `json:"repository"`
```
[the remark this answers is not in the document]
jhrozek: @teodor-yanev might have an opinion since he's working on actually consuming the fields

internal/controlplane/handlers_artifacts.go
jhrozek: So since many calls use `authAndContextValidation` (which is confusingly define in handlers_profile.go but used in more handlers), would it be too sneaky to also get the telemetry store from the context there and log at least the project and provider? Logging the actual entries in a generic way might be tricky as the requests and replies are different for each request...

jhrozek:
> * [ ]   Noticed that the EvalParams we now log use names (not IDs) for referring to the profile and rule type. Should we provide additional parameters for providing the human-readable format and not just the UUID for objects where this is possible? Or perhaps converge on one?

@teodor-yanev might have a better opinion, I'm fine changing the format of what's logged in whatever way makes it easier to actually consume the telemetry. I used names in the engine telemetry just because they were available and easier for me to read. But if it would make the logs inconsistent, I'm fine with reverting to IDs or using both

rdimitrov:
> 
> @teodor-yanev might have a better opinion, I'm fine changing the format of what's logged in whatever way makes it easier to actually consume the telemetry. I used names in the engine telemetry just because they were available and easier for me to read. But if it would make the logs inconsistent, I'm fine with reverting to IDs or using both

I was asking so I can update my PR accordingly, but I was planning to change it all if we settle on what makes sense (perhaps log/store both?).

rdimitrov:
@evankanderson @teodor-yanev - hey, what do you think about the discussion around what to store for objects - whether it's UUIDs or names or perhaps both?

evankanderson:
> @evankanderson @teodor-yanev - hey, what do you think about the discussion around what to store for objects - whether it's UUIDs or names or perhaps both? 

This is a good question, and depends a bit on what we consider the canonical identifier for external users.

You should coordinate with @JAORMX and what he's storing for rule evaluations -- the goal is that we could answer questions like "how many profile updates trigger an actual auto remediation"?

evankanderson:
Thinking a bit more, we might want a struct for the resource, so that we don't end up with disconnected fields later.

rdimitrov:
@evankanderson - can you elaborate more about what you suggest? I'm thinking of a struct like - 
```go
type Resource struct {
  ID uuid.uuid
  Name string
}
```
and use that type for the resources we have in the `TelemetryStore`, i.e.
```go
type TelemetryStore struct {
  ...
  Profile Resource
  RuleType Resource
  Artifact Resource
  Repository Resource
  PR Resource
  ...
}
```
Is this what you have in mind? I did not get the example clearly - my understanding is remediations happen every time as a result of profile evaluation (if a rule is failing and the action is enabled).

internal/controlplane/handlers_artifacts.go
rdimitrov: I thought I had replied to this one. 

There's no unified approach to fetching the provider/project pair so far so I defaulted to explicitly call this on each handler instead of embedding it in an existing function (i.e. `authAndContextValidation`) that is frequently used but not common across all RPC handlers (i.e. `DeleteRepositoryByName`).

My suggestion is aligned with what you say - figure out a way to have a standard way of fetching the provider/project pair for all RPCs, but leave the other fields explicitly called from the handlers since they are custom to their request/response.

evankanderson:
> **Edit:** @evankanderson - I think I got confused because of the wording, because for me "profile update" is the result of - 'profile apply' which is different from profile evaluations. I'll update the code to use a common struct (the example above) across both the changes from this PR but also across the ones for the rule evaluation telemetry so it's consistent across all.

Yes, I meant to use the same field for both rule evaluations and RPCs, sorry for not being clear (I was replying on phone).

internal/controlplane/handlers_artifacts.go
evankanderson: Yes!  This is why we don't want writing TelemetryStore to be a "one and done" from a central place, but we should use this as pressure towards having a common way to get common fields (i.e. doing some code cleanup where it makes sense).

internal/logger/telemetry_store.go
```diff
@@ -72,6 +85,18 @@ type TelemetryStore struct {
 
 	// Rules evaluated during processing
 	Evals []RuleEvalData `json:"rules"`
+
+	// Repository is the repository ID that the request was associated with.
+	Repository *uuid.UUID `json:"repository"`
```
evankanderson, line 90: Do we benefit from storing these as `uuid.UUID`s rather than `string`s?  I see two related problems:

1. It means that we need to call `uuid.Parse` to extract this data from the original protobufs where it makes sense.
2. It adds additional error cases when recording information, because it's possible that some other part of the system interprets UUIDs differently than this code.  (Unlikely in this case, but if we decided to change the format, this is one more place we'd need to update.)

If the data is already generally stored in the code as a uuid.UUID, then this seems fine, but if we're needing to add a bunch of parsing calls, I'd tend to prefer to store this as a `string` and make the logging code as dumb as possible, and then do cleanup elsewhere -- that makes it possible for us to see cases where the cleanup didn't happen correctly when we're observing corrupted data.

evankanderson (approved):
A few minor comments, but I think this is on the right track and we can refactor to reuse more common methods later if we feel motivated.

internal/logger/telemetry_store.go
```diff
@@ -40,10 +41,22 @@ type ActionEvalData struct {
...
+	Name string     `json:"name"`
+	ID   *uuid.UUID `json:"id"`
+}
+
 // RuleEvalData reports
 type RuleEvalData struct {
-	RuleName    string `json:"rule_name"`
-	ProfileName string `json:"profile_name"`
+	RuleType *RuleType `json:"ruletype"`
+	Profile  *Profile  `json:"profile"`
```
evankanderson, line 59: One more note: I don't know if it's easier to have these as pointers (caller may need to allocate structure vs set individual fields) or as embedded objects, but I wouldn't worry about memory overhead costs at this point.  If you use them as structures, I think you can compare with the default to see if they are empty:

```golang
if red.RuleType == RuleType{} {
    ....
}
```

internal/logger/telemetry_store_watermill.go
```diff
@@ -51,7 +51,7 @@ func (m *TelemetryStoreWMMiddleware) TelemetryStoreMiddleware(h message.HandlerF
 		}
 
 		ts := &TelemetryStore{
-			Project:  inf.ProjectID.String(),
+			Project:  inf.ProjectID,
 			Provider: inf.Provider,
 			Resource: resourceFromEntity(typ, ent),
```
evankanderson, line 56: Should this fill in the same `repository`, `artifact`, `profile`, etc fields?  (Or have those use this format?)

Ideally, we should have the same resource keys in the logs for both RPCs that update a resource, and rule evaluations that use that resource.  That will make it easier to ask questions like "how many newly added resources have some action taken on them immediately (rule eval with at least one negative result within 30 minutes)?"

evankanderson (dismissed):

internal/logger/telemetry_store.go
rdimitrov: BTW I'll rebase my changes on top of what Ria made, so I'm available to incorporate more feedback on this before merging 😃

internal/logger/telemetry_store.go
rdimitrov: My motivation was mostly about the memory overhead of that.

internal/logger/telemetry_store_watermill.go
rdimitrov: I'll see what else we can extract from the entity message here so we can put them in the TelemetryStore object 👍 

Depending on the message some of the fields might be missing, but that should be expected.

internal/logger/telemetry_store.go
evankanderson: It's useful if we're mostly storing something that's already a `uuid.UUID` into the record, like we do for Project in some cases.  In other cases (e.g. line 130 of `handlers_artifacts.go`), we end up parsing the UUID particularly for the telemetry recording, introducing a new error case that we previously never cared about.

It looks like we only do that once, though, so I think I over-reacted based on a sample size of 2-3 at the start of the review.

internal/logger/telemetry_store.go
evankanderson: I'd be cautious about optimizing here -- I _think_ you're trading off memory allocations vs additional objects that the GC has to track and collect.  I don't know what the right knobs are, but I'd tend to write this in ways that are easy to read and avoid errors until we have an indication that this is a performance problem (and where the problem is, because we're about to assemble a whole bunch of JSON...).

internal/logger/telemetry_store_watermill.go
evankanderson: Yes, partially-filled structs is expected.  What would be nice would be if fields like `project` (across minder) and `login_sha` (across Minder RPC and Trusty RPC) could be consistent.  Right now, I think Ozz used my `Resource string` formulation, and you've added separate fields for each resource type.  Both are fine, but I'd rather un-tangle the two at the source than later on in Tableau.

rdimitrov:
Rebased the PR with the latest changes and applied the feedback from the last review cycle. 

I decided to leave storing the project/provider pair to still happen explicitly in each handler since it is executed after the request is authenticated and also authorised.

I thought of putting it in the `EntityContextProjectInterceptor` middleware at https://github.com/stacklok/minder/blob/6588180ebcd8647c62fe454b3883ff9db439e9b1/internal/controlplane/handlers_authz.go#L191, but decided not to because:
* This means we'll log everything, even the requests that are not authorised
* There's another mechanism to default the provider if it's missing, so this would not guarantee that what we log here is essentially what will be used later in the request processing.

rdimitrov:
I'll proceed and merge this one 👍 If someone has any feedback please comment and I'll address it in a follow up PR.

PR #2125, opened 2024-01-15
WIP: Fix Rule Evaluation Logic for Handling Multiple Rules of the Same Type

### LINKED ISSUE #1609: Incorrect profile evaluation when using multiples rules of the same rule type
**Describe the bug**

Evaluation is not correct when there are multiple rules in a single profile for the same rule type, i.e. 2 `dependabot_configured` rules.

This results to having:
* a single rule evaluation entry whereas there should be one for each rule type
* the resulting rule eval status is based on the last evaluated rule of this type

**To Reproduce**

1. Create a profile with more than 1 rule type entries
2. List the profile details

Vyom-Yadav:
Fixes #1609 

--- 

Major Changes:

- Handled `rule_type_name` collisions by using rule hash (SHA 256).
- Modified CLI output to use `rule_type_name` + `rule_hash` in case of collisions.


database/migrations/000012_rule_evaluations_rule_hash.up.sql
```diff
@@ -0,0 +1,25 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- Add pull_request_id to rule_evaluations
+ALTER TABLE rule_evaluations
+    ADD COLUMN rule_hash TEXT NOT NULL;
```
JAORMX, line 17: I don't think you'll be able to do this. The `NOT NULL` constraint will come into effect on existing rules in staging/prod and will fail to execute.

I'd suggest this being a three-step process:

* Add the new column without the `NOT NULL` constraint.
* Iterate existing rule evaluations in the db to update the hash (`SELECT FOR UPDATE` is your friend).
* Add the `NOT NULL` constraint.

To test:

* Spawn a local minder instance from `main`.
* Create some sample rules and profiles
* Apply the upgrades (migration + minder restart)
* Verify that the hashes have been added.

If your workflow is based on `make run-docker` and if you want to easily test an upgrade, set the following environment variable:

```
export RUN_DOCKER_NO_TEARDOWN=true
```

This removes the teardown from the `make run-docker` invocation, getting docker to restart only the containers that need rebuilding.

JAORMX (changes requested):

PR #2126, opened 2024-01-16
Change auth scope owner for user requests

eleftherias:
This will be used later to decide if the incoming project should be saved in the context.


JAORMX (approved):
This is fairly safe, as we don't take any user data from these requests and instead rely on the token for self-enrollment/unenrollment

PR #2129, opened 2024-01-16
Add overrides for enabling alerts and remediations when creating profiles

JAORMX:
This adds a couple of simple boolean flags to the midner CLI that allows
a caller to enable alerts and remediations explicitly when doing
`minder profile create`.

Note that this was only added to the `create` sub-command, as it's
imperative and not declarative as is the `apply` sub-command.


jhrozek (approved):
nice!

PR #2130, opened 2024-01-16
Derive ruletype response from DB entry

eleftherias:
Use the DB saved result for the response instead of reusing the request body.


JAORMX:
@eleftherias oh, did we have some inconsistency somewhere?

eleftherias:
> @eleftherias oh, did we have some inconsistency somewhere?

Not yet, it is currently working because `authAndContextValidation` updates the request body with the default provider and project, but that seems quite fragile.
I also have a future change where I'd like to remove `authAndContextValidation` from each individual handler and move the context checking to a middleware.

PR #2131, opened 2024-01-17
Switch to offline Sigstore verification and set TimestampVerificationData to nil

rdimitrov:
The following PR has the following changes:
* Recreates https://github.com/stacklok/minder/pull/2120 switching Minder to using offline sigstore verification. For those that are interested there's a nice and detailed explanation of how this works at https://github.com/stacklok/minder/pull/2120#issuecomment-1892619405 
* Sets the TimestampVerificationData to nil. The reason is it was wrongfully used, but also its value is obtained from the `SignedEntryTimestamp` assuming `verifyWithObserverTimestamps()` is set (we do set it).

Thanks to @haydentherapper for raising this! 🍻

### DIFF
--- a/internal/verifier/sigstore/container/container.go
+++ b/internal/verifier/sigstore/container/container.go
@@ -253,15 +253,9 @@ func getBundleVerificationMaterial(params *verifyResult, manifestLayer *v1.Descr
 	}
 	// 3. Construct the verification material
 	return &protobundle.VerificationMaterial{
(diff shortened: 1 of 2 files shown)


haydentherapper (approved):
Looks great!

jhrozek (approved):
formally rubber-stamping since both @haydentherapper approved this PR earlier but also @rdimitrov asked his original PR earlier.

Issue #520, opened 2023-07-20
proto: consider using continuation tokens rather than offset


evankanderson:
**Describe the bug**
Pulling from https://github.com/stacklok/mediator/pull/488#discussion_r1268616926:

Proto recommends continuation tokens over offset/limit: https://protobuf.dev/programming-guides/api/#define-pagination-api FWIW

> I've seen some APIs that use a "continuation token" rather than an explicit offset parameter, to allow the effect of a server-side cursor. Typically this stores something like the "last row scanned" so the next query can start with key > last-seen. I don't think it's critical right now, but I figured I'd mention the two approaches since this is the first PR I'm reviewing where I see pagination implemented.

Generally, it's easier to convert a continuation token to an offset / limit internally (you can simply serialize the "new offset" into the token), but harder to go the other way.  Depending on the APIs and indexes involved, computing an offset may involve fetching and sorting all the items below `offset`, which is additional work for the database.

This is not high priority, but is a common pattern across our APIs.

Vyom-Yadav:
@evankanderson Please assign this issue to me.

JAORMX:
@Vyom-Yadav done! :smile:

PR #2138, opened 2024-01-17
Update proto files for pagination

eleftherias:
Related to PR #2097

Ref #520


JAORMX (approved):

Vyom-Yadav:
@eleftherias Just a question: the last commit from main on my local branch was https://github.com/stacklok/minder/commit/e370b08cd8472ac4b5dc9145862bab34a97a56af, and `make gen` is clean, i.e. not updating any files. So, I'm assuming this is version-related. 

Local dependency should stick to `protoc-gen-go v1.31.0` right?

PR #2139, opened 2024-01-17
Add middleware that populates project in context

eleftherias:
- For all requests where the authorization depends on the project, populate the project in the context
- Always fetch the project from the context when making authorization decisions

Fix https://github.com/stacklok/minder/issues/2078

### DIFF
--- a/internal/controlplane/handlers_artifacts.go
+++ b/internal/controlplane/handlers_artifacts.go
@@ -29,6 +29,7 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/stacklok/minder/internal/db"
+	"github.com/stacklok/minder/internal/engine"
 	"github.com/stacklok/minder/internal/util"
 	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 )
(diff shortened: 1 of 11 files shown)


JAORMX (approved):
LGTM!

internal/controlplane/handlers_artifacts.go
```diff
@@ -29,17 +29,16 @@ import (
...
 )
 
 // ListArtifacts lists all artifacts for a given project and provider
 // nolint:gocyclo
 func (s *Server) ListArtifacts(ctx context.Context, in *pb.ListArtifactsRequest) (*pb.ListArtifactsResponse, error) {
-	projectID, err := getProjectFromRequestOrDefault(ctx, in)
-	if err != nil {
-		return nil, util.UserVisibleError(codes.InvalidArgument, err.Error())
-	}
+	entityCtx := engine.EntityFromContext(ctx)
```
evankanderson, line 40: The docs on `EntityFromContext` explicitly calls out that `*EntityContext` may be nil.  I'm assuming that we're relying on the unary interceptor (and proto annotations) for that not to be the case here?

Looking more broadly, our usage of `EntityFromContext` is not very well guarded by `nil` checks, which makes me think we might want to return an empty `EntityContext` in those cases...

eleftherias: You are right that we're relying on the unary interceptor to set the `EntityContext` and expect it not to be nil. We should update `EntityFromContext` to return an empty `EntityContext` instead. I'll create a followup issue.

evankanderson: If an empty `EntityContext` is the correct result, I'm happy to do that PR.  I hadn't looked at the code enough to feel confident of its safety.

PR #2142, opened 2024-01-17
Remove EntityContext pointers and replace with struct references

evankanderson:
In https://github.com/stacklok/minder/pull/2139#discussion_r1455758961, I noted that we were using the result from `EntityFromContext` without checking to see if it was `nil`.  @eleftherias suggested that we might as well return a concrete struct rather than littering our code with obnoxious checks.  In doing so, I found a case where we `panic` if `Project` is set and `Provider` is not, so I fixed it and added tests until fear turned to boredom.

I also removed `GetProvider` and `GetProject` methods (and switched to direct struct references), since they didn't seem to add any value.  (If they were defaulting values even if `EntityContext` were `nil`, I'd have kept them, but it seemed superfluous in this case.)

### DIFF
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -178,14 +178,14 @@ func populateEntityContext(ctx context.Context, in HasProtoContext) (context.Con
 	}
 
 	// don't look up default provider until user has been authorized
-	providerName := in.GetContext().Provider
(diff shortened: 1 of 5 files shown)


eleftherias (approved):
Thanks @evankanderson!

rdimitrov (approved):

PR #2158, opened 2024-01-20
rego: Add new function to list files using a glob pattern

JAORMX:
This makes it easy for us to write policies that take `yaml` or `yml`
extensions into account without much hassle.

### DIFF
--- a/internal/engine/eval/rego/lib_test.go
+++ b/internal/engine/eval/rego/lib_test.go
@@ -554,3 +554,84 @@ allow {
 	})
 	require.NoError(t, err, "could not evaluate")
 }
+
+func TestListYamlUsingLSGlob(t *testing.T) {
+	t.Parallel()
+
+	fs := memfs.New()
+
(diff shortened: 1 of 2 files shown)


internal/engine/eval/rego/lib_test.go
```diff
@@ -554,3 +554,80 @@ allow {
...
+	require.NoError(t, fs.MkdirAll(".github/workflows", 0755))
+
+	_, err := fs.Create(".github/workflows/security.yaml")
+	require.NoError(t, err, "could not create sec workflow file")
+
+	_, err = fs.Create(".github/workflows/build.yml")
+	require.NoError(t, err, "could not create build workflow file")
+
+	_, err = fs.Create(".github/workflows/release.yaml")
+	require.NoError(t, err, "could not create release workflow file")
```
jhrozek, line 608: Could you add a file that doesn't match the pattern to make sure it's not listed?

JAORMX: Done

PR #2169, opened 2024-01-22
Remove superadmin role

### LINKED ISSUE #2070: Remove superadmin role
This role is not practically used, since it is never set in our identity provider.
There are still some instances where it's used for testing.

eleftherias:
Fix #2070


JAORMX (approved):
Yesssss

PR #2194, opened 2024-01-24
Add migration step for minder users to FGA tuples

JAORMX:
This adds a final step to the minder migration script to go throught the
existing users and attempt to migrate them into OpenFGA.

Closes: https://github.com/stacklok/minder/issues/1849

### DIFF
--- a/cmd/server/app/migrate_up.go
+++ b/cmd/server/app/migrate_up.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/stacklok/minder/internal/authz"
 	serverconfig "github.com/stacklok/minder/internal/config/server"
+	"github.com/stacklok/minder/internal/db"
 	"github.com/stacklok/minder/internal/logger"
 )
 
@@ -114,15 +115,21 @@ var upCmd = &cobra.Command{
 				return err
 			}
 
-			authzw.GetConfig().StoreID = storeID
-			if err := authzw.ResetAuthzClient(); err != nil {
-				return fmt.Errorf("error while resetting authz client: %w", err)
-			}
+			authzw.GetClient().SetStoreId(storeID)
 		}
 
 		mID, err := authzw.WriteModel(ctx)
 		if err != nil {
 			return fmt.Errorf("error while writing authz model: %w", err)
+		}
+
+		if err := authzw.GetClient().SetAuthorizationModelId(mID); err != nil {
+			return fmt.Errorf("error setting authz model ID: %w", err)
+		}
+
+		store := db.NewStore(dbConn)
+		if err := migratePermsToFGA(ctx, store, authzw, cmd); err != nil {
+			return fmt.Errorf("error while migrating permissions to FGA: %w", err)
 		}
 
 		cmd.Printf("Wrote authz model %s to store.\n", mID)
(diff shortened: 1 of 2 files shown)


eleftherias (dismissed):
Unfortunately openFGA throws an error if we try to write a tuple that already exists. Trying it locally I get:
```
Details: error while migrating permissions to FGA: error writing tuples: Write validation error for POST Write with body {"code":"write_failed_due_to_invalid_input","message":"cannot write a tuple which already exists: user: 'user:a845bf67-ab27-4110-ba7c-7f4e5b9d1664', relation: 'admin', object: 'project:1e4e6f8c-306a-4318-9d90-364519ddb5a9': invalid write input"}
 with error code write_failed_due_to_invalid_input error message: cannot write a tuple which already exists: user: 'user:a845bf67-ab27-4110-ba7c-7f4e5b9d1664', relation: 'admin', object: 'project:1e4e6f8c-306a-4318-9d90-364519ddb5a9': invalid write input
```
One option is to first query openFGA and see if the tuple already exists before attempting to write it.
Another option is to check if they support "force" writing which overwrites existing values (I haven't checked if this is an option).

JAORMX:
@eleftherias Seems that OpenFGA does not have idempotent tuple writes (see https://github.com/openfga/openfga/issues/258 ). To work around this, I now iterate through all the users and projects and do the write operations on each. On each result, we continue if the error was that the tuple already existed.

cmd/server/app/migrate_up.go
```diff
@@ -152,6 +160,43 @@ func ensureAuthzStore(ctx context.Context, cmd *cobra.Command, authzw *authz.Cli
 	return storeID, nil
 }
 
+func migratePermsToFGA(ctx context.Context, store db.Store, authzw *authz.ClientWrapper, cmd *cobra.Command) error {
+	cmd.Println("Migrating permissions to FGA...")
+
+	userlist, err := store.ListUsers(ctx, db.ListUsersParams{
+		Limit:  1000,
+		Offset: 0,
+	})
```
evankanderson, line 169: I'm assuming the limit here is something that can be changed if we need to run this and we have 10k users / projects?

cmd/server/app/migrate_up.go
```diff
@@ -152,6 +160,43 @@ func ensureAuthzStore(ctx context.Context, cmd *cobra.Command, authzw *authz.Cli
...
+		for _, proj := range projs {
+			if err := authzw.Write(
+				ctx, user.IdentitySubject, authz.AuthzRoleAdmin, proj.ProjectID,
+			); err != nil && strings.Contains(err.Error(), "already exists") {
+				cmd.Printf("Skipping user %d since it already has role on project %s\n",
+					user.ID, proj.ProjectID)
+				continue
+			} else if err != nil {
+				return fmt.Errorf("error while writing tuple: %w", err)
+			}
```
evankanderson, line 191: Do we want to skip over the rest of the tuples once we hit a single error, or do we want to try to copy as much as we can?  (I'm really not sure, looking for input here)

evankanderson: In particular, this decision is different than the decision on line 177.

cmd/server/app/migrate_up.go
evankanderson: Alternative would be to use a `for` loop:

```golang
i := 0
for {
  userList, err := store.ListUsers(ctx, db.ListUsersParams{Limit: 100, Offset: i})
  if err != nil {
    // report err, exit
  }
  if len(userList) == 0 {
    break
  }
  for _, user := range userlist {
    // ...
  }
}
```

internal/authz/authz.go
```diff
@@ -265,10 +265,12 @@ func (a *ClientWrapper) Delete(ctx context.Context, user string, role Role, proj
 	return nil
 }
 
-func getUserForTuple(user string) string {
+// GetUserForTuple gets a relevant user entry for writing in a tuple
+func GetUserForTuple(user string) string {
```
evankanderson, line 269: Why make these public?

evankanderson (dismissed):
I'm not thrilled about the calls to `authzw.GetClient()` that seem to use a global FGA client.  IMO, it would be better for `GetClient()` to return a _new_ client on each call, and to store the client in the body of the command.

In particular, I'm worried about unit test contagion if we return a bunch of pointers to a common FGA, but each test needs to do its own FGA setup.

evankanderson:
> @eleftherias Seems that OpenFGA does not have idempotent tuple writes (see [openfga/openfga#258](https://github.com/openfga/openfga/issues/258) ). To work around this, I now iterate through all the users and projects and do the write operations on each. On each result, we continue if the error was that the tuple already existed.

I chimed in on the issue with additional details from the linked issue; it looks like ignoring duplicate write errors (maybe in ClientWrapper?) is the right thing to do in the short term.

cmd/server/app/migrate_up.go
```diff
@@ -152,6 +160,43 @@ func ensureAuthzStore(ctx context.Context, cmd *cobra.Command, authzw *authz.Cli
...
+				user.ID, err)
+			continue
+		}
+
+		for _, proj := range projs {
+			if err := authzw.Write(
+				ctx, user.IdentitySubject, authz.AuthzRoleAdmin, proj.ProjectID,
+			); err != nil && strings.Contains(err.Error(), "already exists") {
+				cmd.Printf("Skipping user %d since it already has role on project %s\n",
+					user.ID, proj.ProjectID)
```
evankanderson, line 187: Do we want to put this error handling into `authzw.Write`?  It seems like other clients might want the same "idempotent write" behavior, even if OpenFGA isn't supporting it natively yet.

JAORMX: That's a good idea

eleftherias (approved):

PR #2195, opened 2024-01-24
Support GitHub attestation endpoint

jhrozek:
In addition to fetching provenance attestations from an OCI image, let's also
support the GitHub attestation endpoint as a fallback (a fallback since it's
not GA yet).

The attestation reply is unmarshalled into a protobuf representation and
passed to sigstore-go.

Fixes: https://github.com/stacklok/epics/issues/174

### DIFF
--- a/internal/verifier/sigstore/container/container.go
+++ b/internal/verifier/sigstore/container/container.go
@@ -26,17 +26,23 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"net/http"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
+	containerdigest "github.com/opencontainers/go-digest"
+	"github.com/rs/zerolog"
 	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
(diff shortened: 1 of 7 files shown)


evankanderson:
I think your `Fixes` is not right...

internal/verifier/sigstore/container/container.go
```diff
@@ -26,37 +26,96 @@ import (
...
+) ([]byte, []byte, error) {
+
 	// create a default verification result
 	params := newVerifyResult(BuildImageRef(registry, owner, artifact, version))
 
+	auth := newContainerAuth(authOpts...)
+
 	// construct the bundle
-	err := bundleFromOCIImage(params, newGithubAuthenticator(owner, accessToken))
+	// TODO: handle no access token?
```
JAORMX, line 104: I think we should error out, we should always have a token

jhrozek:
> I think your `Fixes` is not right...

So we started filing the issues in the epics repo because it's private and we were not sure if we can reference the work publicly as the GH attestations program is not public. We got a green light from GH yesterday to merge the code in a public repo, but for some reason I can't seem to be able to move the issue to the minder repo. For now, I'm going to reference the full URL of [the issue](https://github.com/stacklok/epics/issues/174) in the epics repo. Is there a way to move the issue or do you know what can be preventing the issue from being moved? Is it that the epics repo is private but the minder repo is public?

internal/verifier/sigstore/container/container.go
jhrozek: done, tested by running `mindev container verify` with no token.

rdimitrov (approved):
🚀

JAORMX (approved):
shipit :ship:

PR #2198, opened 2024-01-25
Add ability to evaluate scores based on detailed scores for trusty

JAORMX:
This uses the `description` scores in trusty to be able to evluate
scores using the trusty evaluator. This allows for having more detailed
policies such as the following:

```yaml
version: v1
type: profile
name: acme-github-profile
context:
  provider: github
alert: "on"
remediate: "off"
pull_request:
  - type: pr_trusty_check
    def:
      action: summary
      ecosystem_config:
        - name: npm
          score: 8
          evaluate_score: provenance
        - name: pypi
          score: 8
          evaluate_score: provenance
```

That is, we're able to just care about provenance and not just the
trusty summary.


ethomson:
It looks like this is a subset of https://github.com/stacklok/epics/issues/128 ?

JAORMX:
> It looks like this is a subset of https://github.com/stacklok/epics/issues/128 ?

I guess it is! I hadn't seen that one

jhrozek (approved):

PR #2201, opened 2024-01-26
Make authz tuple deletion idempotent

JAORMX:
This detect if the error for deleting an object is caused by the tuple
not existing, and if so, it discards the error since that's what we
wanted. Thus making calls to the tuple deletion idempotent.

### DIFF
--- a/internal/authz/authz.go
+++ b/internal/authz/authz.go
@@ -255,7 +255,9 @@ func (a *ClientWrapper) Delete(ctx context.Context, user string, role Role, proj
 			Object:   getProjectForTuple(project),
 		},
 	}).Execute()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/authz/authz.go
```diff
@@ -255,7 +255,9 @@ func (a *ClientWrapper) Delete(ctx context.Context, user string, role Role, proj
 			Object:   getProjectForTuple(project),
 		},
 	}).Execute()
-	if err != nil {
+	if err != nil && strings.Contains(err.Error(), "cannot delete a tuple which does not exist") {
```
jhrozek, line 258: I assume there's no Error we could match using errors.Is ?

jhrozek (approved):
approved with a question

internal/authz/authz.go
JAORMX: That's right :cry:

PR #2203, opened 2024-01-26
Add new policy writer role

JAORMX:
This adds a new role:

* `policy_writer`: which can only `create`, `update` and `delete` rule types and profiles.

These new role is meant for CI jobs or automation to populate all the
needed pieces for an organization to set up policy via minder.


eleftherias (changes requested):
Can we scope this down to only `policy_writer`? My thinking is:
1) Rule types and profiles go hand in hand, it's unlikely that a user will need permission to create one and not the other
2) Having fewer roles to start with will make it easier for users to choose the right one
3) Adding `rule_type_writer` and `profile_writer` raises the question of whether we should also add `repo_writer`, `provider_writer` etc

JAORMX:
@eleftherias addressed you feedback. Thanks for the review!

eleftherias (approved):

PR #2205, opened 2024-01-26
Solidify authz Client interface

JAORMX:
This removes the OpenFGA-specific logic from the migration script and
relies instead on a more general interface. This also makes the authz
constructor to return that interface to make this happen. The functions
that no longer need exporting are now private, and unused functions are
removed.

### DIFF
--- a/cmd/server/app/migrate_up.go
+++ b/cmd/server/app/migrate_up.go
@@ -25,6 +25,7 @@ import (
 	"github.com/golang-migrate/migrate/v4"
 	_ "github.com/golang-migrate/migrate/v4/database/postgres" // nolint
 	_ "github.com/golang-migrate/migrate/v4/source/file"       // nolint
+	"github.com/rs/zerolog"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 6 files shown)


cmd/server/app/migrate_up.go
```diff
@@ -103,63 +104,31 @@ var upCmd = &cobra.Command{
...
 		}
 
-		mID, err := authzw.WriteModel(ctx)
-		if err != nil {
-			return fmt.Errorf("error while writing authz model: %w", err)
-		}
-
-		if err := authzw.GetClient().SetAuthorizationModelId(mID); err != nil {
-			return fmt.Errorf("error setting authz model ID: %w", err)
+		if err := authzw.PrepareForRun(ctx); err != nil {
```
eleftherias, line 118: Since `PrepareForRun` sets the `StoreId`, it need to be called before `MigrateUp`.

eleftherias (changes requested):

cmd/server/app/migrate_up.go
JAORMX: `MigrateUp` makes sure the store and model exist. So `MigrateUp` actually needs to go first. Once we have a store and a model, then we can actually prepare the client for operations with OpenFGA.

JAORMX: oh... this used to work because I set the StoreID while doing the migration before and now I don't. Let me fix this.

JAORMX: Fixed

eleftherias (approved):

PR #2206, opened 2024-01-26
Fix Rule Evaluation Logic for Handling Multiple Rules of the Same Type - Part 2 (Data Backfilling)

### LINKED ISSUE #1609: Incorrect profile evaluation when using multiples rules of the same rule type

Vyom-Yadav:
Fixes #1609

---

Changes and implications:

- `entity_profiles.contextual_rules` are modified to have rule name key with `rule_type` or `rule_type-rand_uuid` in case of collisions

- Existing null `rule_evaluations.rule_name` columns are filled by referring to `name` key in `entity_profiles.contextual_rules`

- Handlers and Sql queries are updated to reflect not null constraint on `rule_evaluations.rule_name` column

- `migration_profile_backfill_log` table is created to support migrating down

- The rules are reconciled by the subsequent `InternalProfileInitEventTopic`, and `ExecuteEntityEventTopic` events. Profiles/Repositories not receiving these types of events **won't be reconciled** to update them to the latest state.

- Reconciliation can potentially open new PRs/Security Advisories, etc., as the duplicate rule's row (entry) was overridden by the other rule of the same type.

### DIFF
--- a/database/migrations/000017_rule_evaluations_rule_name_backfilling.up.sql
+++ b/database/migrations/000017_rule_evaluations_rule_name_backfilling.up.sql
@@ -1,0 +1,129 @@
+-- Copyright 2024 Stacklok, Inc
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- Begin transaction
+BEGIN;
+
+-- Function to check whether jsonb array element contain a name key
+CREATE OR REPLACE FUNCTION contains_name_key(input_jsonb jsonb) RETURNS boolean AS
+$$
+DECLARE
+  element jsonb;
+BEGIN
+  FOR element IN SELECT * FROM jsonb_array_elements(input_jsonb)
+    LOOP
+      IF element ? 'name' THEN
+        RETURN true;
+      END IF;
+    END LOOP;
+
+  RETURN false;
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
+
+-- Function to add a descriptive name to each element in a JSONB array
+CREATE OR REPLACE FUNCTION add_descriptive_name_to_jsonb_array(input_jsonb jsonb) RETURNS jsonb AS
+$$
+DECLARE
+  updated_array jsonb;
+  element       jsonb;
+  element_type  text;
+BEGIN
+  updated_array := '[]'::jsonb;
+
+  FOR element IN SELECT * FROM jsonb_array_elements(input_jsonb)
+    LOOP
+      element_type := element ->> 'type';
+
+      IF (SELECT COUNT(*) FROM jsonb_array_elements(input_jsonb) WHERE value ->> 'type' = element_type) > 1 THEN
+        element := jsonb_set(element, '{name}', ('"' || element_type || '_' || gen_random_uuid() || '"')::jsonb);
+      ELSE
+        element := jsonb_set(element, '{name}', ('"' || element_type || '"')::jsonb);
+      END IF;
+
+      updated_array := updated_array || jsonb_build_array(element);
+    END LOOP;
+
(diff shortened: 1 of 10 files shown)


Vyom-Yadav:
```
Error: error while creating authz client: Key: 'AuthzConfig.ApiUrl' Error:Field validation for 'ApiUrl' failed on the 'required' tag
Message: Error executing root command
Details: error while creating authz client: Key: 'AuthzConfig.ApiUrl' Error:Field validation for 'ApiUrl' failed on the 'required' tag
```

CI is broken, OpenFGA config is not present for the migrate container, the openfga container needs to be spinned up alongside. Also, the step passes, so error handling of the action needs to be improved.

database/migrations/000017_rule_evaluations_rule_name_backfilling.up.sql
```diff
@@ -0,0 +1,129 @@
...
+-- Add updated profile_id to migration_profile_backfill_log, don't add if already exists
+WITH updated_profile_ids AS (
+  -- Update existing rules to have the name field, internally this field is mandatory
+  UPDATE entity_profiles
+    SET contextual_rules = add_descriptive_name_to_jsonb_array(entity_profiles.contextual_rules),
+      updated_at = now()
+    WHERE contains_name_key(entity_profiles.contextual_rules) = false
+    RETURNING entity_profiles.profile_id)
+INSERT
+INTO migration_profile_backfill_log (profile_id)
```
evankanderson, line 105: I like this approach, thank you!

database/query/profile_status.sql
```diff
@@ -2,7 +2,7 @@
 -- name: UpsertRuleEvaluations :one
 INSERT INTO rule_evaluations (
     profile_id, repository_id, artifact_id, pull_request_id, rule_type_id, entity, rule_name
-) VALUES ($1, $2, $3, $4, $5, $6, sqlc.narg(rule_name))
+) VALUES ($1, $2, $3, $4, $5, $6, $7)
```
evankanderson, line 5: It took me a bit to verify. this, so I'm going to record it.  I think this is safe because the migration script runs _before_ the server update.

Because migrations run before the server code update, we could have the server attempting to insert `rule_name == NULL` from https://github.com/stacklok/minder/blob/b7fb73a53159a74dd6307a4dd7de4c614ec8aef1/internal/engine/eval_status.go#L129 unless we've backfilled the table.  Once the table is backfilled, that shouldn't happen, and all the following rule evaluations should have name set:

Path:

* `EvalStatusParams` is set [here in `createEvalStatusParams`](https://github.com/stacklok/minder/blob/b7fb73a53159a74dd6307a4dd7de4c614ec8aef1/internal/engine/eval_status.go#L48) from the `Profile.Rule` and not otherwise.
* [`getEvaluator`](https://github.com/stacklok/minder/blob/main/internal/engine/executor.go#L272) is the only caller of `createEvalStatusParams`, which is [a pass-through from the `TraverseRules` in `evalEntityEvent`](https://github.com/stacklok/minder/blob/main/internal/engine/executor.go#L239).
* [`TraverseRules` runs the supplied function over each rule in the `relevant` set](https://github.com/stacklok/minder/blob/main/internal/engine/profile.go#L142) computed by [`GetRulesForEntity`](https://github.com/stacklok/minder/blob/main/internal/engine/profile.go#L101), which selects one of `repository`, `build_environment`, `pull_requests`, or `artifacts` from the lists of rules in the Profile and returns them without edit.
* We get the set of Profiles from [`ListProfileByProjectID`](https://github.com/stacklok/minder/blob/main/internal/engine/executor.go#L224) passed through [`MergeDatabaseListIntoProfiles`](https://github.com/stacklok/minder/blob/main/internal/engine/profile.go#L156)
* While `ListProfilesByProjectID` is auto-generated, `MergeDatabaseListIntoProfiles` [calls `rowInfoToProfileMap`](https://github.com/stacklok/minder/blob/main/internal/engine/profile.go#L187) to fill out the rules in the proto.
* `rowInfoToProfileMap` [unpacks the list of rules stored as JSON in the database](https://github.com/stacklok/minder/blob/main/internal/engine/profile.go#L250), and returns them, which doesn't set rule names if not already present.
* The migration is selecting the rows FOR UPDATE, which should hopefully block any attempts to evaluate rules while the migration is in progress.

evankanderson (approved):

PR #2212, opened 2024-01-28
Upgrade postgres to 16.1-alpine in docker-compose

JAORMX:


Vyom-Yadav:
@JAORMX What is the version used by the production minder server?

JAORMX:
Uhmm... I should double check

rdimitrov (approved):
I'll approve because I think it's worth it to try and develop locally using the latest versions of dependencies since it will allow us for a safer upgrade in staging/prod later.

JAORMX:
I think it's fine to have latest and greatest in development. It'll be easier to know if we gotta watch out of certain things on upgrades.

PR #2218, opened 2024-01-29
Add authorization relations in proto

eleftherias:
In this approach we enumerate all of the roles a user can have with an entity (in this case only project). 

If our API evolves to have finer grained authorization, then we can add values to `ObjectOwner` to describe the entity type we are requesting authorization for.
For example, if we add a relation where a user can "create" a repo, then we can add an `OBJECT_OWNER_REPO` enum value to `ObjectOwner` and reuse the `Relation.CREATE` value.

### Alternatives considered
Only specifying the resource type we are requesting access to in the rpc definition and deriving the relation, either by:
- Getting the HTTP verb and mapping it to the relation (e.g get -> get, post -> create)
- Using our naming convention to derive the relation (e.g if the method starts with `List` then it's a `get` relation)

```
rpc ListArtifacts (ListArtifactsRequest) returns (ListArtifactsResponse) {
    option (google.api.http) = {
        get: "/api/v1/artifacts/{provider}"
        additional_bindings {
            get: "/api/v1/artifacts"
        }
    };

    option (rpc_options) = {
        auth_scope: OBJECT_OWNER_PROJECT
        resource: ARTIFACT
    };
}
```
Both of these seem fragile, because they rely on us following certain conventions. In the case of something as important as authorization, I would rather be explicit, even if it's verbose.


proto/minder/v1/minder.proto
```diff
@@ -31,12 +31,29 @@ enum ObjectOwner {
...
+    UPDATE = 2;
+    DELETE = 3;
+    REPO_GET = 5;
+    REPO_CREATE = 6;
+    REPO_UPDATE = 7;
+    REPO_DELETE = 8;
+    ARTIFACT_GET = 9;
+    ARTIFACT_CREATE = 10;
+    ARTIFACT_UPDATE = 11;
+    ARTIFACT_DELETE = 12;
```
JAORMX, line 46: What about adding annotations as such https://stackoverflow.com/questions/11474416/define-dictionary-in-protocol-buffer/11486640#11486640 to make this mapping easier?

JAORMX:
Giving this some thought, I do like the proposed approach more than the described alternative. Explicit is better than implicit and it seems to me like it's less error-prone. This will also help us handle some edge-cases such as the profile status (which needs a dedicated relation IMO).

proto/minder/v1/minder.proto
```diff
@@ -470,6 +526,7 @@ service ProfileService {
 
         option (rpc_options) = {
             auth_scope: OBJECT_OWNER_PROJECT
+            relation: RELATION_PROFILE_GET
```
JAORMX, line 529: I'd say we should have a dedicated relation for this. e.g. `profile_status_get`. Even if it just gets added to the existing roles, it's handy to have such granularity.

proto/minder/v1/minder.proto
```diff
@@ -459,6 +514,7 @@ service ProfileService {
...
```
JAORMX, line 517: same comment as below. A dedicated relation for `profile_status_get` would be ideal.

proto/minder/v1/minder.proto
```diff
@@ -31,12 +31,49 @@ enum ObjectOwner {
     reserved 1; // deprecated OBJECT_OWNER_ORGANIZATION
 }
 
+enum Relation {
+    RELATION_UNSPECIFIED = 0;
```
JAORMX, line 35: Do we need a dedicated relation to mark that an API endpoint doesn't need to go through authorization? e.g. the User endpoints which only apply to the current user.

eleftherias: I was imagining authorization to depend on the relation and the `ObjectOwner`. In the case of user endpoints, the object owner is the user, so we wouldn't bother looking that the relation.

proto/minder/v1/minder.proto
eleftherias: Are you suggesting we add it to the open fga model? Or just to the proto? I'm fine with either

proto/minder/v1/minder.proto
JAORMX: oh! I was thinking of ditching the `ObjectOwner`, `ProjectOwner` etc. pieces altogether.

eleftherias: We could. I kind of like having the relations closely tied to the openfga model, and having a separate way of deciding if we should even call openfga.

proto/minder/v1/minder.proto
JAORMX: The suggestion is for bothboth. To have a `profile_status_get` relation in the FGA model and add it here as well. what do you think?

proto/minder/v1/minder.proto
JAORMX: The `Object/Project/Org` ownership, while something we have now, seems like could it could be easy to forget. What if we have a separate mechanism and deprecate the other way?

proto/minder/v1/minder.proto
eleftherias: Works for me, I'll update

proto/minder/v1/minder.proto
eleftherias: That works, I'm not attached to the `ObjectOwner` enum.
One option could be adding an enum for a target resource type.
```
enum TargetResource {
    TARGET_RESOURCE_PROJECT = 0;
    TARGET_RESOURCE_NONE = 1;
    TARGET_RESOURCE_USER = 2;
}
```
Then `TARGET_RESOURCE_NONE` could also replace our current usage of `anonymous`.

Do you have any other ideas?

JAORMX: I like your proposal. I think we should go for that. It feels a lot more intuitive.

JAORMX (approved):
awesome! great work!

evankanderson (approved):

proto/minder/v1/minder.proto
```diff
@@ -31,12 +31,62 @@ enum ObjectOwner {
...
+    TARGET_RESOURCE_NONE = 1;
+    TARGET_RESOURCE_USER = 2;
+    TARGET_RESOURCE_PROJECT = 3;
+}
+
 message RpcOptions {
-    bool anonymous = 1;
     bool no_log = 2;
     bool owner_only = 3;
     bool root_admin_only = 4;
```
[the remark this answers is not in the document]
eleftherias: Yes, that's right

proto/minder/v1/minder.proto
```diff
@@ -31,12 +31,62 @@ enum ObjectOwner {
...
-    ObjectOwner auth_scope = 5;
+    TargetResource target_resource = 6;
+    Relation relation = 7;
+    reserved 1, 5; // deprecated anonymous and auth_scope
```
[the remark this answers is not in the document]
eleftherias: Yeah I was wondering if it's even worth reserving the field numbers here, since it's not something the clients use.

PR #2219, opened 2024-01-29
Add permissions API to FGA model

JAORMX:
This adds the permissions API to the FGA model. This API is highly
privileged, so only admins can take it into use. Note that we also added
a permissions manager role that's meant for SaaS-like cases where a
support engineer can provision a project for a user by giving them
access.

Note that uses of these future API endpoints should be heavily audited.


eleftherias (commented):
Do we need the `permissions_manager` role?
I don't think there's a right or wrong answer, just a matter of preference. 
I'd prefer not to add this role and keep the role set limited and add new ones as we discover new personas and use-cases from our customers.

JAORMX:
> Do we need the `permissions_manager` role?
> I don't think there's a right or wrong answer, just a matter of preference. 
> I'd prefer not to add this role and keep the role set limited and add new ones as we discover new personas and use-cases from our customers. 

IMO we do. Once we we introduce the hierarchy, it would be ideal that folks doing support have minimal roles that don't infringe on user's privacy. A common scenario is role provisioning, setup and fix up.

eleftherias (approved):

PR #2227, opened 2024-01-30
Implement role assignment and removal APIs

JAORMX:
This implements the role assignment and removal in the Minder server.
This already takes into account the project hierarchy.


internal/controlplane/handlers_authz.go
```diff
@@ -265,12 +265,121 @@ func (*Server) ListRoleAssignments(
...
+
+	authzrole, err := authz.ParseRole(role)
+	if err != nil {
+		return nil, util.UserVisibleError(codes.InvalidArgument, err.Error())
+	}
+
+	// Determine target project.
+	entityCtx := engine.EntityFromContext(ctx)
+	currentProjectID := entityCtx.Project.ID
+	targetProjectID := currentProjectID
```
eleftherias, line 287: I'm wondering if there's any benefit in differentiating the current project and the target project. 
If we have a command like: `minder role assign --subject user1 --project p1`, then the project in the context (i.e current project) would also be `p1`. At least that's how we do it for other commands.

JAORMX: I think you're right. Let me remove this logic from here.

JAORMX: done

eleftherias (changes requested):

internal/controlplane/handlers_authz.go
```diff
@@ -265,12 +267,92 @@ func (*Server) ListRoleAssignments(
...
+	if _, err := s.store.GetUserBySubject(ctx, sub); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, util.UserVisibleError(codes.NotFound, "User not found")
+		}
+		return nil, status.Errorf(codes.Internal, "error getting user: %v", err)
+	}
+
+	// Determine target project.
+	entityCtx := engine.EntityFromContext(ctx)
+	targetProjectID := entityCtx.Project.ID
```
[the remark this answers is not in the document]
JAORMX: I can add this validation in a separate PR (should probably merge before this).

internal/controlplane/handlers_authz.go
```diff
@@ -265,12 +267,92 @@ func (*Server) ListRoleAssignments(
...
+	// Verify if user exists
+	if _, err := s.store.GetUserBySubject(ctx, sub); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, util.UserVisibleError(codes.NotFound, "User not found")
+		}
+		return nil, status.Errorf(codes.Internal, "error getting user: %v", err)
+	}
+
+	// Determine target project.
+	entityCtx := engine.EntityFromContext(ctx)
```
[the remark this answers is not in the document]
JAORMX: done.

internal/controlplane/handlers_authz.go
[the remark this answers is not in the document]
JAORMX: Actually... we might not need that as the hierarchy is already checked in OpenFGA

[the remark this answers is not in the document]
evankanderson: Also, `targetProjectID` feels a little funny, as there should only be one `projectID` in this context.  There are two users (the _caller_ and the _subject_ user who's being granted permissions), so those names make sense.

internal/controlplane/handlers_authz.go
```diff
@@ -265,12 +267,88 @@ func (*Server) ListRoleAssignments(
...
+	targetProjectID := entityCtx.Project.ID
+
+	if err := s.authzClient.Delete(ctx, sub, authzrole, targetProjectID); err != nil {
+		return nil, status.Errorf(codes.Internal, "error writing role assignment: %v", err)
+	}
```
evankanderson, line 344: Is it possible for `Delete` to return a "Not Found"?  If so, do we want to handle it specially here?

proto/minder/v1/minder.proto
```diff
@@ -1511,6 +1511,6 @@ message RoleAssignment {
     string role = 1;
     // subject is the subject to which the role is assigned.
     string subject = 2;
-    // context is the context in which the role is assigned.
-    optional Context context = 3;
+    // project is the project in which the role is assigned.
+    optional string project = 3;
```
evankanderson, line 1515: Just use tag 4.  There's no prize for keeping consecutive tag numbers.

evankanderson (dismissed):

internal/controlplane/handlers_authz.go
JAORMX: no, I special cased that in the authz wrapper.

evankanderson (approved):

eleftherias (approved):

PR #2229, opened 2024-01-31
Implement CLI for permissions

JAORMX:
This implements role granting, denying and listing.


cmd/cli/app/project/role/role_deny.go
```diff
@@ -0,0 +1,72 @@
...
+
+var denyCmd = &cobra.Command{
+	Use:   "deny",
+	Short: "Deny a role to a subject on a project within the minder control plane",
+	Long: `The minder project role deny command removes a user from a role grant
+on a particular project.`,
+	RunE: cli.GRPCClientWrapRunE(DenyCommand),
+}
+
+// DenyCommand is the command for enrolling a provider
```
rdimitrov, line 37: Comment needs updating

cmd/cli/app/project/role/role_deny.go
```diff
@@ -0,0 +1,72 @@
...
+		RoleAssignment: &minderv1.RoleAssignment{
+			Role:    r,
+			Subject: sub,
+		},
+	})
+	if err != nil {
+		return cli.MessageAndError("Error denying role", err)
+	}
+
+	cmd.Println("Success.")
```
rdimitrov, line 62: Should we make the status messages for denying/granting different so people don't confuse them on accident, i.e. "access granted successfully"?

cmd/cli/app/project/role/role_deny.go
```diff
@@ -0,0 +1,72 @@
...
+	return nil
+}
+
+func init() {
+	RoleCmd.AddCommand(denyCmd)
+
+	denyCmd.Flags().StringP("sub", "s", "", "subject to grant access to")
+	denyCmd.Flags().StringP("role", "r", "", "the role to grant")
+	denyCmd.MarkFlagsOneRequired("sub", "role")
```
rdimitrov, line 71: Do want to have both of these present or it's one or the other?

cmd/cli/app/project/role/role_list.go
```diff
@@ -0,0 +1,94 @@
...
+	RunE: cli.GRPCClientWrapRunE(ListCommand),
+}
+
+// ListCommand is the command for listing roles
+func ListCommand(ctx context.Context, cmd *cobra.Command, conn *grpc.ClientConn) error {
+	client := minderv1.NewPermissionsServiceClient(conn)
+
+	project := viper.GetString("project")
+	format := viper.GetString("output")
+
```
rdimitrov, line 49:
```suggestion
	// Ensure the output format is supported
	if !app.IsOutputFormatSupported(format) {
		return cli.MessageAndError(fmt.Sprintf("Output format %s not supported", format), fmt.Errorf("invalid argument"))
	}

```

rdimitrov (commented):
Added a few comments, but all looks good! 👍

I especially like how much cleaner and consistent writing our CLI commands has become! 🚀

cmd/cli/app/project/role/role_deny.go
JAORMX: We need both to be present.

rdimitrov (approved):

PR #2246, opened 2024-02-01
Make openfga call in authz middleware

### LINKED ISSUE #1847: Implement FGA-based authz-middleware

eleftherias:
Fix #1847

This PR changes the middleware to delegate to OpenFGA for authorization decisions, such as "can user X view the repos in project Y?" and "can user X create a profile in project Y?".

On each user request, the middleware determines 1) The user making the request, 2) the resource being acted upon and 3) the relation / action that user is performing on the project.
The relation is derived from 2 fields in the RPC proto definition. The `target_resource`, which is the type of resource being acted upon, for example `TARGET_RESOURCE_PROJECT`. The `relation`, which is the action the user is performing on a project, for example `RELATION_REPO_GET`.

Once the user, project and relation is determined a call is made to OpenFGA to request authorization, for example "user:123 repo_get project:456".

This PR also removes the prior logic we used to make authorization decisions, where we checked the DB to see if the project belonged to the user.


JAORMX:
@eleftherias can you be a little bit more thorough with the description in the PR? I have the context but it would be good for other folks to be able to easily see what the change was.

internal/controlplane/handlers_authz.go
```diff
@@ -180,24 +88,38 @@ func ProjectAuthorizationInterceptor(ctx context.Context, req interface{}, _ *gr
...
 	}
 
+	relation := opts.GetRelation()
+
+	relationValue := relation.Descriptor().Values().ByNumber(relation.Number())
+	if relationValue == nil {
+		return nil, status.Errorf(codes.Internal, "error reading relation value %v", relation)
+	}
+	extension := proto.GetExtension(relationValue.Options(), minder.E_Name)
+	relationName, ok := extension.(string)
```
eleftherias, line 98: If anyone knows of a nicer way to do this let me know

JAORMX (approved):
The code lgtm and I tested this out:
* basic authorization works as expected.
* A user is not able to reach resources in other projects without authorization
* We get relevant error messages

SHIPIT! :ship:

PR #2250, opened 2024-02-01
Change the database migration CI workflow to check for the expected DB version instead

jhrozek:
Instead of checking for the expected tables which might come and go,
make sure that the version corresponds to the highest version in the
database migration files.

### DIFF
--- a/.github/workflows/compose-migrate.yml
+++ b/.github/workflows/compose-migrate.yml
@@ -34,5 +34,9 @@ jobs:
               docker logs minder_migrate_up
               exit 1
           fi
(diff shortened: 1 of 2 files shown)


rdimitrov (approved):

JAORMX (approved):

jhrozek:
let's hold merging this, I'll try if the CI fix works by dropping a table

jhrozek:
> let's hold merging this, I'll try if the CI fix works by dropping a table

actually nah, the local tests seem to work.

PR #2254, opened 2024-02-01
WIP: Make role grants add users to projects

JAORMX:
This adds the necessary code to ensure that a user belongs to a project
once it gets a role in it.

### DIFF
--- a/internal/authz/authz.go
+++ b/internal/authz/authz.go
@@ -417,7 +417,7 @@ func getUserFromTuple(user string) string {
 	return strings.TrimPrefix(user, "user:")
 }
 
-func getRoleAssignmentsFromReadResones(resp *fgaclient.ClientReadResponse, proj *string) []*minderv1.RoleAssignment {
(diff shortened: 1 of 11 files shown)


internal/authz/authz.go
```diff
@@ -388,3 +416,22 @@ func getProjectForTuple(project uuid.UUID) string {
 func getUserFromTuple(user string) string {
 	return strings.TrimPrefix(user, "user:")
 }
+
+func getRoleAssignmentsFromReadResones(resp *fgaclient.ClientReadResponse, proj *string) []*minderv1.RoleAssignment {
```
eleftherias, line 420: typo
```suggestion
func getRoleAssignmentsFromReadResponse(resp *fgaclient.ClientReadResponse, proj *string) []*minderv1.RoleAssignment {
```

eleftherias (commented):
Makes sense, just a small typo.
It is unfortunate that we need to keep this mapping in 2 places. Did you happen to look at what the impact would be if we remove the `user_projects` table and rely only on OpenFGA?

JAORMX:
@eleftherias I like the idea of relying more on OpenFGA to get projects for a user. I'll try it out in a separate PR.

PR #2271, opened 2024-02-04
Handle rate limiting errors for GitHub API

### LINKED ISSUE #2261: Handle rate limiting errors for GitHub API
### Describe the issue

Currently, we don't handle rate limit exceeded errors from the GitHub API.  From [GitHub API Docs](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit):

> Continuing to make requests while you are rate limited may result in the banning of your integration.

GitHub suggests [Best practices for using the REST API](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28), which mentions handling the rate limit exceeded errors.

Discord Thread: https://discord.com/channels/1184987096302239844/1185287949240242258/1201083912521252885

- [x] Add client caching to handle rate limiting (https://github.com/stacklok/minder/pull/2271)
- [ ] Update all endpoints to handle rate-limiting errors

### To Reproduce

_No response_

main

Vyom-Yadav:
Issue #2261 

* We employ an optimistic concurrency control model to minimize goroutine blocking and reduce rate limit errors, thus preventing integration bans

* Each server replica maintains its own set of rate-limited clients, eliminating the need for synchronization across replicas. If a rate limit error occurs, the client self-blocks and is added to a cache for future requests

* Requests that return rate limit errors are retried following a backoff period

Note:

The current watermill pub-sub implementation being used, i.e. SQL and go channels, are both a single channel with head-of-line blocking, which means that blocking in middleware/handler blocks all processing. So, currently, if a token is blocked, the entire processing will be blocked.

This will change when we shift to other pub-sub implementations and have multiple workers.

### DIFF
--- a/cmd/server/app/serve.go
+++ b/cmd/server/app/serve.go
@@ -37,6 +37,7 @@ import (
 	"github.com/stacklok/minder/internal/engine"
 	"github.com/stacklok/minder/internal/events"
 	"github.com/stacklok/minder/internal/logger"
+	"github.com/stacklok/minder/internal/providers/ratecache"
 	provtelemetry "github.com/stacklok/minder/internal/providers/telemetry"
 	"github.com/stacklok/minder/internal/reconcilers"
 )
@@ -121,11 +122,13 @@ var serveCmd = &cobra.Command{
 
 		serverMetrics := controlplane.NewMetrics()
 		providerMetrics := provtelemetry.NewProviderMetrics()
+		restClientCache := ratecache.NewRestClientCache(ctx)
 
 		s, err := controlplane.NewServer(
 			store, evt, serverMetrics, cfg, vldtr,
 			controlplane.WithProviderMetrics(providerMetrics),
 			controlplane.WithAuthzClient(authzc),
+			controlplane.WithRestClientCache(restClientCache),
 		)
 		if err != nil {
 			return fmt.Errorf("unable to create server: %w", err)
@@ -141,6 +144,7 @@ var serveCmd = &cobra.Command{
 			engine.WithProviderMetrics(providerMetrics),
 			engine.WithMiddleware(aggr.AggregateMiddleware),
 			engine.WithMiddleware(tsmdw.TelemetryStoreMiddleware),
+			engine.WithRestClientCache(restClientCache),
 		)
 		if err != nil {
 			return fmt.Errorf("unable to create executor: %w", err)
@@ -148,7 +152,9 @@ var serveCmd = &cobra.Command{
 
 		s.ConsumeEvents(exec)
 
-		rec, err := reconcilers.NewReconciler(store, evt, &cfg.Auth, reconcilers.WithProviderMetrics(providerMetrics))
+		rec, err := reconcilers.NewReconciler(store, evt, &cfg.Auth,
+			reconcilers.WithProviderMetrics(providerMetrics),
+			reconcilers.WithRestClientCache(restClientCache))
 		if err != nil {
 			return fmt.Errorf("unable to create reconciler: %w", err)
 		}
@@ -175,6 +181,9 @@ var serveCmd = &cobra.Command{
 
 		// Wait for all entity events to be executed
 		exec.Wait()
+
+		// Wait for the client cache background routine to finish
+		restClientCache.Wait()
 
 		return errg.Wait()
 	},
--- a/internal/providers/github/github.go
+++ b/internal/providers/github/github.go
@@ -18,14 +18,19 @@ package github
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 	"time"
 
+	backoffv4 "github.com/cenkalti/backoff/v4"
 	"github.com/google/go-github/v56/github"
+	"github.com/rs/zerolog"
 	"golang.org/x/oauth2"
 
 	"github.com/stacklok/minder/internal/db"
+	"github.com/stacklok/minder/internal/providers/ratecache"
 	"github.com/stacklok/minder/internal/providers/telemetry"
 	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 	provifv1 "github.com/stacklok/minder/pkg/providers/v1"
@@ -34,6 +39,12 @@ import (
 const (
 	// ExpensiveRestCallTimeout is the timeout for expensive REST calls
 	ExpensiveRestCallTimeout = 15 * time.Second
+	// MaxRateLimitWait is the maximum time to wait for a rate limit to reset
+	MaxRateLimitWait = 5 * time.Minute
+	// MaxRateLimitRetries is the maximum number of retries for rate limit errors after waiting
+	MaxRateLimitRetries = 1
+	// DefaultRateLimitWaitTime is the default time to wait for a rate limit to reset
+	DefaultRateLimitWaitTime = 1 * time.Minute
 )
 
 // Github is the string that represents the GitHub provider
@@ -52,6 +63,7 @@ type RestClient struct {
 	client *github.Client
 	token  string
 	owner  string
+	cache  ratecache.RestClientCache
 }
 
 // Ensure that the GitHub client implements the GitHub interface
@@ -62,9 +74,9 @@ var _ provifv1.GitHub = (*RestClient)(nil)
 // endpoint (as is the case with GitHub Enterprise), set the Endpoint field in
 // the GitHubConfig struct
 func NewRestClient(
-	ctx context.Context,
 	config *minderv1.GitHubProviderConfig,
 	metrics telemetry.HttpClientMetrics,
+	restClientCache ratecache.RestClientCache,
 	token string,
 	owner string,
 ) (*RestClient, error) {
@@ -73,7 +85,13 @@ func NewRestClient(
 	ts := oauth2.StaticTokenSource(
 		&oauth2.Token{AccessToken: token},
 	)
-	tc := oauth2.NewClient(ctx, ts)
+
+	tc := &http.Client{
+		Transport: &oauth2.Transport{
+			Base:   http.DefaultClient.Transport,
+			Source: oauth2.ReuseTokenSource(nil, ts),
+		},
+	}
 
 	tc.Transport, err = metrics.NewDurationRoundTripper(tc.Transport, db.ProviderTypeGithub)
 	if err != nil {
@@ -94,6 +112,7 @@ func NewRestClient(
 		client: ghClient,
 		token:  token,
 		owner:  owner,
+		cache:  restClientCache,
 	}, nil
 }
 
@@ -114,4 +133,137 @@ func ParseV1Config(rawCfg json.RawMessage) (*minderv1.GitHubProviderConfig, erro
 	}
 
 	return w.GitHub, nil
+}
+
+// setAsRateLimited adds the RestClient to the cache as rate limited.
+// An optimistic concurrency control mechanism is used to ensure that every request doesn't need
+// synchronization. RestClient only adds itself to the cache if it's not already there. It doesn't
+// remove itself from the cache when the rate limit is reset. This approach leverages the high
+// likelihood of the client or token being rate-limited again. By keeping the client in the cache,
+// we can reuse client's rateLimits map, which holds rate limits for different endpoints.
+// This reuse of cached rate limits helps avoid unnecessary GitHub API requests when the client
+// is rate-limited. Every cache entry has an expiration time, so the cache will eventually evict
+// the rate-limited client.
+func (c *RestClient) setAsRateLimited() {
+	if c.cache != nil {
+		c.cache.Set(c.owner, c.token, db.ProviderTypeGithub, c)
+	}
+}
+
+// waitForRateLimitReset waits for token wait limit to reset. Returns error if wait time is more
+// than MaxRateLimitWait or requests' context is cancelled.
+func (c *RestClient) waitForRateLimitReset(ctx context.Context, err error) error {
+	var rateLimitError *github.RateLimitError
+	isRateLimitErr := errors.As(err, &rateLimitError)
+
+	if isRateLimitErr {
+		return c.processPrimaryRateLimitErr(ctx, rateLimitError)
+	}
+
+	var abuseRateLimitError *github.AbuseRateLimitError
+	isAbuseRateLimitErr := errors.As(err, &abuseRateLimitError)
+
+	if isAbuseRateLimitErr {
+		return c.processAbuseRateLimitErr(ctx, abuseRateLimitError)
+	}
+
+	return nil
+}
+
+func (c *RestClient) processPrimaryRateLimitErr(ctx context.Context, err *github.RateLimitError) error {
+	logger := zerolog.Ctx(ctx)
+	rate := err.Rate
+	if rate.Remaining == 0 {
+		c.setAsRateLimited()
+
+		waitTime := DefaultRateLimitWaitTime
+		resetTime := rate.Reset.Time
+		if !resetTime.IsZero() {
+			waitTime = time.Until(resetTime)
+		}
+
+		logRateLimitError(logger, "RateLimitError", waitTime, c.owner, err.Response)
+
+		if waitTime > MaxRateLimitWait {
+			logger.Debug().Msgf("rate limit reset time: %v exceeds maximum wait time: %v", waitTime, MaxRateLimitWait)
+			return err
+		}
+
+		// Wait for the rate limit to reset
+		select {
+		case <-time.After(waitTime):
+			return nil
+		case <-ctx.Done():
+			logger.Debug().Err(ctx.Err()).Msg("context done while waiting for rate limit to reset")
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *RestClient) processAbuseRateLimitErr(ctx context.Context, err *github.AbuseRateLimitError) error {
+	logger := zerolog.Ctx(ctx)
+	c.setAsRateLimited()
+
+	retryAfter := err.RetryAfter
+	waitTime := DefaultRateLimitWaitTime
+	if retryAfter != nil && *retryAfter > 0 {
+		waitTime = *retryAfter
+	}
+
+	logRateLimitError(logger, "AbuseRateLimitError", waitTime, c.owner, err.Response)
+
+	if waitTime > MaxRateLimitWait {
+		logger.Debug().Msgf("abuse rate limit wait time: %v exceeds maximum wait time: %v", waitTime, MaxRateLimitWait)
+		return err
+	}
+
+	// Wait for the rate limit to reset
+	select {
+	case <-time.After(waitTime):
+		return nil
+	case <-ctx.Done():
+		logger.Debug().Err(ctx.Err()).Msg("context done while waiting for rate limit to reset")
+		return err
+	}
+}
+
+func logRateLimitError(logger *zerolog.Logger, errType string, waitTime time.Duration, owner string, resp *http.Response) {
+	var method, path string
+	if resp != nil && resp.Request != nil {
+		method = resp.Request.Method
+		path = resp.Request.URL.Path
+	}
+
+	event := logger.Debug().
+		Str("owner", owner).
+		Str("wait_time", waitTime.String()).
+		Str("error_type", errType)
+
+	if method != "" {
+		event = event.Str("method", method)
+	}
+
+	if path != "" {
+		event = event.Str("path", path)
+	}
+
+	event.Msg("rate limit exceeded")
+}
+
+func performWithRetry[T any](ctx context.Context, op backoffv4.OperationWithData[T]) (T, error) {
+	exponentialBackOff := backoffv4.NewExponentialBackOff()
+	maxRetriesBackoff := backoffv4.WithMaxRetries(exponentialBackOff, MaxRateLimitRetries)
+	return backoffv4.RetryWithData(op, backoffv4.WithContext(maxRetriesBackoff, ctx))
+}
+
+func isRateLimitError(err error) bool {
+	var rateLimitError *github.RateLimitError
+	isRateLimitErr := errors.As(err, &rateLimitError)
+
+	var abuseRateLimitError *github.AbuseRateLimitError
+	isAbuseRateLimitErr := errors.As(err, &abuseRateLimitError)
+
+	return isRateLimitErr || isAbuseRateLimitErr
 }
--- a/internal/providers/github/github_rest.go
+++ b/internal/providers/github/github_rest.go
@@ -25,6 +25,7 @@ import (
 	"net/url"
 	"strings"
 
+	backoffv4 "github.com/cenkalti/backoff/v4"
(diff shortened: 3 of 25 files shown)


internal/providers/github/github_rest.go
```diff
@@ -609,10 +610,30 @@ func (c *RestClient) ListPullRequests(
...
+
+		// If err is of type RateLimitError or AbuseRateLimitError, don't do permanent backoff
+		if isRateLimitError(err) {
+			return err
+		}
+
+		return backoffv4.Permanent(err)
+	}
+	return performWithRetry(ctx, op)
 }
```
Vyom-Yadav, line 637: `CreateComment` endpoint is modified as an example. If we agree on the implementation, other endpoints will be modified as well.

internal/providers/github/github.go
```diff
@@ -115,3 +126,78 @@ func ParseV1Config(rawCfg json.RawMessage) (*minderv1.GitHubProviderConfig, erro
...
+		return nil
+	}
+
+	select {
+	case <-blockedChan:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-time.After(MaxRateLimitWait):
+		return fmt.Errorf("rate limit wait exceeded for %s", c.GetOwner())
```
jhrozek, line 142: Do you think it would be a good idea to create a timer manually and cancel it the wait finishes sooner than the timer fires to deallocate it?
```go
timer := time.NewTimer(MaxRateLimitWait)
    defer func() {
        if !timer.Stop() {
            <-timer.C
        }
    }()
```

internal/providers/github/github.go
```diff
@@ -65,6 +74,7 @@ func NewRestClient(
 	ctx context.Context,
 	config *minderv1.GitHubProviderConfig,
 	metrics telemetry.HttpClientMetrics,
+	tokenStatStore *tokenstatus.Store,
```
jhrozek, line 77: reading the rest of this module sounds like `tokenstatus.Store` is now required. Is that intended? If yes, then should we error if the store is not provided?

Alternatively, I wonder if it would make sense to rather create a `WithTokenStatus` option and handle its absence in the REST code.

internal/tokenstatus/token_status.go
```diff
@@ -0,0 +1,70 @@
...
+package tokenstatus
+
+import (
+	"sync"
+
+	"github.com/stacklok/minder/internal/db"
+)
+
+// Store stores the rate limit status of token
+type Store struct {
```
jhrozek, line 25: I don't have a great use-case in mind, so this might be an over engineering suggestion, but since the members of store are not exported anyway, I wonder if creating an interface and exposing it would allow for easier testing.

jhrozek (commented):
I started the review, I haven't finished reading all of the PR, so I'm submitting the comments I had in the meantime and will resume the review later today.

Great work!

internal/tokenstatus/token_status.go
```diff
@@ -0,0 +1,70 @@
...
+	status map[key]chan struct{}
```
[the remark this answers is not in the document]
Vyom-Yadav: > Have you considered using `xsync.MapOf`? We already use it in minder, so you wouldn't drag in a new dependency and you wouldn't have to take care of the mutex yourself.

We use `v1.5.2` which is not supported now. and even`v3` had a bug fix a few months back:

> Fix too aggressive Map/MapOf shrinking on deletion leading to out of range panic

First of all, I'd suggest that we update the version as many bugs have been fixed since `v1.5.2`. Secondly, I'm open to using `xsync.MapOf` but I'm not too sure about the stability given that a recent version fixed a out of range panic. Wdyt?

internal/providers/github/github.go
Vyom-Yadav: Yeah, we should stop the timer with defer for better and faster gc. How about
```go
func (c *RestClient) waitIfRateLimited(ctx context.Context) error {
	blockedChan, ok := c.tokenStatStore.GetBlockingChan(c.token, db.ProviderTypeGithub)
	if !ok {
		return nil
	}

	timer := time.NewTimer(MaxRateLimitWait)
	defer timer.Stop()

	select {
	case <-blockedChan:
		return nil
	case <-ctx.Done():
		return ctx.Err()
	case <-timer.C:
		return fmt.Errorf("rate limit wait exceeded for %s", c.GetOwner())
	}
}
```

internal/providers/github/github.go
Vyom-Yadav: > reading the rest of this module sounds like tokenstatus.Store is now required. Is that intended?'

Yes, it is. I'll add a null check.

> Alternatively, I wonder if it would make sense to rather create a WithTokenStatus option and handle its absence in the REST code.

Why would it be absent? The only case where I think it may be absent is when we don't want to handle rate limit errors. Is this something we should support? For me, if we don't want to be blocked, we can pass a context with a short timeout (in case we are blocked by rate limit)

internal/tokenstatus/token_status.go
Vyom-Yadav: I am ok with doing so, but as you said, I also don't have a use case for that. As far as testing is concerned, I think the current (main) implementation is fine, and we wouldn't need any method, unless we want to do something similar to mocking, counting method calls, which I feel is not required.

internal/providers/github/github_rest.go
jhrozek: I like this pattern, but I wonder if we can make this more generic to avoid repeating the closure around all GH calls. Taking your `CreateComment` example into account, could we do something like:
```go
func (c *RestClient) CreateComment(ctx context.Context, owner, repo string, number int, comment string) error {
    createCommentWrapper := func(ctx context.Context) (*github.Response, error) {
        return c.client.Issues.CreateComment(ctx, owner, repo, number, &github.IssueComment{
            Body: &comment,
        })
    }

    return c.rateLimitHandler(ctx, createCommentWrapper)
}
```

and then extract the closure into a non-exported function:
```go
type ghClientWrapper func(ctx context.Context) (*github.Response, error)

func (c *RestClient) rateLimitHandler(ctx context.Context, wrapper ghClientWrapper) error {
    op := func() error {
        // Check if the token is currently rate limited, block the goroutine if it is
        err := c.waitIfRateLimited(ctx)
        if err != nil {
            return err
        }

        _, resp, err := apiCall(ctx)
        if err != nil {
            rateLimCheckErr := c.checkRateLimitResponse(ctx, resp.Response)
            if rateLimCheckErr != nil {
                return rateLimCheckErr
            }

            // If err is of type RateLimitError or AbuseRateLimitError, don't do permanent backoff
            if isRateLimitError(err) {
                return err
            }

            return backoffv4.Permanent(err)
        }

        return nil
    }

    return performWithRetry(ctx, op)
}
```

internal/providers/github/github_rest.go
```diff
@@ -623,3 +644,19 @@ func (c *RestClient) ListEmails(ctx context.Context, opts *github.ListOptions) (
...
+}
+
+func isRateLimitError(err error) bool {
+	var rateLimitError *github.RateLimitError
+	isRateLimitErr := errors.As(err, &rateLimitError)
+
+	var abuseRateLimitError *github.AbuseRateLimitError
+	isAbuseRateLimitErr := errors.As(err, &abuseRateLimitError)
+
+	return isRateLimitErr || isAbuseRateLimitErr
```
jhrozek, line 661: can we just do:
```go
func isRateLimitError(err error) bool {
    var rateLimitError *github.RateLimitError
    var abuseRateLimitError *github.AbuseRateLimitError
    return errors.As(err, &rateLimitError) || errors.As(err, &abuseRateLimitError)
}
```

jhrozek: although on second though, the extra boolean flags might be useful for extra debugging (I'll add another comment with better locality about that)

internal/providers/github/github.go
```diff
@@ -115,3 +126,78 @@ func ParseV1Config(rawCfg json.RawMessage) (*minderv1.GitHubProviderConfig, erro
...
+	// Optimistic concurrency is used to avoid the overhead of a mutex lock for every request.
+	err := c.waitIfRateLimited(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Check if rate limit headers are present
+	remaining, _ := strconv.Atoi(resp.Header.Get("x-ratelimit-remaining"))
+	resetTimeUnix, _ := strconv.ParseInt(resp.Header.Get("x-ratelimit-reset"), 10, 64)
+	retryAfter, _ := strconv.Atoi(resp.Header.Get("retry-after"))
```
jhrozek, line 165: Do you think it might be useful to log these values at a `Debug` level?

internal/providers/github/github.go
```diff
@@ -115,3 +126,78 @@ func ParseV1Config(rawCfg json.RawMessage) (*minderv1.GitHubProviderConfig, erro
...
+	}
+
+	if retryAfter > 0 {
+		waitTime := time.Duration(retryAfter) * time.Second
+
+		logger.Warn().Msgf("Secondary rate limit exceeded by %s. Waiting for %s", c.GetOwner(), waitTime)
+		return c.blockAndWaitWithTimeout(ctx, waitTime, MaxRateLimitWait)
+	}
+
+	logger.Warn().Msgf("Rate limit exceeded by %s. Waiting for 1 minute before retrying", c.GetOwner())
```
jhrozek, line 181: Could we extract `1*time.Minute` to a var or a const to make sure we don't forget updating the log message if we update the wait time?

internal/providers/github/github.go
```diff
@@ -115,3 +126,78 @@ func ParseV1Config(rawCfg json.RawMessage) (*minderv1.GitHubProviderConfig, erro
...
+		return fmt.Errorf("rate limit wait exceeded for %s", c.GetOwner())
+	}
+}
+
+func (c *RestClient) checkRateLimitResponse(ctx context.Context, resp *http.Response) error {
+	if resp.StatusCode != http.StatusForbidden && resp.StatusCode != http.StatusTooManyRequests {
+		return nil
+	}
+
+	logger := zerolog.Ctx(ctx)
```
jhrozek, line 151: could we extract the URL we tried to talk to to get a sense of what call was rate-limited and add that as a `Str` for the logger?

jhrozek (commented):
Thank you I went through the rest of the code and left some comments. Mostly they are nits and I think this is great work.

internal/tokenstatus/token_status.go
jhrozek: OK, let's leave it like this.

internal/tokenstatus/token_status.go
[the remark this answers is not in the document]
jhrozek: I don't have a great answer, I'm even surprised we were using such an old version. I don't see an issue with your implementation, I think the way you used `Lock`/`Rlock` etc is already correct and given that these are used in network traffic I'm not too concerned about lock congestion being a bottleneck.

So let's leave the code as-is and let me see about upgrading `xsync.MapOf`

internal/providers/github/github.go
jhrozek: This looks good!

internal/providers/github/github.go
jhrozek: To be honest I was only wondering if we could decrease the number of places where we pass the `tokenstatus.Store` by using something like the `WithTokenStatus` option and then just have a passthrough implementation as a default opt to avoid increasing the number of required parameters across the codebase. There's no functional reason.

internal/tokenstatus/token_status.go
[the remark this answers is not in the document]
Vyom-Yadav: I wonder if dependabot catches updates when a newer version is released i.e. `yourcoolpackage/v3`.

internal/providers/github/github.go
```diff
@@ -115,3 +126,78 @@ func ParseV1Config(rawCfg json.RawMessage) (*minderv1.GitHubProviderConfig, erro
...
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-time.After(MaxRateLimitWait):
+		return fmt.Errorf("rate limit wait exceeded for %s", c.GetOwner())
+	}
+}
+
+func (c *RestClient) checkRateLimitResponse(ctx context.Context, resp *http.Response) error {
+	if resp.StatusCode != http.StatusForbidden && resp.StatusCode != http.StatusTooManyRequests {
```
jhrozek, line 147: Are there cases where `resp` might be `nil` on a fatal error in which case we'd have to just rely on the error rather then the reply? (I noticed that there's no `err` check between calling `c.client.Issues.CreateComment` and dereferencing `resp`)

```
	_, resp, err := c.client.Issues.CreateComment(ctx, owner, repo, number, &github.IssueComment{
			Body: &comment,
		})

		rateLimCheckErr := c.checkRateLimitResponse(ctx, resp.Response)
		if rateLimCheckErr != nil {
			return rateLimCheckErr
		}
```

Vyom-Yadav:
I just noticed, that the go-github library internally also does something when it encounters rate-limiting errors (It starts sending mock responses stating you've been rate-limited). 

Let me check that out, and after that I'll address the review items.

Vyom-Yadav:
@jhrozek So apparently, `go-github` comprehensively maintains rate limits for various GH endpoints and resources internally once the particular client hits a rate limit. 

E.g., If a req to create a comment hits the rate limit, then it would add that rate limit to that particular category store.  https://github.com/google/go-github/blob/975b2e55cec1dd1a438ed15ec605cc46113506ab/github/github.go#L1323-L1324

`go-github` will not make new requests internally until the timer expires (It does not use a timer, instead till a timestamp). The caller would get rate limiting errors from the library internally instead of GitHub i.e. it won't create repeated requests. This is a really formidable approach.

Here is some code from the lib to make more sense (https://github.com/google/go-github/blob/975b2e55cec1dd1a438ed15ec605cc46113506ab/github/github.go#L936-L990):
```go
// checkRateLimitBeforeDo does not make any network calls, but uses existing knowledge from
// current client state in order to quickly check if *RateLimitError can be immediately returned
// from Client.Do, and if so, returns it so that Client.Do can skip making a network API call unnecessarily.
// Otherwise it returns nil, and Client.Do should proceed normally.
func (c *Client) checkRateLimitBeforeDo(req *http.Request, rateLimitCategory rateLimitCategory) *RateLimitError {
	c.rateMu.Lock()
	rate := c.rateLimits[rateLimitCategory]
	c.rateMu.Unlock()
	if !rate.Reset.Time.IsZero() && rate.Remaining == 0 && time.Now().Before(rate.Reset.Time) {
		// Create a fake response.
		resp := &http.Response{
			Status:     http.StatusText(http.StatusForbidden),
			StatusCode: http.StatusForbidden,
			Request:    req,
			Header:     make(http.Header),
			Body:       io.NopCloser(strings.NewReader("")),
		}
		return &RateLimitError{
			Rate:     rate,
			Response: resp,
			Message:  fmt.Sprintf("API rate limit of %v still exceeded until %v, not making remote request.", rate.Limit, rate.Reset.Time),
		}
	}

	return nil
}

// checkSecondaryRateLimitBeforeDo does not make any network calls, but uses existing knowledge from
// current client state in order to quickly check if *AbuseRateLimitError can be immediately returned
// from Client.Do, and if so, returns it so that Client.Do can skip making a network API call unnecessarily.
// Otherwise it returns nil, and Client.Do should proceed normally.
func (c *Client) checkSecondaryRateLimitBeforeDo(req *http.Request) *AbuseRateLimitError {
	c.rateMu.Lock()
	secondary := c.secondaryRateLimitReset
	c.rateMu.Unlock()
	if !secondary.IsZero() && time.Now().Before(secondary) {
		// Create a fake response.
		resp := &http.Response{
			Status:     http.StatusText(http.StatusForbidden),
			StatusCode: http.StatusForbidden,
			Request:    req,
			Header:     make(http.Header),
			Body:       io.NopCloser(strings.NewReader("")),
		}

		retryAfter := time.Until(secondary)
		return &AbuseRateLimitError{
			Response:   resp,
			Message:    fmt.Sprintf("API secondary rate limit exceeded until %v, not making remote request.", secondary),
			RetryAfter: &retryAfter,
		}
	}

	return nil
}
```
It internally maintains when the client will be ready to make new requests. 

---

These were the features. Now the problem is that we create a new client for every request, making all that code useless. The approach we should be really following is caching the client and checking the cache for a client if it already exists. Wdyt about this approach? 

Also, if we add the client to the cache, we may want to use something more sophisticated than a map with Mutex. We cannot have a constantly growing cache. We would have to remove no longer valid tokens to avoid memory going boom. (Maybe a periodic job within the server to invalidate the cache)

Vyom-Yadav:
We pass in context whenever we initialize the GitHub client, so I don't think caching the client with context is possible.  Instead, we can follow the approach of not passing `ctx` while creating the client. I have not checked all usages, but we are probably not passing `ctx` with a `httpClient` as a value. 

`ctx` is only used to get `ctx` client:
```go
func ContextClient(ctx context.Context) *http.Client {
	if ctx != nil {
		if hc, ok := ctx.Value(HTTPClient).(*http.Client); ok {
			return hc
		}
	}
	if appengineClientHook != nil {
		return appengineClientHook(ctx)
	}
	return http.DefaultClient
}
``` 
If we don't use Google app engine, we get the `DefaultClient`, so I think we can safely remove `ctx` while creating a new rest client. For context cancellation, `ctx` passed to rest API calls would perfectly do that task.

Vyom-Yadav:
Other than that, I don't see any problem caching a `*http.Client`, wdyt?

Vyom-Yadav:
After doing some research, I think we can do this. Istio also caches the HTTP client to reuse the connection in the future. https://github.com/istio/istio/blob/7bd27e8a432bd700f754e54593dc584d4973d376/pilot/cmd/pilot-agent/status/server.go#L266-L266

jhrozek:
Thank you for doing all this research @Vyom-Yadav. Where/how were you thinking about caching the client, in the tokenStatusStore?

One thing that sounds concerning to me is that if we were to cache the http clients globally as opposed to e.g. for a single evaluation of an entity, the number of open connections would grow in a situation like stacklok's SaaS where a single minder instance can server many users and we could exhaust the number of fds available or just grow the memory a lot.

Vyom-Yadav:
> Where/how were you thinking about caching the client, in the tokenStatusStore?

Yes, I was thinking of storing these clients in memory only. A bit different from tokenStatusStore, but yes, the same concept.

> One thing that sounds concerning to me is that if we were to cache the http clients globally as opposed to e.g. for a single evaluation of an entity, the number of open connections would grow in a situation like stacklok's SaaS where a single minder instance can server many users and we could exhaust the number of fds available or just grow the memory a lot.

That's a good concern, but we can do many things to get around it. 
- For instance, we could only cache rate limited clients instead of caching all clients, if a client gets rate limited cache it, and query cache before providing new clients to calling functions. 

- Secondly, we can have a max-age (1 hour looks good as tokens are hourly limited), and automatically prune them from the cache after they expire. We can have periodic expired clients removals (lazy), or we can start a timer for every cache entry and send a cache the deletion signal (eager).

- We can have size based eviction mechanism where if the size of cache reaches a certain threshold, we prune unused clients (user deleted) or even randomly delete clients to get below a certain threshold.

All these optimization should prevent us from exhausting server resources. Sounds good?

Vyom-Yadav:
@jhrozek I implemented a pattern I think should work without hitting any bottleneck; please review it once. Thanks!

jhrozek:
> @jhrozek I implemented a pattern I think should work without hitting any bottleneck; please review it once. Thanks!

Apologies for the delay. Last week, the whole Stacklok company was meeting in-person and the bandwidth for development was limited. I am looking at the code now.

internal/providers/ratecache/store.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (r *restClientCache) evictExpiredEntriesRoutine() {
+	timer := time.NewTimer(r.evictionTime)
+	defer timer.Stop()
+	for {
+		<-timer.C
+		r.evictExpiredEntries()
+
+		// Reset the timer for the next expiration
+		timer.Reset(r.evictionTime)
```
jhrozek, line 100: I wonder if ticking once per eviction time is not too seldom. If a an entry is to be evicted just before a goroutine tick, this might mean that the entry will be in the cache for 30 more minutes. Do you think it would be useful to have the goroutine act more frequently, maybe twice per eviction time?

internal/providers/github/github.go
```diff
@@ -115,3 +134,128 @@ func ParseV1Config(rawCfg json.RawMessage) (*minderv1.GitHubProviderConfig, erro
...
+		method = resp.Request.Method
+		path = resp.Request.URL.Path
+	}
+
+	logger.Debug().
+		Str("owner", owner).
+		Str("method", method).
+		Str("path", path).
+		Str("wait_time", waitTime.String()).
+		Msg("rate limit exceeded")
```
jhrozek, line 244: Do you think it might make sense to also log the two different rate limits (abuse vs primary) using a keyword (e.g. pass another parameter like "type" and set it one of two constant values)?

jhrozek (approved):
this is really nice code, great work.

I left a couple of comments, feel free to make another iteration or a follow up PR or reply if you disagree, but none of my comments are really blocking.

internal/providers/ratecache/store.go
Vyom-Yadav: Yes, it would be better to run the expired check using half the interval. Done.

internal/providers/github/github.go
Vyom-Yadav: Yes, that would be better. Done.

Vyom-Yadav:
@jhrozek When we use `defer ticker.Stop()`, it didn't free the underlying resources as the goroutine would have never got a chance to return, leading to a resource leak. I implemented a pattern with proper wait and context cancellation. Please check the new pattern.

jhrozek (approved):
LGTM!

cmd/server/app/serve.go
```diff
@@ -176,6 +182,9 @@ var serveCmd = &cobra.Command{
 		// Wait for all entity events to be executed
 		exec.Wait()
 
+		// Wait for the client cache background routine to finish
+		restClientCache.Wait()
+
```
Vyom-Yadav, line 187: @jhrozek Question: According to this logic, if the go routines above finish their execution (somehow, though it shouldn't happen for a server), this would mean the context has not been cancelled, and this would lead to an infinite wait as the ticker only stops using defer and context cancellation. 

On the other hand, the go routines started above should keep running as it's a server binary. In case of panic, the whole process would be stopped. Just double-checking, whether this is correct or not :)

JAORMX (approved):

Vyom-Yadav:
@JAORMX I was addressing https://github.com/stacklok/minder/pull/2271#discussion_r1495557451 :)
I think that is incorrect `Wait()`. We want to free the ticker resources, so it would be better to do something like:
```go
func (r *restClientCache) Close() {
	r.evictionTicker.Stop()
}
```

JAORMX:
@Vyom-Yadav let's do it in a subsequent PR. Thanks for all this work!

PR #2274, opened 2024-02-05
Add wait to ensure all entity events are executed

Vyom-Yadav:
From: https://discord.com/channels/1184987096302239844/1185287949240242258/1202878694201237515

The wait group counter was being incremented and decremented, but the server wasn't waiting for all entity event executions to finish. This change ensures that the server waits for all entity events to be executed before exiting.


Vyom-Yadav:
cc @JAORMX

JAORMX (approved):

PR #2292, opened 2024-02-07
Migrate to github.com/puzpuzpuz/xsync/v3

jhrozek:
We were using an old release of `xsync` from 2022 which hasn't been
receiving updates through dependabot because of the `v3` module bump.


stacklokbot (commented):

rdimitrov (approved):

JAORMX (approved):

PR #2299, opened 2024-02-07
Verify all simple signing layers, do not filter events for .sig, enable OIDC identities

rdimitrov:
The following PR updates:
* We now collect and try to create a bundle for all simple signing layers present in the artifact's manifest. This is the use case when we search for signatures in the OCI registry.
* We no longer filter webhook events for `.sig` versions
* Renamed the `workflow_name` parameter we expose to the rule type to something more generic like a `signer_identity` and it nows include the value of the SAN parameter (for example this now allows for users to set the email of the identity that  is expected to sign the artifact through Fulcio)

Fixes: #2298

### DIFF
--- a/internal/engine/ingester/artifact/artifact.go
+++ b/internal/engine/ingester/artifact/artifact.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	"github.com/rs/zerolog"
+	"github.com/sigstore/sigstore-go/pkg/verify"
 	"google.golang.org/protobuf/proto"
 
 	evalerrors "github.com/stacklok/minder/internal/engine/errors"
@@ -56,7 +57,7 @@ type verification struct {
 	IsVerified        bool   `json:"is_verified"`
 	Repository        string `json:"repository"`
 	Branch            string `json:"branch"`
-	WorkflowName      string `json:"workflow_name"`
+	SignerIdentity    string `json:"signer_identity"`
 	RunnerEnvironment string `json:"runner_environment"`
 	CertIssuer        string `json:"cert_issuer"`
 }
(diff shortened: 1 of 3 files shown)


internal/engine/ingester/artifact/artifact.go
```diff
@@ -56,7 +57,7 @@ type verification struct {
 	IsVerified        bool   `json:"is_verified"`
 	Repository        string `json:"repository"`
 	Branch            string `json:"branch"`
-	WorkflowName      string `json:"workflow_name"`
+	SignerIdentity    string `json:"signer_identity"`
```
jhrozek, line 60: In case someone wonders if it's safe to rename these parameters - yes, they haven't made their way outside staging yet.

internal/verifier/sigstore/container/container.go
```diff
@@ -334,61 +338,75 @@ func getDigestFromVersion(version string) ([]byte, error) {
 }
 
 // bundleFromOCIImage returns a ProtobufBundle based on OCI image reference.
-func bundleFromOCIImage(imageRef string, auth githubAuthenticator) ([]sigstoreBundle, error) {
-	// 1. Get the signature manifest from the OCI image reference
+func bundleFromOCIImage(ctx context.Context,
```
jhrozek, line 341: If you end up touching the code again, we might want to rename the function to `bundlesFromOCIImage` (plural)

jhrozek (approved):

internal/verifier/sigstore/container/container.go
rdimitrov: ah, you're right, same goes for the other one too 😃

PR #2301, opened 2024-02-07
Display name of profile when running `profile create`

### LINKED ISSUE #2282: Print the name of the newly created profile
It's slightly irritating to not see the name of a profile that was just created. A typical workflow is "create a profile and then check its status", but if you create a profile:
```
./bin/minder profile create -f /Users/jakub/devel/playground/good-repo-go-provenance.yaml
WARNING: This build uses a test environment and may not be stable
+----------+--------------------+-------------------------------------+----------------------------------------------------------+
|  ENTITY  |        RULE        |             RULE PARAMS             |                     RULE DEFINITION                      |
+----------+--------------------+-------------------------------------+----------------------------------------------------------+
| artifact | artifact_signature | name: good-repo-go                  | branch: main                                             |
|          |                    | sigstore: tuf-repo-cdn.sigstore.dev | cert_issuer: https://token.actions.githubusercontent.com |
|          |                    | tags:                               | is_signed: true                                          |
|          |                    | - latest                            | is_verified: true                                        |
|          |                    |                                     | repository: https://github.com/jakubtestorg/good-repo-go |
|          |                    |                                     | runner_environment: github-hosted                        |
|          |                    |                                     | workflow_name: build-image-signed-ghat.yml               |
|          |                    |                                     |                                                          |
+----------+--------------------+-------------------------------------+----------------------------------------------------------+
```

then you need to run `profile list` to see what profiles are available, pick the right one and pass it to `profile status`.

It would be much nicer if the profile create command printed the name of the newly created profile.

UX-wise it would be nice if we could print a row that would span above (across?) the rest of the rows that we already print, but I don't know if our table library allows that. Printing the name as an additional column might be OK, too or even a separate line.

This should all be printed only with `-otable`.

dmjb:
Fixes #2282

When `profile create` runs successfully, the name of the new profile will now be displayed above the table of rule information.

Alternatives Considered:

* Add profile name to each row of the table:
  - The profile name will get repeated for each rule in the table, and this does not look great.
  - Adding in the profile name makes the table wide, and render poorly on smaller terminals.
* Listing ID alongside name:
  - After talking to @jhrozek, I learned that we are de-emphasizing the profile ID in the CLI in future.

Update: Have also made the same change to the `profile apply` command


cmd/cli/app/profile/table_render.go
```diff
@@ -81,38 +81,33 @@ func NewProfileTable() table.Table {
...
-	renderEntityRuleSets(minderv1.ArtifactEntity, p.Artifact, t)
+	renderProfileRow(minderv1.ArtifactEntity, p.Artifact, t)
 
 	// pull request
-	renderEntityRuleSets(minderv1.PullRequestEntity, p.PullRequest, t)
+	renderProfileRow(minderv1.PullRequestEntity, p.PullRequest, t)
 }
 
-func renderEntityRuleSets(entType minderv1.EntityType, rs []*minderv1.Profile_Rule, t table.Table) {
+func renderProfileRow(entType minderv1.EntityType, rs []*minderv1.Profile_Rule, t table.Table) {
```
dmjb, line 96: Some refactoring changes:

1) I renamed the function. I understand the logic of the old name, but I feel like the new name is a little more explicit as to what the is doing
2) I collapsed two small functions into one: the `renderRuleTable` function was only ever used by `renderEntityRuleSets`, and I feel like combining them together makes the code easier to follow.

Open to feedback/comments on this one.

rdimitrov (commented):
I few small nits, but overall looks good! 🚀 👍 
* @jhrozek - Perhaps we should have this for `profile apply` too? `apply` effectively is creating a profile first and if it gets `already exists`, then tries to update it.
* What do you think about updating `ExecOnOneProfile` to return the profile object alongside the error and use it directly to print its name?

dmjb:
@rdimitrov Thank you for the feedback. I have implemented both changes.

rdimitrov (approved):
Awesome! 🚀 Looks great! 🙏

PR #2310, opened 2024-02-08
Add provider listing functionality

JAORMX:
This gets us started with the Providers API by adding a listing
functionality. This allows us to list all providers within a given
project.

Since we're ont allowing folks to create their own providers yet, this
will only show the default one we pre-populate for them.

The intention is to slowly start to build the given functionality we
need on top of this.


eleftherias (approved):
Looks good, just needs a rebase

PR #2327, opened 2024-02-09
Update CLI table outputs for ruletype and profile

rdimitrov:
The following PR:
* Removes the use of emojis in profile and remediation statuses - they were nice, but they always messed up the table rendering 
image

* Moved the Guidance column from the detailed `profile status list` table and put it in the `ruletype get` output
* Added a Skipped text for remediations that skipped, prior to that it was empty looking and it looked like there could have been an error
* Added a function that concatenates and wraps a string. Used that in the `ruletype create/apply` output so it renders on a smaller footprint. 
* Shortened some of the headers too because of that.


jhrozek (approved):
I will miss the unicode blink, but these tables look much better (no annoying wrapping) on my laptop screen.

PR #2330, opened 2024-02-10
Add project list RPC (no client changes)

evankanderson:
Since permissions are associated with projects, now that we have a ProjectsService, it seemed like it _might_ make sense to move the permissions endpoints under projects.  Unfortuantely, the shape of the `Context` object (with an `optional project`) means that we can't interpolate the project into the URL path without doing more work.  I'd love to get opinions on those endpoints as well.

I included a simple test and also tested manually with `grpcurl`


proto/minder/v1/minder.proto
```diff
@@ -1453,6 +1516,14 @@ message Profile {
     string version = 11;
 }
 
+message ListProjectsRequest {
+
```
JAORMX, line 1520: Should this add the context? In case the user wants to list a sub tree?

Issue #2314, opened 2024-02-09
Fix Trusty support for Go


rdimitrov:
### Describe the issue

Enabling the support for Go in Trusty is not working out of the box in Minder, so we have to fix the issues - goParser needs to be updated, etc.

PR #2332, opened 2024-02-10
Fix a few Trusty issues that enable support for Go in Minder

rdimitrov:
The following PR:
* Updates the go parser extracting the new packages from the `go.mod` diff
* Updates the go parser unit tests, it seemed previously they were addressing `go.sum` patch diffs
* Fixes the Trusty hyperlink for the alternative packages and now shows `No alternative found` when there's no package alternative found.
* Renames the `Ecosystem` column to `Pkg` so there's no horizontal scrolling because of this
* Fixes a bug in Minder which was not rendering to the user low-score packages that had no alternatives. 

Fixes: #2314 

**Before:**
image
image
**After:**
image

### DIFF
--- a/internal/engine/ingester/diff/parse.go
+++ b/internal/engine/ingester/diff/parse.go
@@ -19,6 +19,7 @@ import (
 	"bufio"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strings"
 
 	"github.com/stacklok/minder/internal/util"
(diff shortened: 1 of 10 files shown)


JAORMX (approved):

internal/engine/ingester/diff/parse.go
```diff
@@ -137,31 +138,52 @@ func pyNormalizeName(pkgName string) string {
 }
 
 func goParse(patch string) ([]*pb.Dependency, error) {
```
dmjb, line 140: General comment: in future, we may want to consider splitting each of these parsers (and their tests) out into separate files. This file is going to keep growing and growing otherwise.

rdimitrov: Yeah, now that you mention it the way I see it is number of supported ecosystems == number of parsers, but I agree, let's see how this grows and we'll revisit 👍

PR #2344, opened 2024-02-13
Revert "Enable Coveralls for coverage reporting"

evankanderson:
Reverts stacklok/minder#2342

The YAML here is not valid (wrong indentation), and this causes all testing to break.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,29 +27,6 @@ jobs:
       - name: Run tests
         run: make test-silent
 
-    cover:
-      name: Coverage
-      permissions:
-        contents: read
-      runs-on: ubuntu-latest
-      steps:
-        - name: Checkout
-          uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
-        - name: Set up Go
-          uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5
-          with:
-            go-version-file: 'go.mod'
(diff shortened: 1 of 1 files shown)


JAORMX (approved):

Vyom-Yadav:
@evankanderson @JAORMX Just a suggestion. I would pick mutation testing over simple coverage testing. Most mutation-introducing frameworks check code coverage too (with mutations as an addition). I haven't used any go mutation testing frameworks but I extensively worked with [pitest java](https://pitest.org/) in one of the projects, and it really surpasses checking for line coverage (logic coverage > line coverage).

evankanderson:
> @evankanderson @JAORMX Just a suggestion. I would pick mutation testing over simple coverage testing. Most mutation-introducing frameworks check code coverage too (with mutations as an addition). I haven't used any go mutation testing frameworks but I extensively worked with [pitest java](https://pitest.org/) in one of the projects, and it really surpasses checking for line coverage (logic coverage > line coverage).

I hear you.  😁

This is about having a tool to _track_ coverage over time, rather than a tool to compute coverage.

For computing coverage for go, `go cover` is the standard tool which we already have hooked up.  We can add more tools if they produce a standard coverage file, and coveralls should be able to report on the combined coverage result.

Vyom-Yadav:
Cool, I have seen coveralls but never configured it myself. With mutation testing, you should get it as early as possible. Otherwise, you may end up with a pile of technical debt that consumes a lot of time.

PR #2343, opened 2024-02-13
Add test in CLI directory

evankanderson:
Test that https://github.com/stacklok/minder/pull/2342 worked for adding coverage data.

PR #2351, opened 2024-02-14
Re-enable coverage reporting to coveralls

evankanderson:
See https://coveralls.io/builds/65695104 from https://github.com/stacklok/minder/pull/2343 for an example

### DIFF
--- a/.mk/test.mk
+++ b/.mk/test.mk
@@ -26,8 +26,8 @@ test-silent: clean init-examples ## run tests in a silent mode (errors only outp
 	go test -json -race -v ./... | gotestfmt -hide "all"
 
 .PHONY: cover
-cover: ## display test coverage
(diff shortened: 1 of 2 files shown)


.mk/test.mk
```diff
@@ -26,8 +26,8 @@ test-silent: clean init-examples ## run tests in a silent mode (errors only outp
 	go test -json -race -v ./... | gotestfmt -hide "all"
 
 .PHONY: cover
-cover: ## display test coverage
-	go test -v -race $(shell go list ./... | grep -v /vendor/) -v -coverprofile=coverage.out
+cover: init-examples ## display test coverage
```
JAORMX, line 29: Is this to be kept?

evankanderson: We don't have a `/vendor/` directory, so it does nothing.

Note that we don't exclude this for the `test` target.

Yeah, I slipped in some cleanup.

JAORMX (approved):

PR #2358, opened 2024-02-15
Fix TUF root init to bump to sigstore-go to v0.2.0

puerco:
This PR updates the sigstore init code to address a breaking change in the sisgstore-go API.

Supersedes #2348 

 Signed-off-by: Adolfo Garcia Veytia (puerco) <puerco@stacklok.com>


internal/verifier/sigstore/sigstore.go
```diff
@@ -70,12 +70,11 @@ func New(trustedRoot string, authOpts ...container.AuthMethod) (*Sigstore, error
 	}
 
 	// init sigstore's verifier
-	trustedrootJSON, err := tuf.GetTrustedrootJSON(trustedRoot, cacheDir)
-	if err != nil {
-		return nil, err
-	}
+	tufOpts := tuf.DefaultOptions()
+	tufOpts.Root = []byte(trustedRoot)
```
jhrozek, line 74: I don't think this works (but at the same time the variable is misnamed and our tests don't catch this).

our trusted root is just the name of the sigstore instance (e.g. `"tuf-repo-cdn.sigstore.dev`). If you want to pass the TUF root bytes, we can instead just pass in what `embeddedRootJson` returns or set `RepositoryBaseURL` to a URL of the sigstore instance, maybe defaulting to `https:// + trustedRoot`

jhrozek (changes requested):

internal/verifier/sigstore/sigstore.go
puerco: OK, I factored out the root init and added tests. Also, I added a `readRootJson()` utility func to handle looking up cached root data (to pair it with the existing `writeRootJson`).

PTAL @jhrozek  :)

internal/verifier/sigstore/sigstore.go
```diff
@@ -52,50 +49,85 @@ const (
...
+	tufOpts := tuf.DefaultOptions()
+	tufOpts.DisableLocalCache = true
+
+	// Set the repository base URL, fix the scheme if not provided
+	tufURL, err := url.Parse(sigstoreTUFRepoURL)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error parsing sigstore TUF repo URL: %w", err)
+	}
+	if tufURL.Scheme == "" {
```
jhrozek, line 114: Do we want to even allow a scheme that is not https?

jhrozek (commented):
Left one question inline, but that can be addressed in a follow-up.

More interestingly, I can't seem to verify a signature that is minted by using the PGI using my personal google account which is something that @rdimitrov mentioned earlier today and I'm unsure if this is a bug in sigstore-go.

what I'm seeing in the logs is:
```
{"level":"error","exception.message":"when verifying a certificate identity, there must be subject alternative name criteria","Timestamp":1708634646262101000,"message":"error getting certificate identity"}
```
so I'm not going to ack now to avoid breaking master.

That said, the code looks really good to me. Thank you for the refactor and adding the tests, that's something we should have done some time ago.

jhrozek (approved):
The part where we sign with a PGI sigstore needs to be addressed in a separate patch due to changes in sigstore-go.

rdimitrov (approved):
Go Go sigstore! 🚀

PR #2365, opened 2024-02-19
Use latest golangci-lint in CI

JAORMX:
Instead of having to manually maintain the golangci-lint version in our
job, let's instead default to the latest one. This will get us the
latest and greatest CI checks provided by the tool and reduce the
maintenance burden of manual upgrades to this file. Unfortunately, this
is something that dependabot does not support updating.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,8 +17,6 @@ jobs:
           cache: false
       - name: golangci-lint
         uses: golangci/golangci-lint-action@3cfe3a4abbb849e10058ce4af15d205b6da42804 # v3
-        with:
-          version: v1.56.1  
 
   mod-tidy:
     name: Go mod tidy


jhrozek:
Interesting that just bumping the CI version decreases the coveralls coverage? Maybe just the PR needs rebasing?

jhrozek:
> Interesting that just bumping the CI version decreases the coveralls coverage? Maybe just the PR needs rebasing?

Probably a coveralls fart. We should change the blocking condition to block only if the decrease is at least 2% or such CC @evankanderson

jhrozek (approved):

evankanderson:
https://docs.coveralls.io/build-types#common-issue-drifted-builds--two-dot-vs-three-dot-diff-comparisons-in-coverage-reports

It seems like there are possibly some issues comparing builds with the appropriate base.  In any case, we should probably only fail if the coverage gap is at least a certain absolute size (e.g. 50 LoC)

PR #2367, opened 2024-02-19
Remove `internal/smoke` package

JAORMX:
These smoke tests will be covered elsewhere.


jhrozek (approved):
I don't have an issue with removing the test, but could you write to the PR opening or as a comment where would these tests live eventually?

JAORMX:
@jhrozek I don't think that has been decided just yet. To some extent I rather have the smoke tests be part of a separate repo, but that decision is yet to be made.

PR #2372, opened 2024-02-19
Drop the artifact_versions table

jhrozek:
The artifact versions are no longer used and the code is also pushed to
the prod SaaS. Let's just remove the table and the associated API.

Signed-off-by: Radoslav Dimitrov <radoslav@stacklok.com>


JAORMX (approved):
die artifact_versions! die!

PR #2375, opened 2024-02-19
Add test coverage for `internal/eea` `FlushAll`

JAORMX:
This adds basic tests for the happy path of `FlushAll`

### DIFF
--- a/internal/eea/eea_test.go
+++ b/internal/eea/eea_test.go
@@ -24,10 +24,12 @@ import (
 	"time"
 
 	"github.com/ThreeDotsLabs/watermill/message"
+	"github.com/golang/mock/gomock"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	mockdb "github.com/stacklok/minder/database/mock"
 	serverconfig "github.com/stacklok/minder/internal/config/server"
 	"github.com/stacklok/minder/internal/db"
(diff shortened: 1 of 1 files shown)


internal/eea/eea_test.go
```diff
@@ -169,6 +171,194 @@ func createNeededEntities(ctx context.Context, t *testing.T) (projID uuid.UUID,
...
 
+func TestFlushAll(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name        string
+		mockDBSetup func(context.Context, *mockdb.MockStore)
+	}{
+		{
+			name: "flushes one repo",
```
dmjb, line 182: I don't want to block this PR, but it seems like we should have some more tests about the negative scenarios. For example, if I take a look at the code for `FlushAll` I think we should cover scenarios such as:

1) Asserting that calling the method when the cache is empty returns no error.
2) Returns an error when `GetRepositoryByID` fails
3) Does nothing when `GetRepositoryByID` cannot find a project for the repository.
4) Does nothing when `buildEntityWrapper` cannot find any matching data in the DB

JAORMX: @dmjb was planning to submit that in a separate PR as to not overwhelm the reader. But yeah, I agree we need negative tests as well.

evankanderson (approved):
I'm not thrilled by how verbose the database mocks are, but not going to block this on that but.

dmjb (approved):

PR #2377, opened 2024-02-19
When the `context` of a request is nil, return InvalidArgument/400

### LINKED ISSUE #2376: If no Context is provided, minder returns a 500, 400 might be more appropriate
### Describe the issue

In case the minder server receives a client call that doesn't specify a context where a context is required (e.g. ListRepositories), currently the error returned to the client is 500 Server Error. Those should only be used for cases of an internal error and for this case where an expected parameter is missing we could use 400 Bad Request instead or in case the parameter is there but is malformed, maybe 422 might be better.

### To Reproduce

Call the Minder REST API but do not include a Context.

_No response_

dmjb:
Fixes #2376

When the context of a request is missing, return a user-visible "InvalidArgument" response. Previously, this returned a regular Go error which caused the GRPC server to return a 500 response to the client.

### DIFF
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+
 	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


JAORMX:
Seems golangci-lint is not happy. might wanna fix that. We also don't allow it to properly comment on the PR... this should fix it https://github.com/stacklok/minder/pull/2378

rdimitrov (approved):

PR #2385, opened 2024-02-20
Fix infinite wait for ratecache using non blocking Close() func

Vyom-Yadav:
Issue #2261 (From https://github.com/stacklok/minder/pull/2271#discussion_r1495557451)

* Using Wait() in server.go introduced the bug where if the context is not cancelled, the server will continue to wait even if other go routines finished

* Replaces Wait() with non blocking Close() operation

* All this is done to free the resources associated with ticker

### DIFF
--- a/cmd/server/app/serve.go
+++ b/cmd/server/app/serve.go
@@ -124,6 +124,7 @@ var serveCmd = &cobra.Command{
 		serverMetrics := controlplane.NewMetrics()
 		providerMetrics := provtelemetry.NewProviderMetrics()
 		restClientCache := ratecache.NewRestClientCache(ctx)
+		defer restClientCache.Close()
 
 		s, err := controlplane.NewServer(
 			store, evt, serverMetrics, cfg, vldtr,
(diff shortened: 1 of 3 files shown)


Vyom-Yadav:
cc @JAORMX

JAORMX (dismissed):
LGTM @jhrozek can you also take a look?

Vyom-Yadav:
Hmm, why is https://coveralls.io/builds/65803176/source?filename=internal%2Fproviders%2Fratecache%2Fstore.go#L126 out of coverage. There is a test for context cancellation 🤔

JAORMX:
> Hmm, why is https://coveralls.io/builds/65803176/source?filename=internal%2Fproviders%2Fratecache%2Fstore.go#L126 out of coverage. There is a test for context cancellation 🤔

... interesting, sometimes `select` calls may lag as they're waiting for conditions to fulfill. You could try to test it by not calling a `defer` `Close` in that test and verifying if the channel is indeed closed.

Vyom-Yadav:
> sometimes select calls may lag as they're waiting for conditions to fulfill

That seems to be the case. Select statement blocks, but the execution is so fast that it unblocks on `closeChan`. Removing the defer close call works in this case.

JAORMX (approved):

jhrozek (approved):

PR #2391, opened 2024-02-21
Add skeleton of IDP interface for resolving friendly names

evankanderson:
See discussion in #2326.  The goal here is to enable the following:

1. Preserve existing relations in OpenFGA and avoid needing a migration.
2. Support `minder project role grant -r viewer -u JAORMX` (which should look up the "JAORMX" `preferred_username` in Keycloak (populated from GitHub), and then store the (stable) Keycloak ID in the OpenFGA grant.  We don't want to store `JAORMX`, because he might change his account name later.
3. Support listing `JAORMX` as the username coming back from ListRoleAssignments, so that we can read the output.  We can also include the Keycloak userids in the RPC response, but no one will read them.  :grin:
4. Add an IdentityProvider for KeyCloak + GitHub that will call [getUsers with a filter](https://www.keycloak.org/docs-api/23.0.6/rest-api/#_get_adminrealmsrealmusers) and/or [create a user](https://www.keycloak.org/docs-api/23.0.6/rest-api/#_post_adminrealmsrealmusers) and [set their GitHub federated identity information](https://www.keycloak.org/docs-api/23.0.6/rest-api/#_post_adminrealmsrealmusersidfederated_identityprovider)
   * This will allow us to resolve users to KeyCloak identities when the role grant is made, rather than at a possible later date when GitHub identities have shifted.
5. In the future, add an `actions` IdentityProvider for GitHub actions which would be distinct from KeyCloak + GitHub users in 2
   * The syntax for this would be something like `minder project role grant -r editor -u actions/repo:stacklok/minder-rules-and-profiles:ref:refs/heads/main`.  (Yeah, action names are pretty awesome...)

To fit the interface into existing code, we'll need to replace:
* `ParseAndValidate` (4 calls, `handlers_token.go` and 3x `handlers_user.go`)
* Calls around `authzClient` (`default_project.go`, `handlers_authz.go`, `handlers_user.go`) which will need to use the canonical `Identity.String()` for `Write`, `Check`, and `ProjectsForUser`, and `Identity.Human()` for `AssignmentsToProject`.


internal/auth/interface.go
```diff
@@ -0,0 +1,197 @@
...
+	// "alexsmith" to "alexawesome" after a life change, and someone else might
+	// enroll the "alexsmith" handle.  If you are storing data, you want UserID,
+	// not HumanName.  If you are presenting data, you probably want HumanName.
+	//
+	// For KeyCloak, this is `preferred_username`.
+	HumanName string
+	// Provider is the identity provider that vended this identity.  Note that
+	// UserID and HumanName are only unique within the context of a single
+	// identity provider.
+	Provider IdentityProvider
```
JAORMX, line 49: Do we want to add email to this? For some IdPs we wouldn't quite have a preferred user name and the email would be more relevant.

internal/auth/interface.go
```diff
@@ -0,0 +1,197 @@
...
+	// the identity in the remote identity provider.
+	//
+	// For Keycloak + GitHub, this may define a new user in Keycloak based on
+	// GitHub user data if the user is not already known to Keycloak.
+	Resolve(ctx context.Context, id string) (*Identity, error)
+}
+
+// IdentityProvider provides an abstract interface for looking up identities
+// in a remote identity provider.
+type IdentityProvider interface {
```
JAORMX, line 101: is IdentityProvider assumed to be OIDC-compatible? if so, let's mark that in the docs.

internal/auth/interface.go
```diff
@@ -0,0 +1,197 @@
...
+	String() string
+	// URL returns the `iss` URL of the identity provider.
+	URL() url.URL
+}
+
+// IdentityClient supports the ability to look up identities in one or more
+// IdentityProviders.
+type IdentityClient struct {
+	// This map is a bit overloaded; it maps both short provider names
+	providers map[string]IdentityProvider
```
JAORMX, line 117: you could use `xsync.MapOf` which we already use in other parts of the code-base.

JAORMX (commented):
This abstraction makes sense to me. Is the idea also to abstract the OIDC token validation into the implementations here?

internal/auth/interface.go
eleftherias: If it is, could we use zitadel's `RelyingParty` interface instead of creating our own abstraction?

eleftherias: We're already using it in our code base when we create the Keycloak provider

internal/auth/interface.go
evankanderson: I think email _is_ the preferred username for some providers.  For others (e.g. GitHub Actions) users may not have email addresses.

internal/auth/interface.go
evankanderson: I'm not sure that we want to require it.  I imagined this as being decoupled from specific identity provider implementations, so we'd still need something like OIDC (or x.509 client certs) to assert the user's identity.  Once we have their identity, we can then go through this interface to map it to/from database representation.

internal/auth/interface.go
JAORMX: Makes sense, I think that's fine

evankanderson (commented):
I got some email comments from Ria that didn't show up here, adding them and then I'm going to polish this a little tonight and finish tomorrow.

evankanderson:
We did this in #3155

Issue #2262, opened 2024-02-01
Add Background Reconciliation for Entities


Vyom-Yadav:
### Please describe the enhancement

Minder relies purely on event-based processing triggered by webhooks/user actions. Apart from edge-based triggering, Minder should also reconcile the entities in the background. Reconcilers under the scope of this issue are:

- [ ] Continuously update the evaluation status for entities against registered rules. (Task - I)
- [ ] Continuously evaluate the state of the webhook created by the minder on GitHub. (As the user may accidentally delete it) (Task - II)

These reconcilers are required to ensure all existing rule evaluations in minder are updated according to any newly added logic. See #1609 for example.

### Solution Proposal

Continuously running the reconcilers is not feasible due to resource constraints and the provider's API limits. Reconcilers would run based on a cron job with a small enough batch size not to hit rate-limiting errors for any provider.

### Additional context

For more details see: https://discord.com/channels/1184987096302239844/1201083912521252885

jhrozek:
Hi @Vyom-Yadav I was thinking about this feature some time ago (though I haven't gotten far, thank you very much for picking it up) here are my notes from that time. Feel free to take anything you like from them or not if otherwise :-). A lot of it is very hand-wavy, as I said, I didn't get too deep.

- I was thinking about having a separate process, a singleton as opposed to be a part of the potentially replicated minder instances. The idea behind this was that presumably we'd use a queue of entities to be updated and having a singleton writing to the queue just seemed simpler.
- this new component (please name it `reminder` :-)) would periodically (say once an hour?) traverse the database, picking up projects that contain entities that haven't been updated in `$TIME_PERIOD`.
- each project would get a time slot and could update up to N entities from the project within that time slot. Individual updates within that interval would be spaced apart with a jitter to avoid spikes. An entry to update an entity would be written to a work queue. We use watermill with an SQL back end, without thinking about the details too deep I was wondering if we could make this `reminder` a topic publisher and `minder` workers subscribers to the periodical update topics.
- the minder processes being subscribers to the periodical update table would then issue an `ExecuteEntityEventTopic`

Vyom-Yadav:
Good points, @jhrozek ! I'm also thinking about having this as a separate process. Based on the number of replicas, we can process an entity repeatedly, leading to rate-limiting errors.

The Singleton Client Binary would have the logic for scheduling reconciliation events. The approach I had in my mind is:

[...]

PR #2394, opened 2024-02-21
Refactor project structure: Move cursor file to utils package and migrate common config struct to common.go

Vyom-Yadav:
Pre-Requisites for #2262 

Changes:

- Migrated `server/db.go` -> `config/common.go` as both server and reminder service would need to connect to the DB.
- Migrated `controlplane/cursor.go` -> `util/cursor.go` as `reminder` package shouldn't depend on `controlplane` pkg.
- Migrated `GRPCClientConfig` struct and associated CLI method to `config/common.go` as reminder service would need to connect to the GRPC server (for sending reminders)


internal/util/cursor/cursor.go
```diff
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package controlplane
+// Package cursor provides a way to encode and decode cursors for paginated queries
+package cursor
```
dmjb, line 16: Are the cursors ever used outside of the control plane? If not, I think it may make sense to keep them as a sub-package of controlplane.

dmjb: Yes, they should probably go into a less-generic package than `util`, but that is something we can worry about when you introduce your next PR(s). I'm happy to leave them in util for now.

dmjb (approved):

JAORMX (approved):

PR #2413, opened 2024-02-23
Add provider callback handler for web app

### LINKED ISSUE #2404: Add provider redirect handler for web based apps
With the current CLI provider enrolment process, the Minder backend handles the redirect from the provider (e.g GitHub) and generates and stores a user token.
The Minder CLI queries the backend, looking for a confirmation that the token has been saved.

From a  web based app perspective, it would be nicer if the user was redirected back to frontend app after the provider has been authorized. This way the web app doesn't have to continuously query the Minder backend.

eleftherias:
# Summary

- Remove callback handler from rpc definition since it's always an HTTP request
- Add optional redirect_url parameter

Fixes #2404

## Change Type

# Testing

# Review Checklist:


internal/controlplane/server.go
```diff
@@ -316,6 +317,12 @@ func (s *Server) StartHTTPServer(ctx context.Context) error {
 
 	mw := otelhttp.NewMiddleware("webhook")
 
+	// Use Gorilla Mux router to handler path variables
+	gorillaMux := gorillamux.NewRouter()
```
JAORMX, line 321: aren't we able to also handle variables using the `mux` from the GRPC gateway runtime we already have? Why do we need to introduce a new dependency?

https://grpc-ecosystem.github.io/grpc-gateway/docs/operations/inject_router/

internal/controlplane/handlers_oauth.go
```diff
@@ -115,108 +126,150 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
-
-	// Check the nonce to make sure it's valid
-	valid, err := mcrypto.IsNonceValid(in.State, s.cfg.Auth.NoncePeriod)
-
-	if err != nil {
-		return nil, status.Errorf(codes.Unknown, "error checking nonce: %s", err)
-	}
+// HandleProviderCallback handles the OAuth 2.0 authorization code callback from the enrolled
+// provider. This function gathers the state from the database and compares it to the state
+// passed in. If they match, the provider code is exchanged for a provider token.
```
JAORMX, line 131: you might want to indicate that this is handle by the gorilla Mux router, given that it's the only instance where it differs

internal/controlplane/handlers_oauth.go
```diff
@@ -115,108 +126,150 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
+//
+//nolint:gocyclo
```
JAORMX, line 133: maybe we should listen to this warning and address the cyclomatic complexity by splitting this function.

JAORMX (changes requested):

internal/controlplane/server.go
eleftherias: Thanks for pointing that out! Updated

internal/controlplane/handlers_oauth.go
eleftherias: Done

internal/controlplane/handlers_oauth.go
eleftherias: Removed usage of gorilla mux, but added a comment saying this is an HTTP only handler

internal/controlplane/handlers_oauth.go
```diff
@@ -92,6 +95,13 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
 		owner = sql.NullString{Valid: true, String: *req.Owner}
 	}
 
+	var redirectUrl sql.NullString
+	if req.RedirectUrl == nil {
+		redirectUrl = sql.NullString{Valid: false}
+	} else {
+		redirectUrl = sql.NullString{Valid: true, String: *req.RedirectUrl}
```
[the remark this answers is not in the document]
eleftherias: In this case the redirect URL will be something like `minder.com/onboarding/repo-registration` so that the frontend user can continue where they left off before the redirect to GitHub. It won't have any additional query params like `code` etc. Do you think it still needs to be encrypted?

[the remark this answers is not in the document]
JAORMX: gotcha, it wasn't entirely clear to me what it would contain. If it's just the host and path it's fine.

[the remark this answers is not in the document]
JAORMX: Should we add URL validation so we make sure such things don't get stored here then?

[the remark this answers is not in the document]
eleftherias: I'd be more inclined to encrypt it just in case, rather than restrict what can be passed in. The frontend team hasn't determined what the structure will be, and I'd like to allow the flexibility of sending `minder.com/onborading?step=repo-registration` if that ends up being more convenient for them. What do you think?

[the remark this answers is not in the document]
JAORMX: Can you give a little bit more information on what the proposed flow and usage for this is? Just wanna make sure we don't over-complicate things while keeping the security concerns in check.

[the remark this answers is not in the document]
eleftherias: For sure, this is the intended flow:
- User wants to enrol a provider from the frontend Minder app
- FE queries BE to get the github authorization URL providing the URL where the Minder backend will send the user when enrolment is complete, for example `minder.com/onboarding/repo-registration`
- BE responds with the GitHub authorization URL and the frontend redirects the user there
- When GitHub authorization is complete, GitHub redirects to Minder BE with the oauth code and state param. Minder BE uses the code to generate an OAuth token, and stores the token. It uses the state to figure out 1) which project to associate the token with and 2) to get the redirect back to the frontend
- Minder BE redirects the user back to the frontend

The key difference between the web flow and the CLI flow is that the CLI flow shows the "Minder enrollment complete" page, while the web flow takes the user back to the frontend.

With the web flow and the CLI flow, none of the OAuth data (code or token) is passed to the frontend. It's all handled exclusively by the backend.

evankanderson:
I checked out v0.0.24 and verified that we do not call `ExchangeCodeForTokenCLI` anywhere in that release, so we should be good to go from a compatibility point of view.  I'm reading the code now.

dmjb (approved):

internal/controlplane/handlers_oauth.go
```diff
@@ -115,55 +127,93 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
-// This function is used by the CLI client.
-func (s *Server) ExchangeCodeForTokenCLI(ctx context.Context,
-	in *pb.ExchangeCodeForTokenCLIRequest) (*httpbody.HttpBody, error) {
+// HandleProviderCallback handles the OAuth 2.0 authorization code callback from the enrolled
+// provider. This function gathers the state from the database and compares it to the state
+// passed in. If they match, the provider code is exchanged for a provider token.
+// note: this is an HTTP only (not RPC) handler
+func (s *Server) HandleProviderCallback() runtime.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
+		ctx := context.Background()
```
evankanderson, line 136: This should use [`r.Context()`](https://pkg.go.dev/net/http#Request.Context), not the background context.  (Or it can just pass along `r`.)

internal/controlplane/handlers_oauth.go
```diff
@@ -115,55 +127,93 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
 	span := trace.SpanFromContext(ctx)
-	span.SetName("server.ExchangeCodeForTokenCLI")
-	span.SetAttributes(attribute.Key("code").String(in.Code))
+	span.SetName("server.HandleProviderCallback")
+	span.SetAttributes(attribute.Key("code").String(code))
 	defer span.End()
 
-	// Check the nonce to make sure it's valid
-	valid, err := mcrypto.IsNonceValid(in.State, s.cfg.Auth.NoncePeriod)
+	provider := pathParams["provider"]
```
evankanderson, line 157: Do we want to check that provider is non-empty?

internal/controlplane/handlers_oauth.go
```diff
@@ -115,55 +127,93 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
+
+	if stateData.RedirectUrl.Valid {
+		parsedURL, err := url.Parse(stateData.RedirectUrl.String)
+		if err != nil {
+			return fmt.Errorf("error parsing redirect URL: %w", err)
+		}
+		http.Redirect(w, r, parsedURL.String(), http.StatusTemporaryRedirect)
+		if err != nil {
+			return fmt.Errorf("error redirecting to URL: %w", err)
+		}
```
evankanderson, line 191: Should we `return` at this point?  We've sent a redirect, so the ResponseWriter isn't going to be good for much else.  In fact, attempting to return an error here will probably screw up the ResponseWriter if it tries to send a second response code.

internal/controlplane/handlers_oauth.go
```diff
@@ -115,55 +127,93 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
-// This function gathers the state from the database and compares it to the state
-// passed in. If they match, the code is exchanged for a token.
-// This function is used by the CLI client.
-func (s *Server) ExchangeCodeForTokenCLI(ctx context.Context,
-	in *pb.ExchangeCodeForTokenCLIRequest) (*httpbody.HttpBody, error) {
+// HandleProviderCallback handles the OAuth 2.0 authorization code callback from the enrolled
+// provider. This function gathers the state from the database and compares it to the state
+// passed in. If they match, the provider code is exchanged for a provider token.
+// note: this is an HTTP only (not RPC) handler
+func (s *Server) HandleProviderCallback() runtime.HandlerFunc {
```
evankanderson, line 134: Do we want to try to enforce that this is HTTPS-only by checking the [`x-forwarded-proto` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) or [`Forwarded: ...proto`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded)?  (Yeah, that's kinda crummy, but is mostly-standardized, I think)

internal/controlplane/handlers_oauth.go
```diff
@@ -115,55 +127,93 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
+	}
+
+	return nil
+}
+
+func (s *Server) generateOAuthToken(ctx context.Context, provider string, code string,
+	stateData db.GetProjectIDPortBySessionStateRow) error {
 	// generate a new OAuth2 config for the given provider
-	oauthConfig, err := auth.NewOAuthConfig(provider.Name, true)
+	oauthConfig, err := auth.NewOAuthConfig(provider, true)
```
evankanderson, line 206: We used to use `getProviderFromRequestOrDefault`, but now we're just taking the path parameter as-is and feeding it in to `NewOAuthConfig`.  I think we need to do some validation here.

internal/controlplane/handlers_oauth.go
```diff
@@ -115,55 +127,93 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
+		_, err = w.Write(auth.OAuthSuccessHtml)
+		if err != nil {
+			return fmt.Errorf("error writing OAuth success page: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (s *Server) generateOAuthToken(ctx context.Context, provider string, code string,
```
evankanderson, line 203: I am sad that the existing code is entwined with the `Server` and doesn't take the cryptoEngine and a slice of the Store interface.  It would be pretty easy to test on its own if we did that.

We could also replace `DeleteAccessToken` / `CreateAccessToken` with `UpsertAccessToken`, which would do an [`INSERT ... ON CONFLICT DO UPDATE`](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT), which would be more atomic than what we do today (two HTTP requests could race each other).

On the other hand, you may not be looking to do a big refactor, so I'm not going to block on it.

evankanderson (commented):
A few comments on how you could slip some testing in if you want, but I'm not going to block on it.

Musts:

1. Use `r.Context`, not `context.Background`.
2. Validate the `provider` parameter
3. (maybe) handle the `.../callback/cli` path if it is used.

internal/controlplane/handlers_oauth.go
eleftherias: We should check that the provider is not empty, but `getProviderFromRequestOrDefault` never made sense here because this endpoint was only ever called from GitHub which never had a user context.

internal/controlplane/handlers_oauth.go
eleftherias: From what I can tell the handlers that are generated from our rpc routes don't enforce HTTPS-only (but maybe I didn't look in the right places), so I don't think we need to enforce it here.

internal/controlplane/handlers_oauth.go
eleftherias: I was struggling with testing this because `oauthConfig.Exchange` makes HTTP calls.
I was thinking eventually we can have a factory on the `Server` struct that generates the oauth config, similar to what we have for `auth.JwtValidator`. That would make this entire handler testable. 
I just wasn't ready to take on that refactor now. 
Do you have any other ideas for making this testable considering the calls in `oauthConfig.Exchange`?

internal/controlplane/handlers_oauth.go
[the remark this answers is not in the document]
JAORMX: The redirect URI is stored in the DB, which is what I suggested encrypting.

internal/controlplane/handlers_oauth.go
eleftherias: I agree

internal/controlplane/handlers_oauth.go
eleftherias: I like the idea, where would `Register(string, *auth.OAuthConfig)` be called? I imagine it could be on startup, since we have (almost) all the data we need to construct it. The only thing that's dynamic is the `cli` boolean which changes the redirect URL based on whether the CLI or web app initiated the enrolment, and I would like to get rid of that differentiation anyway.

internal/controlplane/handlers_oauth.go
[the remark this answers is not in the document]
evankanderson: Ah - I'd missed that because it was off the diff. Carry on!

internal/controlplane/handlers_oauth.go
```diff
@@ -115,55 +128,94 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
 	if err != nil {
-		return nil, providerError(err)
+		return err
 	}
 
-	// generate a new OAuth2 config for the given provider
-	oauthConfig, err := auth.NewOAuthConfig(provider.Name, true)
+	// Telemetry logging
+	logger.BusinessRecord(ctx).Provider = provider
+	logger.BusinessRecord(ctx).Project = stateData.ProjectID
```
evankanderson, line 185: Can we move these before `s.generateOAuthToken, so that we include the provider and project information even on errors?

internal/controlplane/handlers_oauth.go
eleftherias: I'll create a followup issue to refactor this and make it testable

JAORMX (approved):
LGTM

eleftherias:
Changed the DB migration number to 22, anticipating that https://github.com/stacklok/minder/pull/2452 will be merged first

JAORMX (dismissed):

eleftherias:
Had to resolve conflicts

JAORMX (approved):

PR #2415, opened 2024-02-23
Upgrade repoID to int64, because that's the size from GitHub

evankanderson:
# Summary

Changes the stored and protocol repoID to be `int64` size instead of `int32` size.  [`go-github` uses an `int64` for this field](https://pkg.go.dev/github.com/google/go-github/v31/github#Repository), and GitHub doesn't guarantee the int size, which appears to be sequential and was 734857594 for a recently-created repo.

## Change Type

# Testing

Ran `go test ./...` and ensured no compilation errors.  This should include an up-migration, but not a down-migration (which seems scary to me).

# Review Checklist:


puerco (approved):

JAORMX (approved):
Let's go for this now. Though I'm the future we'd anyway switch the repo id to be a string to match formats for other providers

dmjb (approved):

rdimitrov (approved):

PR #2446, opened 2024-02-27
Promote Github PR ID to int64

### LINKED ISSUE #2443: Certain Github IDs are downcast to int32
### Describe the issue

After the recent issue involving repository IDs being downcast, I have noticed that other Github IDs are being downcast. For example:

Here's one for webhooks: https://github.com/stacklok/minder/blob/main/internal/controlplane/handlers_repositories.go#L100
Here's one for PR IDs: https://github.com/stacklok/minder/blob/main/internal/controlplane/handlers_githubwebhooks.go#L738

These seem to be related to protobuf and DB schemas.

Need to:

* Audit the different IDs relating to Github to find any other IDs which may have this downcast problem.
* Change the relevant schemas
* Change the code to use int64

### To Reproduce

_No response_

_No response_

dmjb:
# Summary

Fixes #2443

PR ID and line number are were previously downcasted to int32 in a number of places despite being mostly treated as an int64 type. While the likelihood of either of these values exceeding a 32-bit MAX_INT are minimal, this change makes them 64-bit to err on the side of caution.

EDIT: After discussion with the team, we have decided to leave the line numbers as int32. The rationale is that if the diffs ever get that large, we're going to have a whole bunch of other problems to deal with.

## Change Type

# Testing

# Review Checklist:


eleftherias (dismissed):
The code change looks good! For the commit message, our standard is to capitalise the subject line, so this one should be "Promote...". You can change the subject when merging.

proto/minder/v1/minder.proto
```diff
@@ -253,7 +253,7 @@ message PrContents {
         repeated Line patch_lines = 3;
 
         message Line {
-            int32 line_number = 1;
+            int64 line_number = 1;
```
evankanderson, line 256: I'd leave this as an int32, because I'm terrified of needing more than 2B line numbers for a patch.

internal/engine/ingester/diff/diff.go
```diff
@@ -216,7 +216,7 @@ func ingestFileForFullDiff(filename, patch, patchUrl string) (*pb.PrContents_Fil
 		} else if strings.HasPrefix(line, "+") {
 			result = append(result, &pb.PrContents_File_Line{
 				Content:    line[1:],
-				LineNumber: int32(currentLineNumber),
+				LineNumber: currentLineNumber,
```
evankanderson, line 219: We could just declare `currentLineNumber` as int32 here on line 206, particularly since we parse it as an int32 on line 212.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -735,7 +735,7 @@ func getPullRequestInfoFromPayload(
 
 	return &pb.PullRequest{
 		Url:      prUrl,
-		Number:   int32(prNumber),
+		Number:   int64(prNumber),
```
evankanderson, line 738: I'm assuming we can't use `JQReadFrom[int64]` on line 721 because the underlying API understands that JSON numbers are floats?

evankanderson (dismissed):
I wouldn't expand the line_number counter, that seems overly optimistic.

proto/minder/v1/minder.proto
teodor-yanev: Yeah, if there's a PR file exceeding 2^32 number of lines, this counter is the last thing to worry about 😄

internal/engine/ingester/diff/diff.go
dmjb: Tried this, but since `strconv.ParseInt` returns `(int64, error)` we can't just wrap the return value in `int32()`.

JAORMX (approved):

PR #2458, opened 2024-02-28
Ensure user is logged in upon going through the quickstart flow

rdimitrov:
# Summary

The following PR:
* Ensures that the user is logged in upon going through the quickstart flow
* If they are logged out, the CLI will prompt the user to log in, recreate the gRPC clients and the user can continue their flow uninterruptedly
* We do that at the beginning and not for every step of the quickstart command, as we discussed we'll be extending the refresh token to 10hrs (as oppose to what we have now - 30mins). In both cases there should be enough time for the user to complete their quickstart flow.
* Updated the message that we render at the end to match the two use cases - the profile is created vs the profile was already there. Now it should look much better since we tailor the message instead of appending another note like we did before.

Fixes #(related issue)

## Change Type

# Testing

Tested this locally.
**Prompt for logging in:** 
image
**User decided they don't want to log in:**
image

# Review Checklist:


cmd/cli/app/quickstart/quickstart.go
```diff
@@ -137,9 +153,12 @@ var cmd = &cobra.Command{
 // quickstartCommand is the quickstart command
 //
 //nolint:gocyclo
-func quickstartCommand(_ context.Context, cmd *cobra.Command, conn *grpc.ClientConn) error {
+func quickstartCommand(_ context.Context, cmd *cobra.Command, inConn *grpc.ClientConn) error {
+	var err error
+	var conn *grpc.ClientConn
+	conn = inConn
```
puerco, line 159: What's the purpose of copying the connection here? Can't we just reuse the variable?

rdimitrov: We can, yes 👍 I have no other reason other than I thought it'd better to be more explicit to say we are creating a new connection, but I'm happy to change this 👍 

As for the safety of not leaving a connection open, my understanding is we should be okay even if we overwrite the incoming variable as long as we do a `defer conn.Close()` on both and we do that right after the create part.

rdimitrov: Fixed it in https://github.com/stacklok/minder/pull/2458/commits/d1d74f8926b95df44986c06f5db9d315f9e72cc3 👍

JAORMX (approved):

PR #2459, opened 2024-02-28
Move webhook create/delete into separate interface

dmjb:
Part of ongoing test improvement work

This started out as an attempt to refactor
`registerWebhookForRepository` (which currently is the least-tested method in the codebase by # of untested lines). I discovered that breaking it up will require further rounds of refactoring, but I started by moving webhook logic behind a separate interface, since it was probably the easiest part of the logic to isolate.

Existing tests for `handlers_repository` were altered to use the new interface which simplified those tests somewhat.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/controlplane/handlers_repositories_test.go
```diff
@@ -221,18 +234,7 @@ func TestServer_RegisterRepository(t *testing.T) {
 			}
 
 			if tt.repoErr == nil {
-				// The last part of the hook URL is a UUID which is generated iternally.
```
dmjb, line 224: Since we can now control the value of the UUID, this logic is no longer needed

internal/controlplane/handlers_repositories_test.go
```diff
@@ -305,16 +307,8 @@ func (*StubGitHub) CreateComment(context.Context, string, string, int, string) e
 }
 
 // CreateHook implements v1.GitHub.
-func (s *StubGitHub) CreateHook(_ context.Context, owner string, repo string, hook *github.Hook) (*github.Hook, error) {
```
dmjb, line 308: These mock methods are no longer needed: the test relies on the stub of `WebhookManager` instead

evankanderson (changes requested):
I think you need to hook up the webhook client in the server before this can go live.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -266,64 +259,10 @@ func (s *Server) registerWebhookForRepository(
 		return regResult, nil
 	}
 
-	urlUUID := uuid.New().String()
-
-	webhookUrl := fmt.Sprintf("%s/%s", url, urlUUID)
-	parsedOriginalURL, err := urlparser.Parse(webhookUrl)
+	hookUUID, githubHook, err := s.webhookManager.CreateWebhook(ctx, client, repo.Owner, repo.Name)
```
JAORMX, line 262: I agree that we should move webhook creation to a dedicated interface, but I'm not sure I'm too keen on this way of doing it. I was thinking more along the lines of having a dedicated "webhook-setter" provider type or trait which would be instantiated by the provider builder, and we'd then shove the github-related logic in there. The idea is that there are other systems (e.g. DockerHub) which support webhook installation, so this is something that belongs more on the provider side as opposed to logic that should go in the server.

internal/controlplane/handlers_repositories_test.go
```diff
@@ -221,18 +234,7 @@ func TestServer_RegisterRepository(t *testing.T) {
...
-					t.Fatalf("expected webhook to be registered once, got %+v", stubClient.NewHooks)
-				}
-				hookURL := stubClient.NewHooks[0].GetURL()
-				parts := strings.Split(hookURL, "/")
-				got.Result.Repository.HookUuid = parts[len(parts)-1]
-			}
-
-			if !reflect.DeepEqual(got, tt.want) {
-				t.Errorf("Server.RegisterRepository() = %v, want %v", got, tt.want)
+				got.Result.Repository.HookUuid = hookUUID
```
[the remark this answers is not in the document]
dmjb: This change got accidentally added to my commit, fixing...

internal/controlplane/handlers_repositories_test.go
dmjb: Yeah, `GetRepository` is still used.

internal/controlplane/server.go
```diff
@@ -85,6 +86,7 @@ type Server struct {
 	// inject more entity-specific interfaces. For example, we may want to
 	// consider having a struct per grpc service
 	profileValidator *profiles.Validator
+	webhookManager   webhooks.WebhookManager
```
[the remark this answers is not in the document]
dmjb: Found the same issue while doing some manual testing of Minder... fixed now.

internal/repositories/github/webhooks/manager.go
```diff
@@ -0,0 +1,181 @@
...
+			logger.Error().Msg("error deleting hook")
+			return err
+		}
+	}
+
+	return nil
+}
+
+func isMinderHook(hook *github.Hook, hostURL string) (bool, error) {
+	configURL := hook.Config["url"].(string)
```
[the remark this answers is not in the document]
dmjb: This code is mostly copy-paste from what was in the controlplane. I was also wondering whether it was a good idea to silently ignore the missing URL or not... happy to change it to throw an error in this case.

internal/repositories/github/webhooks/manager.go
```diff
@@ -0,0 +1,181 @@
...
+			"content_type": "json",
+			"ping_url":     ping,
+			"secret":       secret,
+		},
+		Events: targetedEvents,
+	}
+
+	webhook, err := client.CreateHook(ctx, repoOwner, repoName, newHook)
+	if err != nil {
+		logger.Error().Msg("error creating hook")
```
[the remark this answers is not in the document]
dmjb: This log and error behaviour was copied from the original code, so I assumed there might have been a specific reason for it. Happy to change it.

internal/controlplane/handlers_githubwebhooks.go
dmjb: I think this will be a good idea once we build another provider which needs to manage webhooks. At that point in time, it will become much more clear what a generic webhook manager interface may look like. Until then, I think we should have a GitHub-specific interface, and we can refactor the logic to work with the generic interface once we build it.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -266,64 +259,11 @@ func (s *Server) registerWebhookForRepository(
...
-		},
-		Events: ghEvents,
-	}
-
-	// if we have an existing hook for same repo, delete it
-	hooks, err := client.ListHooks(ctx, repo.Owner, repo.Name)
-	if errors.Is(err, githubprovider.ErrNotFound) {
-		logger.Debug().Msg("no hooks found")
-	} else if err != nil {
+		log.Printf("error while creating webhook: %v", err)
```
evankanderson, line 264: Can we use the `logger` here (and maybe put it back onto `ctx`, so this `"error while creating webhook"` message includes the repo name and org?

evankanderson: Alternatively, you can point me to where in the caller the same thing is set, and use the logger from there.

internal/controlplane/handlers_githubwebhooks.go
evankanderson: I think in the future, when we

```
// TODO: the remaining logic should be refactored into a repository
// registration interface
```

The webhookManager would be part of the repository registration interface, not part of `Server`, correct?  If so, this seems like an okay intermediate step.

internal/repositories/github/webhooks/manager.go
```diff
@@ -0,0 +1,179 @@
...
+func (w *webhookManager) cleanupStaleHooks(
+	ctx context.Context,
+	client GitHubWebhookClient,
+	repoOwner string,
+	repoName string,
+	webhookHost string,
+) error {
+	logger := zerolog.Ctx(ctx).With().
+		Str("repoName", repoName).
+		Str("repoOwner", repoOwner).Logger()
```
evankanderson, line 140: If you do this out in `registerWebhookForRepository` and store it back in `ctx`, you can simply use `logger := zerolog.Ctx(ctx)` here.

evankanderson (approved):

internal/controlplane/handlers_githubwebhooks.go
dmjb: @evankanderson Yes, and I even have a draft PR sitting around locally where I have made this change already.

internal/controlplane/handlers_githubwebhooks.go
dmjb: I think I got confused by the mix and match of logging styles in the control plane... I'll change it back to zerolog.

evankanderson: Yeah, zerolog got put in partway through, and we haven't fully standardized.  In the future, I might want us to move back towards `slog`, but it hasn't been a priority.  Maybe a lazy August afternoon switchover.

evankanderson (approved):

PR #2468, opened 2024-02-29
Support for offline tokens

JAORMX:
# Summary

This adds new sub-commands to `minder auth` which enable the usage and installation
of offline tokens. These tokens enable you to create long-lived credentials which are handy
for CI and automation cases.

Co-Authored-By: Eleftheria Stein-Kousathana <eleftheria@stacklok.com>

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/cmd/cli/app/auth/offline_get.go
+++ b/cmd/cli/app/auth/offline_get.go
@@ -1,0 +1,81 @@
+//
+// Copyright 2024 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/stacklok/minder/internal/config"
+	clientconfig "github.com/stacklok/minder/internal/config/client"
+	"github.com/stacklok/minder/internal/util/cli"
+)
+
+// offlineTokenGetCmd represents the offline-token get command
+var offlineTokenGetCmd = &cobra.Command{
+	Use:   "get",
+	Short: "Retrieve an offline token",
+	Long: `The minder auth offline-token get command project lets you retrieve an offline token
+for the minder control plane.
+
+Offline tokens are used to authenticate to the minder control plane without
+requiring the user's presence. This is useful for long-running processes
+that need to authenticate to the control plane.`,
+
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		ctx, cancel := cli.GetAppContext(cmd.Context(), viper.GetViper())
+		defer cancel()
+
+		clientConfig, err := config.ReadConfigFromViper[clientconfig.Config](viper.GetViper())
+		if err != nil {
+			return fmt.Errorf("error reading config: %w", err)
+		}
+
+		f := viper.GetString("file")
+
+		// No longer print usage on returned error, since we've parsed our inputs
+		// See https://github.com/spf13/cobra/issues/340#issuecomment-374617413
+		cmd.SilenceUsage = true
+
+		// wait for the token to be received
+		token, err := login(ctx, cmd, clientConfig, []string{"offline_access"})
+		if err != nil {
+			return err
+		}
+
+		// write the token to the file
+		if err := os.WriteFile(f, []byte(token.RefreshToken), 0600); err != nil {
+			return fmt.Errorf("error writing offline token to file: %w", err)
+		}
+
+		cmd.Printf("Offline token written to %s\n", f)
+
+		return nil
+	},
+}
+
+func init() {
+	OfflineTokenCmd.AddCommand(offlineTokenGetCmd)
+
+	offlineTokenGetCmd.Flags().StringP("file", "f", "offline.token", "The file to write the offline token to")
+
+	if err := viper.BindPFlag("file", offlineTokenGetCmd.Flag("file")); err != nil {
+		panic(err)
+	}
+}
--- a/cmd/cli/app/auth/offline_use.go
+++ b/cmd/cli/app/auth/offline_use.go
@@ -1,0 +1,108 @@
+//
+// Copyright 2024 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/stacklok/minder/internal/config"
+	clientconfig "github.com/stacklok/minder/internal/config/client"
+	"github.com/stacklok/minder/internal/util"
+)
+
+// offlineTokenUseCmd represents the offline-token use command
+var offlineTokenUseCmd = &cobra.Command{
+	Use:   "use",
+	Short: "Use an offline token",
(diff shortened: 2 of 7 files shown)


cmd/cli/app/auth/offline_use.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil
+	},
+}
+
+func init() {
+	OfflineTokenCmd.AddCommand(offlineTokenUseCmd)
+
+	offlineTokenUseCmd.Flags().StringP("file", "f", "offline.token", "The file that contains the offline token")
+	offlineTokenUseCmd.Flags().BoolP("from-env", "e", false, "Use the offline token from the environment")
+	offlineTokenUseCmd.Flags().StringP("env-var", "v", "MINDER_OFFLINE_TOKEN", "The environment variable to use for the offline token")
```
rdimitrov, line 106: Perhaps there's no need to have an explicit env var and another flag that says to read from it. Instead we can simplify this to have 2 flags:
* One that says read the token from this path(`--tokenpath offline.token`
* Another that says here's the token, i.e. `--token $MY_TOKEN_VAR`
* And the two should exclude each other.

cmd/cli/app/auth/offline_use.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		f := cmd.Flags().Lookup("file").Value.String()
+		fromEnv := cmd.Flags().Lookup("from-env").Value.String() == "true"
+		env := cmd.Flags().Lookup("env-var").Value.String()
+		if f == "" && !fromEnv {
+			return cli.MessageAndError("file or from-env is required", errors.New("file or from-env is required"))
+		} else if env == "" && fromEnv {
+			return cli.MessageAndError("env-var is required", errors.New("env-var is required"))
+		}
```
rdimitrov, line 55: Let's use cobra to handle that

cmd/cli/app/auth/offline_get.go
```diff
@@ -0,0 +1,78 @@
...
+
+		clientConfig, err := config.ReadConfigFromViper[clientconfig.Config](viper.GetViper())
+		if err != nil {
+			return cli.MessageAndError("Unable to read config", err)
+		}
+
+		f := cmd.Flags().Lookup("file").Value.String()
+		if f == "" {
+			return cli.MessageAndError("file is required", errors.New("file is required"))
+		}
```
rdimitrov, line 51: Again let's use cobra to handle that

cmd/cli/app/auth/offline_get.go
```diff
@@ -0,0 +1,78 @@
...
+		cmd.Printf("Offline token written to %s\n", f)
+
+		return nil
+	},
+}
+
+func init() {
+	OfflineTokenCmd.AddCommand(offlineTokenGetCmd)
+
+	offlineTokenGetCmd.Flags().StringP("file", "f", "offline.token", "The file to write the offline token to")
```
rdimitrov, line 77: We can set the flag as Required here

cmd/cli/app/auth/offline_use.go
```diff
@@ -0,0 +1,107 @@
...
+requiring the user's presence. This is useful for long-running processes
+that need to authenticate to the control plane.`,
+
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		clientConfig, err := config.ReadConfigFromViper[clientconfig.Config](viper.GetViper())
+		if err != nil {
+			return cli.MessageAndError("Unable to read config", err)
+		}
+
+		f := cmd.Flags().Lookup("file").Value.String()
```
rdimitrov, line 48: We should be able to use something like `fileFlag := viper.GetString("file")`

cmd/cli/app/auth/offline_get.go
```diff
@@ -0,0 +1,78 @@
...
+	Short: "Retrieve an offline token",
+	Long: `The minder auth offline-token get command project lets you retrieve an offline token
+for the minder control plane.
+
+Offline tokens are used to authenticate to the minder control plane without
+requiring the user's presence. This is useful for long-running processes
+that need to authenticate to the control plane.`,
+
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		ctx := cmd.Context()
```
rdimitrov, line 41: I realise this was not changed in this PR but should we use the `GRPCClientWrapRunE` where we already get the correct app context, i.e. `ctx, cancel := GetAppContext(cmd.Context(), viper.GetViper())`?

rdimitrov (commented):
Added a few comments mostly around keeping the way we read/set flags consistent with the rest of the CLI commands.

cmd/cli/app/auth/offline_use.go
JAORMX: I tried and it didn't set the default I needed. I could use some help in figuring out why

cmd/cli/app/auth/offline_get.go
JAORMX: That wrapper also wraps authentication using an existing token, in this case we want to customize that. So the wrapper is not applicable here.

rdimitrov: Ah, you're right. Well, at least we can switch to using the proper context, right?

cmd/cli/app/auth/offline_use.go
rdimitrov: That's strange, I did a quick look at the `profile` command as an example for the `provider` flag and we don't do anything different there. If you can't make it work we can go like that and figure this out later.

cmd/cli/app/auth/offline_get.go
JAORMX: Yes we can!

cmd/cli/app/auth/offline_use.go
JAORMX: Lol, that's exactly what I looked at as well... yeah, just couldn't figure it out

rdimitrov: Yeah, leave it like that. I recall that the flags should be bound to Viper for this to work but funny why it's working in one place and not on another.

cmd/cli/app/auth/offline_get.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+// offlineTokenGetCmd represents the offline-token get command
+var offlineTokenGetCmd = &cobra.Command{
+	Use:   "get",
+	Short: "Retrieve an offline token",
+	Long: `The minder auth offline-token get command project lets you retrieve an offline token
+for the minder control plane.
+
+Offline tokens are used to authenticate to the minder control plane without
```
eleftherias, line 37: I'm wondering if this should say "to the minder CLI", since the tokens are scoped to the CLI client. In theory tokens generated by different clients can have different characteristics, although in our current implementation they don't.

eleftherias (commented):
IMO if we're going to expose offline tokens to external users we need to show them how to revoke the tokens.
In theory it's possible now through the keycloak account management,  but we don't direct users there.

JAORMX:
> IMO if we're going to expose offline tokens to external users we need to show them how to revoke the tokens. In theory it's possible now through the keycloak account management, but we don't direct users there.

I agree. And Keycloak has a token revocation endpoint we could use, want me to include that as part of this PR? Was thinking of adding a `minder auth offline-token revoke` command.

eleftherias:
> > IMO if we're going to expose offline tokens to external users we need to show them how to revoke the tokens. In theory it's possible now through the keycloak account management, but we don't direct users there.
> 
> I agree. And Keycloak has a token revocation endpoint we could use, want me to include that as part of this PR? Was thinking of adding a `minder auth offline-token revoke` command.

Yeah I think it should be part of this PR so we don't accidentally release a CLI version without the revocation option

JAORMX:
@eleftherias sounds good! I'll work on that tomorrow.

cmd/cli/app/auth/offline_use.go
```diff
@@ -0,0 +1,112 @@
...
+	Short: "Use an offline token",
+	Long: `The minder auth offline-token use command project lets you install and use an offline token
+for the minder control plane.
+
+Offline tokens are used to authenticate to the minder control plane without
+requiring the user's presence. This is useful for long-running processes
+that need to authenticate to the control plane.`,
+
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		if err := viper.BindPFlags(cmd.Flags()); err != nil {
```
rdimitrov, line 43: Is it needed to bind this twice - on init and also here?

cmd/cli/app/auth/offline_use.go
```diff
@@ -0,0 +1,112 @@
...
+
+		return nil
+	},
+}
+
+func init() {
+	OfflineTokenCmd.AddCommand(offlineTokenUseCmd)
+
+	offlineTokenUseCmd.Flags().StringP("offline-token-path", "f", "offline.token", "The file that contains the offline token")
+	offlineTokenUseCmd.Flags().StringP("offline-token", "t", "",
```
rdimitrov, line 96: Should we rename these to `token` and `token-path` instead? The command is already called "offline-token" and they are available only in it, so we can avoid repeating `offline` in the flag names.

JAORMX:
Just rebased

eleftherias (approved):

PR #2470, opened 2024-03-01
Add new endpoint for fetching rule evaluations

evankanderson:
# Summary

Defines a new API for fetching rule evaluation results, rather than overloading `GetProfileStatusByName`.  Does not provide an implementation, though hopefully the implementation is straightforward.

## Change Type

# Testing

```
make gen
make test
make build
```

# Review Checklist:


dmjb (dismissed):

proto/minder/v1/minder.proto
```diff
@@ -1128,6 +1149,45 @@ message DeleteRuleTypeResponse {
...
+        // Filter profiles to only those matching the specified labels.
+        // 
+        // The default is to return all user-created profiles; the string "*" can
+        // be used to select all profiles, including system profiles.  This syntax
+        // may be expanded in the future.
+        string label_filter = 3;
+    }
+
+    // If set, only return evaluation results for the named rules.
+    // If empty, return evaluation results for all rules
```
rdimitrov, line 1173: I guess it's not intentional but the comment here is the same as the next one

proto/minder/v1/minder.proto
```diff
@@ -1438,6 +1498,11 @@ message Profile {
     // name is the name of the profile instance.
     string name = 3;
 
+    // labels are a set of system-provided attributes which can be used to
+    // filter profiles and status results.  Labels cannot be set by the user,
+    // but are returned in ListProfiles.
+    repeated string labels = 12;
```
JAORMX, line 1504: Are we gonna have labels be a fixed list of strings? or are we going to have labels be a keypair like k8s? This is something I'd like to get consensus on.

eleftherias (commented):
Is this in preparation for label-based profiles? Or something else?

JAORMX:
> Is this in preparation for label-based profiles? Or something else?

yep

evankanderson (commented):
> Are we gonna have labels be a fixed list of strings? or are we going to have labels be a keypair like k8s? This is something I'd like to get consensus on. 

Yeah, that's exactl the sort of thing I wanted to highlight with this API proposal.

I was thinking bag-of-strings, but we could do keypairs if that seems more useful.

JAORMX:
@evankanderson thinking about this further, I don't particularly have a use-case for key-value style labels. So, an array of strings is just as good in this case.

evankanderson:
My first thought was to start with dns-style labels, but I'm wondering if that's too constricting.

JAORMX:
> My first thought was to start with dns-style labels, but I'm wondering if that's too constricting.

I'd say DNS + a minimal subset of characters like "/" and "-" should be fine to start with

rdimitrov (approved):

proto/minder/v1/minder.proto
evankanderson: Added details on the allowed strings.

proto/minder/v1/minder.proto
evankanderson: Oops, thanks, fixed!

evankanderson:
Okay, I updated the comments based on our discussion about K:V maps vs "bag of strings".  We ended up going with the bag of strings approach because all our "values" ended up being singletons (e.g. "true" or "hidden", but no other value).  Also clarified max size and allowed characters for the labels.

proto/minder/v1/minder.proto
JAORMX: For the record, we've agreed that an array of labels was the way to go and we'd have a DNS-style restricted character set to begin with. Thanks @evankanderson for driving this!

JAORMX (approved):

PR #2490, opened 2024-03-05
Cast sql limit to bigint to force sqlc to generate NullInt64

Vyom-Yadav:
# Summary

It becomes tedious designing applications tossing int32 around. The limit sql allows can be an int64. 

## Change Type

# Testing

# Review Checklist:


Vyom-Yadav:
I still see https://github.com/stacklok/minder/blob/e68fcce36136bd2a9822846458282e26a13619bf/proto/buf.yaml#L23-L26
#2415 has been merged, works for this PR too

proto/minder/v1/minder.proto
```diff
@@ -793,7 +793,7 @@ message DeleteRepositoryByNameResponse {
 
 message ListRepositoriesRequest {
     string provider = 1 [deprecated=true];
-    int32 limit = 3;
+    int64 limit = 3;
```
jhrozek, line 796: This /seems/ to be OK based on https://protobuf.dev/programming-guides/proto3/#updating which says "int32, uint32, int64, uint64, and bool are all compatible – this means you can change a field from one of these types to another without breaking forwards- or backwards-compatibility. If a number is parsed from the wire which doesn’t fit in the corresponding type, you will get the same effect as if you had cast the number to that type in C++ (for example, if a 64-bit number is read as an int32, it will be truncated to 32 bits)." but I want to add an explicit comment and leave the PR around for a little bit so that someone else with a better protobuf fu can take a look, too.

database/query/repositories.sql
```diff
@@ -30,7 +30,7 @@ SELECT * FROM repositories
 WHERE provider = $1 AND project_id = $2
   AND (repo_id >= sqlc.narg('repo_id') OR sqlc.narg('repo_id') IS NULL)
 ORDER BY project_id, provider, repo_id
-LIMIT sqlc.narg('limit');
+LIMIT sqlc.narg('limit')::bigint;
```
jhrozek, line 33: @dmjb can you double check this change please?

proto/minder/v1/minder.proto
Vyom-Yadav: I'll also add: https://buf.build/docs/breaking/rules#field_wire_json_compatible_type

Vyom-Yadav: Hmm, we never invoke `buf breaking` in the CI https://buf.build/docs/reference/cli/buf/breaking. Let me check whether this is ok or not (will add to CI too)

https://protobuf.dev/programming-guides/proto3/#json:

> | proto3 | JSON | JSON example	 | Notes	 |
> |--------|--------|--------|--------|
> | int32, fixed32, uint32 | number | `1, -10, 0` | JSON value will be a decimal number. Either numbers or strings are accepted. |
> | int64, fixed64, uint64 | string | `"1", "-10"` | JSON value will be a decimal string. Either numbers or strings are accepted. |

Vyom-Yadav: It fails _without_ `FIELD_WIRE_JSON_COMPATIBLE_TYPE` :

```
$ buf breaking proto/minder/v1/minder.proto --against '.git#branch=main'
proto/minder/v1/minder.proto:796:5:Field "3" on message "ListRepositoriesRequest" changed type from "int32" to "int64". See https://developers.google.com/protocol-buffers/docs/proto3#updating for wire compatibility rules and https://developers.google.com/protocol-buffers/docs/proto3#json for JSON compatibility rules.
```
So this is a breaking change.

jhrozek: oh is it a breaking change because of the change in JSON? I was going based on the proto3-updating link which says "int32, uint32, int64, uint64, and bool are all compatible – this means you can change a field from one of these types to another without breaking forwards- or backwards-compatibility. " so I'm confused.

@evankanderson thoughts?

Vyom-Yadav: It mentions on top (https://protobuf.dev/programming-guides/proto3/#updating):

> Note
> If you use JSON or [proto text format](https://protobuf.dev/reference/protobuf/textformat-spec) to store your protocol buffer messages, the changes that you can make in your proto definition are different.

But IMO the docs should be improved, it should define what is _different_.

dmjb: I believe @evankanderson enabled the `FIELD_WIRE_JSON_COMPATIBLE_TYPE` - he may thoughts on this thread.

dmjb (approved):

dmjb (changes requested):
We're going to discuss this internally today to ensure that there are no backwards incompatibility issues.

Vyom-Yadav:
> We're going to discuss this internally today to ensure that there are no backwards incompatibility issues.

I'm okay with just changing it for the DB and not the API. We can upcast internally.

proto/minder/v1/minder.proto
evankanderson: When converting from int32 to int64, the proto form of the _output_ will change from string to int.  It will accept both JSON numbers (IEEE 64-bit ints) and numeric strings as input.  Proto does this for JSON serialization because JSON numbers will lose precision around 2^53, and the proto team feels that is not acceptable.

Given that we've got other limits that probably mean limit should be < 2**16, I'm indifferent as to whether this is an int64 or stays an int32.

internal/controlplane/handlers_repositories.go
```diff
@@ -165,13 +165,13 @@ func (s *Server) ListRepositories(ctx context.Context,
...
 
-	limit := sql.NullInt32{Valid: false, Int32: 0}
+	limit := sql.NullInt64{Valid: false, Int64: 0}
 	reqLimit := in.GetLimit()
 	if reqLimit > 0 {
 		if reqLimit > maxFetchLimit {
 			return nil, util.UserVisibleError(codes.InvalidArgument, "limit too high, max is %d", maxFetchLimit)
 		}
-		limit = sql.NullInt32{Valid: true, Int32: reqLimit + 1}
+		limit = sql.NullInt64{Valid: true, Int64: reqLimit + 1}
```
evankanderson, line 174: Does this mean that a user who doesn't pass in `limit` gets an unlimited list, but if they pass in a `limit`, we limit them to 100?

It feels like we should _always_ enforce a limit, so maybe the following code would be simpler?

```golang
limit := maxFetchLimit
if in.GetLimit() > 0 && in.getLimit < maxFetchLimit {
	limit = in.GetLimit()
}
```

And then later on, we could just check `if len(repos) == limit` on line 204.

evankanderson (commented):
I'm not going to block this, though if you want to get past the `buf` checks, you'll need to temporarily disable the FIELD_WIRE_JSON_COMPATIBLE_TYPE check, and then re-enable it in a subsequent PR.  Since this is an in-value, the output change shouldn't be a big deal.

On the other hand, returning more than 2^31 rows is probably a problem _anyway_, and it might make more sense to enforce a limit on our `List` queries rather than using `sql.narg`.  That would simplify our types and allow us to just use regular `int` types without a specific size constraint.

database/query/repositories.sql
Vyom-Yadav: > do we ever want to do a single select that fetches all repositories

Technically, we shouldn't, but the minder CLI uses this endpoint, and the CLI still needs to support paginating results.

> do we want to force clients to paginate?

We don't force the client to paginate.

internal/controlplane/handlers_repositories.go
Vyom-Yadav: I agree, but the minder CLI doesn't support paginating results. `minder repo list` is why we check that way.

database/query/repositories.sql
evankanderson: > > do we want to force clients to paginate?
>
> We don't force the client to paginate.

My point was that we _could_ enforce this in the API.

But, we'll do that in a subsequent PR.

evankanderson (approved):

Vyom-Yadav:
@evankanderson Resolved the merge conflicts, so it dismissed your review.

evankanderson (approved):

PR #2505, opened 2024-03-05
Add repo reconcile command to minder CLI

Vyom-Yadav:
# Summary

Part of #2262. Adds a one-shot reminder command for CLI users. Only an `admin` can create a repo reconciliation request. 

## Change Type

# Testing

# Review Checklist:


Vyom-Yadav:
> coverage: 39.19% (-0.2%) from 39.348%

I'll add tests. Marked as a draft for now, but still open to reviews or suggestions.

Vyom-Yadav:
> === NAME  TestVerifyMultipleProjects
>     authz_test.go:142: creating connection to address 0.0.0.0:40555
>     authz_test.go:148: 
>         	Error Trace:	/home/runner/work/minder/minder/internal/authz/authz_test.go:148
>         	Error:      	Received unexpected error:
>         	            	error while listing authz stores: Get "http://0.0.0.0:40451/stores": dial tcp 0.0.0.0:40451: connect: connection

Failing quite a lot, I pushed 3-4 times, so I'll leave this as it is.

internal/authz/model/minder.generated.json
evankanderson, line 1: We should add this file to `.gitattributes` as `linguist-generated=true`, so it's hidden from diffs by default.

proto/minder/v1/minder.proto
```diff
@@ -674,6 +675,21 @@ service ProvidersService {
...
+            post: "/api/v1/reconciliation/repository"
+            body: "*"
+        };
+
+        option (rpc_options) = {
+            target_resource: TARGET_RESOURCE_PROJECT
+            relation: RELATION_REPO_RECONCILIATION_TASK_CREATE
+        };
+    }
+}
```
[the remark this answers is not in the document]
evankanderson: If we put this under `ProfileService`, it would be more clear that the reconciliation tasks were associated with profiles and executing the policy -- the fact that this doesn't have a clear "parent" in the service suggests that maybe we're not clear about how this is associated with the different resources.

On the other hand, if this executes multiple Profiles, we might want to put it directly under `ProjectsService`, which I think we recently introduced.

evankanderson (commented):
Sorry about the delay on review for this; I think we all were looking at #2475 

We've talked a lot about having a one-shot "evaluate" command for policies, and this looks like a good step in the right direction.  Thanks!

I have a few concrete concerns about _where_ we put this API in our API space, but I'm perfectly okay with having a lot of restrictions on the initial implementation -- for example, that it only works on repos, and that it applies all profiles to that one repo object.

(I'm thinking about how you can avoid needing to add a lot of code to the `internal/controlplane` part of this PR, while still giving us room to grow this later.)

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
Vyom-Yadav: Good points. The current eventing system we have in place does not support reconciling on a single profile or rule, so if we want to go that route, we would have to make changes to the main eventing system in order to reduce duplication. 

The current model runs _all profiles in a project_ on the entity we supply. I moved it under `ProjectsService` with a new relation `entity_reconciliation_task_create: editor`.

internal/authz/model/minder.fga
```diff
@@ -90,4 +90,6 @@ type project
     define profile_update: editor or policy_writer
     define profile_delete: editor or policy_writer
 
-    define profile_status_get: viewer
\ No newline at end of file
+    define profile_status_get: viewer
+
+    define repo_reconciliation_task_create: admin
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

cmd/cli/app/repo/repo_reconcile.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	fmt.Printf("Reconciliation task created: %s\n", resp.GetTaskId())
+	return nil
+}
+
+func init() {
+	RepoCmd.AddCommand(reconcileCmd)
+	reconcileCmd.Flags().StringP("name", "n", "", "Name of the repository (owner/repo)")
+	reconcileCmd.Flags().StringP("id", "i", "", "ID of the repository")
```
Vyom-Yadav, line 94: I have added reconciliation to the existing `minder repo` cmd instead of creating a new command. If we had gone with something like:

`minder entity reconcile`: This would break how our CLI is structured right now. We usually follow `minder <entity> <action>` type of structure. Having a generic `entity` subcommand didn't look good to me.

`minder reconcile --entity --id`: I am not a big fan of this structure. We have designated commands for entities, like `repo`, `artifacts`, so if we follow this, the user can expect `minder list repo/artifacts` to work, which is not how our CLI works.

---

Later we can add it to `minder artifacts`, etc.

cmd/cli/app/reminder/reminder_create.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/stacklok/minder/cmd/cli/app"
+	"github.com/stacklok/minder/internal/util/cli"
+	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+)
+
+var createCmd = &cobra.Command{
+	Use:   "create",
+	Short: "Creates a reminder",
+	Long:  `The reminder create subcommand is used to create a reminder (reconciliation request).`,
```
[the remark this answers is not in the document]
Vyom-Yadav: I went with `minder repo reconcile`. See https://github.com/stacklok/minder/pull/2505#discussion_r1521455172 for rationale.

proto/minder/v1/minder.proto
```diff
@@ -1694,6 +1707,19 @@ message DeleteProjectResponse {
...
+    // entity is the entity to be reconciled.
+    EntityTypedId entity = 1;
+
+    // context is the context in which the entity reconciliation task is created.
+    Context context = 2;
+}
+
+message CreateEntityReconciliationTaskResponse {
+    // task_id is the id of the task that was created.
+    string task_id = 1;
```
evankanderson, line 1720: Do we provide any use for this `task_id` handle?  I'd rather not provide a handle to internal state if there's nothing to be done with it.

internal/controlplane/handlers_reconciliationtasks.go
```diff
@@ -0,0 +1,102 @@
...
+	// TODO: Support other entity types, replace with switch
+	if entity.GetType() == pb.Entity_ENTITY_REPOSITORIES {
+		msg, err = s.getRepositoryReconciliationMessage(ctx, entity)
+		if err != nil {
+			return nil, err
+		}
+		topic = reconcilers.InternalReconcilerEventTopic
+	} else {
+		return nil, status.Errorf(codes.InvalidArgument, "entity type %s is not supported", entity.GetType())
+	}
```
evankanderson, line 64: Non-blocking, but you might want to use a `switch` statement here.

internal/controlplane/handlers_reconciliationtasks.go
```diff
@@ -0,0 +1,102 @@
...
+	var msg *message.Message
+	var topic string
+
+	// TODO: Support other entity types, replace with switch
+	if entity.GetType() == pb.Entity_ENTITY_REPOSITORIES {
+		msg, err = s.getRepositoryReconciliationMessage(ctx, entity)
+		if err != nil {
+			return nil, err
+		}
+		topic = reconcilers.InternalReconcilerEventTopic
```
evankanderson, line 61: I think we need to check that `entity` is owned by `entityCtx`, or we get a repeat of https://github.com/stacklok/minder/security/advisories/GHSA-v627-69v2-xx37.  (Yeah, this is trickier than it should be...)

evankanderson: We probably also want to check that `entity` exists here, so at least the query for ownership also lets us alert if the target doesn't exist.

evankanderson: Oh, it turns out that `getRepositoryReconciliationMessage` actually uses the server's database handle to fetch the repository!

Can we pass the `store` in explicitly, and make `getRepositoryReconciliationMessage` a free function (not associated with the `Server` struct)?  Right now, there's a lot of functionality tied up in `Server`, and I'd like to push towards _less_ coupling.

internal/controlplane/handlers_reconciliationtasks.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, status.Errorf(codes.InvalidArgument, "error parsing repository id: %v", err)
+	}
+
+	repo, err := s.store.GetRepositoryByID(ctx, repoUUID)
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil, status.Errorf(codes.NotFound, "repository not found")
+	} else if err != nil {
+		return nil, status.Errorf(codes.Internal, "cannot read repository: %v", err)
+	}
+
```
evankanderson, line 90: You could compare `repo.ProjectID` with `entityCtx.Project.ID` here to check the ownership (see my other comment).

evankanderson (commented):
Adding the stand-along `reconcile` command to `repo` makes sense to me for the time being.

Two other small comments, but this looks pretty much good to go.  Thank you!

internal/controlplane/handlers_reconciliationtasks.go
Vyom-Yadav: > we get a repeat of [GHSA-v627-69v2-xx37](https://github.com/stacklok/minder/security/advisories/GHSA-v627-69v2-xx37)

I also thought of that vulnerability with `GetRepositoryByID(ctx, repoUUID)`. But the only reason why I didn't check the owner was assuming that `id` is an internal state (sort of secretive), unlike the public repo name. If we want to verify that only the owner can fetch the repo using `id`, we should _add the project and provider to the query_ itself and not hack around.

internal/controlplane/handlers_reconciliationtasks.go
dmjb: It should be possible to use the `GetRepositoryByIDAndProject` query on line 84

Vyom-Yadav: Updated to use that

internal/controlplane/handlers_reconciliationtasks.go
Vyom-Yadav: It's already in todo because lint complains and asks for other cases. I don't want to add all that fallthrough and all cases. Else if fine for now

proto/minder/v1/minder.proto
Vyom-Yadav: Removed `task_id`

Vyom-Yadav:
Updated, PTAL

evankanderson (approved):

evankanderson:
Sorry about the delay; we had a few items to iron out outside of Minder itself yesterday, and I didn't get to the code reviews I wanted to.

PR #2546, opened 2024-03-07
[WIP] Move repo creation/deletion into separate interface

dmjb:
EDIT: Based on feedback from @evankanderson, I have decided to break this PR up into smaller chunks. Hold off on review for now.

# Summary

This relates to ongoing test/refactoring work.

Move the methods for repo creation and deletion behind their own
interfaces, and add test for the new code and the controlplane endpoints
which use those methods.

In future, we may want to incorporate this Repository interface into the
providers somehow so that the controlplane can create repos for
different types of source control systems without needing to know which
exact type they are.

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/repositories/github/service.go
```diff
@@ -0,0 +1,284 @@
...
+
+func (r *repositoryService) DeleteRepositoryByName(
+	ctx context.Context,
+	client ghclient.GitHubRepoClient,
+	projectID uuid.UUID,
+	repoOwner string,
+	repoName string,
+) error {
+	// assumption: provider name should always be `ghprovider.Github` for this Github-specific code
+	params := db.GetRepositoryByRepoNameParams{
```
dmjb, line 154: I am making the explicit assumption that this code only works with Github (after all, it only accepts intstances of the Github client from the Github provider). If we add additional providers in future (e.g. Gitlab) we should have different interfaces for interacting with other providers, and any shareable code should be factored out.

internal/controlplane/handlers_repositories.go
```diff
@@ -483,30 +393,69 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
 	return out, nil
 }
 
-func (s *Server) deleteRepositoryAndWebhook(
+// TODO: this probably can probably be used elsewhere
```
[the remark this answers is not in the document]
evankanderson: Thank you, though I still see a diff removing this function.  (And maybe GitHub is trying too hard to line up unrelated changes?)

internal/controlplane/handlers_repositories.go
```diff
@@ -483,30 +393,68 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+			codes.InvalidArgument,
+			"provider %s is not supported for github webhook",
+			p.GetName(),
+		)
+	}
+
+	client, err := p.GetGitHub()
+	if err != nil {
+		return uuid.Nil, nil, status.Errorf(codes.Internal, "error creating github provider: %v", err)
+	}
```
evankanderson, line 431: Do we need both the `!p.Implements(...)` check _and_ the error handling from `p.GetGitHub`?  I'd hope that the second would return error if unsupported.

evankanderson (commented):
A few comments (and I shared them on Discord), but GitHub does not make it easy to review large diffs like this that include both moves and refactoring.  If you can get it down to <600 combined `+` and `-` in a change (ignoring generated), I'll probably have an easier time.

One way to do this would be to just extract one or two methods at a time -- for example, the `registerWebhookForRepository` that's in `handlers_githubwebhooks.go` is a prime candidate to be moved elsewhere, possibly into `webhookManager`, for example.

PR #2565, opened 2024-03-08
Implement a HTTP PATCH method for updating profiles

jhrozek:
# Summary

Updating the whole profile by POST-ing the profile contents wouldn't
work well for the UI which would have to keep the whole profile around.
Instead, let's implement partial updates with PATCH.

At the moment, only remediate and alert settings updates are
implemented.

Fixes: https://github.com/stacklok/epics/issues/236

## Change Type

# Testing

You can use the following curl to test:
```
curl -XPATCH -H "Authorization: Bearer $MINDER_BEARER_TOKEN" 'http://localhost:8080/api/v1/profile?context.provider=github&context.project=a34dc21e-0164-4508-ada1-32419bc72d65&id=67c550c7-74cd-4bd6-a6f4-78ffc65d6af9' -d '{   "alert": "off" }'
```

the returned profile is the fully updated one

# Review Checklist:


jhrozek:
cc @davolokh

proto/minder/v1/minder.proto
```diff
@@ -920,6 +933,25 @@ message UpdateProfileResponse {
     Profile profile = 1;
 }
 
+message PatchProfileRequest {
+    // The context in which the patch is applied. Provided explicitly
+    // so that the patch itself can be minimal and contain only
+    // the attribute to set, e.g. remediate=true
+    Context context = 1;
```
eleftherias, line 940: If the context is required, isn't it the same if the frontend passes in the nested context under profile?
Or is it not enforced if it's nested?

jhrozek: I thought it would be confusing to include it in the body that represents what is going to be patched (changed), same as the ID.

btw @evankanderson pointed me to [https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/patch_feature/](the grpc PATCH) docs that say that the method is confusingly not supposed to be `patch` but `put`. I'll test how the method behaves then.

(I'll welcome another opinion on the separate context and ID in the meantime of course)

jhrozek: I shouldn't reply on a Friday evening, sorry

> I thought it would be confusing to include it in the body that represents what is going to be patched (changed), same as the ID.
> 
> btw @evankanderson pointed me to [grpc patch docs](https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/patch_feature) that say that the method is confusingly not supposed to be `patch` but `put`. I'll test how the method behaves then.
>

I misread Evan's link, the method should be patch, what the example really had was:
```
rpc UpdateV2(UpdateV2Request) returns (google.protobuf.Empty) {
 option (google.api.http) = {
   put: "/v2/example/a_bit_of_everything/{abe.uuid}"
   body: "abe"
   additional_bindings {
     patch: "/v2/example/a_bit_of_everything/{abe.uuid}"
     body: "abe"
   }
 };
}
```

but both your comment and Evan's link make me think that just reusing the Update method and require an update field might be better. I'll try going down that route and push a new patch version.

internal/controlplane/handlers_profile.go
```diff
@@ -577,6 +577,54 @@ func (s *Server) GetProfileStatusByProject(ctx context.Context,
...
+		return nil, status.Errorf(codes.InvalidArgument, "error in entity context: %v", err)
+	}
+
+	if ppr.GetId() == "" {
+		return nil, util.UserVisibleError(codes.InvalidArgument, "profile ID must be specified")
+	}
+
+	profileID, err := uuid.Parse(ppr.GetId())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Malformed UUID")
```
JAORMX, line 596: Should this be a user visible error?

internal/controlplane/handlers_profile.go
```diff
@@ -577,6 +577,54 @@ func (s *Server) GetProfileStatusByProject(ctx context.Context,
...
+		params.Remediate = validateActionType(patch.GetRemediate())
+	}
+	if patch.Alert != nil {
+		params.Alert = validateActionType(patch.GetAlert())
+	}
+
+	// Update top-level profile db object
+	_, err = s.store.UpdateProfile(ctx, params)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error updating profile: %v", err)
```
JAORMX, line 613: Should this be use visible?

proto/minder/v1/minder.proto
jhrozek: In the end I settled on `"/api/v1/profile/{id}` handling PATCH and accepting the context in a separate part of the call such as query parameters.

I still think the explicit context is somehow cleaner than the alternative which would be to pass both the context and the patch in the same message:
```
curl -XPATCH -H "Authorization: Bearer $MINDER_BEARER_TOKEN" 'http://localhost:8080/api/v1/profile' -d '{ "patch": {"context": { "provider": "github", "project": "a34dc21e-0164-4508-ada1-32419bc72d65" },"id": "9ad30fe9-d1ba-495c-9ca4-f99a97374d61", "remediate":"on" }}'
```
basically it seems like the case with the embedded context is useful for a git-ops flow like you said on Slack the other day, but I'm not sure if the patch method would be a part of such workflow, it seems really more like an imperative call. That said, I do have a local branch implementing that.

JAORMX (approved):

jhrozek:
@JAORMX I just resolved a conflict, can you re-approve? No other changes.

jhrozek:
done!

internal/controlplane/handlers_profile.go
```diff
@@ -578,6 +578,54 @@ func (s *Server) GetProfileStatusByProject(ctx context.Context,
...
+	if ppr.GetId() == "" {
+		return nil, util.UserVisibleError(codes.InvalidArgument, "profile ID must be specified")
+	}
+
+	profileID, err := uuid.Parse(ppr.GetId())
+	if err != nil {
+		return nil, util.UserVisibleError(codes.InvalidArgument, "Malformed UUID")
+	}
+
+	params := db.UpdateProfileParams{ID: profileID}
```
JAORMX, line 600: where do we set the `project` for the given profile? Seems to me like we should add it to the `UpdateProfileParams` to ensure that we only update the profile if it belongs to the project in the context.

JAORMX (approved):

PR #2566, opened 2024-03-09
Store expected GitHub userid in database during enroll

evankanderson:
# Summary

It turns out that it's possible (though unintended) to hand off an enrollment URL from the CLI to another user or profile than the one logged in to Minder.  Prevent this by storing the GitHub ID of the active Minder user as part of the OAuth enrollment state.

As part of this change, we need KeyCloak to fetch and store the GitHub user ID (unchanging).  For later use in improving the auth grant flow, I also added the GitHub user login (human-readable).  As these made the keycloak configuration much more complex, I borrowed `keycloak-config-cli` used in the Bitnami chart to simplify moving configuration from dev to production.  I can separate this change if needed.

I also adjusted a few interfaces that were either overly constrained (getting token information in the context was broadened to be able to get other claims, and a database row that was part of the Provider interface was simplifed to a `sql.NullString`, though maybe it should simply be a `string`).

## Change Type

# Testing

Manual testing, though I intend to add unit tests to raise coverage for `handlers_oauth.go` to at least 60%.

# Review Checklist:


.mk/identity.mk
```diff
@@ -22,8 +22,12 @@ ifndef KC_GITHUB_CLIENT_SECRET
 	$(error KC_GITHUB_CLIENT_SECRET is not set)
 endif
 	@echo "Setting up GitHub login..."
-	@$(CONTAINER) exec -it keycloak_container /opt/keycloak/bin/kcadm.sh config credentials --server http://localhost:8080 --realm master --user admin --password admin
+#	@$(CONTAINER) exec -it keycloak_container /opt/keycloak/bin/kcadm.sh config credentials --server http://localhost:8080 --realm master --user admin --password admin
```
evankanderson, line 25: Since the health check runs `config credentials` now, we don't need this line anymore.

database/query/session_store.sql
```diff
@@ -1,20 +1,14 @@
 -- name: CreateSessionState :one
-INSERT INTO session_store (provider, project_id, session_state, owner_filter, redirect_url) VALUES ($1, $2, $3, $4, $5) RETURNING *;
-
--- name: GetSessionState :one
-SELECT * FROM session_store WHERE id = $1;
-
--- name: GetSessionStateByProjectID :one
-SELECT * FROM session_store WHERE project_id = $1;
```
evankanderson, line 8: These queries were unused.

internal/auth/jwtauth.go
```diff
@@ -105,14 +105,33 @@ var userSubjectContextKey struct{}
...
+	data, ok := token.Get(claim)
+	if !ok {
+		return ret, false
+	}
+	ret, ok = data.(T)
+	return ret, ok
+}
+
+// WithAuthTokenContext stores the specified user subject in the context.
+func WithAuthTokenContext(ctx context.Context, subject openid.Token) context.Context {
```
eleftherias, line 135: Since we now store the token and not just the subject, the godocs, parameter and function names should also change. For example, `userSubjectContextKey` should be `userTokenContextKey`

internal/controlplane/handlers_oauth.go
```diff
@@ -218,6 +243,20 @@ func (s *Server) generateOAuthToken(ctx context.Context, provider string, code s
 		return fmt.Errorf("error exchanging code for token: %w", err)
 	}
 
+	// Older enrollments may not have a RemoteUser stored; these should age out fairly quickly.
+	tokenMatches.Add(ctx, 1, metric.WithAttributes(
```
eleftherias, line 247: Should this come after we validate the user is correlated successfully? (after line 258). If we keep it here we're incrementing even if the user attempting to use the state doesn't match the user who initiated the flow.

eleftherias (changes requested):

internal/controlplane/handlers_oauth.go
evankanderson: I wanted to know how many times when we attempt to validate a token against a user's ID.  If that's 100% of the time, then we can make the `remoteUser` field mandatory.  If it's less than that, I'll need to figure out why.

internal/auth/jwtauth.go
evankanderson: I did change the function name, I think, except that we now have one "store" and two "loads":

* `GetUserSubjectFromContext` returns the `sub` claim
* `GetUserClaimFromContext` returns other claims, or `"", false` if not present.

internal/controlplane/handlers_oauth.go
evankanderson: I added a counter on the GetAuthorizationURL RPC as well, so we can correlate the number of requested tokens that store `gh_id` with the number that get claimed later.

eleftherias (dismissed):

internal/auth/jwtauth.go
```diff
@@ -101,18 +101,37 @@ func NewJwtValidator(ctx context.Context, jwksUrl string) (JwtValidator, error)
...
+var userTokenContextKey struct{}
 
 // GetUserSubjectFromContext returns the user subject from the context, or nil
 func GetUserSubjectFromContext(ctx context.Context) string {
-	subject, ok := ctx.Value(userSubjectContextKey).(string)
+	token, ok := ctx.Value(userTokenContextKey).(openid.Token)
 	if !ok {
+		fmt.Printf("***\n")
+		fmt.Printf("no token in coGntext\n")
+		fmt.Printf("***\n")
```
JAORMX, line 112: I'm guessing this is a leftover from debugging?

internal/controlplane/handlers_oauth.go
```diff
@@ -39,9 +43,25 @@ import (
...
 	"github.com/stacklok/minder/internal/engine"
 	"github.com/stacklok/minder/internal/logger"
+	"github.com/stacklok/minder/internal/providers"
 	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 )
 
+// Track how often users who register a token are correlated with the
+// GitHub user from GetAuthorizationURL
+var tokenMeter = otel.Meter("enrolled-token")
+var tokenMatches metric.Int64Counter
```
JAORMX, line 53: we have a metrics construct for the controlplane https://github.com/stacklok/minder/blob/main/internal/controlplane/metrics/metrics.go should this be moved there?

internal/providers/providers.go
```diff
@@ -55,14 +56,14 @@ func GetProviderBuilder(
...
-	return NewProviderBuilder(&prov, encToken, decryptedToken.AccessToken, opts...), nil
+	return NewProviderBuilder(&prov, encToken.OwnerFilter, decryptedToken.AccessToken, opts...), nil
 }
 
 // ProviderBuilder is a utility struct which allows for the creation of
 // provider clients.
 type ProviderBuilder struct {
 	p               *db.Provider
-	tokenInf        db.ProviderAccessToken
+	ownerFilter     sql.NullString // NOTE: we don't seem to actually use the null-ness anywhere.
```
JAORMX, line 66: we'll get rid of ownerFilter at some point. This is more of a provider specific configuration rather than for the token, but it ended up there because the token table came first.

internal/controlplane/handlers_repositories_test.go
```diff
@@ -236,6 +236,7 @@ type EventPayload struct {
 	Repository int
 }
 
+// NOTE: this stub is also used by handlers_oauth_test.go ¯\_(ツ)_/¯
```
JAORMX, line 239: seems we have a bunch of these all over the place. Besides this we have:
* https://github.com/stacklok/minder/blob/main/internal/eea/eea_test.go#L508-L535
* https://github.com/stacklok/minder/blob/main/internal/util/testqueue/passthroughqueue.go

Not as part of this PR, but it would be good to push these test utilities to a package that's easier to find.

JAORMX (changes requested):
Seems you gotta rebase and increase the migration number

internal/auth/jwtauth.go
evankanderson: Oops, thanks!

internal/providers/providers.go
evankanderson: I suspect we want to remove SQL-isms from the provider interface at some point, but I didn't want to boil that ocean here.  At the same time, the `ownerFilter` was the only part of the access token which was being used, so I wanted to narrow that interface to exclude the other rows.

internal/controlplane/handlers_oauth.go
evankanderson: I put this here in hopes of avoiding as much entanglement across objects, but I've moved it to metrics.  When in Rome...

internal/controlplane/handlers_repositories_test.go
evankanderson: I'd love to move much of the implementations out of `controlplane/handlers_`, and I expect that normalizing these (or reducing their scope) will be part of that.

internal/providers/providers.go
JAORMX: that's fine, it is what it is.

evankanderson:
Updated; PTAL.

JAORMX (approved):

JAORMX:
aaaagh, that linter lol

evankanderson:
> aaaagh, that linter lol

No, I screwed up the tests, too.

evankanderson:
Okay, fixed the test failures... I'd missed adding the metric method to the interface (another 4 lines of code!) and then needed to change `GetProviderByName`, whose type signature had changed.

evankanderson:
Merged again after @dmjb 's changes produced a conflict.

I'd like to get this in before I end up in perpetual merge-cycles.  I think this is safe all the way out to production as currently writ.  (I had a different opinion this morning, but realized I was incorrect.)

eleftherias (changes requested):
This breaks the `minder auth delete` command. I haven't looked into which line change causes this, but the error is:
```
{"level":"error","Resource":{"service":"minder.v1.UserService","method":"DeleteUser"},"login_sha":"0b063c7365d9e6ac462a119b005bdb173a8ce2bb9c30753891819a5d45448fda","telemetry":"true","Attributes":{"http.code":"Internal","http.content-type":["application/grpc"],"http.duration":"20.078208ms","http.forwarded":[],"http.user_agent":["minder-cli/v0.0.34+ref.188ff0637 (darwin) arm64 (go1.21.7) grpc-go/1.62.1"],"exception.message":"rpc error: code = Internal desc = failed to get client access token: oauth2: \"unauthorized_client\" \"Invalid client or Invalid client credentials\""},"Timestamp":1710320062941465585}
```

JAORMX:
> This breaks the `minder auth delete` command. I haven't looked into which line change causes this, but the error is:
> 
> ```
> {"level":"error","Resource":{"service":"minder.v1.UserService","method":"DeleteUser"},"login_sha":"0b063c7365d9e6ac462a119b005bdb173a8ce2bb9c30753891819a5d45448fda","telemetry":"true","Attributes":{"http.code":"Internal","http.content-type":["application/grpc"],"http.duration":"20.078208ms","http.forwarded":[],"http.user_agent":["minder-cli/v0.0.34+ref.188ff0637 (darwin) arm64 (go1.21.7) grpc-go/1.62.1"],"exception.message":"rpc error: code = Internal desc = failed to get client access token: oauth2: \"unauthorized_client\" \"Invalid client or Invalid client credentials\""},"Timestamp":1710320062941465585}
> ```

just pushed a fix.

eleftherias (approved):

Issue #2518, opened 2024-03-06
Use suitable enrollment flow based on the provider


eleftherias:
Depends on: #2517

The current mechanism for enrolling a provider and setting up authorization gives the user 2 options. They can either supply a personal access token, or they go through an OAuth 2 authorization flow.

The CLI should not hardcode these steps and should instead query the provider to see which authorization flows it supports.

PR #2601, opened 2024-03-12
Actually check for auth flows in provider enrollment

JAORMX:
# Summary

This modifies the provider enrollment to actually verify that the provider supports the given auth flows before attempting them.

This is enforced both in the CLI and the server.

Closes: https://github.com/stacklok/minder/issues/2518

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_oauth.go
+++ b/internal/controlplane/handlers_oauth.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"slices"
 
 	"github.com/google/uuid"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
@@ -39,6 +40,7 @@ import (
 	"github.com/stacklok/minder/internal/db"
 	"github.com/stacklok/minder/internal/engine"
 	"github.com/stacklok/minder/internal/logger"
+	"github.com/stacklok/minder/internal/util"
 	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 )
 
@@ -54,6 +56,11 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
 	provider, err := getProviderFromRequestOrDefault(ctx, s.store, req, projectID)
 	if err != nil {
 		return nil, providerError(err)
+	}
+
+	if !slices.Contains(provider.AuthFlows, db.AuthorizationFlowOauth2AuthorizationCodeFlow) {
+		return nil, util.UserVisibleError(codes.InvalidArgument,
+			"provider does not support authorization code flow")
 	}
 
 	// Configure tracing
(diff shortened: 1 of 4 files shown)


internal/controlplane/handlers_oauth.go
```diff
@@ -288,6 +295,11 @@ func (s *Server) StoreProviderToken(ctx context.Context,
 		return nil, providerError(err)
 	}
 
+	if !slices.Contains(provider.AuthFlows, db.AuthorizationFlowUserInput) {
```
jhrozek, line 298: why not call p.SupportsAuthFlow() here? is it because one is using a protobuf constant and the other a db constant? Since we're using the constant verbatim and not from a variable, could we use the protobuf constant here as well?

jhrozek (approved):
it would be nice to reuse the utility function but that's a non-blocking comment

internal/controlplane/handlers_oauth.go
evankanderson: I think `getProviderFromRequestOrDefault` actually returns a `db.Provider`, not a proto object mentioned in `SupportsAuthFlow`.

With that said, I'd prefer to have it return something (maybe a `ProviderBuilder`?) that encapsulates this check more than `slices.Contains` on a returned database row.

internal/controlplane/handlers_oauth.go
```diff
@@ -56,6 +58,11 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
 		return nil, providerError(err)
 	}
 
+	if !slices.Contains(provider.AuthFlows, db.AuthorizationFlowOauth2AuthorizationCodeFlow) {
```
evankanderson, line 61: Can we use `provider.SupportsAuthFlow(...)` here?  I think this is the same comment jakub had down below.

Ideally, `provider` returned from `getProviderFromRequestOrDefault` would be either an actual Provider instance, or at least a ProviderBuilder instance.  But that's a larger refactor.

pkg/api/protobuf/go/minder/v1/providers.go
```diff
@@ -23,3 +25,8 @@ func (provt ProviderType) ToString() string {
 func (a AuthorizationFlow) ToString() string {
 	return enumToStringViaDescriptor(a.Descriptor(), a.Number())
 }
+
+// SupportsAuthFlow returns true if the provider supports the given auth flow
+func (p *Provider) SupportsAuthFlow(flow AuthorizationFlow) bool {
+	return slices.Contains(p.GetAuthFlows(), flow)
+}
```
evankanderson, line 32: I don't see where this is used?

JAORMX: It is in the CLI

internal/controlplane/handlers_oauth.go
JAORMX: Should we have a separate PR with that refactor?

jhrozek: I'm fine with a separate PR if it comes soon and we don't propagate code patters like this over time

JAORMX: on it.

PR #2608, opened 2024-03-12
Move repo deletion into a separate interface

dmjb:
This is the first of a two part process of moving repo management logic into a separate interface. This PR moves the logic for the two repo deletion endpoints behind an interface, and provides tests for both the deletion logic as well as the endpoints which call it.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/repositories/github/service.go
```diff
@@ -0,0 +1,136 @@
...
+	webhookManager webhooks.WebhookManager
+	store          db.Store
+	eventProducer  events.Interface
+}
+
+// NewRepositoryService creates an instance of the RepositoryService interface
+func NewRepositoryService(
+	webhookManager webhooks.WebhookManager,
+	store db.Store,
+	eventProducer events.Interface,
```
dmjb, line 64: The event producer is not used in this PR, it will be used in the next PR.

internal/controlplane/handlers_repositories.go
```diff
@@ -312,84 +313,47 @@ func (s *Server) GetRepositoryByName(ctx context.Context,
...
-		return nil, status.Errorf(codes.Internal, "cannot read repository: %v", err)
-	}
-
-	err = s.deleteRepositoryAndWebhook(ctx, repo, provider)
 	if err != nil {
 		return nil, err
 	}
 
-	// Telemetry logging
-	logger.BusinessRecord(ctx).Provider = repo.Provider
```
dmjb, line 346: Telemetry now lives in the service layer. This is because some of the attributes being tracked are only available inside the service (e.g. the ID of the repo when deleting by name). I considered returning the data needed for telemetry, but this made the abstraction very leaky.

internal/controlplane/handlers_repositories.go
```diff
@@ -491,28 +455,62 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+	ctx context.Context,
+	request HasProtoContext,
+	deletionMethod func(v1.GitHub, uuid.UUID) error,
+) error {
+	projectID, client, err := s.getProjectIDAndClient(ctx, request)
+	if err != nil {
 		return err
 	}
 
-	if err := s.store.Commit(tx); err != nil {
```
dmjb, line 514: The old code deleted the repo from the DB, then deleted the webhook. The new code deletes the webhook, and then deletes from the DB. Since the deletion of the webhook is idempotent, I decided not to use a transaction.

internal/repositories/github/service.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/stacklok/minder/internal/db"
+	"github.com/stacklok/minder/internal/events"
+	"github.com/stacklok/minder/internal/logger"
+	ghprovider "github.com/stacklok/minder/internal/providers/github"
+	ghclient "github.com/stacklok/minder/internal/repositories/github/clients"
+	"github.com/stacklok/minder/internal/repositories/github/webhooks"
+)
+
+// RepositoryService encapsulates logic related to registering and deleting repos
+type RepositoryService interface {
```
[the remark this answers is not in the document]
evankanderson: It's also interesting that we only have a `github` directory under `repositories/`.  I'm willing to use `Service` for now -- I suspect over time we're going to want to be able to move the different providers out of the Minder core to allow for third-party provider integrations, at which point these _will_ be some form of gRPC service.

internal/repositories/github/service.go
```diff
@@ -0,0 +1,136 @@
...
+	// DeleteRepositoryByName removes the webhook and deletes the repo from the
+	// database. The repo is identified by its name and project.
+	DeleteRepositoryByName(
+		ctx context.Context,
+		client ghclient.GitHubRepoClient,
+		projectID uuid.UUID,
+		repoOwner string,
+		repoName string,
```
evankanderson, line 42: I like how this interface is only in terms of other interfaces or dumb data objects.  Nice!

internal/repositories/github/service.go
evankanderson: I was going to suggest simply allowing people to use the struct initialization form, but I realized that we have a really bad track record at using the interfaces after that point, so let's hide the implementation for now.

internal/repositories/github/service.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// then remove the entry in the DB
+	if err = r.store.DeleteRepository(ctx, repo.ID); err != nil {
+		return fmt.Errorf("error deleting repository from DB: %w", err)
+	}
+
+	// Telemetry logging
+	logger.BusinessRecord(ctx).Provider = repo.Provider
+	logger.BusinessRecord(ctx).Project = repo.ProjectID
+	logger.BusinessRecord(ctx).Repository = repo.ID
```
evankanderson, line 133: Minor note -- it would be good to set these as early as possible, so that early exits still capture this telemetry.

I haven't made a big deal about it yet, but since you're cleaning things up, let's move this up to the point where we know `repo` (or even store the data that's in the proto, even if it's not normalized).

When we store this data only at the end of the operation, it means that we can't correlate e.g. 70 errors in the logs with the project that's having a bad day.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -296,42 +294,6 @@ func (s *Server) registerWebhookForRepository(
...
-	pbOpts := []providers.ProviderBuilderOption{
-		providers.WithProviderMetrics(s.provMt),
-		providers.WithRestClientCache(s.restClientCache),
-	}
-	providerBuilder, err := providers.GetProviderBuilder(ctx, provider, s.store, s.cryptoEngine, pbOpts...)
-	if err != nil {
-		return status.Errorf(codes.Internal, "cannot get provider builder: %v", err)
-	}
-
-	client, err := providerBuilder.GetGitHub()
```
evankanderson, line 313: What happened to this providers-finding stuff?  Did moving it behind `repositories/github` mean that we already know the provider?

internal/controlplane/handlers_repositories.go
evankanderson: We should be extracting the telemetry from the request earlier than we were anyway -- right now most of our errors don't include much telemetry about the resources involved.

internal/controlplane/handlers_repositories.go
evankanderson: +1 to deleting the webhook first.  Otherwise, we could delete our record that the webhook existed before we'd succeeded at deleting the webhook itself.

internal/controlplane/handlers_repositories.go
```diff
@@ -491,28 +455,54 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+	if err != nil {
+		return uuid.Nil, nil, providerError(err)
+	}
+
+	pbOpts := []providers.ProviderBuilderOption{
+		providers.WithProviderMetrics(s.provMt),
+		providers.WithRestClientCache(s.restClientCache),
+	}
+
+	p, err := providers.GetProviderBuilder(ctx, provider, s.store, s.cryptoEngine, pbOpts...)
```
evankanderson, line 477: I'd love to get this code abstracted into a function -- it seems like we have it in a lot of places.

But not in this PR!

It almost feels like `getProviderFromRequestOrDefault()` should return a `ProviderBuilder` (messy name for a provider with unknown aspects) rather than a `db.Provider`, which is leaking the database layer into other logic.

internal/controlplane/handlers_repositories.go
```diff
@@ -491,28 +455,54 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+	deletionMethod func(v1.GitHub, uuid.UUID) error,
+) error {
+	projectID, client, err := s.getProjectIDAndClient(ctx, request)
+	if err != nil {
 		return err
 	}
 
-	if err := s.store.Commit(tx); err != nil {
-		return status.Errorf(codes.Internal, "error deleting repository")
+	err = deletionMethod(client, projectID)
```
evankanderson, line 501: The indirection here seems less useful than the other parts of this PR, but I may be missing something.

It feels like the two methods should be:

```
DeleteRepoByName() {
  repo := getRepoByName()
  deleteRepo(repo.id)
}

DeleteRepoById() {
  deleteRepo(req.id)
}
```

evankanderson: While we certainly can pass one function to another, it feels like it might be simpler to just pass data around and have one "standard" form.

evankanderson: (if `deleteRepo` checks project ownership, the first form doesn't even need to check that the name lines up, though that may be the course of least resistance.)

internal/repositories/github/service.go
```diff
@@ -0,0 +1,136 @@
...
+		return err
+	}
+	return r.deleteRepository(ctx, client, &repo)
+}
+
+func (r *repositoryService) deleteRepository(ctx context.Context, client ghclient.GitHubRepoClient, repo *db.Repository) error {
+	var err error
+
+	// Cleanup any webhook we created for this project
+	// `webhooksManager.DeleteWebhook` is idempotent, so there are no issues
```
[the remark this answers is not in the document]
dmjb: Yes.

internal/controlplane/handlers_githubwebhooks.go
dmjb: The provider-getting logic lives in the unexported method I added into `handlers_repositories`

evankanderson (approved):

PR #2609, opened 2024-03-12
TEMPORARY: log where do we push and using what token

jhrozek:
This patch IS TO BE REVERTED SOON. It should only provide more debugging
into our staging environment while we figure out some very strange auth
issues.


rdimitrov (dismissed):

dmjb (approved):

dmjb (approved):

jhrozek:
right now the tests are working so I'm going to close this PR so that it doesn't get merged by accident.

Issue #2543, opened 2024-03-07
Track the projects which are subscribed to the healthcheck, and which version they are on


dmjb:
### Please describe the enhancement

We need some state to know which projects have the healthcheck, and which version of the healthcheck they are on.

### Solution Proposal

We need the following changes to the database schema:
1) A bundles table which tracks bundle name, namespace, version.
2) A subscriptions table which is a join table from project to bundle
3) A nullable foreign key reference from profiles and rule types to bundle.

PR #2670, opened 2024-03-15
Define DB schema for bundles and subscriptions

dmjb:
Relates to #2543

Define the subscriptions and bundles table. Add a foreign key reference to subscriptions in both the profiles and rule type tables.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/database/migrations/000029_bundle_subscription.up.sql
+++ b/database/migrations/000029_bundle_subscription.up.sql
@@ -1,0 +1,43 @@
+-- Copyright 2024 Stacklok, Inc
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


database/migrations/000029_bundle_subscription.up.sql
```diff
@@ -0,0 +1,43 @@
...
+    name        TEXT NOT NULL,
+    UNIQUE      (namespace, name)
+);
+
+CREATE TABLE subscriptions(
+    -- same comment as for PK of `bundles`
+    id                UUID NOT NULL DEFAULT gen_random_uuid() PRIMARY KEY,
+    project_id        UUID NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
+    -- this FK is not ON DELETE CASCADE
+    -- this prevents deletion of bundles which are still in use
```
JAORMX, line 29: uhm... wait, what would happen is that if a bundle gets deleted and a subscription is still around, the bundle will get deleted, but bundles that used to use that subscription will not be. I think we want the `ON DELETE CASCADE` in that case.

What we would want though is a trigger or something else to check that we don't delete a bundle if it's still in use. right?

dmjb: As per the chat on Slack - my plan is to have code which will clean up unused bundles when subscriptions are deleted or updated. I am open to considering some kind of DB driven cleanup, but I generally prefer to do it explicitly in code.

JAORMX (approved):

JAORMX:
This was indeed the behavior we wanted

PR #2675, opened 2024-03-15
Add alert info to RuleEvaluationStatuses

rdimitrov:
# Summary

The following PR populates the Alert field for each rule evaluation status result as it was missing so far. It also generates a valid URL for the potential security advisory that was opened.

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_profile.go
+++ b/internal/controlplane/handlers_profile.go
@@ -17,6 +17,7 @@ package controlplane
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"sort"
@@ -373,9 +374,11 @@ func (s *Server) getRuleEvaluationStatuses(
 		// make sure all fields are valid
 		if !dbRuleEvalStat.EvalStatus.Valid ||
 			!dbRuleEvalStat.EvalDetails.Valid ||
+			!dbRuleEvalStat.EvalLastUpdated.Valid ||
 			!dbRuleEvalStat.RemStatus.Valid ||
 			!dbRuleEvalStat.RemDetails.Valid ||
-			!dbRuleEvalStat.EvalLastUpdated.Valid {
+			!dbRuleEvalStat.AlertStatus.Valid ||
+			!dbRuleEvalStat.AlertDetails.Valid {
 			log.Print("error rule evaluation value not valid")
 			continue
 		}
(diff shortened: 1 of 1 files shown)


internal/controlplane/handlers_profile.go
```diff
@@ -367,52 +369,102 @@ func (s *Server) getRuleEvaluationStatuses(
...
+	if dbRuleEvalStat.AlertLastUpdated.Valid {
+		st.Alert.LastUpdated = timestamppb.New(dbRuleEvalStat.AlertLastUpdated.Time)
+	}
 
-		ruleEvaluationStatuses = append(ruleEvaluationStatuses, st)
+	// If the alert metadata is valid, parse it and set the URL
+	if dbRuleEvalStat.AlertMetadata.Valid {
+		// Define a struct to match the JSON structure
+		data := struct {
+			GhsaId string `json:"ghsa_id"`
```
JAORMX, line 451: can we move this to a function so we don't hardcode github-specific pieces into this logic?

internal/controlplane/handlers_profile.go
```diff
@@ -367,52 +369,102 @@ func (s *Server) getRuleEvaluationStatuses(
...
+			l.Err(err).Msg("error getting rule evaluation status")
 			continue
 		}
+		// Append the rule evaluation status to the list
+		ruleEvaluationStatuses = append(ruleEvaluationStatuses, st)
+	}
+	return ruleEvaluationStatuses
+}
 
-		if dbRuleEvalStat.EvalStatus.EvalStatusTypes == db.EvalStatusTypesFailure ||
```
rdimitrov, line 383: @JAORMX wdyt about this, should I revert it? I thought to remove it as it might be needed in the frontend but then again would it be needed in case the rule evaluation is successful?

internal/controlplane/handlers_profile.go
rdimitrov: good point 👍 move it out to a separate function.

JAORMX (approved):

JAORMX (changes requested):
wait, re-reading

dmjb (changes requested):
I appreciate that there is already a bunch of logic in this file for getting rule statuses, but I have two concerns:

1) There's no tests for the new code.
2) We are piling more logic into the profiles endpoints, which already have a huge amount of logic (this file is one of the least tested files in the codebase)

I wonder if we could have some kind of incremental step where the `getRuleEvaluationStatuses` function, and the new functions added in this PR can be moved into the `ProfileService` interface, or perhaps even into a new `RuleEvaluationService` interface.

This will require some extra work in this PR, but will at least continue down the road of decoupling business logic out of the controlplane, and making it easier to write tests for.

internal/controlplane/handlers_profile.go
rdimitrov: We decided to revert that check 👍

JAORMX:
So, to reflect what I said in slack. We should only be showing `guidance` if there's an failure. No use in telling people what they can do to fix an issue if the rule is succeeding.

dmjb (dismissed):
After chatting to @rdimitrov offline, I'm withdrawing my earlier feedback for now so that we can get the smoke tests working.

JAORMX (approved):

PR #2687, opened 2024-03-16
Implement EvalResultsService/ListEvaluationResults()

### LINKED ISSUE #2525: Implement the API for querying profile/rule status per given entity type and name
This actually implements the API created in https://github.com/stacklok/minder/issues/2524

- Returns a list of entities and their respected profile statuses, including ruletype details like evaluation status, remediation status and alert status.
- That is for a given entity_type and entity_id/name.
- For example repository and github.com/stacklok/minder or artifact and stacklok/minder-server
- If the entity_id/name is empty, we should return all entities, i.e all repositories.
- All of this is scoped to the project it executes on.
- Probably it should be part of the ProfileService proto service

puerco:
# Summary

This PR implements the handler for `minder.v1.EvalResultsService/ListEvaluationResults()` to query the evaluation statuses of profiles and entities. It adds a field to include the ruletype guidance to the sql query behind `ListRuleEvaluationsByProfileId`.

/cc @rdimitrov @kantord 

Fixes #2525

## Change Type

# Testing

# Review Checklist:


internal/controlplane/handlers_evalstatus.go
```diff
@@ -17,220 +17,331 @@ package controlplane
...
+		return nil, status.Errorf(codes.Internal, "Eerror getting user details: %v", err)
+	}
+
+	if _, err := uuid.Parse(in.GetProfile()); err != nil && in.GetProfile() != "" {
+		return nil, status.Error(codes.InvalidArgument, "Error invalid profile ID")
+	}
 
-	err := entityCtx.Validate(ctx, s.store)
+	// Read the list if projects the user has access to
+	projects, err := s.authzClient.ProjectsForUser(ctx, userInfo.IdentitySubject)
```
JAORMX, line 48: This part throws me off, normally, we want to encompass a certain project or project tree for a certain operation. This diverges away from that and does a request for all projects a user has access to. While this is a possible way to go... I personally think we should keep operations project-scoped, as that's what we've done for our authorization logic, and it is our tenancy model. Let's use the project from the entity context.

If we need an endpoint for all projects for a user, let that be a new API call with the user scope.

JAORMX (changes requested):

internal/controlplane/handlers_evalstatus.go
puerco: OK, I'm fine restricting it. Any objections to not listing all projects @rdimitrov ? We discussed returning everything, but I also think the function should behave as the others.

kantord:
Hi @puerco, I noticed that the remediation status is an arbitrary string in the JSON schema:

```json
        "remediationStatus": {
          "type": "string",
          "title": "remediation_status is the status of the remediation"
        },
```

this makes it difficult to correctly display this value in the FE, because this value is not actually displayed as a free-form string. For example each value is displayed with a different icon and color.

It would be really useful to have the list of all possible values restricted in the JSON schema using `enum` or sth similar.

If that is not possible then at the very least I will need the list of possible values in order to make sure that all values are displayed properly :pray:

internal/controlplane/handlers_evalstatus.go
```diff
@@ -17,220 +17,324 @@ package controlplane
...
-	entityCtx := engine.EntityFromContext(ctx)
+	if _, err := uuid.Parse(in.GetProfile()); err != nil && in.GetProfile() != "" {
+		return nil, status.Error(codes.InvalidArgument, "Error invalid profile ID")
+	}
 
-	err := entityCtx.Validate(ctx, s.store)
+	projectID, err := uuid.Parse(*in.GetContext().Project)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "error in entity context: %v", err)
+		return nil, fmt.Errorf("error parsing project identifier")
```
JAORMX, line 43: why not just use:

```
```

??

internal/controlplane/handlers_evalstatus.go
kantord: at the moment the only FE consuming this is project scoped. this endpoint is already returning a lot of data anyways, so probably the future improvements lie in somehow restricting the amount of data it returns rather than expanding it even further

internal/controlplane/handlers_evalstatus.go
puerco: I was trying to validate the input but it's already checked elsewhere I can drop this check :)

puerco:
@kantord here are the eval statuses currently supported:

https://github.com/stacklok/minder/blob/main/internal%2Fengine%2Ferrors%2Ferrors.go#L111

Since this is a new API call I think we can safely merge it soon if the approach is right, and then follow up with any changes, I can check how to have an enum.if the strings.

kantord:
> @kantord here are the eval statuses currently supported:
> 
> https://github.com/stacklok/minder/blob/main/internal%2Fengine%2Ferrors%2Ferrors.go#L111
> 
> Since this is a new API call I think we can safely merge it soon if the approach is right, and then follow up with any changes, I can check how to have an enum.if the strings.

sure, I agree, it's not a blocker for testing so let's merge

JAORMX (approved):
fair enough, let's update context handling in a subsequent PR.

PR #2688, opened 2024-03-17
Default explicitly to remediate:off and alert:on

### LINKED ISSUE #2674: Get rid of the "unset" default from the
### Describe the issue

For some reason we defaulted to having "no value" as the default for both alert and remediate and we handle the actual default in minder code. (remediate defaults to off, alert to on). Then we also store and let the user retrieve the profile with these unset remediations, which is breaking the UI. On the CLI side we actually also print the hardcoded defaults if nothing is set.

I don't recall the reason for the unset default, I /think/ we were trying to let us change the default without a migration. But we've settled on remediate:off and alert:on as reasonable defaults in the meantime.

Because the unset string as a default doesn't actually make a whole lot of sense, let's just migrate our db to having the proper settings and move the defaults to insert and upsert instead.

### To Reproduce

Create a profile with the remediate and alert unset
Retrieve the profile

The values are explicitly unset.

_No response_

jhrozek:
# Summary

For some reason we defaulted to having "no value" as the default for both
alert and remediate and we handle the actual default in minder code.
(remediate defaults to off, alert to on). Then we also store and let the user
retrieve the profile with these unset remediations, which is breaking the UI.
On the CLI side we actually also print the hardcoded defaults if nothing is
set.

Instead, let's assign the defaults as values into minder profiles.

The full fix will be a two step process. The first step is this PR where we
make sure that minder doesn't store any new NULL values into those columns and
handles retrieving NULL values on the server side.

The next step, once this fix is promoted to production, will be to actually
migrate the columns to the `ActionType` type instead of the current `NullActionType`.
But we can't do that at the same time, because otherwise we end up in a situation
where the database only access non-Null types but the server can accept and try
to store Null types.

Fixes #2674

## Change Type

# Testing

# Review Checklist:


jhrozek:
cc @davolokh this should fix the UI empty output for profiles

JAORMX (approved):

Issue #2520, opened 2024-03-06
Create a GitHub App provider


eleftherias:
Create a GitHub App provider that implements the same traits as the current GitHub OAuth provider.

Below are some key implementation differences

### Acting user attributes
In the OAuth provider, the user details to include when creating a commit or PR come from the PAT.
In the GitHub App provider, the user details should correspond to the apps details, as described [in this discussion](https://github.com/orgs/community/discussions/24664) and [this section of the design doc](https://docs.google.com/document/d/1SvXgkaB8ffgD2VH9pQq8rabgV-8ow3T9oBRtgZFPGVY/edit#heading=h.40mrvec3qfxm).

### Listing repositories
To get a list of all repositories, the GitHub app will [fetch the list of repos](https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-repositories-accessible-to-the-app-installation) available to the installations associated with a given project.

### Credentials
A GitHub app will need to generate tokens based on the following:
- The installation ID of the GitHub organization we are querying.This will be stored in the Minder DB.
- The GitHub app ID for the Minder app. We get this when we create the GitHub app.
- The client secret (PEM file) which we will also generate when we create the GitHub app.
The instructions for generating an installation access token are available in the [GitHub documentation](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app).

There may be other APIs that are different between the OAuth app and the GitHub app. 

note: This issue does not include supporting the "Marketplace flow". For the scope of this issue, we should only add support for the UI / CLI initiated flow, where we know how to associate an installation ID to a project.

[Design document](https://docs.google.com/document/d/1SvXgkaB8ffgD2VH9pQq8rabgV-8ow3T9oBRtgZFPGVY/edit)

evankanderson:
The provider will need a new table to store the installation IDs, and we'll need to be able to indicate the different providers.

If we associate each provider with a specific GitHub installation, we may need multiple providers in the same project, and a way to select which provider to use when registering a repo.

PR #2713, opened 2024-03-19
Fill in GitHub App provider

eleftherias:
# Summary

This PR fills in the functions for the GitHub App provider delegate. This is the functionality that the provider will use once enrolled.

This PR does not add the functionality to enroll the GitHub App provider, or the functionality to generate an installation token credential.

Ref #2520

## Change Type

# Testing

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -1232,6 +1232,18 @@ message GitHubProviderConfig {
...
+  string endpoint = 1;
+  // AppName is the GitHub App Name
+  string app_name = 2;
+  // AppID is the GitHub App ID
+  string app_id = 3;
+  // UserID is the GitHub App User ID
+  int64 user_id = 4;
+}
+
 // Context defines the context in which a rule is evaluated.
```
eleftherias, line 1247: I'm unsure if all these fields belong in the provider definition. In the GitHub OAuth definition we only include the `endpoint`, but my assumption is if we're interacting with multiple GitHub endpoints (e.g public vs private), then each one will have it's own GitHub App.

JAORMX: So, for the OAuth2 app, we're able to point minder to an enterprise github endpoint. I'm not sure we can do the same with a github app flow? Is that a GitHub cloud-only kind of thing or is this something that may also work in on-prem installations? Would be good to figure out. Else, we can remove the `endpoint` for now and add it when we see the use-case.

JAORMX (commented):
mostly lgtm, just got one comment.

proto/minder/v1/minder.proto
eleftherias: GitHub Apps are available on GitHub Enterprise Server as well https://docs.github.com/en/apps/sharing-github-apps/making-your-github-app-available-for-github-enterprise-server#each-github-enterprise-server-instance-must-register-their-own-github-app

JAORMX: ack, let's include the endpoint then.

JAORMX (approved):

jhrozek (approved):

Issue #2541, opened 2024-03-07
Build logic to add healthcheck profiles/rules to a project


dmjb:
### Please describe the enhancement

Create methods/functions for adding/updating the rules and profiles from the healthcheck bundle to a project. These need to:
1) Create/Apply the rules/profiles from the bundle files on disk instead of via an API request.
2) Ensure that the namespace and versioning information is correctly applied to the profiles and rules.

### Solution Proposal

As part of this work, we should refactor the existing profile and rule creation code so that it is not coupled to the controlplane

PR #2714, opened 2024-03-19
Create interface for reading from mindpak bundles

dmjb:
Relates to #2541

Create an interface for reading the contents of the bundle, and provide an implementation and tests.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


pkg/mindpak/testdata/t2/manifest.json
```diff
@@ -1,16 +1,14 @@
 {
```
dmjb, line 1: I repurpose this test data for my own tests. The changes did not affect any existing tests.

pkg/mindpak/testdata/t2/manifest.json
```diff
@@ -1,16 +1,14 @@
...
-      "seconds": 1709881152,
-      "nanos": 854906417
-    }
+    "date": null
   },
   "files": {
     "profiles": [
       {
-        "name": "branch-protection.yaml",
+        "name": "branch-protection-github-profile.yaml",
```
dmjb, line 11: Changed the filename to match the profile name. I think this might be a useful constraint to enforce on bundles since it simplifes the process of mapping a profile to its path inside the bundle

internal/marketplace/bundles/bundle.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (b *bundle) GetMetadata() *mindpak.Metadata {
+	return b.original.Manifest.Metadata
+}
+
+func (b *bundle) GetProfile(name string) (*v1.Profile, error) {
+	// ensure name has file extension
+	name = ensureYamlSuffix(name)
+	// validate that profile exists
+	_, ok := b.profiles[name]
```
puerco, line 73: One thing that got me thinking after writing the library was that `name` may be overloaded in the bundle, leading to ambiguity. Here (as in `pkg/mindpak`) we mean the file name, while the profiles and  rules types [have their own name string](https://github.com/stacklok/minder-rules-and-profiles/blob/6f1bb40c4394d30dd2d837da99cfd52a39d20842/profiles/github/dependabot_go.yaml#L5). Should we rename the name in the bundle to filename? Index by its true name?

puerco (commented):
Aweome @dmjb , this is exactly how I was imagining the mindpak API. I think these should live in the bundle directly,

pkg/mindpak/testdata/t2/manifest.json
puerco: Hehe this is dancing around the same issue I noted below. Sorry, I missed it before sending my comment :)

internal/marketplace/bundles/bundle.go
JAORMX: It may indeed be overloaded by the bundle, we may have a single OCI image reference instead of a file. Ideally, the method of reading rule types and profiles should be pluggable.

dmjb: From an API viewpoint, I think it makes the most sense to refer to a rule type or profile by its actual name (i.e. the name defined inside the profile or rule type). The filename is just an implementation detail.

The way I see it, there are two possible options here:
1. We ensure when building the bundle that the file name matches the profile/rule type which is stored in the file.
2. We add a separate file name or reference field.

Option 2) seems like the most flexible, but I'm not sure if we need that flexibility at the moment.

JAORMX: Let's go for simplicity now. You're right that it's an implementation detail.

JAORMX (approved):

internal/marketplace/bundles/bundle.go
puerco: The only tricky part here is that if we unify name + fielame, then how do we specify namespace rule types and profiles?

PR #2725, opened 2024-03-20
Introduce GitHub App installation access token

eleftherias:
# Summary
This introduces the `GitHubInstallationTokenCredential`, which represents a new way of authenticating with GitHub using an installation access token.
The token is generated when a new `ProviderBuilder` is created and then used for all interactions in that `ProviderBuilder`.

### Token Generation
The process for generating the installation access token is described by GitHub, [here](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app). Essentially, Minder first generates a signed JWT using the private key that is specific to our GitHub App. Minder then calls GitHub to exchange the JWT for a installation access token that has permissions to a specific installation (i.e GitHub org).

### Token lifetime
The installation access token is valid for 1 hour. That gives us plenty of time to respond to an event. A new token is generated for every new `ProviderBuilder` creation, to avoid 1) storing tokens and 2) dealing with token expiry.

### Rate limiting
In contrast to the oauth tokens, it's not enough to track the rate limit on the token, we instead need to track the rate limit on the installation ID. That's why the key in the rate limiting queue is the installation ID, rather than the token.

### Differentiating between token types
In the `ProviderBuilder` the credential type is determined by the data in the DB. If the provider has an access token, then we will use the access token. If there is no access token found for the provider, we will look for an installation ID.
This makes the credential more generic and hopefully more reusable as we add more providers.

note: In practice this credential is not used yet, because we don't have a mechanism for creating a new GitHub App installation.

Ref #2520

## Change Type

# Testing

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -1244,6 +1244,8 @@ message GitHubAppProviderConfig {
   string app_id = 3;
   // UserID is the GitHub App User ID
   int64 user_id = 4;
+  // PrivateKeyPath is the path to the GitHub App Private Key
+  string private_key_path = 5;
```
JAORMX, line 1248: Do we want to make this configurable? Should this be an implementation detail?

internal/providers/providers.go
```diff
@@ -287,11 +290,44 @@ func getCredentialForProvider(
...
 	return nil, fmt.Errorf("error getting credential: %w", err)
 }
 
+// getInstallationTokenCredential returns a GitHub installation token credential if the provider has an installation ID
+func getInstallationTokenCredential(ctx context.Context, prov db.Provider, store db.Store) (provinfv1.Credential, error) {
+	installation, err := store.GetInstallationIDByProviderID(ctx, uuid.NullUUID{
+		UUID:  prov.ID,
+		Valid: true,
+	})
+	if err == nil {
```
JAORMX, line 306: nit: Let's first check no rows, then throw the error and do the logic in the bottom of the function. That way we reduce indentation and cognitive load.

internal/providers/providers.go
```diff
@@ -287,11 +290,44 @@ func getCredentialForProvider(
...
+		if err != nil {
+			return nil, fmt.Errorf("error parsing github app config: %w", err)
+		}
+
+		privateKeyBytes, err := os.ReadFile(cfg.PrivateKeyPath)
+		if err != nil {
+			return nil, fmt.Errorf("error reading private key: %w", err)
+		}
+
+		privateKey, err := jwt.ParseRSAPrivateKeyFromPEM(privateKeyBytes)
```
JAORMX, line 317: Are github credentials always RSA? is this something we can configure?

proto/minder/v1/minder.proto
eleftherias: Similar to the app ID and endpoint, if we have we're talking to multiple GitHubs, private and public, then we'll need multiple GitHub Apps and each app will have it's own private key. The private key is generated from GitHub, we don't choose our own, so we couldn't reuse the same key for multiple apps.

internal/providers/providers.go
eleftherias: They'll always be RSA. From the docs here: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/managing-private-keys-for-github-apps#generating-private-keys
> Note: If you're using a library that requires a specific file format, the PEM file you download will be in PKCS#1 RSAPrivateKey format.

internal/providers/providers.go
eleftherias: Done!

JAORMX:
Per the slack conversation, we've agreed to rework the private details (like key path) off this configuration).

JAORMX (approved):

PR #2729, opened 2024-03-20
 Create Subscription service interface

### LINKED ISSUE #2543: Track the projects which are subscribed to the healthcheck, and which version they are on

dmjb:
Fixes #2543

Provide an interface for managing subscriptions from projects to bundles. This includes methods for adding bundle profiles and rule types to projects. Methods to update and delete subscriptions will be added in future PRs.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


pkg/mindpak/reader/reader.go
```diff
@@ -67,6 +67,11 @@ func (b *bundleReader) GetMetadata() *mindpak.Metadata {
 }
 
 func (b *bundleReader) GetProfile(name string) (*v1.Profile, error) {
+	// if called with a namespace prefix, remove it
```
JAORMX, line 70: All of the PR looks good but this throws me off a little. Why strip the namespace when returning this? Wouldn't we want to show this anyway? We also now have a `display name` for profiles, maybe we want to strip the namespace and set it there.

dmjb: This variable is used to find the profile's file inside the bundle, and does not remove anything from the profile's YAML. You'll see that the next step is to add `.yaml` to the end of the name.

The set of assumptions which led me down this road:

1) The rule type and profile names (i.e. the `name` field inside the YAML, or in the database) include the namespace.
1) From an interface/API point of view - when fetching a profile from the bundle, the name with namespace makes the most sense for anyone using that interface to get a specific profile.
3) The filenames of the YAML files holding those profiles/rule types do not include the namespace. It would be awkward to do this because we currently use `/` as a namespace separator, and that would imply that we need to create a directory inside the bundle for that namespace.

Solutions:
1) Do not put the namespace in the filename inside the bundle, strip it out.
2) Use a different namespace separator like `:` which would allow us to avoid this problem.

I picked 1), but if we want to consider using a different namespace separator, now is the time to decide that.

JAORMX: ah! I see what you're saying. I think your initial decision for option 1 was correct. LGTM. Thanks for the explanation!

... coffee also helped

JAORMX (approved):
shipit!

PR #2730, opened 2024-03-20
Add GitHub App data to configuration

eleftherias:
The introduces a `provider` section in the configuration that allows configuring provider settings.

It also allows for optional fields in the config, by adding pointers to the allowed types.


internal/providers/providers.go
```diff
@@ -88,14 +84,10 @@ func WithRestClientCache(cache ratecache.RestClientCache) ProviderBuilderOption
...
-func NewProviderBuilder(
-	p *db.Provider,
-	ownerFilter sql.NullString,
-	credential provinfv1.Credential,
-	opts ...ProviderBuilderOption,
-) *ProviderBuilder {
+func NewProviderBuilder(p *db.Provider, ownerFilter sql.NullString, credential provinfv1.Credential, cfg *serverconfig.Config, opts ...ProviderBuilderOption) *ProviderBuilder {
 	pb := &ProviderBuilder{
 		p:           p,
+		cfg:         cfg,
```
eleftherias, line 90: We'd have to pass the config everywhere when we create a new `ProviderBuilder`.

internal/providers/providers.go
```diff
@@ -210,7 +202,7 @@ func (pb *ProviderBuilder) GetGitHub() (provinfv1.GitHub, error) {
 		return nil, fmt.Errorf("error parsing github app config: %w", err)
 	}
 
-	cli, err := githubapp.NewGitHubAppProvider(cfg, pb.metrics, pb.restClientCache, gitHubCredential)
+	cli, err := githubapp.NewGitHubAppProvider(cfg, pb.cfg.GitHubApp, pb.metrics, pb.restClientCache, gitHubCredential)
```
eleftherias, line 205: It seems weird to pass both the config from the definition and from the `config.yaml`. If we go down this route, should we also move the endpoint to the `config.yaml` and remove it from the provider definition?

internal/providers/github/app/app.go
```diff
@@ -84,20 +85,26 @@ func NewGitHubAppProvider(
...
 		if err != nil {
 			return nil, err
 		}
 		ghClient.BaseURL = parsedURL
 	}
 
+	appName := appConfig.AppName
+	userId := appConfig.UserID
+
+	// Alternatively user viper to get the values directly
```
eleftherias, line 99: We also have the option of getting the values directly from viper but that makes testing harder.

eleftherias (commented):
@JAORMX I added some comments about what this approach would look like.

internal/providers/providers.go
JAORMX: Yeah, I don't think there's away around that.... unless we end up creating a `ProviderBuilderFactory` that has a set of persistent values like metrics and configuration. (screams in Java)

internal/providers/providers.go
JAORMX: I'm not sure about the endpoint, could we have a minder instance that has tenants that use a private installation and other that point to the public github endpoint?

eleftherias: I think that's a valid scenario. I also think we don't need to support it at this moment.

internal/config/server/config.go
```diff
@@ -43,6 +43,7 @@ type Config struct {
 	WebhookConfig WebhookConfig         `mapstructure:"webhook-config"`
 	Events        EventConfig           `mapstructure:"events"`
 	Authz         AuthzConfig           `mapstructure:"authz"`
+	GitHubApp     GitHubAppConfig       `mapstructure:"github-app"`
```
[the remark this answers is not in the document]
eleftherias: I like it

JAORMX (approved):

JAORMX (approved):

Issue #2722, opened 2024-03-20
Rotate webhook secrets


jhrozek:
### Please describe the enhancement

There's currently no way to rotate minder webhook secrets in case the secret is compromised or needs to be changed for other reason.

### Solution Proposal

We should:
 - add a CLI akin to the migrate CLI that lets a user update webhook secrets
 - provide a way for minder to have a list of fallback secrets in case the one in the config is set to the rotated secret, but not all webhooks could be rotated for one reason or another

PR #2734, opened 2024-03-20
Provide an array of fallback secrets to allow secret migration

jhrozek:
# Summary

In order to provide for a clean webhook secret migration, we need to
provide a window of using the old secret. For this, we add an array of
PreivousWebhookSecrets to the WebhookConfig structure. If validating the
payload with the WebhookConfig.WebhookSecret fails, we fall back to
validating using the previous secrets.

Because validating the request reads the request body, we need to read
the body first and save in a byte slice in case we do have the fallbacks
configured.

Related: #2722

## Change Type

# Testing

Tested by setting the correct shared secrets in the previousWebhookSecrets
array and junk in WebhookConfig.WebhookSecret as well as the usual case
of WebhookConfig.WebhookSecret containing the correct secret.

# Review Checklist:

### DIFF
--- a/config/server-config.yaml.example
+++ b/config/server-config.yaml.example
@@ -65,6 +65,7 @@ webhook-config:
   external_webhook_url: "https://example.com/api/v1/webhook/github"
   external_ping_url: "https://example.com/api/v1/health"
   webhook_secret: "your-password"
+# previous_webhook_secret_file: ./previous_secrets
 
 # OAuth2 Configuration (used during enrollment)
 # These values are to be set within the GitHub OAuth2 App page
--- a/internal/config/server/webhook.go
+++ b/internal/config/server/webhook.go
@@ -15,6 +15,12 @@
 
 package server
 
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
 // WebhookConfig is the configuration for our webhook capabilities
 type WebhookConfig struct {
 	// ExternalWebhookURL is the URL that we will send our webhook to
@@ -22,6 +28,22 @@ type WebhookConfig struct {
 	// ExternalPingURL is the URL that we will send our ping to
 	ExternalPingURL string `mapstructure:"external_ping_url"`
 	// WebhookSecret is the secret that we will use to sign our webhook
-	// TODO: Check if this is actually used and needed
 	WebhookSecret string `mapstructure:"webhook_secret"`
+	// PreviousWebhookSecretFile is a reference to a file that contains previous webhook secrets. This is used
+	// in case we are rotating secrets and the external service is still using the old secret. These will not
+	// be used when creating new webhooks.
+	PreviousWebhookSecretFile string `mapstructure:"previous_webhook_secret_file"`
+}
+
+// GetPreviousWebhookSecrets retrieves the previous webhook secrets from a file specified in the WebhookConfig.
+// It reads the contents of the file, splits the data by whitespace, and returns it as a slice of strings.
+func (wc *WebhookConfig) GetPreviousWebhookSecrets() ([]string, error) {
+	data, err := os.ReadFile(wc.PreviousWebhookSecretFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read previous webhook secrets from file: %w", err)
+	}
+
+	// Split the data by whitespace and return it as a slice of strings
+	secrets := strings.Fields(string(data))
+	return secrets, nil
 }
--- a/internal/controlplane/handlers_githubwebhooks.go
+++ b/internal/controlplane/handlers_githubwebhooks.go
@@ -17,11 +17,14 @@
 package controlplane
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"mime"
 	"net/http"
 	"sort"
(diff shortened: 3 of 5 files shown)


eleftherias (approved):

dmjb (approved):

JAORMX (changes requested):
Uhm... how will this array be configured in the helm chart? I think want to reference file paths and not secrets from an array since it won't be straight forward to inject those from k8s

config/server-config.yaml.example
```diff
@@ -65,6 +65,9 @@ webhook-config:
   external_webhook_url: "https://example.com/api/v1/webhook/github"
   external_ping_url: "https://example.com/api/v1/health"
   webhook_secret: "your-password"
+#  previous_webhook_secrets:
+#   - "under-migration"
+#   - "should-be-rotated"
```
evankanderson, line 70: Should these be strings in configuration, or references to either files on disk or environment variables?

What do you think of having a single file with a list of accepted webhook secrets?

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -186,6 +188,49 @@ func (s *Server) HandleGitHubWebHook() http.HandlerFunc {
...
+	// if there are any previous secrets we need to save the body
+	// because every call to ValidatePayload will consume the body
+	if len(wc.PreviousWebhookSecrets) > 0 {
+		bodyBytes, err = getAndDuplicateBody(r)
+		if err != nil {
+			return
+		}
+	}
+
+	payload, err = github.ValidatePayload(r, []byte(wc.WebhookSecret))
```
evankanderson, line 203: There's a [`ValidatePayloadFromBody`](https://pkg.go.dev/github.com/google/go-github/v43/github#ValidatePayloadFromBody) which takes an `io.Reader` and the two header fields needed for the validation which might be simpler to use:

```suggestion
	bodyBytes, err := io.ReadAll(r.Body)
	if err != nil {
		return nil, err
	}
	body := io.NopCloser(bytes.NewReader(bodyBytes))
	signature := r.Header.Get(github.SHA256SignatureHeader)
	if signature == "" {
		signature = r.Header.Get(github.SHA1SignatureHeader)
	}

	contentType, _, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
	if err != nil {
		return nil, err
	}

	payload, err = github.ValidatePayload(contentType, body, signature, []byte(wc.WebhookSecret))
	
	body.Seek(0, io.SeekStart)
```

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -186,6 +188,49 @@ func (s *Server) HandleGitHubWebHook() http.HandlerFunc {
...
+	if err == nil {
+		return
+	}
+
+	for _, prevSecret := range wc.PreviousWebhookSecrets {
+		r.Body = io.NopCloser(bytes.NewBuffer(bodyBytes))
```
evankanderson, line 209: bytes.NewBuffer() is supposed to take ownership of its argument; you probably want `bytes.NewReader()` instead.

config/server-config.yaml.example
jhrozek: Good point (both you and @JAORMX). I'll change the option to be akin `auth.token_key` to reference a file

internal/controlplane/handlers_githubwebhooks.go
jhrozek: good catch, thanks.

internal/controlplane/handlers_githubwebhooks.go
jhrozek: yes, the reason I didn't use it was that I didn't want to copy reading the signature and the content type headers. I'll experiment with how the code feels before the next push along with the other changes.

config/server-config.yaml.example
jhrozek: done!

internal/controlplane/handlers_githubwebhooks.go
jhrozek: taken into account, it's not like the headers and the mime type are going to change, WDYT?

internal/controlplane/handlers_githubwebhooks.go
jhrozek: fixed!

internal/config/server/webhook.go
```diff
@@ -15,13 +15,35 @@
...
 // WebhookConfig is the configuration for our webhook capabilities
 type WebhookConfig struct {
 	// ExternalWebhookURL is the URL that we will send our webhook to
 	ExternalWebhookURL string `mapstructure:"external_webhook_url"`
 	// ExternalPingURL is the URL that we will send our ping to
 	ExternalPingURL string `mapstructure:"external_ping_url"`
 	// WebhookSecret is the secret that we will use to sign our webhook
-	// TODO: Check if this is actually used and needed
 	WebhookSecret string `mapstructure:"webhook_secret"`
+	// PreviousSecrets is an array of the previous secrets that we will use to verify signatures on webhook messages
```
JAORMX, line 32: This docs are no longer applicable

JAORMX (changes requested):

JAORMX (approved):

PR #2735, opened 2024-03-20
Add a CLI for rotating webhook secrets

jhrozek:
# Summary

Adds a minder-server subcommand that allows you to change minder webhook secrets
in github as a way of rotating the secret

Fixes: #2722

## Change Type

# Testing

```
minder-server webhook update -p github
```

# Review Checklist:

### DIFF
--- a/cmd/server/app/webhook_update.go
+++ b/cmd/server/app/webhook_update.go
@@ -1,0 +1,202 @@
+//
+// Copyright 2023 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package app
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"net/url"
+	"os"
+	"strings"
+
+	"github.com/rs/zerolog"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+
+	"github.com/stacklok/minder/internal/config"
+	serverconfig "github.com/stacklok/minder/internal/config/server"
+	"github.com/stacklok/minder/internal/crypto"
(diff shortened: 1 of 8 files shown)


JAORMX:
Will we need another k8s job to run this?

cmd/server/app/webhook_update.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	defer func(dbConn *sql.DB) {
+		err := dbConn.Close()
+		if err != nil {
+			zerolog.Ctx(ctx).Error().Err(err).Msg("error closing database connection")
+		}
+	}(dbConn)
+
+	store := db.NewStore(dbConn)
+	allProviders, err := store.GlobalListProvidersByName(ctx, providerName)
```
JAORMX, line 87: Should we instead iterate all providers, check that they implement github, and then attempt the migration? We're moving away from predetermined names and more into traits and classes

JAORMX (changes requested):

docs/docs/run_minder_server/run_the_server.md
```diff
@@ -249,6 +249,16 @@ webhook-config:
 
 After these steps, your Minder server should be ready to receive webhook events from GitHub, and add webhooks to repositories.
 
+In order to rotate webhook secrets, you can use the `minder-server` CLI tool to update the webhook secret.
+
+```bash
+minder-server webhook update -p github
+```
+
+Note that the command simply replaces the webhook secret in the configuration
```
eleftherias, line 258: Isn't it the opposite? It seems like this command updates the secret in GitHub, but you will need to restart your Minder server to pick up the secret in the configuration.

eleftherias (commented):
One comment on the documentation, but otherwise I tested it and it works great.

cmd/server/app/webhook_update.go
jhrozek: We can do that, sure.

docs/docs/run_minder_server/run_the_server.md
jhrozek: gah, of course when I wrote configuration I meant github..

jhrozek:
> Will we need another k8s job to run this?

yes, that's how I was thinking the rotation would work:
 - generate a new secret
 - set the new secret as the current one and the older one as a fallback
 - run this command as a job
 - monitor staging logs for the log message that says that a fallback secret was used to verify a webhook signature
 - if we see none, remove the fallback secret finally

cmd/server/app/webhook_update.go
jhrozek: done!

JAORMX (approved):

PR #2750, opened 2024-03-21
Remove built-in github provider

eleftherias:
# Summary

This removes the built-in GitHub provider that gets created when a user registers.

Instead, a provider is created after the enrolment flow is complete and we have the access token.
If a provider with the same name already exists, the behaviour doesn't change, it just updates the credential for that provider.

## Change Type

# Testing

Tested Scenarios:
- Built-in project exists from previous version and enrolment happens with new version
- Provider already enrolled with credentials in previous version, then enroll with new version updates the credentials
- Enroll with / without owner filter
- Account is created on new version, credentials update on every enrolment
- Tested with old CLI version 0.0.36

# Review Checklist:

### DIFF
--- a/cmd/cli/app/provider/provider_enroll.go
+++ b/cmd/cli/app/provider/provider_enroll.go
@@ -53,7 +53,6 @@ actions such as adding repositories.`,
 // EnrollProviderCommand is the command for enrolling a provider
 func EnrollProviderCommand(ctx context.Context, cmd *cobra.Command, conn *grpc.ClientConn) error {
 	client := minderv1.NewOAuthServiceClient(conn)
-	provcli := minderv1.NewProvidersServiceClient(conn)
 
 	provider := viper.GetString("provider")
 	project := viper.GetString("project")
@@ -83,25 +82,8 @@ func EnrollProviderCommand(ctx context.Context, cmd *cobra.Command, conn *grpc.C
 		}
 	}
 
-	prov, err := provcli.GetProvider(ctx, &minderv1.GetProviderRequest{
-		Context: &minderv1.Context{Provider: &provider, Project: &project},
-		Name:    provider,
-	})
-	if err != nil {
-		return cli.MessageAndError("Error getting provider", err)
-	}
-
 	if token != "" {
-		if !prov.Provider.SupportsAuthFlow(minderv1.AuthorizationFlow_AUTHORIZATION_FLOW_USER_INPUT) {
-			return fmt.Errorf("provider %s does not support token enrollment", provider)
-		}
-
 		return enrollUsingToken(ctx, cmd, client, provider, project, token, owner)
-	}
-
-	if !prov.Provider.SupportsAuthFlow(
-		minderv1.AuthorizationFlow_AUTHORIZATION_FLOW_OAUTH2_AUTHORIZATION_CODE_FLOW) {
-		return fmt.Errorf("provider %s does not support OAuth2 enrollment", provider)
 	}
 
 	// This will have a different timeout
--- a/internal/controlplane/handlers_oauth.go
+++ b/internal/controlplane/handlers_oauth.go
@@ -24,11 +24,8 @@ import (
 	"net/http"
 	"net/url"
 	"slices"
-	"strconv"
 
-	"github.com/google/uuid"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
-	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
@@ -42,10 +39,12 @@ import (
 	"github.com/stacklok/minder/internal/engine"
 	"github.com/stacklok/minder/internal/logger"
 	"github.com/stacklok/minder/internal/providers"
-	"github.com/stacklok/minder/internal/providers/credentials"
+	"github.com/stacklok/minder/internal/providers/github/oauth"
 	"github.com/stacklok/minder/internal/util"
 	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 )
+
+const defaultProvider = oauth.Github
 
 // GetAuthorizationURL returns the URL to redirect the user to for authorization
 // and the state to be used for the callback. It accepts a provider string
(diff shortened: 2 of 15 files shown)


cmd/cli/app/provider/provider_enroll.go
```diff
@@ -83,24 +84,22 @@ func EnrollProviderCommand(ctx context.Context, cmd *cobra.Command, conn *grpc.C
...
+	prov, err := provcli.GetProviderDefinition(ctx, &minderv1.GetProviderDefinitionRequest{
+		Class: provider,
 	})
 	if err != nil {
 		return cli.MessageAndError("Error getting provider", err)
 	}
 
 	if token != "" {
-		if !prov.Provider.SupportsAuthFlow(minderv1.AuthorizationFlow_AUTHORIZATION_FLOW_USER_INPUT) {
+		if !slices.Contains(prov.AuthFlows, minderv1.AuthorizationFlow_AUTHORIZATION_FLOW_USER_INPUT) {
```
JAORMX, line 95: why not implement a `SupportsAuthFlow` in the provider definition protobuf?

internal/controlplane/handlers_oauth.go
```diff
@@ -55,13 +58,20 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
 	entityCtx := engine.EntityFromContext(ctx)
 	projectID := entityCtx.Project.ID
 
+	var provider string
+	if req.GetContext().GetProvider() == "" {
+		provider = defaultProvider
+	} else {
+		provider = req.GetContext().GetProvider()
+	}
```
JAORMX, line 66: If we're removing the default provider, wouldn't it make sense that we first do a `CreateProvider` call and THEN do the `GetAuthorizationURL`? The logic then wouldn't change here.

internal/controlplane/handlers_oauth.go
```diff
@@ -231,10 +241,23 @@ func (s *Server) generateOAuthToken(ctx context.Context, provider string, code s
...
+		return status.Errorf(codes.Internal, "error starting transaction: %v", err)
+	}
+	defer s.store.Rollback(tx)
+
+	qtx := s.store.GetQuerierWithTransaction(tx)
+
+	provider, err := getOrCreateProvider(ctx, qtx, providerName, stateData)
+	if err != nil {
+		return fmt.Errorf("error getting provider: %w", err)
+	}
```
JAORMX, line 255: not too keen on this, I rather have an explicit `CreateProvider` API call rather than embedding more complex logic in here.

internal/controlplane/handlers_oauth.go
eleftherias: My thinking is that if a provider requires credentials, then we shouldn't save the provider in the DB until we have credential for it. Otherwise we'll end up with providers that aren't usable.

internal/controlplane/handlers_providers.go
```diff
@@ -47,7 +46,17 @@ func (s *Server) GetProvider(ctx context.Context, req *minderv1.GetProviderReque
 	})
 	if err != nil {
 		if errors.Is(err, sql.ErrNoRows) {
-			return nil, util.UserVisibleError(codes.NotFound, "provider not found")
+			// FIXME: This is temporary for backwards compatibility with old CLI versions
```
eleftherias, line 49: The CLI previously queried the DB to get the authorization flows for the provider. This is to keep the server backwards compatible with older CLI versions.

internal/controlplane/handlers_oauth.go
JAORMX: I like this idea. Let's go for that. Thanks!

proto/minder/v1/minder.proto
```diff
@@ -1990,6 +2001,24 @@ message ListProviderClassesResponse {
...
+    Context context = 1;
+    // class is the class of the provider to get the definition for.
+    string class = 2;
+}
+
+message GetProviderDefinitionResponse {
+    ProviderDefinition definition = 1;
+}
+
+message ProviderDefinition {
```
JAORMX, line 2015: Let's rename this to `ProviderClassDefinition`

proto/minder/v1/minder.proto
```diff
@@ -1990,6 +2001,24 @@ message ListProviderClassesResponse {
...
 }
 
+message GetProviderDefinitionRequest {
+    // context is the context in which the provider definition is evaluated.
+    Context context = 1;
+    // class is the class of the provider to get the definition for.
+    string class = 2;
+}
+
+message GetProviderDefinitionResponse {
```
JAORMX, line 2011: `ClassDefinition`

dmjb (changes requested):

cmd/cli/app/provider/provider_enroll.go
```diff
@@ -83,24 +83,25 @@ func EnrollProviderCommand(ctx context.Context, cmd *cobra.Command, conn *grpc.C
...
+		},
+		Class: provider,
 	})
 	if err != nil {
 		return cli.MessageAndError("Error getting provider", err)
 	}
 
 	if token != "" {
-		if !prov.Provider.SupportsAuthFlow(minderv1.AuthorizationFlow_AUTHORIZATION_FLOW_USER_INPUT) {
+		if !prov.Definition.SupportsAuthFlow(minderv1.AuthorizationFlow_AUTHORIZATION_FLOW_USER_INPUT) {
```
JAORMX, line 97: if we're creating the provider in one go with the right auth flow, why check it here even? Let's just try to use the provider and fail from the server side. It should just display to the user as an error message. that would also simplify the logic on the CLI.

internal/controlplane/handlers_oauth.go
```diff
@@ -231,6 +241,43 @@ func (s *Server) generateOAuthToken(ctx context.Context, provider string, code s
 		return fmt.Errorf("error exchanging code for token: %w", err)
 	}
 
+	tx, err := s.store.BeginTransaction()
```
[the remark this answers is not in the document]
dmjb: `RuleTypeService` is a good reference point for this: https://github.com/stacklok/minder/blob/main/internal/ruletypes/service.go

I would start off by creating a `ProviderService` interface with a `CreateProvider` method. Other methods can be moved incrementally over to the interface in future.

Happy to pair on this if you want.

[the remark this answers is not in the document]
JAORMX: Can we do this as part of a subsequent PR? It's a good refactoring, but it can be a subsequent enhancement.

PR #2770, opened 2024-03-25
Store labels for profiles in the database

### LINKED ISSUE #2764: Store labels for profiles in the database
In order to be able to [filter labels](https://github.com/stacklok/minder/issues/2759) we need to first store them in the DB.

We can start simple with an additional column with the `TEXT[]` data type. If we have more use-cases for labels, we can then normalize the database to include a separate table with labels an a many-to-many join table, but currently I don't think it's needed.

The only interface that will support adding and updating these labels will be the one with subscription ID, not the general CreateProfile/Update profile exposed by our API. Since internally also the general CreateProfile/UpdateProfile API calls to the Subscription API, we need to add additional checks to the Subscription API as well.

jhrozek:
# Summary

Extends the `profiles` table with an additional column that stores profile
labels. This was one way I was considering, the other was to normalize the tables
and have a separate profile_labels table and then a "join table" linking profiles with
labels.

The labels column is not indexed. After having experimented with a GIN index, it was apparent that the cost/benefit ratio is not there. Also, we found out that we don't index the `entity_profiles.profile_id` column which was actually more costly than scanning the individual rows for labels.

I decided to go with a simple additional column for now because while the indexing
would have been easier with a separate table, the application logic for creating
and updating profiles would be much involved as we'd have to store the labels and
the link in a transaction. We might want to do that later, but I'm not sure if adding
database complexity when not needed is wise. At the same time, migration from the single
column to the normalized tables should be possible as well.

The second patch only allows setting labels on profiles that have a
subscription and disallows changing labels. We can change the validation logic, but IIRC
for now we're going to only have a single label and we already check the presence
of a subscription earlier, so this should be enough for now and we can change
the validation later when we actually have a better use-case.

Finally the third patch changes CreateProfile and UpdateProfile to handle labels.

Fixes #2764

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/database/migrations/000036_profile_label.down.sql
+++ b/database/migrations/000036_profile_label.down.sql
@@ -1,0 +1,17 @@
+-- Copyright 2024 Stacklok, Inc
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+ALTER TABLE profiles DROP COLUMN labels;
+
+DROP INDEX IF EXISTS idx_entity_profiles_profile_id;
--- a/database/migrations/000036_profile_label.up.sql
+++ b/database/migrations/000036_profile_label.up.sql
@@ -1,0 +1,20 @@
+-- Copyright 2024 Stacklok, Inc
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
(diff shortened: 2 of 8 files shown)


evankanderson:
I did some investigation, including reading https://pganalyze.com/blog/gin-index , and I think we don't want to use a GIN index.  From a discussion elsewhere, and EXPLAIN testing:

> The particular scenario I'm worried about is if we have e.g. 100k projects, and each of them has 5 profiles, one of which is labelled `stacklok:health` (for example).  The index for that particular label will contain 100k entries; each time we create a new project, we'll need to update that index entry, and at the end of the day, we're scanning the 100k-item index to avoid reading 4 items found by the B-Tree that don't have that label.

jhrozek:
> I did some investigation, including reading https://pganalyze.com/blog/gin-index , and I think we don't want to use a GIN index. From a discussion elsewhere, and EXPLAIN testing:
> 
> > The particular scenario I'm worried about is if we have e.g. 100k projects, and each of them has 5 profiles, one of which is labelled `stacklok:health` (for example).  The index for that particular label will contain 100k entries; each time we create a new project, we'll need to update that index entry, and at the end of the day, we're scanning the 100k-item index to avoid reading 4 items found by the B-Tree that don't have that label.

I pushed a new patch that drops the GIN and instead indexes the profile_id on entity_profiles as discussed on Slack.

jhrozek:
@dmjb PTAL

evankanderson (approved):
Do we need any tests for this right now, or are they coming in a future PR?

JAORMX (approved):
lgtm

PR #2771, opened 2024-03-25
Modify operations used by marketplace to accept transactions

dmjb:
Relates to #2541

The creation/update of profiles, rules and subscriptions need to happen inside transactions. Modify the relevant services to accept an ExtendedQuerier so the transaction can be created outside the code.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/profiles/service.go
```diff
@@ -47,45 +47,29 @@ import (
 type ProfileService interface {
 	// CreateProfile creates the profile in the specified project
 	// returns the updated profile structure on successful update
+	// subscriptionID should be set to nil when not calling
 	CreateProfile(
-		ctx context.Context,
```
dmjb, line 51: As part of this work, I decided to simplify the interface for profiles and rules to get rid of the dedicated subscription methods - now there is a single method and you pass a nil UUID if you do not need a subscription to be set.

internal/ruletypes/service.go
```diff
@@ -44,64 +44,44 @@ type RuleTypeService interface {
...
 		provider *db.Provider,
 		subscriptionID uuid.UUID,
 		ruleType *pb.RuleType,
+		qtx db.ExtendQuerier,
 	) error
 }
 
 type ruleTypeService struct {
-	store db.Store
+	//store db.Store
```
jhrozek, line 79: did you mean to leave this commented out?

internal/db/store.go
```diff
@@ -110,3 +110,24 @@ func (q *Queries) GetRuleEvaluationByProfileIdAndRuleType(
 	return ListRuleEvaluationsByProfileIdRow{},
 		fmt.Errorf("GetRuleEvaluationByProfileIdAndRuleType - expected 1 row, got %d", len(res))
 }
+
+// WithTransaction wraps an operation in a new DB transaction.
+// Ideally this would be a method of the Store interface, but Go's generics do
+// not allow for generic methods :(
+func WithTransaction[T any](store Store, fn func(querier ExtendQuerier) (T, error)) (result T, err error) {
```
jhrozek, line 117: Not your fault, but I wish we had a better name for the type we return from GetQuerierWithTransaction than ExtendQuerier..

internal/profiles/service.go
jhrozek: Do you suppose we will move more database methods here, e.g. Delete? If we do do you think we should rather have CreateWithTransaction accepting the db.ExtendQuerier interface and the simple functions being wrappers around the WithTransaction using the db.WithTransaction wrapper?

jhrozek (commented):
looks good, some comments.

internal/db/store.go
JAORMX: the name is funky, but remember you don't need to pass `ExtendedQuerier` though, we could pass `Store` as well which is a more general interface.

internal/marketplaces/subscriptions/service.go
```diff
@@ -36,51 +36,51 @@ import (
...
 ) SubscriptionService {
 	return &subscriptionService{
 		profiles: profiles,
 		rules:    rules,
-		store:    store,
 	}
 }
 
-func (s *subscriptionService) Subscribe(ctx context.Context, project types.ProjectContext, bundle reader.BundleReader) error {
+func (s *subscriptionService) Subscribe(
```
dmjb, line 76: Other change: I merged `Subscribe` and `CreateRuleTypes` together since they are always called together. Now that we have transactions, we can apply them atomically as well.

JAORMX (approved):

Issue #2769, opened 2024-03-25
Allow listing remote repositories without specifying provider


eleftherias:
The CLI command `minder repo register` and the RPC `ListRemoteRepositoriesFromProvider` should allow the option of not specifying a provider.
If a user does not specify a provider, then Minder should return all repositories from providers that implement the `repo-lister` trait.

PR #2802, opened 2024-03-26
Allow `ListRemoteRepositoriesFromProvider` to gather from multiple providers

JAORMX:
# Summary

If the provider is left empty in that API call, this iterates over all
providers and will do the call if they implement the right provider
trait.

The result is then re-used to register repositories, thus allowing us to
register repos from multiple providers seamlessly.

Fixes https://github.com/stacklok/minder/issues/2769

## Change Type

# Testing

# Review Checklist:


cmd/cli/app/repo/repo_register.go
```diff
@@ -125,8 +126,15 @@ func getUnregisteredInputRepos(inputRepoList string, alreadyRegisteredRepos sets
 
 func fetchRemoteRepositoriesFromProvider(ctx context.Context, provider, project string, client minderv1.RepositoryServiceClient) (
 	[]*minderv1.UpstreamRepositoryRef, error) {
+	var provPtr *string
+	if provider != "" {
+		provPtr = &provider
+	}
```
evankanderson, line 132: Can we make `""` also mean "choose any provider"?

evankanderson: (If we can do that, we might not need to change the client at all)

internal/controlplane/handlers_repositories.go
```diff
@@ -308,29 +309,61 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
 	entityCtx := engine.EntityFromContext(ctx)
 	projectID := entityCtx.Project.ID
 
-	provider, err := getProviderFromRequestOrDefault(ctx, s.store, in, projectID)
+	provs, err := listProvidersOrInferDefault(ctx, s.store, in, projectID)
```
evankanderson, line 312: What about `providersFromContext` for this method?

internal/controlplane/handlers_repositories.go
```diff
@@ -308,29 +309,61 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
 
-	if !p.Implements(db.ProviderTypeRepoLister) {
-		return nil, util.UserVisibleError(codes.Unimplemented, "provider does not implement repository listing")
+	failIfListerNotImplemented := len(provs) == 1
+
+	for _, provider := range provs {
+		zerolog.Ctx(ctx).Debug().
+			Str("provider", provider.Name).
+			Str("projectID", projectID.String()).
+			Msg("listing repositories")
```
evankanderson, line 327: This feels like it would be better as a trace rather than a debug message -- what do you think?

internal/controlplane/handlers_repositories.go
```diff
@@ -308,29 +309,61 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+
+		pbOpts := []providers.ProviderBuilderOption{
+			providers.WithProviderMetrics(s.provMt),
+			providers.WithRestClientCache(s.restClientCache),
+		}
+		p, err := providers.GetProviderBuilder(ctx, provider, s.store, s.cryptoEngine, &s.cfg.Provider, pbOpts...)
```
evankanderson, line 333: Not for this PR, but it feels like most (5 of 7) of the arguments to `GetProviderBuilder` are actually associated with the server, so we should think about having a "ProviderBuilderFactory" (or some better name, like `ProviderClass`) that takes the `store`, `cryptoEngine`, etc arguments, and is stored as a field on the `server`.

WDYT?

internal/controlplane/handlers_repositories.go
```diff
@@ -308,29 +309,61 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+			return nil, status.Errorf(codes.Internal, "cannot get provider builder: %v", err)
+		}
+
+		// check if the provider implements the repo lister interface
+		if failIfListerNotImplemented && !p.Implements(db.ProviderTypeRepoLister) {
+			return nil, util.UserVisibleError(codes.Unimplemented, "provider does not implement repository listing")
+		} else if !p.Implements(db.ProviderTypeRepoLister) {
+			// skip the provider if it does not implement the repo lister interface
+			continue
+		}
```
evankanderson, line 344: I don't feel like this is correct... what if I have 2 providers, neither of which implements `ProviderTypeRepoLister`?

It feels like we might want to have a counter `eligibleProviders` that's incremented if the provider is a repo lister, and then return the error at the end of the loop if `eligibleProviders` is 0.

internal/controlplane/handlers_repositories.go
```diff
@@ -308,29 +309,61 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+
+		results, err := s.listRemoteRepositoriesForProvider(ctx, provider.Name, p, projectID)
+		if err != nil {
+			return nil, err
+		}
+
+		out.Results = append(out.Results, results...)
+
+		// Telemetry logging
+		logger.BusinessRecord(ctx).Project = projectID
```
evankanderson, line 354: I don't think we want to repeatedly set the projectID to be the same thing on each call.

Additionally, I think we want to set `.Project` _before_ this loop (as early as we know it), so that we get RPC telemetry even for RPC failures.  (I missed this during the initial telemetry pass, and I've been commenting as we refactor the RPC methods.)

internal/controlplane/handlers_repositories.go
```diff
@@ -308,29 +309,61 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
 	}
 
+	return out, nil
+}
+
+func (s *Server) listRemoteRepositoriesForProvider(
+	ctx context.Context,
+	provName string,
+	p *providers.ProviderBuilder,
```
evankanderson, line 363: Again, not for this PR, but it feels like the `providers.ProviderBuilder` is actually a provider instance at this point, since it includes the `db.Provider` row.

internal/controlplane/common.go
```diff
@@ -67,13 +67,45 @@ func getProviderFromRequestOrDefault(
...
+	// Allows us to take into account the hierarchy to find the provider
+	parents, err := store.GetParentProjects(ctx, projectId)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "cannot retrieve parent projects: %s", err)
+	}
+	providers, err := store.ListProvidersByProjectID(ctx, parents)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "cannot retrieve providers: %s", err)
+	}
+	// if we do not have a provider name, check if we can infer it
```
evankanderson, line 88: I don't think this comment is correct anymore -- we're not trying to infer the provider here, we're returning all matching providers.

internal/controlplane/common.go
```diff
@@ -67,13 +67,45 @@ func getProviderFromRequestOrDefault(
...
+	if in.GetContext().GetProvider() == "" {
+		return providers, nil
+	}
+
+	prov, err := findProvider(in.GetContext().GetProvider(), providers)
```
evankanderson, line 93: Again, not for this PR, but this feels massively inefficient from a data transfer point of view -- we ignore the provider argument and fetch _all possible providers_, only to filter through all of them looking for one that matches the provider argument, rather than passing the provider as a `sql.NullString` to a `store` method.

But, don't change that for this PR, we should do it in a separate refactor.

evankanderson (approved):
I'd really love it if `RegisterRepository` could do the right thing with an empty provider argument, rather than needing to rely on the client fetching and filtering through the results when requesting a specific repository to be registered.

(Ideally, if I run `minder repo register --repo foo/baz`, that would correspond to one RPC to the server and wouldn't call `ListRemoteRepositories...` at all.)

With all that said, I'm not going to block this from going through, though existing client installs will need a `--provider` argument with this approach.

cmd/cli/app/repo/repo_register.go
JAORMX: That is already the case with this PR. I'm anyway refining the approach.

internal/controlplane/handlers_repositories.go
JAORMX: Good idea, let's do that in another PR.

cmd/cli/app/repo/repo_register.go
evankanderson: I was suggesting that we don't need to make the changes on 129-137, and we can just pass a pointer to `""` to the server, which would be less code and the same meaning.

internal/controlplane/common.go
```diff
@@ -36,44 +36,113 @@ type HasProtoContext interface {
...
+
+// getNameFilterParam allows us to build a name filter for our provider queries
+func getNameFilterParam(in *pb.Context) sql.NullString {
+	if in.GetProvider() == "" {
+		return sql.NullString{}
+	}
+	return sql.NullString{
+		String: in.GetProvider(),
+		Valid:  true,
+	}
```
evankanderson, line 135:
```suggestion
	return sql.NullString{
		String: in.GetProvider(),
		Valid:  in.GetProvider() != "",
	}
```

eleftherias (approved):
Looks good! I tested for backwards compatibility and that works as well.

I agree that we should have a followup story to derive the provider from the repository owner, especially to avoid making the frontend redo their API calls.

internal/controlplane/common.go
```diff
@@ -36,44 +36,113 @@ type HasProtoContext interface {
...
+	if err != nil {
+		return db.Provider{}, err
+	}
+
+	p, err := inferProvider(providers, name)
+	if err != nil {
+		return db.Provider{}, err
+	}
+
+	return p, nil
```
evankanderson, line 77:
```suggestion
	return inferProvider(providers, name)
```

internal/controlplane/common.go
```diff
@@ -36,44 +36,113 @@ type HasProtoContext interface {
...
-	matchesName := func(provider db.Provider) bool {
-		return provider.Name == in.GetContext().GetProvider()
-	}
-
-	i := slices.IndexFunc(providers, matchesName)
-	if i == -1 {
-		return db.Provider{}, util.UserVisibleError(codes.InvalidArgument, "invalid provider name: %s", in.GetContext().GetProvider())
-	}
-	return providers[i], nil
+	return providers[0], nil
```
evankanderson, line 147: I'm not sure I understand the logic here -- when `nameFilter` is valid, we'll return the first item, even if several were found?

What if zero rows were found, will this panic?  (I realize that right now, we only feed it results from `findProvider` which won't return a 0-length or `nil` array, but I worry about the function being re-used somewhere.  I think we could safely fold it into `getProviderFromRequestOrDefault`.

dmjb (approved):

evankanderson (approved):

internal/controlplane/common.go
JAORMX: this is not called if zero are found

JAORMX: https://github.com/stacklok/minder/pull/2812

internal/controlplane/common.go
JAORMX: https://github.com/stacklok/minder/pull/2813

internal/controlplane/common.go
JAORMX: https://github.com/stacklok/minder/pull/2814

PR #2822, opened 2024-03-27
Revert default config

eleftherias:
# Summary

Revert PRs that default the config value.
This is to continue to get the default port for the SQL connection for the events.

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/config/server/config.go
+++ b/internal/config/server/config.go
@@ -89,14 +89,8 @@ func setViperStructDefaults(v *viper.Viper, prefix string, s any) {
 		}
 		valueName := strings.ToLower(prefix + field.Tag.Get("mapstructure"))
 
-		// Extract a default value the `default` struct tag
-		// we don't support all value types yet, but we can add them as needed
-		value := field.Tag.Get("default")
-
 		if field.Type.Kind() == reflect.Struct {
-			if value != "{}" {
-				setViperStructDefaults(v, valueName+".", reflect.Zero(field.Type).Interface())
-			}
+			setViperStructDefaults(v, valueName+".", reflect.Zero(field.Type).Interface())
 			continue
 		}
 
@@ -105,6 +99,9 @@ func setViperStructDefaults(v *viper.Viper, prefix string, s any) {
 			continue
 		}
 
+		// Extract a default value the `default` struct tag
+		// we don't support all value types yet, but we can add them as needed
+		value := field.Tag.Get("default")
 		defaultValue := reflect.Zero(field.Type).Interface()
 		var err error // We handle errors at the end of the switch
 		fieldType := field.Type.Kind()
--- a/internal/config/server/config.go
+++ b/internal/config/server/config.go
@@ -19,7 +19,10 @@ package server
 
 import (
 	"fmt"
+	"reflect"
+	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/spf13/viper"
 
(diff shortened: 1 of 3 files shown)


teodor-yanev (approved):

JAORMX (approved):

Vyom-Yadav:
I get that this was done to maybe fix production immediately. But isn't the logically correct thing to remove `default:"{}"` from wherever it was causing an issue? https://github.com/stacklok/minder/blob/66179b892f89e053aa0766916b03f5088e686b5f/internal/config/server/events.go#L47-L51

What's the point of having `default:"{}"` if it's not honoured? That's just leaving a bug in.

evankanderson:
(But, as it is, the values in `EventConfig` are dangerous in combination with the code to not initialize items with `default:"{}"` tags.)

evankanderson:
I think the right answer is to remove those (ignored) defaults, and then add a complaint if they sneak back in...

evankanderson:
Going through this code again, I think the exciting thing that got missed in https://github.com/stacklok/minder/pull/2633 is that if you _skip_ calling `setViperStructDefaults` on the sub-message, fields don't merge properly between configs if you have multiple layers, _and_ can't be set using environment variables.  (See the comment at the top of `setViperStructDefaults`.)

Since this just bit us, I'll be working on a set of patches to roll this safely forward again.  Did you need the ability to clear out the defaults from the structs that you added in #2633, or were you just trying to be complete?

Vyom-Yadav:
> I think the exciting thing that got missed in https://github.com/stacklok/minder/pull/2633 is that if you skip calling setViperStructDefaults on the sub-message, fields don't merge properly between configs if you have multiple layers, and can't be set using environment variables

Right, with that, we dropped the ability to set things using env variables for struct fields with `default:"{}"`. 

For testing, if possible, I'd suggest having a test that replicates non-confidential server configuration options to ensure that the input configuration gives the expected _complete (all options)_ output. That way, we can bridge some gaps between SAAS and OSS. 

> Did you need the ability to clear out the defaults from the structs that you added in https://github.com/stacklok/minder/pull/2633, or were you just trying to be complete?

I sort of needed that. The problem is that types like db connection are shared by multiple fields, and having the same defaults doesn't make sense. So zeroing them using `default:"{}"` looked like a better option to have the user explicitly specify that using config. Also, now that you mentioned env vars above, I think validation in https://github.com/stacklok/minder/pull/2638 doesn't check env vars. It gives an error based on just the config file. I'll fix that portion.

JAORMX:
> The symptom (which we apparently don't have test coverage for) was that the `DatabaseConfig.port` for Minder was being zeroed in staging and production, which caused Minder to fail to start.
> 
> I think we need to figure out a _different_ way to configure this, and add a test that will prevent the misconfiguration from causing crashing servers.
> 
> We _also_ have an action item to investigate why our alerting didn't immediately complain when we had crashing servers, so we only noticed it several hours later.

Another issue was the EEA, which ended configured with a zero value in the aggregator interval. Which ended up effectively disabling the lock. We really need better testing in general.

evankanderson:
See https://github.com/stacklok/minder/pull/2836 for a fix to flag the `default:"{}"` and remove them, and then https://github.com/stacklok/minder/pull/2839 to [enable this case](https://github.com/stacklok/minder/pull/2822#issuecomment-2024391594):

> For testing, if possible, I'd suggest having a test that replicates non-confidential server configuration options to ensure that the input configuration gives the expected _complete (all options)_ output. That way, we can bridge some gaps between SAAS and OSS.

I _think_ the test added in #2836 effectively does this, for EEA.  I'm not sure that testing each option's pass-through in all configurations is going to be a good use of effort -- as an example, [Coveralls currently reports 60% coverage for `internal/config`](https://coveralls.io/github/stacklok/minder).  At the same time, we _do_ need a mechanism to check that a prototypical configuration produces a reasonable outcome.

Maybe it makes sense to [extend the `--dump_config`](https://github.com/stacklok/minder/blob/main/cmd/server/app/serve.go#L58) option to dump something more human-readable and then have a test that runs `go run ./cmd/server serve --dump_config $OTHER_FLAGS` and compares the output to what we expect?

> > Did you need the ability to clear out the defaults from the structs that you added in #2633, or were you just trying to be complete?
> 
> I sort of needed that. The problem is that types like db connection are shared by multiple fields, and having the same defaults doesn't make sense. So zeroing them using `default:"{}"` looked like a better option. The user explicitly specifies that using config. Also, now that you mentioned env vars above, I think validation in #2638 doesn't check env vars. It gives an error based on just the config file. I'll fix that portion.

I think the two PRs above (particularly #2839) should cover your use case.

Issue #2825, opened 2024-03-27
Create GitHub App manifest for quick creation


eleftherias:
For users that want to run their own Minder instance, or run Minder locally for testing, they will need to create a GitHub App.
We can create a [GitHub App manifest](https://docs.github.com/en/apps/sharing-github-apps/registering-a-github-app-from-a-manifest) to make the creation process faster and less error prone.

This is a nice to have for https://github.com/stacklok/epics/issues/154.

evankanderson:
@jhrozek may have one of these sitting around

evankanderson:
This happens once in the development setup, and no one's been excited to do it.

Issue #2522, opened 2024-03-06
Support GitHub App provider in the Minder CLI


eleftherias:
Depends on https://github.com/stacklok/minder/issues/2520

Add the GitHub app provider to the list of available providers for enrollment in the Minder CLI.

We need to be careful here not to break functionality for existing users.
For operations like registering repositories, the GitHub app should not be the default provider if the user doesn't have an enrolled GitHub app provider.

PR #2827, opened 2024-03-27
Add enrolment for GitHub App provider

### LINKED ISSUE #2520: Create a GitHub App provider

eleftherias:
# Summary

This add the last step for supporting a GitHub App provider on the Minder server.

Fixes #2520

### Out of scope for this PR

- Although the CLI can be used for testing this functionality, note that the CLI work for supporting the GitHub App provider is not complete and is a separate issue, https://github.com/stacklok/minder/issues/2522.

- This does not include the work for fetching the list of artifacts from a repo, that is separated into https://github.com/stacklok/epics/issues/280

## Change Type

# Testing

### Create a GitHub App
The first step is to create a GitHub App in your personal account.
(This can be tedious and I've created an issue that will make this quicker https://github.com/stacklok/minder/issues/2825.)

1) In your GitHub profile, go to Setting -> Developer Settings -> GitHub Apps
2) Select "New GitHub App"
3) Fill in the GitHub App name, this name must be unique among other GitHub Apps
4) Set the Callback URL to be `http://localhost:8080/api/v1/auth/callback/github-app/app`
5) Check the box "Request user authorization (OAuth) during installation"
6) Under Permissions, select the Repository Permissions: Administration (read and write), Contents (read and write), Metadata (read only), Packages (read and write), Pull requests (read and write), Repository security advisories (read and write), Webhooks (read and write), Workflows (read and write). You should have 8 Repository Permissions selected. Also select the following Organization Permission: Members (read-only)
7) Under "Where can this GitHub App be installed?", select "any account" if you want to test with organizations or accounts outside your own. Otherwise leave the default.

### Set up `server-config.yaml`
Once your GitHub App is created, you need to set up any App related configuration options.

In your `server-config.yaml` add the following section:
```
github-app:
  client_id: <client-id>
  client_secret: <client-secret>
  redirect_uri: "http://localhost:8080/api/v1/auth/callback/github-app/app" # This needs to match the registered callback URL in the GitHub App
```
Replace `<client-id>` with the client ID of your GitHub App, which is found in the General -> About section of your GitHub App on GitHub.
Replace `<client-secret>` with a client secret that you generate by clicking "Generate a new client secret"

Then, add the following section to your `server-config.yaml`:
```
provider:
  github-app:
    app_name: <app-name>
    app_id: <app-id>
    user_id: <user-id>
    private_key: ".secrets/github-app.pem"
```
Replace `<app-name>` with the name of your app, which you can get by looking at the GitHub URL when editing your GitHub App. For example, if the URL is `https://github.com/settings/apps/eleftherias-test-app`, then your app name is `eleftherias-test-app`.
Replace `<app-id>` with the app ID of your GitHub App, which is found in the General -> About section of your GitHub App on GitHub.
Replace `<user-id>` with the result of running this command `curl https://api.github.com/users/<app-name>%5Bbot%5D | jq ".id"`, where `<app-name>` is the App name you used above.

Finally, in the section General -> Private keys, click "Generate a private key". Rename the downloaded key to `github-app.pem` and add it to the `.secrets` directory in your Minder App.

### Actual Testing
Once you've set up your configuration and started Minder, you can enroll a GitHub App provider.
```
minder provider enroll --provider github-app
```
This will open a browser and take you through the authorization flow, where you will grant the App access to your account or organization.

You can then list the providers
```
minder provider list
```
You should see a provider named `github-app-<owner>`, where `<owner>` is the account or organization that you installed the App on.

All other commands are the same, as long as you specify the provider for each command.
For example, to register repos:
```
minder repo register --provider github-app-eleftherias
```

To create ruletypes:
```
minder ruletype create -f ./my-rules --provider github-app-eleftherias
```

To create a profile:
```
minder profile apply -f profile.yaml --provider github-app-eleftherias
```

To see the status of a profile:
```
minder profile status list -n artifact-profile --provider github-app-eleftherias
```

This is not the final state of the CLI, but it does allow for testing at this point.

# Review Checklist:


internal/controlplane/handlers_oauth.go
```diff
@@ -221,34 +263,110 @@ func (s *Server) processCallback(ctx context.Context, w http.ResponseWriter, r *
...
+	err = s.providers.ValidateGitHubInstallationId(ctx, token, installationID)
 	if err != nil {
-		if errors.Is(err, providers.ErrInvalidTokenIdentity) {
-			return newHttpError(http.StatusForbidden, "User token mismatch").SetContents(
-				"The provided login token was associated with a different GitHub user.")
+		return newHttpError(http.StatusForbidden, "User installation ID mismatch").SetContents(
+			"The GitHub user does not have access to the requested installation.")
+	}
+
+	if state != "" {
```
JAORMX, line 298: nit: Let's move this block to its own function for readability.

internal/controlplane/handlers_oauth.go
```diff
@@ -221,34 +263,110 @@ func (s *Server) processCallback(ctx context.Context, w http.ResponseWriter, r *
...
+		_, err = s.providers.CreateGitHubAppProvider(ctx, db.ProviderClassGithubApp, *token, stateData, installationID)
+		if err != nil {
+			if errors.Is(err, providers.ErrInvalidTokenIdentity) {
+				return newHttpError(http.StatusForbidden, "User token mismatch").SetContents(
+					"The provided login token was associated with a different GitHub user.")
+			}
+			return fmt.Errorf("error creating GitHub App provider: %w", err)
+		}
+
+		if stateData.RedirectUrl.Valid {
```
JAORMX, line 325: can you add a comment about what happens if the `stateData.RedirectUrl` is null or not valid?

internal/providers/credentials/github_app_jwt.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/golang-jwt/jwt/v4"
+)
+
+// CreateGitHubAppJWT creates a JWT token for a GitHub App
+func CreateGitHubAppJWT(appId int64, privateKey *rsa.PrivateKey) (string, error) {
+	// Create the Claims
+	claims := jwt.MapClaims{
+		"iat": time.Now().Unix(),
+		"exp": time.Now().Add(time.Minute * 10).Unix(),
+		"iss": appId,
```
JAORMX, line 32: isn't the `iss` supposed to be a string? even checking https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app suggests so.

internal/providers/service.go
```diff
@@ -157,6 +165,123 @@ func (p *providerService) CreateGitHubOAuthProvider(ctx context.Context, provide
...
+	installationOwner, err := p.getInstallationOwner(ctx, installationID)
+	if err != nil {
+		return nil, fmt.Errorf("error getting installation: %w", err)
+	}
+
+	tx, err := p.store.BeginTransaction()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error starting transaction: %v", err)
+	}
+	defer p.store.Rollback(tx)
```
JAORMX, line 185: @dmjb added a `db.WithTransaction` Let's use that instead.

https://github.com/stacklok/minder/blob/main/internal/db/store.go#L118

internal/providers/service.go
```diff
@@ -157,6 +165,123 @@ func (p *providerService) CreateGitHubOAuthProvider(ctx context.Context, provide
...
+
+	qtx := p.store.GetQuerierWithTransaction(tx)
+
+	// Save the installation ID and create a provider
+	savedProvider, err := qtx.CreateProvider(ctx, db.CreateProviderParams{
+		Name:       fmt.Sprintf("%s-%s", providerClass, installationOwner.GetLogin()),
+		ProjectID:  stateData.ProjectID,
+		Class:      db.NullProviderClass{ProviderClass: providerClass, Valid: true},
```
JAORMX, line 193: isn't the class something we can just set here?

internal/providers/service.go
```diff
@@ -157,6 +165,123 @@ func (p *providerService) CreateGitHubOAuthProvider(ctx context.Context, provide
...
+		Implements: app.Implements,
+		Definition: json.RawMessage(`{"github-app": {}}`),
+		AuthFlows:  app.AuthorizationFlows,
+	})
```
JAORMX, line 197: nit: This `CreateProvider` looks like something we could have as a constructor specific for the class.

eleftherias: Are you thinking we could add a `CreateProvider` to `github/app.go` and then the `Implements`, `AuthorizationFlows` and `Class` would be static? Then here we would call `app.CreateProvider(projectId, owner)`.

evankanderson:
> # Summary
> This add the last step for supporting a GitHub App provider on the Minder server.
> 
> Fixes #2520
> 
> ### Out of scope for this PR
> * Although the CLI can be used for testing this functionality, note that the CLI work for supporting the GitHub App provider is not complete and is a separate issue, [Support GitHub App provider in the Minder CLI #2522](https://github.com/stacklok/minder/issues/2522).

I think this is awesome to do separately.  ❤️ 

Thank you!

internal/controlplane/handlers_oauth.go
eleftherias: I gave it a try, but it ends up looking weird, because this part creates the provider and extracts the redirect URL and then redirects. So I had a function called `enrollProviderAndGetRedirectURL` which was really strange. I'm open to other suggestions, maybe breaking out a subsection of this block.

evankanderson (commented):
(I got through the refactor of the existing OAuth provider callback, will pick up in about 90 minutes)

docker-compose.yaml
```diff
@@ -47,6 +47,8 @@ services:
       # config file, point to them here:
       # - ./.github_client_id:/secrets/github_client_id:z
       # - ./.github_client_secret:/secrets/github_client_secret:z
+      # If you're using a GitHub App, you'll need to provide the private key:
+      - ./.secrets/github-app.pem:/app/.secrets/github-app.pem:z
```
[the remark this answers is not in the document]
eleftherias: Yeah I added that already in a separate PR https://github.com/stacklok/minder/pull/2811/files

internal/controlplane/handlers_oauth.go
```diff
@@ -162,6 +175,26 @@ func (s *Server) HandleProviderCallback() runtime.HandlerFunc {
...
+// HandleGitHubAppCallback handles the authorization callback from the GitHub App. This function validates the
+// GitHub user has access to the installation. It also gathers the state from the database and compares it to
+// the state passed in, if present. If they match a new GitHub App provider is created with the installation ID.
+// note: this is an HTTP only (not RPC) handler
+func (s *Server) HandleGitHubAppCallback() runtime.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
+		ctx := r.Context()
+
+		if err := s.processAppCallback(ctx, w, r, pathParams); err != nil {
+			if httpErr, ok := err.(*httpResponseError); ok {
```
jhrozek, line 187: should this be an `errors.Is` instead in case `processAppCallback` returns a wrapped error (I know we have the same issue in `HandleProviderCallback`

internal/providers/credentials/github_app_jwt.go
eleftherias: Both work, but I made it a string to be more correct

internal/controlplane/handlers_oauth.go
evankanderson: I think you mean `errors.As`, since we'd want to extract the wrapped error as well in that case in order to be able to call `WriteResponse`.

internal/controlplane/handlers_oauth.go
```diff
@@ -136,21 +138,32 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
 		return nil, err
 	}
 
+	var authorizationURL string
+	if slices.Contains(providerDef.AuthorizationFlows, db.AuthorizationFlowGithubAppFlow) {
+		gitHubAppConfig := s.cfg.Provider.GitHubApp
+		if gitHubAppConfig == nil {
+			return nil, status.Errorf(codes.Internal, "error getting GitHub App config: %s", err)
+		}
+		authorizationURL = fmt.Sprintf("https://github.com/apps/%v/installations/new?state=%v", gitHubAppConfig.AppName, state)
```
jhrozek, line 147: Just a question here, is the GH app provider something we want to eventually support for e.g. enterprise github installations? I guess when we do this URL should come from the GH instance URL which would be set in the provider settings?

internal/controlplane/handlers_oauth.go
```diff
@@ -221,34 +264,111 @@ func (s *Server) processCallback(ctx context.Context, w http.ResponseWriter, r *
...
+		}
+		if !valid {
+			return errors.New("invalid nonce")
+		}
+
+		// get projectID from session along with state nonce from the database
+		stateData, err := s.store.GetProjectIDBySessionState(ctx, state)
+		if err != nil {
+			return fmt.Errorf("error getting project ID by session state: %w", err)
+		}
```
evankanderson, line 313: This block seems like it could be a sub-block:

```golang
// returns non-nil error if state is not valid.  Free function, could live wherever.
stateData, err := getValidSessionState(ctx, state, s.cfg.Auth)
```

internal/controlplane/handlers_oauth.go
evankanderson: Yeah, the flow here is a bit odd:

```
if state:
  [some stuff]
  if have redirect:
    [ more stuff ]
    early return  // case A
  // no else
  // case B
else:
  [ some stuff]  // case C

common return from B and C, but not A
```

evankanderson: I'm inclined to share _less_ here, and have each branch do its own returns for clarity, possibly with a common helper method for each.

(especially since the common code is about 4 lines that we could make a helper in the `auth` module next to the `OAuthSuccessHtml` constant.)

internal/controlplane/server.go
```diff
@@ -339,14 +339,18 @@ func (s *Server) StartHTTPServer(ctx context.Context) error {
...
+	err := gwmux.HandlePath(http.MethodGet, "/api/v1/auth/callback/{provider}/cli", s.HandleOAuthCallback())
 	if err != nil {
 		return fmt.Errorf("failed to register provider callback handler: %w", err)
 	}
-	err = gwmux.HandlePath(http.MethodGet, "/api/v1/auth/callback/{provider}/web", s.HandleProviderCallback())
+	err = gwmux.HandlePath(http.MethodGet, "/api/v1/auth/callback/{provider}/web", s.HandleOAuthCallback())
 	if err != nil {
 		return fmt.Errorf("failed to register provider callback handler: %w", err)
 	}
+	err = gwmux.HandlePath(http.MethodGet, "/api/v1/auth/callback/{provider}/app", s.HandleGitHubAppCallback())
```
evankanderson, line 350: This is just (part of) a string we put in to the GitHub API, right?

jhrozek (dismissed):
The code reads well to me and seems to work well, too.
(Ozz and Evan are deeper in the GH app knowledge than I though so they might want to give their approvals too..)

internal/controlplane/handlers_oauth.go
eleftherias: I think so. I was spinning on this for a while, because at this section in the code the provider doesn't exist yet in the DB, so we would need to fetch the enterprise URL from somewhere else. I figured we could cross that bridge when we come to it and keep the flow simpler for now.

internal/providers/service.go
```diff
@@ -157,6 +165,116 @@ func (p *providerService) CreateGitHubOAuthProvider(ctx context.Context, provide
...
+) (*db.Provider, error) {
+	installationOwner, err := p.getInstallationOwner(ctx, installationID)
+	if err != nil {
+		return nil, fmt.Errorf("error getting installation: %w", err)
+	}
+
+	savedProvider, err := db.WithTransaction(p.store, func(qtx db.ExtendQuerier) (*db.Provider, error) {
+		// Save the installation ID and create a provider
+		savedProvider, err := qtx.CreateProvider(ctx, db.CreateProviderParams{
+			Name:       fmt.Sprintf("%s-%s", db.ProviderClassGithubApp, installationOwner.GetLogin()),
```
evankanderson, line 183: To make sure I'm following this right, the name will be `github-app-stacklok` or `github-app-StacklokLabs`, for example?

internal/providers/service.go
```diff
@@ -157,6 +165,116 @@ func (p *providerService) CreateGitHubOAuthProvider(ctx context.Context, provide
...
+		}
+
+		_, err = qtx.UpsertInstallationID(ctx, db.UpsertInstallationIDParams{
+			ProviderID: uuid.NullUUID{
+				UUID:  savedProvider.ID,
+				Valid: true,
+			},
+			OrganizationID:    installationOwner.GetID(),
+			AppInstallationID: strconv.FormatInt(installationID, 10),
+			EnrollingUserID:   sql.NullString{},
```
evankanderson, line 211: This is the default, so we don't need to specify it, right?

internal/providers/service.go
```diff
@@ -157,6 +165,116 @@ func (p *providerService) CreateGitHubOAuthProvider(ctx context.Context, provide
...
+		if err != nil {
+			return nil, err
+		}
+		return &savedProvider, nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return savedProvider, nil
```
evankanderson, line 222: These could just be `return db.WithTransaction(...)`, right?

evankanderson (commented):
A bunch of small comments, but this is looking good to me!

internal/controlplane/server.go
evankanderson: What I mean is that we don't need this to link up with anything else except our docs.

internal/controlplane/handlers_oauth.go
evankanderson: Yeah - I think when we cross that bridge, we add the URL to the GHE server to the app provider-specific table, and write some more code.  😁

evankanderson (dismissed):
The only one I feel kind of strongly about is @JAORMX 's https://github.com/stacklok/minder/pull/2827#discussion_r1541392596

But I don't feel strongly enough that the current code couldn't go in and be refactored to that in short order.  (i.e. the changes wouldn't cross a function boundary, so ¯\_(ツ)_/¯

JAORMX (approved):
I'm up for letting this in an optimizing in subsequent PRs

internal/controlplane/server.go
eleftherias: We also need to put it in our `server-config.yaml` because the oauth client library that we use requires a callback URL and it is required to match the callback for the GitHub App.

internal/providers/service.go
eleftherias: That's right

internal/controlplane/handlers_oauth.go
eleftherias: That's a good point. Maybe in case 1, where `state == ""` we redirect to the Minder frontend, since that will be the next step for them if they want to actually use Minder?

internal/controlplane/handlers_oauth.go
```diff
@@ -136,21 +138,32 @@ func (s *Server) GetAuthorizationURL(ctx context.Context,
...
+	var authorizationURL string
+	if slices.Contains(providerDef.AuthorizationFlows, db.AuthorizationFlowGithubAppFlow) {
+		gitHubAppConfig := s.cfg.Provider.GitHubApp
+		if gitHubAppConfig == nil {
+			return nil, status.Errorf(codes.Internal, "error getting GitHub App config: %s", err)
+		}
+		authorizationURL = fmt.Sprintf("https://github.com/apps/%v/installations/new?state=%v", gitHubAppConfig.AppName, state)
+	} else if slices.Contains(providerDef.AuthorizationFlows, db.AuthorizationFlowOauth2AuthorizationCodeFlow) {
+		authorizationURL = oauthConfig.AuthCodeURL(state, oauth2.AccessTypeOffline)
+	}
```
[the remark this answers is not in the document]
eleftherias: Go for it! I like the idea

internal/controlplane/handlers_oauth.go
evankanderson: 👍

internal/controlplane/server.go
evankanderson: Ah, okay.  I was trying to figure out if this should share constants, and now I know what to tie together.

evankanderson (approved):

Issue #2526, opened 2024-03-06
Use the correct remediation status while waiting for PR to be accepted


rdimitrov:
Currently when we have a PR remediation and it opens a PR, the remediation status is `Success` which is a bit confusing since we haven't really fixed the issue. Perhaps a more appropriate value is something like `Pending` or `Opened`.

* Note that alerts rely on this status, so updating it means we should also consider how this affects the actions engine and potentially the alerts executions too.

puerco:
Fixing this in #2530 2530

Issue #1207, opened 2023-10-15
remediate: store the existing PR remediation in the database


jhrozek:
At the moment the PR remediator is trying to figure out whether an existing remediation was opened by listing the opened pull requests. It might be better to instead track the PR in a database entry. This might give us:

- fewer API calls
- not relying on a magic comment in the PR
- be able to track closed PRs without listing them, too

Issue #1201, opened 2023-10-14
remediate: automatically close a remediation if it's no longer relevant


rdimitrov:
This enhancement is a result of the implementation and discussions in #1192.

<img width="929" alt="image" src="https://github.com/stacklok/mediator/assets/16540482/9f43deb3-9fa6-47ff-aefb-acc95acebde3">

**Brain dump:**

This _seems_ achievable if we use:
* current evaluation status `enginerr.ErrorAsEvalStatus(evalErr)` and 
* last remediate status `prevEvalFromDb.RemediateStatus.RemediateStatusTypes` 
* and based on that set
* the `interfaces.ActionCmd` that the remediate action should execute when doing `remediate.Do(..., cmd, ...)`, i.e. open(`ActionCmdOn`), close(`ActionCmdOff`) or do nothing.

cc: @jhrozek

PR #2833, opened 2024-03-27
Add a pending remediation status and support tracking opened PRs

rdimitrov:
# Summary

The following PR:

- Adds a new metadata column in the `rule_details_remediate` table for storing the opened PR number
- Refactors the way we process remediations through PRs - we no longer use magic comments to keep track of opened ones.
- If the issue is fixed out of band (through another PR for example) we'll now automatically close the PR along with the opened alert too
- Status is now `pending` when evaluation failed and remediation opened a PR
- Status is `skipped` if we don't need remediation
- Status is `success` if remediation was considered instant (i.e. `rest`) but we haven't re-evaluated yet (so rule is still failing). 
- Once we re-evaluate and the issue is fixed, the remediation status will become `skipped`.
- Adds a new `ClosePullRequest()` handler
- Removes the part where we were not updating the database for alerts with skipped status
- Refactors the part where we make the decision of whether we should remediate or not to happen in one place in the actions engine
- Fixes another bug where the remediation for pinned actions actually opens another PR after merging the first one

**What's left (can be done in a follow-up PR):**
- The PR adds more log messages, but I will revisit/update their usefulness 
- Currently if there's nothing to do, we return the previous action state as we do upsert at the end. The good thing is it updates the timestamp, but not sure how wasteful it is. Perhaps it's better to return a custom ErrActionDidNothing and skip upserting if this is the case. 
- Test this more with other ruletypes that remediate through a PR
- Add the pull request URL to be returned by the API
- Support rebasing/updating the PR if main changes while the PR is opened

Supersedes https://github.com/stacklok/minder/pull/2530
Fixes https://github.com/stacklok/minder/issues/2526
Fixes https://github.com/stacklok/minder/issues/1207
Fixes https://github.com/stacklok/minder/issues/1201

## Change Type

# Testing

Tested locally with a demo repository using the `dependabot_configured` ruletype:

**Start with a failing repo:**
* evaluation is failure
* rem is pending
* alert is on
* We have an opened PR and alert

**Case 1 - Merged the PR**
* eval is success
* rem is skipped
* alert is off
* alert is closed automatically

**Case 2 - Fixed the issue through another PR**
* Eval is ok
* rem is skipped
* alert is off
* both PR and alert are closed automatically

# Review Checklist:


internal/providers/github/common.go
```diff
@@ -564,6 +564,32 @@ func (c *GitHub) CreatePullRequest(
 	return pr, nil
 }
 
+// ClosePullRequest closes a pull request in a repository.
+func (c *GitHub) ClosePullRequest(
+	ctx context.Context,
+	owner, repo, number string,
+) error {
+	u := fmt.Sprintf("repos/%v/%v/pulls/%v", owner, repo, number)
```
jhrozek, line 572: Instead of hand-crafting the close you can use PullRequest.Edit and set the state to closed I think

database/query/profile_status.sql
```diff
@@ -50,9 +52,9 @@ INSERT INTO rule_details_alert (
 VALUES ($1, $2, $3, sqlc.arg(metadata)::jsonb, NOW())
 ON CONFLICT(rule_eval_id)
     DO UPDATE SET
-                  status = CASE WHEN $2 != 'skipped' THEN $2 ELSE rule_details_alert.status END,
-                  details = CASE WHEN $2 != 'skipped' THEN $3 ELSE rule_details_alert.details END,
-                  metadata = CASE WHEN $2 != 'skipped' THEN sqlc.arg(metadata)::jsonb ELSE rule_details_alert.metadata END,
+                  status = $2,
+                  details = $3,
+                  metadata = sqlc.arg(metadata)::jsonb,
```
jhrozek, line 57: I think I understand what you are doing now but I can't read the previous code anymore :-) why did we code it like this?

internal/engine/actions/actions.go
```diff
@@ -138,22 +138,82 @@ func (rae *RuleActionsEngine) processAction(
...
+	if prevEvalFromDb.AlertStatus.Valid {
+		prevRemediation = prevEvalFromDb.RemStatus.RemediationStatusTypes
+	}
+
+	// Start evaluation scenarios
+
+	// Case 1 - Do nothing if the evaluation status has not changed and remediation did not errored-out
+	if newEval == prevEval && prevRemediation != db.RemediationStatusTypesError {
+		return engif.ActionCmdDoNothing
+	}
```
jhrozek, line 188: aren't there legitimate cases for the PR evaluator where we want to push a new version of the PR even if the evaluation was failure before and is still failing?
e.g. I want frizbee to unroll tags to shas. The evaluation is failing and there is a PR. While the PR is open, someone pushes another workflow with more actions..

internal/engine/actions/actions.go
```diff
@@ -138,22 +138,82 @@ func (rae *RuleActionsEngine) processAction(
...
+	case db.EvalStatusTypesFailure:
+		// Case 4 - Evaluation has changed from something else to FAILED -> Remediation should be ON
+		// We should do nothing if the Remediation is already pending or successful
+		if db.RemediationStatusTypesPending == prevRemediation || db.RemediationStatusTypesSuccess == prevRemediation {
+			return engif.ActionCmdDoNothing
+		}
+		// The Remediation should be turned ON (if it wasn't already)
+		return engif.ActionCmdOn
+	case db.EvalStatusTypesSkipped:
+	case db.EvalStatusTypesPending:
```
jhrozek, line 211: can we even get pending eval here (its fine to leave it there, just asking)

database/query/profile_status.sql
evankanderson: Previous comments assume a bunch more context than you might have now:

https://github.com/stacklok/minder/pull/1228/files#r1365307972

internal/engine/actions/actions.go
```diff
@@ -138,22 +138,82 @@ func (rae *RuleActionsEngine) processAction(
 	params engif.ActionsParams,
 	metadata *json.RawMessage,
 ) (json.RawMessage, error) {
+	logger := zerolog.Ctx(ctx).Debug().
+		Str("action", string(actionType)).
+		Str("cmd", string(cmd))
+
+	// Exit early in case we shouldn't do anything, i.e., cmd != ActionCmdOn or cmd != ActionCmdOff
```
evankanderson, line 145: This comment doesn't quite line up with the code below, and the code below seems clear without it.  Remove the comment?

internal/engine/actions/actions.go
```diff
@@ -138,22 +138,82 @@ func (rae *RuleActionsEngine) processAction(
...
+		switch actionType {
+		case remediate.ActionType:
+			// Return the previous remediation status.
+			err := enginerr.RemediationStatusAsError(params.GetEvalStatusFromDb().RemStatus.RemediationStatusTypes)
+			return *(getMeta(params.GetEvalStatusFromDb().RemMetadata)), err
+		case alert.ActionType:
+			// Return the previous alert status.
+			err := enginerr.AlertStatusAsError(params.GetEvalStatusFromDb().AlertStatus.AlertStatusTypes)
+			return *(getMeta(params.GetEvalStatusFromDb().AlertMetadata)), err
+		}
```
evankanderson, line 157: Do we need a default case?

evankanderson, line 157: The above might read more easily if you extract the `params.GetEvalStatusFromDb()` that's used 4 times into an 'evalStatus' temporary.

```suggestion
		logger.Msg("skipping action")
		evalStatus := params.GetEvalStatusFromDb()
		switch actionType {
		case remediate.ActionType:
			// Return the previous remediation status.
			err := enginerr.RemediationStatusAsError(evalStatus.RemStatus.RemediationStatusTypes)
			return *(getMeta(evalStatus.RemMetadata)), err
		case alert.ActionType:
			// Return the previous alert status.
			err := enginerr.AlertStatusAsError(evalStatus.AlertStatus.AlertStatusTypes)
			return *(getMeta(evalStatus.AlertMetadata)), err
		}
```

evankanderson: In particular, if there's a different action type, it drops out of the `if` and goes on to the rest of the body.

internal/engine/actions/actions.go
```diff
@@ -195,65 +255,70 @@ func shouldAlert(
...
-			return engif.ActionCmdOn
-		}
-		// We should do nothing if the Alert is already ON
+	case db.EvalStatusTypesSkipped:
+	case db.EvalStatusTypesPending:
 		return engif.ActionCmdDoNothing
 	}
+
 	// Default to do nothing
 	return engif.ActionCmdDoNothing
```
evankanderson, line 283: Is this logic the same as 191-216?

internal/engine/actions/actions.go
rdimitrov: That's a good point, I haven't thought about the frizbee action as I assumed we are proposing only new files so far (dependabot, codeql, trivy, etc). I'll try to think about how to implement that 👍 Do you know if it's working like that right now and this will break it or?

internal/providers/github/common.go
rdimitrov: ah, missed that, thanks! 👍

database/query/profile_status.sql
rdimitrov: yeah, I think it was probably wrong to do it like that in the first place. Now it should act the same for remediations and alerts and the logs should reflect the exact state we processed 👍

internal/engine/actions/actions.go
rdimitrov: I don't think so too, but the lint check said it's a good practice so thus why I added it 👍

internal/engine/actions/actions.go
rdimitrov: good point, I'll remove it 👍

internal/engine/actions/actions.go
rdimitrov: My assumption was every action should be able to act correctly when it's called with the `ActionCmdDoNothing ` command, so even if we add a new one later and we miss to add it to this part there shouldn't be a problem. But now that you asked it it got me thinking that if we end up going to the action while skipping this part for some reason currently this will not preserve the old state.

I'll remove the `switch` here and add each case to be handled explicitly by each action. This way it's up to the action to implement what it should do on `ActionCmdDoNothing`, i.e. some may decide to preserve old state, some may choose something else.

Thanks!

internal/engine/actions/actions.go
jhrozek: with frizbee in particular I think the additional changes are the way it works now, yes, because unllike dependabot and codeql where we just put in the content, with frizbee you feed the content you get from evaluation to the frizbee library which modifies it and gives you back the modified content.

internal/engine/actions/actions.go
rdimitrov: Yes, I decided to refactor the separate `if` statements into a single switch now that we have duplicate behaviour for different evaluation statuses, i.e. `db.EvalStatusTypesError` and `db.EvalStatusTypesFailure`

rdimitrov:
These are the rest of the changes I think we can do in a follow-up PR:
* The PR adds more log messages, but I will revisit/update their usefulness
* Add the pull request URL to be returned by the API
* Support rebasing/updating the remediation PR if main changes while the PR is opened

database/migrations/000040_remediate_metadata.down.sql
```diff
@@ -0,0 +1,17 @@
+-- Copyright 2024 Stacklok, Inc
```
jhrozek, line 1: please rename the migrations to 41

internal/engine/actions/actions.go
```diff
@@ -115,7 +115,7 @@ func (rae *RuleActionsEngine) DoActions(
 		cmd := shouldRemediate(params.GetEvalStatusFromDb(), params.GetEvalErr())
 		// Run remediation
 		result.RemediateMeta, result.RemediateErr = rae.processAction(ctx, remediate.ActionType, cmd, ent, params,
-			nil)
+			getMeta(params.GetEvalStatusFromDb().RemMetadata))
```
jhrozek, line 118: since we always pass metadata for both remediate and alerts, would it make sense to change the interface for `actions.Do` to accept a `json.RawMessage`, not a pointer to it to avoid crashes in some corner cases?

internal/engine/actions/actions.go
```diff
@@ -138,22 +138,64 @@ func (rae *RuleActionsEngine) processAction(
...
+
+	// Get previous evaluation status
+	prevEval := db.EvalStatusTypesPending
+	if prevEvalFromDb.EvalStatus.Valid {
+		prevEval = prevEvalFromDb.EvalStatus.EvalStatusTypes
+	}
+
+	// Get previous Remediation status
+	prevRemediation := db.RemediationStatusTypesSkipped
+	if prevEvalFromDb.AlertStatus.Valid {
```
jhrozek, line 161: Should this say `RemStatus.Valid`?

database/migrations/000040_remediate_metadata.down.sql
rdimitrov: Done 👍

internal/engine/actions/actions.go
jhrozek: (not in this PR)

internal/engine/actions/actions.go
rdimitrov: Good catch! 💯

internal/engine/actions/actions.go
rdimitrov: good point, we certainly can 👍

internal/engine/actions/actions.go
jhrozek: I'm fine fixing this in a follow-up, but could this be solved by:
 - if the old status is failure always defer to the remediator
 - in the PR remediator, check what HEAD the PR points to. If it is the same as the hash of the commit we created locally, shortcut, otherwise go forward and force-push
 - the other remediators we have at least for now are "atomic" so this should not happen in the first place and if it does, then we should have anotoher mechanism that prevents an infinite remediate loop anyway

jhrozek (approved):
Thank you the logic reads much better now. I'm fine merging this and following up on this work with figuring out how to update the pending remediations.

rdimitrov:
Thanks! 

For anyone else looking at the PR, let's wait a bit until we merge this so we have the smoke test updates in a PR ready for merging too 👍

PR #2839, opened 2024-03-27
Allow overriding config struct defaults

### LINKED ISSUE #2838: Allow overriding config struct defaults
We re-use some types (notably, \'config.DatabaseConfig\') multiple times in our configuration tree.  It would be nice to be able to have different defaults for the different instances of these structs.  One way to do this would be to allow storing the overrides as JSON in the "defaults" tag, which would override the defaults of the inner type.

evankanderson:
# Summary

Follow-on to #2836, adds the ability to change defaults in fields from a parent struct.  This probably only works one-level-deep, but that seems sufficent.

Fixes #2838

## Change Type

# Testing

Updated test to take advantage of struct override; some additional manual tests with different values in the test (not included in the final PR).

# Review Checklist:


jhrozek:
What is the difference between this PR and PR #2836 ?

evankanderson:
> What is the difference between this PR and PR #2836 ?

This incorporated #2836, but adds support for the following:

```golang
type SQLEventConfig struct {
	// InitSchema is whether or not to initialize the schema
	InitSchema bool                  `mapstructure:"init_schema" default:"true"`
	Connection config.DatabaseConfig `mapstructure:"connection" default:"{\"dbname\":\"watermill\"}"`
}
```

Where we override the `dbname` field of the Connection, since `config.DatabaseConfig` is used several places.

evankanderson (commented):
@Vyom-Yadav PTAL.

@jhrozek , this is ready for review now.

internal/config/server/config.go
```diff
@@ -137,3 +139,16 @@ func setViperStructDefaults(v *viper.Viper, prefix string, s any) {
 		v.SetDefault(valueName, defaultValue)
 	}
 }
+
+func overrideViperStructDefaults(v *viper.Viper, prefix string, newDefaults string) {
+	overrides := map[string]any{}
+	if err := json.Unmarshal([]byte(newDefaults), &overrides); err != nil {
+		panic(fmt.Sprintf("Failed to parse overrides in %q: %v", prefix, err))
+	}
```
Vyom-Yadav, line 147: So this doesn't allow zeroing the config i.e. `default:"{}"`, is this by design? 
Following the current approach zeroing can be achieved by something like:
```go
Aggregator AggregatorConfig `mapstructure:"aggregator" default:"{\"lock_interval\":0}"`
```

internal/config/server/config.go
```diff
@@ -137,3 +139,16 @@ func setViperStructDefaults(v *viper.Viper, prefix string, s any) {
...
+
+	for key, value := range overrides {
+		// TODO: we don't do any fancy type checking here, so this could blow up later.
+		// I expect it will blow up at config-parse time, which should be earlier enough.
+		v.SetDefault(prefix+"."+key, value)
```
Vyom-Yadav, line 152: Yeah, this can blow up if types don't match, but it will blow up while unmarshalling from Viper (when we`ReadConfigFromViper`), so this seems alright.

Vyom-Yadav (approved):
Just a few comments:

internal/config/server/config.go
```diff
@@ -100,14 +105,11 @@ func setViperStructDefaults(v *viper.Viper, prefix string, s any) {
 			// TODO: we could allow `default` tags on structs to override the defaults inside the type.
 			// for now, we just warn about setting the tag, because it's ignored.
 			if _, ok := field.Tag.Lookup("default"); ok {
-				panic(fmt.Sprintf("Unsupported default on struct field: %q", valueName))
+				overrideViperStructDefaults(v, valueName, value)
```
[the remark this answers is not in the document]
evankanderson: Oops, thanks!

internal/config/server/config.go
evankanderson: Yes, that is by design.  In most cases empty defaults are actually useless (as noticed when the `events.sql.connection` defaults for port and TLS were zeroed).  Making this zeroing explicit reduces the effort to produce a _valid_ configuration leveraging defaults.  (At a slightly higher cost of possible misconfiguration by omitting a value.)

internal/config/server/config.go
evankanderson: Given the current use-cases, this seemed like the right balance between a large amount of type-matching code and deferring the error slightly later.

rdimitrov (approved):

PR #2841, opened 2024-03-28
Replace use of ProviderBuilder in controlplane

dmjb:
Relates to #2845

Create a new `TraitInstantiator` interface. This is essentially a dedicated factory interface with a similar interface to `ProviderBuilder`. This change allows us to stub out the process of creating instances of the Provider interfaces/traits from a `db.Provider`. Furthermore, because many of the dependencies which previously had to be passed to `ProviderBuilder` are now passed to `TraitInstantiator` during application wireup, this means that fewer dependencies have to be passed around in the code, and less boilerplate is required to create an instance of a Provider interface.

I changed the name since A) we are moving towards a model where Providers have "traits", and thus the responsibility is to instantiate instances of those traits and B) it always seemed strange to me that `ProviderBuilder` needed a `Provider` as an input - it creates the impression that there are two (or more) things called "Provider" in the codebase.

Note that I have not replaced all uses of ProviderBuilder yet. I'd like to get this change merged before converting the engine code to use this new interface.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/controlplane/handlers_repositories.go
```diff
@@ -322,21 +322,24 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
 	}
 
 	for _, provider := range provs {
+		// keep linter happy
+		// TODO: this should not be necessary post Go 1.22 - update linter
+		p := provider
```
JAORMX, line 327: Don't we already use and enforce go 1.22?

internal/providers/factory.go
```diff
@@ -0,0 +1,243 @@
+// Copyright 2024 Stacklok, Inc.
```
dmjb, line 1: In a future PR, I'll write unit tests for these methods. Right now - their behaviour is validated through pre-existing tests which exercise these methods.

internal/providers/factory.go
```diff
@@ -0,0 +1,243 @@
...
+	GetGit(ctx context.Context, provider *db.Provider) (provinfv1.Git, error)
+	GetHTTP(ctx context.Context, provider *db.Provider) (provinfv1.REST, error)
+	GetGitHub(ctx context.Context, provider *db.Provider, options *GitHubOptions) (provinfv1.GitHub, error)
+	GetRepoLister(ctx context.Context, provider *db.Provider) (provinfv1.RepoLister, error)
+}
+
+// GitHubOptions contains options used when creating a GitHub instance
+// If omitted, these will be looked up in the database
+type GitHubOptions struct {
+	Credential  provinfv1.GitHubCredential
```
JAORMX, line 47: We'll use credentials in all providers, why is this specific to GitHub? Did you check the OCI Providers POC?

internal/controlplane/handlers_githubwebhooks_test.go
```diff
@@ -239,27 +236,6 @@ func (s *UnitTestSuite) TestHandleWebHookRepository() {
 	repositoryID := uuid.New()
 	projectID := uuid.New()
 
-	// create a test oauth2 token
```
dmjb, line 242: These are no longer needed. Previously, these operations get called to create a ProviderBuilder. In the TraitInstantiator, they are only called when a dependent attempts to instantiate a provider traits. Since these test cases do not need a GitHub instance to be instantiated, they are unused.

internal/providers/factory.go
dmjb: In most cases, the credentials are loaded from the database. However, there are a few places where the Github trait is instantiated with credentials which are not in the DB - one is in `internal/providers/service.go` (which is in this PR), and there are two in the CLI (not in this PR). In those cases, it is necessary to provide some Github specific details, namely the owner filter and the credentials used.

Your comment did make me realize that there was a bug in my change where I assumed Github credentials in more places than I should have. I fixed that in the most recent commit.

internal/controlplane/handlers_repositories.go
dmjb: See the comment I left in this code in my most recent commit - the gosec authors have decided not to get rid of this warning until all versions prior to 1.22 are unsupported.

go.mod
```diff
@@ -62,7 +62,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.24.0
 	go.uber.org/mock v0.4.0
 	golang.org/x/crypto v0.21.0
-	golang.org/x/exp v0.0.0-20240222234643-814bf88cf225
```
dmjb, line 65: I updated `github.com/golangci/golangci-lint/cmd/golangci-lint` in an attempt to fix a false positive from gosec. I later discovered that I need to explicitly disable that linter warning, but I decided to leave these updates in place.

JAORMX: Let's do that upgrade but in another PR.

internal/db/domain.go
```diff
@@ -0,0 +1,24 @@
...
+package db
+
+import "slices"
+
+// This file contains domain-level methods for db structs
+
+// CanImplement returns true if the provider implements the given type.
+func (p *Provider) CanImplement(impl ProviderType) bool {
+	return slices.Contains(p.Implements, impl)
+}
```
evankanderson, line 24: I was just thinking about this last night.  I think this is a good first start, but I'm wondering whether it would make sense to rename `db.Provider` to `db.ProviderData` to emphasize that it's the data-at-rest representation rather than the active object representing a provider.  Either that, or we just accept that a `db.Provider` and a `provider.Provider` are unrelated to each other, and maybe use the same name more places.

sqlc renaming feature: https://docs.sqlc.dev/en/stable/howto/rename.html#tables

internal/providers/instantiator.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/stacklok/minder/internal/providers/ratecache"
+	"github.com/stacklok/minder/internal/providers/telemetry"
+	provinfv1 "github.com/stacklok/minder/pkg/providers/v1"
+)
+
+// TraitInstantiator is responsible for creating instances of the Profile
+// interfaces defined in `github.com/stacklok/minder/pkg/providers/v1` from
+// the providers stored in the database.
+type TraitInstantiator interface {
+	GetGit(ctx context.Context, provider *db.Provider) (provinfv1.Git, error)
```
evankanderson, line 37: Should this be part of the `Provider` interface?

evankanderson, line 37: What do you think of `AsGit` rather than `GetGit`, to make it more clear that this is effectively a down-casting operation?

internal/providers/instantiator.go
```diff
@@ -0,0 +1,249 @@
...
+	GetGitHub(ctx context.Context, provider *db.Provider, options *GitHubOptions) (provinfv1.GitHub, error)
+	GetRepoLister(ctx context.Context, provider *db.Provider) (provinfv1.RepoLister, error)
+}
+
+// GitHubOptions contains options used when creating a GitHub instance
+// If omitted, these will be looked up in the database
+type GitHubOptions struct {
+	Credential  provinfv1.GitHubCredential
+	OwnerFilter *string
+}
```
evankanderson, line 48: I'm not a fan of having these on `GetGitHub`, rather than having them be part of the TraitInstantiator object state.  Do you have a use case beyond `verifyProviderTokenIdentity`?

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -318,15 +317,6 @@ func (s *Server) parseGithubEventForProcessing(
...
 	}
 
-	pbOpts := []providers.ProviderBuilderOption{
-		providers.WithProviderMetrics(s.provMt),
-		providers.WithRestClientCache(s.restClientCache),
-	}
-	provBuilder, err := providers.GetProviderBuilder(ctx, prov, s.store, s.cryptoEngine, &s.cfg.Provider, pbOpts...)
-	if err != nil {
-		return fmt.Errorf("error building client: %w", err)
-	}
```
evankanderson, line 328: I like removing this factory from all the calls, but I feel like I want two layers of objects:

1. A `ProviderFactory` which takes all the current arguments except `ctx` and `prov`, and has `Build(ctx, currentProject, providerName) (Provider, error)`.  This combines lines 313-328 into a function, and doesn't directly expose `db.Provider` elsewhere.  As we discover other cases where we need to find a provider, the `ProviderFactory` would get new methods, such as `BuildForEntity(ctx, currentProject, entityType)`

3. `fooProvider` objects which wrap the `db.Provider` type with the selected implementation behavior, e.g. `ghAppProvider` and `ghOAuthProvider`.  Both of these would implement `v1.Provider`, which would have `As$Interface` (`AsGit`, `AsGitHub`, `AsREST`, etc) methods for down-casting as well as `Implements` if we need the checks.

I'm a little concerned with passing db row objects everywhere in that we end up making it hard to separate the schema from the business logic; ideally the db queries and token / crypto pieces would live in `ghAppProvider` and `ghOAuthProvider` and we have a central place to implement the mapping from user-provider context -> provider, rather than going directly to the database layer.

dmjb: I completely agree with you, but I also feel like it's a good idea to incrementally move towards that sort of model. Right now, too much knowledge of the creation of profiles exist in too many parts of the codebase. I want to reduce that coupling/lack of cohesion before introducing a new abstraction.

internal/providers/instantiator.go
dmjb: Good call, will make that change.

internal/providers/instantiator.go
dmjb: Yeah, the more I think about the approach I took, the less I like it. I might try another approach to avoid this hack.

internal/db/domain.go
evankanderson: > Agreed; I'd hold off on the rename until after #2841, #2842, and the follow-on that mostly hides `db.Provider` from controlplane before doing the rename.

internal/controlplane/handlers_githubwebhooks.go
evankanderson: I'm fine with incremental progress, as long as we agree on the destination.  I've just had bad experiences with incremental progress hiding disagreement on the final destination.

internal/providers/instantiator.go
evankanderson: Looking more at the code, I wonder if `verifyProviderTokenIdentity` should really be `verifyGitHubTokenIdentity`, and it should simply instantiate a `github.Client` itself.

In both cases, that call is being made before the provider is created, so it's a bit of an odd case that might be be addressed by not trying to re-use the rest of the framework.

dmjb:
Decided to take a different approach

Issue #2544, opened 2024-03-07
Include healthcheck bundle as part of the deploy of Minder


dmjb:
### Please describe the enhancement

Once we create the healthcheck bundle, it needs to be included in the image of the Minder deploy.

Constraint: For April, we only need to support the Healthcheck in Stacklok-hosted Minder deployments. We do not need to worry about how to solve this problem for self-hosted instances.

### Solution Proposal

@evankanderson has suggested using ko static assets to include the files in the bundle in the Minder server image.

PR #2849, opened 2024-03-28
Include healthcheck bundle in repo

dmjb:
Relates to #2544

Add the .tar.gz containing the healthcheck bundle. This is built from the profile/rule types repo, with modifications made to the contents to add the `stacklok` namespace to the name of each rule and profile. It contains:

1) The `stacklok-health-check` profile.
2) All rule types.

The healthcheck profile has a `healthcheck` label.

Tested/validated locally.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


JAORMX (commented):
Should we instead install this via an init container?

JAORMX (changes requested):
Let's not add a tarball to the repo. Instead, let's install the tarball via an init container on a shared volume and configure that container, volume and the needed config on our deployment side.

evankanderson:
My thought was to roll this out in a few stages.

1. (Checking in a binary artifact that we have a clear trail of provenance on -- this is not that, yet)
2. Building the `.tgz` as part of the build process.
3. Moving the build of the `.tgz` to someplace else, and pulling in an artifact.
4. Using an init container to put the `.tgz` in place (possibly via S3 pull + checksum or the like)

However, on reflection, I'm concerned about step 2 -- since we're pulling the rules from a separate repo, we may be forced to choose between:

1. Losing image build reproducibility.  This would make me very sad -- being able to get the same image bits in a year is a pretty strong defense against any kind of build hijinks.
2. Keeping some sort of cross-repo commit identifier in `stacklok/minder`, and then updating it regularly.  This would be be annoying from an operational and code review point of view, but probably the better of the two options.

However!  If we move more directly to step 4, we might be able to avoid the uncomfortable steps 2 and 3.... so I'm willing to at least consider it.

PR #2850, opened 2024-03-28
Remove provider from profile

### LINKED ISSUE #2848: Decouple profiles from provider instance
Currently profiles are tied to a specific provider within a project.
When we have multiple providers in a project, this means the same profile will have to be created for each provider.
Instead, we can tie profile to a project only, and apply them for every provider.

Related to https://github.com/stacklok/minder/issues/2522 and #2831

eleftherias:
# Summary

This removes the provider from the profile, so that profiles can be created once and then used by all providers.

Fixes  #2848

## Change Type

# Testing

# Review Checklist:


internal/marketplaces/subscriptions/service.go
```diff
@@ -136,7 +136,7 @@ func (s *subscriptionService) CreateProfile(
 		return fmt.Errorf("error while retrieving profile from bundle: %w", err)
 	}
 
-	_, err = s.profiles.CreateProfile(ctx, project.ID, project.Provider, subscription.ID, profile, qtx)
```
dmjb, line 139: Check the `ProjectContext` struct in this package (the type of `project` in this method). Now that providers are removed from both rules and profiles, I think we may be able to get rid of that struct, and just pass around a project ID to the methods in this interface.

internal/controlplane/handlers_profile.go
```diff
@@ -240,17 +231,15 @@ func getRuleEvalEntityInfo(
 	entityType *db.NullEntities,
 	selector *uuid.NullUUID,
 	rs db.ListRuleEvaluationsByProfileIdRow,
-	providerName string,
 ) map[string]string {
 	l := zerolog.Ctx(ctx)
-	entityInfo := map[string]string{
-		"provider": providerName,
-	}
+	entityInfo := map[string]string{}
```
evankanderson, line 236: I'm assuming this won't break existing clients?

internal/controlplane/handlers_repositories.go
```diff
@@ -98,18 +98,13 @@ func (s *Server) ListRepositories(ctx context.Context,
...
-		return nil, providerError(err)
-	}
-
 	reqRepoCursor, err := cursorutil.NewRepoCursor(in.GetCursor())
 	if err != nil {
 		return nil, util.UserVisibleError(codes.InvalidArgument, err.Error())
 	}
 
 	repoId := sql.NullInt64{}
-	if reqRepoCursor.ProjectId == projectID.String() && reqRepoCursor.Provider == provider.Name {
```
evankanderson, line 112: Why are we removing the provider from the repo cursor?

internal/controlplane/handlers_repositories.go
```diff
@@ -98,18 +98,13 @@ func (s *Server) ListRepositories(ctx context.Context,
 	entityCtx := engine.EntityFromContext(ctx)
 	projectID := entityCtx.Project.ID
 
-	provider, err := getProviderFromRequestOrDefault(ctx, s.store, in, projectID)
-	if err != nil {
-		return nil, providerError(err)
-	}
-
```
evankanderson, line 105: This is removing the ability to list repositories by provider.  Intentional?

evankanderson: (If repos are associated with providers, it seems useful to be able to limit the listo to one provider, to answer "what happens if I delete this" if nothing else.)

evankanderson (approved):
One question about ListRepositories, but the rest LGTM!  Thanks!

internal/controlplane/handlers_profile.go
eleftherias: The provider still gets added later in the flow, and we use the provider from the repo

internal/controlplane/handlers_repositories.go
eleftherias: If we don't get the provider from the context, then we can't compare it to what we have in the cursor

internal/controlplane/handlers_repositories.go
eleftherias: Yeah I can see us wanting to do that in the CLI. I'll update this to use the provider if it's in the context.

internal/marketplaces/subscriptions/service.go
```diff
@@ -43,14 +42,14 @@ type SubscriptionService interface {
 	// and bundle. It is a no-op if the project is already subscribed.
 	Subscribe(
 		ctx context.Context,
-		project types.ProjectContext,
+		projectID uuid.UUID,
```
dmjb, line 45: 🎉

dmjb (dismissed):

internal/controlplane/handlers_repositories.go
evankanderson: See my next comment -- I think the cursor could contain `""` or an actual provider name, since this is ListRepositories.

JAORMX (approved):

eleftherias:
Waiting for https://github.com/stacklok/minder/pull/2871 to be merged first, because it addresses [this comment](https://github.com/stacklok/minder/pull/2850#discussion_r1543413771)

dmjb (approved):

Issue #2727, opened 2024-03-20
Handle GitHub App installation deleted event


eleftherias:
When we setup our GitHub App, we will start receiving Webhook events when something changes with an App installation.

We should handle the "installation deleted" event by deleting the provider associated with the installation and removing any other data associated with that provider, such as registered repositories and installation IDs. 
Similarly if the installation is deleted and we don't have a provider, we should delete the installation ID from the installations table, if present

GitHub describes what the event payload looks like [here](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=deleted#installation).

The endpoint that receives these events on Minder should be different than the current `HandleGitHubWebHook` handler. This will give us the flexibility to use different secrets for the webhook that is implicitly created with the GitHub App and the webhooks that we create when a repository is registered.

I also believe we should create a separate event topic to publish the installation related events to.
```
InstallationEventTopic = "execute.installation.event"
```
```
r.Register(events.InstallationEventTopic, e.HandleInstallationEvent, e.mdws...)
```

<details>
  <summary>Some code from an old POC that may be useful (click to expand):</summary>
```
func (s *Server) HandleGitHuAppWebHook() http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		rawWBPayload, err := github.ValidatePayload(r, []byte(viper.GetString("webhook-config.webhook_secret")))
		if err != nil {
			fmt.Printf("Error validating webhook payload: %v", err)
			w.WriteHeader(http.StatusBadRequest)
			return
		}

		var payload struct {
			Action       string `json:"action"`
			Installation struct {
				ID int64 `json:"id"`
			} `json:"installation"`
		}

		if err := json.Unmarshal(rawWBPayload, &payload); err != nil {
			http.Error(w, "Error decoding JSON", http.StatusBadRequest)
			return
		}

		event, err := github.ParseWebHook(github.WebHookType(r), rawWBPayload)
		switch event.(type) {
		case *github.InstallationEvent:
			err := s.handleInstallationEvent(rawWBPayload)
			if err != nil {
				w.WriteHeader(http.StatusBadRequest)
			}
		}

	}
}

func (s *Server) handleInstallationEvent(payload []byte) error {
	var event github.InstallationEvent

	// unmarshal the payload into the event
	if err := json.Unmarshal(payload, &event); err != nil {
		return err
	}

	// get the Installation ID
	installationID := event.GetInstallation().GetID()

[...]

PR #2870, opened 2024-03-29
Handle GitHub App installation deleted event

jhrozek:
# Summary

Adds a new HTTP handler for handling the github app. The handler fires an event
to a newly added provider installation handler which in turns talks to the
provider service that handles the deletion.

There's some additional refactoring as well as questions about provider and installation
relationships that I'll point out inline in the PR.

Fixes: #2727

## Change Type

# Testing

I tested two scenarios:
1) "claimed" installation that I created with `minder provider enroll --provider=github-app --owner=jakubtestorg`
and then deleted from the org page. This hits the codepath that deletes both the installation and the provider.
2) "unclaimed" installation that I created by navigating to `https://github.com/apps/jakub-s-local-minder-test/installations/new` where `jakub-s-local-minder-test` is my gh app name and then deleting the app from the org page. This only hits the codepath that deletes the installation.

# Review Checklist:

### DIFF
--- a/database/query/provider_github_app_installations.sql
+++ b/database/query/provider_github_app_installations.sql
@@ -1,6 +1,9 @@
 -- name: GetInstallationIDByProviderID :one
 SELECT * FROM provider_github_app_installations WHERE provider_id = $1;
 
+-- name: GetInstallationIDByAppID :one
+SELECT * FROM provider_github_app_installations WHERE app_installation_id = $1;
+
 -- name: UpsertInstallationID :one
 INSERT INTO provider_github_app_installations
     (provider_id, app_installation_id, organization_id, enrolling_user_id, enrollment_nonce, project_id)
@@ -18,3 +21,6 @@ WHERE provider_github_app_installations.provider_id = $1
 
 -- name: GetInstallationIDByEnrollmentNonce :one
 SELECT * FROM provider_github_app_installations WHERE project_id = $1 AND enrollment_nonce = $2;
+
+-- name: DeleteInstallationIDByAppID :exec
+DELETE FROM provider_github_app_installations WHERE app_installation_id = $1;
--- a/database/query/providers.sql
+++ b/database/query/providers.sql
@@ -58,4 +58,4 @@ UPDATE providers
     WHERE id = sqlc.arg('id') AND project_id = sqlc.arg('project_id');
 
 -- name: DeleteProvider :exec
-DELETE FROM providers WHERE id = $1 AND project_id = $2;
+DELETE FROM providers WHERE id = $1;
--- a/internal/controlplane/handlers_githubwebhooks.go
+++ b/internal/controlplane/handlers_githubwebhooks.go
@@ -113,6 +113,69 @@ func entityFromWebhookEventTypeKey(m *message.Message) pb.Entity {
 	return pb.Entity_ENTITY_UNSPECIFIED
 }
 
+// HandleGitHubAppWebhook handles incoming GitHub App webhooks
+func (s *Server) HandleGitHubAppWebhook() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		wes := &metrics.WebhookEventState{
+			Typ:      "unknown",
+			Accepted: false,
+			Error:    true,
+		}
+		defer func() {
+			s.mt.AddWebhookEventTypeCount(r.Context(), wes)
+		}()
+
+		rawWBPayload, err := validatePayloadSignature(r, &s.cfg.WebhookConfig)
+		if err != nil {
+			log.Printf("Error validating webhook payload: %v", err)
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+
+		wes.Typ = github.WebHookType(r)
+		if wes.Typ == "ping" {
+			log.Printf("ping received")
+			wes.Error = false
+			return
+		}
+
+		m := message.NewMessage(uuid.New().String(), nil)
+		m.Metadata.Set(events.ProviderDeliveryIdKey, github.DeliveryID(r))
+		m.Metadata.Set(events.ProviderSourceKey, "https://api.github.com/") // TODO: handle other sources
+		m.Metadata.Set(events.GithubWebhookEventTypeKey, wes.Typ)
+
+		ctx := r.Context()
+		l := zerolog.Ctx(ctx).With().
+			Str("webhook-event-type", m.Metadata[events.GithubWebhookEventTypeKey]).
+			Str("providertype", m.Metadata[events.ProviderTypeKey]).
+			Str("upstream-delivery-id", m.Metadata[events.ProviderDeliveryIdKey]).
+			Logger()
+
+		if err := s.parseGithubAppEventForProcessing(rawWBPayload, m); err != nil {
+			wes = handleParseError(wes.Typ, err)
+			if wes.Error {
+				w.WriteHeader(http.StatusInternalServerError)
+			} else {
+				w.WriteHeader(http.StatusOK)
+			}
+			return
+		}
+
+		wes.Accepted = true
+		l.Info().Str("message-id", m.UUID).Msg("publishing event for execution")
+
+		if err := s.evt.Publish(providers.ProviderInstallationTopic, m); err != nil {
+			wes.Error = true
+			log.Printf("Error publishing message: %v", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		wes.Error = false
+		w.WriteHeader(http.StatusOK)
+	}
+}
+
 // HandleGitHubWebHook handles incoming GitHub webhooks
 // See https://docs.github.com/en/developers/webhooks-and-events/webhooks/about-webhooks
 // for more information.
@@ -280,6 +343,51 @@ func handleParseError(typ string, parseErr error) *metrics.WebhookEventState {
 	}
 	log.Print(logMsg)
 	return state
+}
+
+func (_ *Server) parseGithubAppEventForProcessing(
+	rawWHPayload []byte,
+	msg *message.Message,
+) error {
+	var event github.InstallationEvent
+
+	if msg.Metadata.Get(events.GithubWebhookEventTypeKey) != "installation" {
+		return newErrNotHandled("github app event %s not handled", msg.Metadata.Get(events.GithubWebhookEventTypeKey))
+	}
+
+	err := json.Unmarshal(rawWHPayload, &event)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling payload: %w", err)
+	}
+
+	action := event.GetAction()
+	if action == "" {
+		return fmt.Errorf("action is empty")
+	}
+
+	if action != "deleted" {
+		return newErrNotHandled("event %s with action %s not handled",
+			msg.Metadata.Get(events.GithubWebhookEventTypeKey), action)
+	}
+
+	installationID := event.GetInstallation().GetID()
+	if installationID == 0 {
+		return fmt.Errorf("installation ID is 0")
+	}
+
+	payloadBytes, err := json.Marshal(providers.GitHubAppInstallationDeletedPayload{
+		InstallationID: installationID,
+	})
+	if err != nil {
+		return fmt.Errorf("error marshalling payload: %w", err)
+	}
+
+	providers.ProviderInstanceRemovedMessage(
+		msg,
+		db.ProviderClassGithubApp,
+		payloadBytes)
+
+	return nil
 }
 
 func (s *Server) parseGithubEventForProcessing(
--- a/internal/providers/installations.go
+++ b/internal/providers/installations.go
@@ -49,22 +49,16 @@ const (
 
 // InstallationManager is a struct representing the installation manager
 type InstallationManager struct {
-	evt     events.Publisher
-	svc     ProviderService
-	querier db.Querier
+	svc ProviderService
 }
 
 // NewInstallationManager creates a new installation manager
 func NewInstallationManager(
-	evt events.Publisher,
-	querier db.Querier,
 	svc ProviderService,
-) (*InstallationManager, error) {
+) *InstallationManager {
 	return &InstallationManager{
-		evt:     evt,
-		svc:     svc,
-		querier: querier,
-	}, nil
+		svc: svc,
+	}
 }
 
 // Register implements the Consumer interface.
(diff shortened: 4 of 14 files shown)


database/query/providers.sql
```diff
@@ -58,4 +58,4 @@ UPDATE providers
     WHERE id = sqlc.arg('id') AND project_id = sqlc.arg('project_id');
 
 -- name: DeleteProvider :exec
-DELETE FROM providers WHERE id = $1 AND project_id = $2;
\ No newline at end of file
+DELETE FROM providers WHERE id = $1;
```
jhrozek, line 61: Note that this function was unused before. Given that we already are passing UUID I hope that it should be possible to delete without passing the project as well?

@JAORMX WDYT?

internal/providers/service.go
```diff
@@ -269,6 +270,31 @@ func (_ *providerService) ValidateGitHubInstallationId(ctx context.Context, toke
 	return nil
 }
 
+// GitHubAppInstallationDeletedPayload represents the payload of a GitHub App installation deleted event
+type GitHubAppInstallationDeletedPayload struct {
+	InstallationID int64 `json:"installation_id"`
+}
+
+func (p *providerService) DeleteGitHubAppInstallation(ctx context.Context, installationID int64) error {
+	installation, err := p.store.GetInstallationIDByAppID(ctx, strconv.FormatInt(installationID, 10))
```
jhrozek, line 279: @eleftherias since the table seems to be GH app specific and GH app installation IDs are Int64, why was the installation ID column set as TEXT?

```sql
CREATE TABLE provider_github_app_installations (
    app_installation_id TEXT PRIMARY KEY,
```

internal/providers/installations.go
```diff
@@ -0,0 +1,113 @@
...
+	if event == ProviderInstanceRemovedEvent {
+		return im.handleProviderInstanceRemovedEvent(ctx, msg)
+	}
+	return nil
+}
+
+func (im *InstallationManager) handleProviderInstanceRemovedEvent(ctx context.Context, msg *message.Message) error {
+	class := msg.Metadata.Get(ClassKey)
+
+	if db.ProviderClass(class) != db.ProviderClassGithubApp {
```
jhrozek, line 89: I'm not thrilled with leaking both the GH app class here and the GH app specific payload struct. I was thinking about having a callback registered on creating a GH app provider with the installation manager instead so that the installation manager would then just register class->callback map and pass on the payload as `[]byte` to the callback. But I didn't want to over-engineer without at least one review pass ...

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -113,6 +113,69 @@ func entityFromWebhookEventTypeKey(m *message.Message) pb.Entity {
 	return pb.Entity_ENTITY_UNSPECIFIED
 }
 
+// HandleGitHubAppWebhook handles incoming GitHub App webhooks
+func (s *Server) HandleGitHubAppWebhook() http.HandlerFunc {
```
jhrozek, line 117: There is a fair bit of code duplication between this webhook and the repo webhook. I can either:
1) do a quick refactor now to split the common code into functions
2) have the providers register a handler with the server

Ideally I'd like all github specific code out of controlplane, but maybe that's work for another day?

internal/providers/service.go
eleftherias: We created the tables before actually using them, and we missed this detail. I'm happy to create a migration to make it a BIGINT.

internal/providers/service.go
```diff
@@ -269,6 +270,31 @@ func (_ *providerService) ValidateGitHubInstallationId(ctx context.Context, toke
...
+	if err != nil {
```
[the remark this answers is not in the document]
eleftherias: I think it's ok to return an error. In `DeleteProfile` we also don't treat `sql.ErrNoRows` as a special case.

database/query/provider_github_app_installations.sql
```diff
@@ -1,6 +1,9 @@
 -- name: GetInstallationIDByProviderID :one
 SELECT * FROM provider_github_app_installations WHERE provider_id = $1;
 
+-- name: GetInstallationIDByAppID :one
+SELECT * FROM provider_github_app_installations WHERE app_installation_id = $1;
```
JAORMX, line 5: Do we have an index for this?

internal/controlplane/handlers_githubwebhooks.go
JAORMX: Work for another day, but we should target moving it at some point

internal/providers/installations.go
eleftherias: I say we skip the check on the provider class. This handler is github specific, as with our other event handlers. I think if we want to reuse them for other providers in the future, then we'll need to come up with a broad approach and apply it across the board.

database/query/provider_github_app_installations.sql
jhrozek: we don't. I'll do that along with the int64 conversion.

internal/providers/installations.go
```diff
@@ -0,0 +1,113 @@
+//
```
dmjb, line 1: I see from the coverage report that there's no coverage here. Can we get some unit tests for this?

internal/providers/service.go
```diff
@@ -269,6 +270,31 @@ func (_ *providerService) ValidateGitHubInstallationId(ctx context.Context, toke
 	return nil
 }
 
+// GitHubAppInstallationDeletedPayload represents the payload of a GitHub App installation deleted event
+type GitHubAppInstallationDeletedPayload struct {
+	InstallationID int64 `json:"installation_id"`
+}
+
+func (p *providerService) DeleteGitHubAppInstallation(ctx context.Context, installationID int64) error {
```
dmjb, line 278: Can we get some unit test coverage for this? It does not look like there is pre-existing coverage for this.

dmjb (changes requested):

internal/providers/service.go
jhrozek: Yes fair points I will add coverage. I wanted to get some feedback on the approach first - maybe I should have opened a draft to not confuse people

dmjb: That's fair - just want to make sure some more test coverage is added before merge to main.

jhrozek: I opened https://github.com/stacklok/minder/pull/2885 but because there are quite a few PRs touching that logic I would prefer to unblock this PR in the meantime and work on merging the tests as we touch services.go.

eleftherias (approved):
Approving to unblock, since there are lots of PRs in parallel

internal/providers/service.go
jhrozek: let's do it in separate patch - I initially added one to the branch but as discussed elsewehere it would take too long to merge this PR. I'll send that as a separate PR and it lands when it lands.

internal/providers/installations.go
jhrozek: I removed the provider class check (but I kept it in the message, the class might come handy in the message and it's just one set - if you prefer to nuke it altogether, I'm OK with it too)

database/query/provider_github_app_installations.sql
jhrozek: oops, sorry, actually we do - it's a primary key so it's indexed by default (found this out while writing the conversion to int64)

internal/providers/installations.go
jhrozek: added some basic coverage, thanks for reminding me!

jhrozek:
> Approving to unblock, since there are lots of PRs in parallel

@dmjb needs to remove his changes requested, too

internal/providers/service.go
jhrozek: also added tests

database/query/providers.sql
jhrozek: also added tests

internal/providers/service.go
jhrozek: > also added tests

sorry wrong thread (unresolving so this one keeps being visible)

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -113,6 +113,69 @@ func entityFromWebhookEventTypeKey(m *message.Message) pb.Entity {
...
+		m.Metadata.Set(events.ProviderDeliveryIdKey, github.DeliveryID(r))
+		m.Metadata.Set(events.ProviderSourceKey, "https://api.github.com/") // TODO: handle other sources
+		m.Metadata.Set(events.GithubWebhookEventTypeKey, wes.Typ)
+
+		ctx := r.Context()
+		l := zerolog.Ctx(ctx).With().
+			Str("webhook-event-type", m.Metadata[events.GithubWebhookEventTypeKey]).
+			Str("providertype", m.Metadata[events.ProviderTypeKey]).
+			Str("upstream-delivery-id", m.Metadata[events.ProviderDeliveryIdKey]).
+			Logger()
```
evankanderson, line 152: Do you want extract these values on 129, and be able to use `l` on 137 and beyond?

Also, a slight preference for `l := zerolog.Ctx(r.Context())...`, rather than adding a temp value whose lifetime I need to reason about.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -282,6 +345,51 @@ func handleParseError(typ string, parseErr error) *metrics.WebhookEventState {
...
+		return fmt.Errorf("error unmarshalling payload: %w", err)
+	}
+
+	action := event.GetAction()
+	if action == "" {
+		return fmt.Errorf("action is empty")
+	}
+
+	if action != "deleted" {
+		return newErrNotHandled("event %s with action %s not handled",
```
evankanderson, line 369: The error handling / control flow here sets my teeth on edge, but I don't have a better alternative yet.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -282,6 +345,51 @@ func handleParseError(typ string, parseErr error) *metrics.WebhookEventState {
...
+		InstallationID: installationID,
+	})
+	if err != nil {
+		return fmt.Errorf("error marshalling payload: %w", err)
+	}
+
+	providers.ProviderInstanceRemovedMessage(
+		msg,
+		db.ProviderClassGithubApp,
+		payloadBytes)
```
evankanderson, line 388: This feels weird... the function seems to be acting on a side-effect, but the method name is "Provider...Message", not "Send..." or "Publish..." or the like.

internal/providers/installations.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (im *InstallationManager) handleProviderInstallationEvent(msg *message.Message) error {
+	ctx := msg.Context()
+	zerolog.Ctx(ctx).Info().Msg("Handling provider installation event")
+
+	event := ProviderInstallationEvent(msg.Metadata.Get(InstallationEventKey))
+	if event == ProviderInstanceRemovedEvent {
+		return im.handleProviderInstanceRemovedEvent(ctx, msg)
+	}
+	zerolog.Ctx(ctx).Error().Msgf("Unknown event: %s", event)
```
evankanderson, line 77: This feels like it might want to be a `switch` with a `default`, but this is okay for now.

internal/providers/installations.go
```diff
@@ -0,0 +1,101 @@
...
+
+	err := json.Unmarshal(msg.Payload, &payload)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal payload: %w", err)
+	}
+
+	return im.svc.DeleteGitHubAppInstallation(ctx, payload.InstallationID)
+}
+
+// ProviderInstanceRemovedMessage returns the provider installation event from the message
```
evankanderson, line 92: This doesn't seem to return anything.  Also, the method starts with a noun; can this be `FillInstallationRemoved(msg, ...)` or the like?

Also, it feels weird to pass a `payload []byte` in here -- can we pass in a struct and have this do the marshalling?

Alternatively, I could see this growing up to actually create the message and set most of the fields on it.

internal/providers/service.go
evankanderson: I used `TEXT` despite it being GitHub-specific because I didn't want to attach too firmly to the implementation detail that these are int64 (for example, if we needed something different for GH Enterprise), but I'm happy to switch to int64 if we prefer.

I do think that there are a bunch of details to GitHub Apps that are probably unique, so using a column type that matches GitHub Apps closely doesn't really bother me here.

internal/providers/service.go
```diff
@@ -269,6 +270,31 @@ func (_ *providerService) ValidateGitHubInstallationId(ctx context.Context, toke
...
+			Int64("installationID", installationID).
+			Msg("Installation not claimed, deleting the installation")
+		return p.store.DeleteInstallationIDByAppID(ctx, strconv.FormatInt(installationID, 10))
+	}
+
+	zerolog.Ctx(ctx).Info().
+		Int64("installationID", installationID).
+		Str("providerID", installation.ProviderID.UUID.String()).
+		Msg("Deleting claimed installation")
+	return p.store.DeleteProvider(ctx, installation.ProviderID.UUID)
```
evankanderson, line 295: I think this is okay for now, but we should make sure that the flow here and in the `DeleteProvider` gRPC endpoint end up staying the same.

internal/controlplane/handlers_githubwebhooks.go
evankanderson: To be specific: it seems like `errNotHandled` is a success-error, which then gets consumed and transformed in some way by `handleParseError`, and the result of that interacts with GitHub's webhook retry-on-error logic.

evankanderson (approved):
No blocking comments, but the most serious is the removal of needing to know the project context from `DeleteProvider`, simply due to our experience of previous security issues there.  I might almost be tempted to create a `DeleteProviderUnconditional` or `DeleteProviderWithoutProject` to discourage people from using the non-project form where possible.

dmjb (approved):

jhrozek:
@eleftherias may I ask for another approve? I just rebased atop the current main. Sorry for the churn, I will try to split my PRs more granularly in the future.

Also @evankanderson you raised really good points and I will fix them, but let's get this merged so we avoid more back-and-forth at this moment. Thank you for the review, I was glad someone called me out on the project-less delete!

database/query/provider_github_app_installations.sql
```diff
@@ -1,6 +1,9 @@
 -- name: GetInstallationIDByProviderID :one
 SELECT * FROM provider_github_app_installations WHERE provider_id = $1;
 
+-- name: GetInstallationIDByAppID :one
```
eleftherias, line 4: non-blocking: A better name for this would be `GetInstallationByInstallationID`, since app ID is something else.

eleftherias (approved):
Added one more comment for the follow up PRs

database/query/provider_github_app_installations.sql
jhrozek: thank you I will take this into account

PR #2871, opened 2024-03-29
Make provider an optional filter

eleftherias:
Ref #2522

# Summary

This makes passing `--provider` optional. If no provider is specified then the command will apply for all providers in the project.

Ref #2522

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_profile.go
+++ b/internal/controlplane/handlers_profile.go
@@ -58,7 +58,7 @@ func (s *Server) CreateProfile(ctx context.Context,
 	}
 
 	// TODO: This will be removed once we decouple providers from profiles
-	provider, err := getProviderFromRequestOrDefault(ctx, s.store, in, entityCtx.Project.ID)
+	provider, err := getProviderFromRequestOrDefault(ctx, s.store, entityCtx.Provider.Name, entityCtx.Project.ID)
 	if err != nil {
 		return nil, providerError(err)
 	}
@@ -634,7 +634,7 @@ func (s *Server) UpdateProfile(ctx context.Context,
 	}
 
 	// TODO: This will be removed once we decouple providers from profiles
-	provider, err := getProviderFromRequestOrDefault(ctx, s.store, in, entityCtx.Project.ID)
+	provider, err := getProviderFromRequestOrDefault(ctx, s.store, entityCtx.Provider.Name, entityCtx.Project.ID)
 	if err != nil {
 		return nil, providerError(err)
 	}
--- a/internal/controlplane/handlers_repositories.go
+++ b/internal/controlplane/handlers_repositories.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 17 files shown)


internal/controlplane/handlers_repositories.go
```diff
@@ -401,60 +421,98 @@ func (s *Server) listRemoteRepositoriesForProvider(
...
+
+	err = s.repos.DeleteRepository(ctx, client, &repo)
+	if err != nil {
+		return status.Errorf(codes.Internal, "unexpected error deleting repo: %v", err)
+	}
+	return nil
+}
+
+// inferProviderByOwner returns the provider to use for a given repo owner
+func (s *Server) inferProviderByOwner(ctx context.Context, owner string, projectID uuid.UUID, providerName string,
```
eleftherias, line 456: @evankanderson This was something we had talked about. Do you have any ideas on how else we can do this?

internal/controlplane/handlers_repositories.go
```diff
@@ -288,8 +295,21 @@ func (s *Server) DeleteRepositoryByName(
...
+	providerFilter := sql.NullString{
+		String: providerName,
+		Valid:  providerName != "",
+	}
+	params := db.GetRepositoryByRepoNameParams{
+		Provider:  providerFilter,
+		RepoOwner: fragments[0],
+		RepoName:  fragments[1],
+		ProjectID: projectID,
+	}
```
dmjb, line 309: Move this query into RepositoryService - ideally we would leave DB queries out of the controlplane.

dmjb (changes requested):

internal/controlplane/handlers_repositories.go
evankanderson: I think we want to do something like this:

```golang
opts, err := s.getProvidersByTrait(ctx, s.store, nil, projectID, db.ProviderTypeGithub)
if err != nil {
  return nil, err
}

slices.SortFunc(opts, func(a, b db.Provider) int {
  // Sort GitHub provider last
  if a.Class == db.ProviderClassGithub && b == db.ProviderClassGithubApp {
    return 1
  }
  if a.Class = db.ProviderClassGithubApp && b.Class == db.ProviderClassGitHub {
    return -1
  }
  return 0
})

for _, prov := range slices {
  if github.CanHandleOwner(ctx, prov, owner) {  // wrapper around checking if the creds can access the repo
    return prov, nil
  }
}

return nil, fmt.Errorf("No providers can handle repo owned by %s", owner)
```

And we could start with the name comparison for ProviderName you have here, but eventually switch to calling `ListHooks` or the like to make sure that we have admin permissions.

evankanderson (commented):
Not sure if this is the best approach -- another option would be to include a "hint" as to the org name in the app installs table, though the org name _can_ change.

internal/controlplane/handlers_profile.go
```diff
@@ -58,7 +58,7 @@ func (s *Server) CreateProfile(ctx context.Context,
 	}
 
 	// TODO: This will be removed once we decouple providers from profiles
-	provider, err := getProviderFromRequestOrDefault(ctx, s.store, in, entityCtx.Project.ID)
+	provider, err := getProviderFromRequestOrDefault(ctx, s.store, entityCtx.Provider.Name, entityCtx.Project.ID)
```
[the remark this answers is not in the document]
eleftherias: I'm removing this line in the next PR #2850, so I'm not going to bother refactoring.

internal/controlplane/handlers_repositories.go
eleftherias: I moved all the DB queries to the service, but I left `DeleteRepository` as it was, because it also needs the repo ID to delete the repository.

evankanderson (approved):

PR #2882, opened 2024-04-01
Add rpc that checks if enrollment is successful based on state

eleftherias:
# Summary

This PR adds an more general rpc that the CLI can query to find out if the provider enrollment was successful, based on the state that was created and returned in the authorization URL.
This also saves the enrollment nonce (aka state) along with the installation ID and access token.
It also also adds the project ID to the installation IDs table. This was done to make collisions less likely between the enrollment nonces.

Ref #2522

## Change Type

# Testing

Run
```
minder provider enroll --provider github-app
```

Ensure that the CLI command does not finish, until the browser authorization flow is complete.

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -990,6 +1004,15 @@ message VerifyProviderTokenFromResponse {
     string status = 1;
 }
 
+message VerifyProviderCredentialRequest {
+    Context context = 1;
+    string enrollment_nonce = 2;
```
JAORMX, line 1009: let's add some docs to this entry

proto/minder/v1/minder.proto
```diff
@@ -990,6 +1004,15 @@ message VerifyProviderTokenFromResponse {
...
+}
+
+message VerifyProviderCredentialResponse {
+    string status = 1;
```
JAORMX, line 1013: let's add some docs about what this is supposed to be

internal/controlplane/handlers_oauth.go
```diff
@@ -470,6 +473,47 @@ func (s *Server) VerifyProviderTokenFrom(ctx context.Context,
...
+				EnrollmentNonce: sql.NullString{Valid: true, String: enrollmentNonce},
+			},
+		)
+
+		if errors.Is(err, sql.ErrNoRows) {
+			return &pb.VerifyProviderCredentialResponse{Status: "KO"}, nil
+		} else if err != nil {
+			return nil, status.Errorf(codes.Internal, "error getting installation ID: %v", err)
+		}
+	} else if err != nil {
```
JAORMX, line 510: nit: for readability, let's move error handling to the top. This function is quite indented that it gets hard to read.

eleftherias: I moved the `if` blocks around, hopefully it's more readable now, but let me know if you see a better ordering of the `if` statements.

JAORMX (dismissed):

cmd/cli/app/provider/provider_enroll.go
```diff
@@ -171,15 +170,15 @@ func enrollUsingOAuth2Flow(
 
 // callBackServer starts a server and handler to listen for the OAuth callback.
 // It will wait for either a success or failure response from the server.
-func callBackServer(ctx context.Context, cmd *cobra.Command, provider string, project string, port string,
-	done chan bool, client minderv1.OAuthServiceClient, openTime time.Time) {
+func callBackServer(ctx context.Context, cmd *cobra.Command, project string, port string, done chan bool,
```
evankanderson, line 173: Why is port a `string` and not an `int`?  (I'm generally suspicious of multiple `string`-type arguments next to each other -- I've seen them get swapped more often than I'd like.

database/migrations/000042_credential_enrollment_nonce.up.sql
```diff
@@ -0,0 +1,28 @@
...
+BEGIN;
+
+ALTER TABLE provider_github_app_installations ADD COLUMN enrollment_nonce TEXT;
+
+ALTER TABLE provider_github_app_installations ADD COLUMN project_id UUID;
+
+ALTER TABLE provider_github_app_installations
+    ADD CONSTRAINT provider_github_app_installations_project_id_fkey
+        FOREIGN KEY (project_id)
+            REFERENCES projects(id);
```
evankanderson, line 24: Do we want `ON DELETE CASCADE`?

proto/minder/v1/minder.proto
```diff
@@ -990,6 +1004,20 @@ message VerifyProviderTokenFromResponse {
...
+// VerifyProviderCredentialRequest contains the enrollment nonce (aka state) that was used when enrolling the provider
+message VerifyProviderCredentialRequest {
+    Context context = 1;
+
+    // enrollment_nonce is the state parameter returned when enrolling the provider
+    string enrollment_nonce = 2;
+}
+
+// VerifyProviderCredentialRequest responds with an OK status if the credential was found, KO if the credential was not
+// found or an error if an error occurred during lookup
```
evankanderson, line 1016: The string values for `VerifyProviderTokenFrom` are a bit funky -- it feels like we might want to change the `Status` to be something like `ProviderName`, with the following states:

* `""` if the supplied nonce is not active
* the string name of the provider if the nonce has been activated.

internal/controlplane/handlers_oauth.go
```diff
@@ -470,6 +473,48 @@ func (s *Server) VerifyProviderTokenFrom(ctx context.Context,
...
+	// check if a token has been created matching the nonce
+	_, err := s.store.GetAccessTokenByEnrollmentNonce(ctx, db.GetAccessTokenByEnrollmentNonceParams{
+		ProjectID:       projectID,
+		EnrollmentNonce: sql.NullString{Valid: true, String: enrollmentNonce},
+	})
+
+	if err != nil {
+		if !errors.Is(err, sql.ErrNoRows) {
+			return nil, status.Errorf(codes.Internal, "error getting access token: %v", err)
+		}
```
evankanderson, line 496: This might read more clearly as:

```suggestion
	if !errors.Is(err, sql.ErrNoRows) {
		if err == nil {
			return &pb.VerifyProviderCredentialResponse{Status:"OK"}, nil
		} else {
			return nil, status.Errorf(codes.Internal, "error getting access token: %v", err)
		}
	}
```

And just early-exit if there's a PAT row or a hard error.  Then the second search could also early-exit if it finds success, and the default case would be to return "KO".

evankanderson (commented):
(I hit the same thing and did #2884, which is in the same vein but a bit smaller.

internal/controlplane/handlers_oauth.go
evankanderson: I put in a suggestion to have two early-exit cases (found successful PAT and found successful install), and then have the default case to be return `KO`.

database/migrations/000042_credential_enrollment_nonce.up.sql
eleftherias: I also thought about this. We have an `ON DELETE CASCADE` on the provider, which implicitly means that if a project is deleted, then the provider is deleted, then the installation is deleted. I'm not sure if it's better to make it explicit here.

proto/minder/v1/minder.proto
eleftherias: That's true, I updated this to return a `created` boolean and the `providerName`. Eventually we can use the provider name to print "Provide XYZ successfully enrolled" from the CLI, but I haven't made that change in this PR.

cmd/cli/app/provider/provider_enroll.go
eleftherias: No idea why it was like that. I changed it to be an int.

JAORMX (approved):

PR #2883, opened 2024-04-01
Disable ingest caching for artifact entities and add a wait period

rdimitrov:
# Summary

The following PR:
* **Disables ingest caching for Artifact entities.** 

The reasoning is artifacts are a bit different than the other entities. During the ingest step we actually check if there's provenance information for this artifact version (i.e. `stacklok/minder:latest`) and if the signature is not yet at the time of this evaluation, we ingest a result that says there's no signature found. 

The issue comes when the 2nd event comes in (the one about the actual signature) before the ingest expire time, we start to evaluate it but use the ingested cache from the 1st one which says - for this version, there's no provenance information available.

* **Adds a wait period for evaluating artifact entity events**

This is an after-effect from the 1st one. Now that we don't cache anything this might result in a short flick of an alert being opened and then closed because:
* On the 1st evaluation we'll see there's no signature yet and we'll open an alert. 
* On the 2nd evaluation we'll see there's a signature and if it satisfies the profile parameters it will close the alert.

Adding a wait period for the signature flow to complete should help for this issue. This should help only for the case of online signing. Of course in cases where the signing takes more than the configured time (starting with 10s) or it happens offline, the alert will stay open until it is completed.

That said, perhaps the wait period alone should be enough but it feels right to do both. Happy to discuss this more if needed 👍 

## Change Type

# Testing

The issue was not reproducible locally but we can test that in staging where it happens ~90% of the time.

# Review Checklist:

### DIFF
--- a/internal/engine/executor.go
+++ b/internal/engine/executor.go
@@ -45,6 +45,9 @@ const (
 	// DefaultExecutionTimeout is the timeout for execution of a set
 	// of profiles on an entity.
 	DefaultExecutionTimeout = 5 * time.Minute
+	// ArtifactSignatureWaitPeriod is the waiting period for potential artifact signature to be available
+	// before proceeding with evaluation.
+	ArtifactSignatureWaitPeriod = 10 * time.Second
 )
 
 // Executor is the engine that executes the rules for a given event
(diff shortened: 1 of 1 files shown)


JAORMX (approved):
shipit!

PR #2892, opened 2024-04-02
Add remediation_url to RuleEvaluationStatus

puerco:
# Summary

This PR adds a field to capture the remediation URL to the evaluation status struct to make it possible to serve it in the rule eval response. 

Follow up to https://github.com/stacklok/minder/issues/2879

## Change Type

# Testing

# Review Checklist:


puerco:
This change is intended to return the URL of the remediation to surface it in clients, but I'm thinking we should instead create a RemediationStatus block to avoid adding fields to the `RuleEvaluationStatus` struct. We already have a field of type `EvalResultAlert` in the Rule Evaluation:

https://github.com/stacklok/minder/blob/c7bf284183c1d36c9fb3a9e98be1860698887acd/proto/minder/v1/minder.proto#L1175-L1205

I think it makes sense to factor out the Remediation status too.  /cc @rdimitrov

JAORMX (approved):

JAORMX:
FWIW, I think factoring it out is a good idea

rdimitrov (approved):
I agree on the refactoring too 👍 I don't have a strong preferences though on where we should do it now or probably as part of the v2 proto 👍 Perhaps the latter one is better since we'll go through all of the use cases and figure out the proper objects layout based on that

Issue #2845, opened 2024-03-28
Create interfaces for Provider instantiation


dmjb:
### Please describe the enhancement

Right now, any part of the code which needs to interact with Providers ends up repeating the same set of logic. This makes it difficult to stub out provider operations for testing, and creates coupling between the logic for creating providers, and the business logic which uses providers.

Define interfaces for creating/getting providers to reduce coupling and ease testing.

### Solution Proposal

At minimum, make ProviderBuilder an interface, and create an interface for pulling `db.Provider` out of the database. We may want some higher level abstractions to combine retrieving a `db.Provider`, and instantiating a particular Provider interface.

PR #2911, opened 2024-04-02
Move more provider access methods to store

dmjb:
Relates to #2845

Migrate more queries for providers to ProviderStore. This will make it easier to refactor away direct access to the provider table in subequent PRs.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_githubwebhooks.go
+++ b/internal/controlplane/handlers_githubwebhooks.go
@@ -412,16 +412,8 @@ func (s *Server) parseGithubEventForProcessing(
 		return fmt.Errorf("error getting repo information from payload: %w", err)
 	}
 
-	ph, err := s.store.GetParentProjects(ctx, dbRepo.ProjectID)
-	if err != nil {
-		return fmt.Errorf("error getting project hierarchy: %w", err)
-	}
-
 	// get the provider for the repository
-	prov, err := s.store.GetProviderByName(ctx, db.GetProviderByNameParams{
-		Name:     dbRepo.Provider,
-		Projects: ph,
-	})
+	prov, err := s.providerStore.GetByName(ctx, dbRepo.ProjectID, dbRepo.Provider)
 	if err != nil {
 		return fmt.Errorf("error getting provider: %w", err)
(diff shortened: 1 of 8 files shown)


internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -412,16 +412,8 @@ func (s *Server) parseGithubEventForProcessing(
...
-	if err != nil {
-		return fmt.Errorf("error getting project hierarchy: %w", err)
-	}
-
 	// get the provider for the repository
-	prov, err := s.store.GetProviderByName(ctx, db.GetProviderByNameParams{
-		Name:     dbRepo.Provider,
-		Projects: ph,
-	})
+	prov, err := s.providerStore.GetByName(ctx, dbRepo.ProjectID, dbRepo.Provider)
```
jhrozek, line 416: Genuine question: The old code was calling GetParentProjects and then getting the provider by name in parent. The new code goes to GetByName, which calls findProjects which calls getProjectHierarchy. Are those equivalent?

dmjb: `getProjectHierarchy` also calls `GetParentProjects`. In fact, I may get rid of the separate `getProjectHierarchy` function since it's only called in one places (originally in this branch it was called more than once)

dmjb: That being said - one important difference is that the `GetByName` will check to see if there are any duplicate results and return an error if there is ambiguity about being what is returned, whereas the code I have replaced will not (see the implementation of `GetByName`).

@eleftherias - since I think you wrote this code: is there any possibility that the original query could match more than one provider?

eleftherias: It is possible, but only in a strange corner case where someone renamed their GitHub organization to have the same name as a GitHub organization that they previously installed Minder on. We generally don't handle renames yet, so I think we can ignore that case for now.

jhrozek (approved):

PR #2917, opened 2024-04-02
Don't mark a PR remediation as error if no previous result exists

jhrozek:
# Summary

If there was no previous remediation status the Null DB type would
have had Valid:false and the actual value set to an empty string which
would trigger the default case of returning a generic error in
`RemediationStatusAsError`.

## Change Type

# Testing

create a profile with remediate: on that uses a pull request remediation but
is already passing. It would end up in error.

# Review Checklist:


jhrozek:
@rdimitrov PTAL do you think this is correct?

puerco (approved):

evankanderson (approved):

PR #2922, opened 2024-04-03
Add delete provider command

### LINKED ISSUE #2897: Add provider delete command
We currently don't have support for deleting providers. We should add a CLI command:
```
minder provider delete --id XYZ
```

eleftherias:
# Summary

This adds the `minder provider delete` command, with the option of passing a provider ID or name.
The command will delete the provider from the DB, delete all associated repos, remove the repo webhooks and uninstall the GitHub App from the GitHub org, if it is a GitHub App provider.

Fixes #2897

## Change Type

# Testing

# Review Checklist:


internal/repositories/github/service.go
```diff
@@ -211,6 +218,33 @@ func (r *repositoryService) DeleteRepository(ctx context.Context, client ghclien
 	return nil
 }
 
+func (r *repositoryService) DeleteRepositoriesByProvider(
```
dmjb, line 221: Could you add a unit test for this?

dmjb (changes requested):

internal/repositories/github/service.go
eleftherias: On it

cmd/cli/app/provider/provider_delete.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// DeleteProviderCommand deletes the provider in a specific project
+func DeleteProviderCommand(ctx context.Context, cmd *cobra.Command, conn *grpc.ClientConn) error {
+	client := minderv1.NewProvidersServiceClient(conn)
+
+	project := viper.GetString("project")
+	name := viper.GetString("name")
+	id := viper.GetString("id")
+
```
evankanderson, line 53: Set `cmd.SilenceUsage = true` here to avoid printing out the command usage if the RPC to the server fails.

See e.g. [line 71 of `provider_enroll.go`](https://github.com/stacklok/minder/blob/main/cmd/cli/app/provider/provider_enroll.go#L71) for an example.

internal/controlplane/handlers_providers.go
```diff
@@ -156,6 +158,103 @@ func (_ *Server) ListProviderClasses(
...
+		if err != nil {
+			return status.Errorf(codes.Internal, "error creating github provider: %v", err)
+		}
+
+		// Delete all repositories associated with the provider and remove the webhooks
+		err = s.repos.DeleteRepositoriesByProvider(ctx, client, provider.Name, projectID)
+		if err != nil {
+			return status.Errorf(codes.Internal, "error deleting repositories: %v", err)
+		}
+	}
```
evankanderson, line 248: This feels like it should be a part of the DeleteProvider implementation, rather than in `ControlPlane`.  e.g. GitHub providers should understand that they need to clean up webhooks, but other types of providers might have other types of cleanup they need to do.

(e.g. we talked about `DeleteProvider` for GHA un-installing the app from the org.)

internal/providers/service.go
```diff
@@ -390,13 +391,57 @@ func (p *providerService) DeleteGitHubAppInstallation(ctx context.Context, insta
...
+	}
+
+	privateKey, err := p.config.GitHubApp.GetPrivateKey()
+	if err != nil {
+		return fmt.Errorf("error getting GitHub App private key: %w", err)
+	}
+	jwt, err := credentials.CreateGitHubAppJWT(p.config.GitHubApp.AppID, privateKey)
+	if err != nil {
+		return fmt.Errorf("error creating GitHub App JWT: %w", err)
+	}
```
evankanderson, line 436: These lines are shared with `getInstallationOwner` on 430 -- it feels like this should be a method.

`getInstallationTokenCredential` has a slightly different path to the same outcome, but might be the right tool for the job as it covers 417-436, with the chance of returning an EmptyCredential.

evankanderson: (This is a small nit, and we can fix it later)

evankanderson (commented):
My only real question / concern is whether we want the logic of deleting dependent repositories (but not other types of things a provider might be responsible for) to live in `controlplane`, or to push that down into the `providers` implementation.

internal/controlplane/handlers_providers.go
eleftherias: I had written it like that initially, but then `providerService` needed to use the `repositoryService` which seemed wrong. We could also call the `store` directly to delete the repositories, but then there will be duplicate logic in the services.

dmjb: I think it's totally reasonable for different services to call eachother - some domain concepts are at a higher level of abstraction than others, and thus delegate to the services for the lower-level domain concepts.

internal/repositories/github/service.go
```diff
@@ -65,6 +65,13 @@ type RepositoryService interface {
 		projectID uuid.UUID,
 		providerName string,
 	) (db.Repository, error)
+	// DeleteRepositoriesByProvider deletes all repositories associated with a provider in a project
+	DeleteRepositoriesByProvider(
```
JAORMX, line 69: I wonder why we need this function at all, shouldn't this just happen via an `ON DELETE CASCADE`?

internal/controlplane/handlers_providers.go
eleftherias: It seems there were some recent changes that also now create an import cycle when I try that:
```
package github.com/stacklok/minder/internal/providers
	imports github.com/stacklok/minder/internal/repositories/github
	imports github.com/stacklok/minder/internal/logger
	imports github.com/stacklok/minder/internal/engine/actions/alert
	imports github.com/stacklok/minder/internal/engine/actions/alert/security_advisory
	imports github.com/stacklok/minder/internal/providers: import cycle not allowed
```

eleftherias: I feel like we can resolve the cycle, but it's not something I want to spend time too much time on now. Can we live with this logic in the controlplane for now and do a followup PR to refactor?

evankanderson: Yes, I'm okay with that.

internal/repositories/github/service.go
evankanderson: Yeah, I had that comment, then I dug in and hit the answer Ria gave.  😁

internal/controlplane/handlers_providers.go
evankanderson: Looking at the import cycle, I'd probably break the `logger` --> `actions/alert` import path, but happy to do that as a follow-up.

evankanderson (approved):

internal/repositories/github/service.go
eleftherias: Updated

dmjb (approved):

PR #2928, opened 2024-04-03
Skip erroring providers when listing repos

eleftherias:
# Summary

Now that we have multiple providers, it's possible one will fail. That should not stop the execution if the other providers succeed. 
This updates the list remote repositories function to error only if all providers error. If one errors, then it logs the error and moves on to the next one.

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_repositories.go
+++ b/internal/controlplane/handlers_repositories.go
@@ -331,6 +331,8 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
 		Results: []*pb.UpstreamRepositoryRef{},
 	}
 
+	var erroringProviders []string
+
 	for _, provider := range provs {
 		zerolog.Ctx(ctx).Trace().
 			Str("provider", provider.Name).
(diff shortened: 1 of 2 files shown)


internal/controlplane/handlers_repositories.go
```diff
@@ -343,17 +345,30 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+			// skip over this provider
+			continue
 		}
 
 		out.Results = append(out.Results, results...)
 	}
 
+	// If all providers failed, return an error
+	if len(erroringProviders) > 0 && len(out.Results) == 0 {
+		return nil, util.UserVisibleError(codes.Internal, "cannot list repositories for providers: %v", erroringProviders)
```
JAORMX, line 369: The logic LGTM, the main caveat I see is that folks will now miss some error messages that were intended to be guidance. e.g. `util.UserVisibleError(codes.PermissionDenied, "cannot get credential for provider: did you run `minder provider enroll`?")` is now lost.

eleftherias: That's true, I can add the "did you run `minder provider enroll`?" back. With the new provider implementation there shouldn't be a scenario where people have a provider without a credential, so it would only happen when a user registered in the past and didn't enroll a provider, and then they return later. I figured that wasn't common enough to keep the message.

JAORMX: Let's keep this for now and fix later. I was thinking already of other providers

JAORMX (approved):

PR #2936, opened 2024-04-04
Support empty profiles by utilizing views

### LINKED ISSUE #2925: Allow creating an empty profile
Currently our validators fail if a profile doesn't contain at least one rule for a supported entity. We should allow creating an empty profile that can be updated later.

jhrozek:
# Summary

Minder UI would benefit if minder backend supported empty profiles to cover cases where
the last rule is removed from a profile, the profile is saved and only then another rule is added.

In contrast to PR #2929, this version creates a LEFT JOIN in a view and
unconditionally adds the profile ID even if the join is empty. This allows the
LEFT JOIN to be then JOIN-ed with the profiles and get a Null EntityProfile
row which is then in turn used to represent profiles with no rules in minder.

Fixes #2925

## Change Type

# Testing

I tested:

- creating an empty profile
- making it non-empty with rules for different entities
- making it empty again
- retrieving an empty profile to make sure the dummy entity_profile record is not displayed
- listing and deleting the empty profile

I did not add many new unit tests or a smoke test yet.

# Review Checklist:


evankanderson (approved):
I think this turned out rather elegantly, nice!

PR #2972, opened 2024-04-05
Add remediation URL to all evaluation responses

puerco:
# Summary

This PR adds the remediation responses to the evaluation statuses in the API. 

It also moves the common metadata parsing functions to the `common.go` file and adds tests 

Fixes https://github.com/stacklok/minder/issues/2879

/cc @rdimitrov 

## Change Type

# Testing

# Review Checklist:


internal/controlplane/common.go
```diff
@@ -27,7 +29,12 @@ import (
 	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 )
 
-const defaultProvider = oauth.Github
+const (
+	defaultProvider = oauth.Github
+	githubURL       = "https://github.com"
+)
+
+var validRepoSlugRe = regexp.MustCompile(`(?i)^[a-z0-9_]+\/[a-z0-9_]+$`)
```
evankanderson, line 37: I think a repo name can also contain hyphens (`-`) and period (`.`).

internal/controlplane/common.go
```diff
@@ -49,3 +56,59 @@ func getNameFilterParam(name string) sql.NullString {
...
+	}
+	prData := &struct {
+		Number int `json:"pr_number"`
+	}{}
+
+	if err := json.Unmarshal(data, prData); err != nil {
+		return "", fmt.Errorf("unmarshalling pull request metadata: %w", err)
+	}
+
+	if repoSlug == "" && prData.Number == 0 {
```
evankanderson, line 76: Do we need the second condition?  Is it ever valid to proceed with `repoSlug == ""`?

internal/controlplane/common.go
```diff
@@ -49,3 +56,59 @@ func getNameFilterParam(name string) sql.NullString {
...
+	}
+
+	// No pull equest found
+	if prData.Number == 0 {
+		return "", nil
+	}
+
+	if !validRepoSlugRe.MatchString(repoSlug) {
+		return "", fmt.Errorf("invalid repository slug")
+	}
```
evankanderson, line 87: Do we need both this check and the check for `""`?

internal/controlplane/common.go
```diff
@@ -49,3 +56,59 @@ func getNameFilterParam(name string) sql.NullString {
...
+
+	return fmt.Sprintf("%s/%s/pull/%d", githubURL, repoSlug, prData.Number), nil
+}
+
+// getAlertURLFromMetadata is a helper function to get the alert URL from the alert metadata
+func getAlertURLFromMetadata(data []byte, repo string) (string, error) {
```
evankanderson, line 93: we call this `repoSlug` above.  Is there a difference?  Does this (or the other) include the owner part?

internal/controlplane/common.go
```diff
@@ -49,3 +56,59 @@ func getNameFilterParam(name string) sql.NullString {
...
+	if repo == "" {
+		return "", fmt.Errorf("cannot form alert URL, no repository defined")
+	}
```
evankanderson, line 96: We do this before `json.Unmarshal` here, but _after_ in the PR case.  Any particular reason for the difference?

internal/controlplane/handlers_evalstatus.go
```diff
@@ -126,7 +126,10 @@ func sortEntitiesEvaluationStatus(
 				profileStatuses[p.Profile.ID] = buildProfileStatus(&p, profileStatusList)
 			}
 
-			stat := buildRuleEvaluationStatusFromDBEvaluation(ctx, &p, e)
+			stat, err := buildRuleEvaluationStatusFromDBEvaluation(ctx, &p, e)
+			if err != nil {
+				return nil, nil, nil, fmt.Errorf("building rule evaluation status: %w", err)
+			}
```
evankanderson, line 132: Do we want to have an error generating a single remediation URL cause the entire status list to fail?

I suspect we want to log and swallow the error in `buildRuleEvaluationStatusFromDBEvaluation`, so that the listing degrades gracefully in the face of malformed data / errors.

internal/controlplane/handlers_profile.go
```diff
@@ -422,6 +422,16 @@ func getRuleEvalStatus(
...
 	}
+	remediationURL := ""
+	if dbRuleEvalStat.Entity == db.EntitiesRepository {
+		remediationURL, err = getRemediationURLFromMetadata(
+			dbRuleEvalStat.RemMetadata.RawMessage,
+			fmt.Sprintf("%s/%s", dbRuleEvalStat.RepoOwner, dbRuleEvalStat.RepoName),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("parsing remediation pull request data: %w", err)
+		}
```
evankanderson, line 433: Again, I suspect we'd prefer an empty URL and a log message to an entirely failed list, so I'd log the error rather than return it here.

internal/controlplane/common.go
puerco: Nice, thanks for catching this

internal/controlplane/common.go
puerco: This was actually a bug here and in the test. I've modified both URL functions to fail if no slug is set.

internal/controlplane/common.go
puerco: I was trying to allow for an empty slig, but I've modified the functions to require a valid repo.

internal/controlplane/common.go
puerco: I've evened the param names. Now both are `repoSlug`.

internal/controlplane/common.go
puerco: I changed both functions to follow the same algorithm. Now they will fail on an invalid repo slug early.

internal/controlplane/handlers_profile.go
puerco: So the idea behind this was to avoid sending an incomplete response, in my mind it was _inconsistent_ instead of _degraded_ :)   but ok, I've changed them to log in case of an error.

PR #2973, opened 2024-04-05
Create Getting Started > Overview page

meganbruce:
Adds a new page to Minder documentation called Overview, in the Getting Started section, where we list the different options for using Minder.


meganbruce:
This is part of docs cleanup for Monday docs days.

PR #2976, opened 2024-04-06
A nicer error message when listing remote repos without an enrolled provider

jhrozek:
# Summary

Our error message when trying to enroll a repository without a provider was less than ideal.

Before:
```
Message: Error getting list of remote repos
Details: NotFound means some requested entity (e.g., file or directory) was
not found.
```

After:
```
Message: Error getting list of remote repos
Details: no suitable provider found, please enroll a provider
```

## Change Type

# Testing

minder repo register without a provider

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_repositories.go
+++ b/internal/controlplane/handlers_repositories.go
@@ -73,6 +73,10 @@ func (s *Server) RegisterRepository(
 
 	provider, err := s.inferProviderByOwner(ctx, githubRepo.GetOwner(), projectID, providerName)
 	if err != nil {
+		pErr := providers.ErrProviderNotFoundBy{}
+		if errors.As(err, &pErr) {
+			return nil, util.UserVisibleError(codes.NotFound, "no suitable provider found, please enroll a provider")
+		}
 		return nil, status.Errorf(codes.Internal, "cannot get provider: %v", err)
 	}
 
@@ -323,6 +327,10 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
 	providerName := in.GetContext().GetProvider()
 	provs, err := s.providerStore.GetByNameAndTrait(ctx, projectID, providerName, db.ProviderTypeRepoLister)
 	if err != nil {
+		pErr := providers.ErrProviderNotFoundBy{}
+		if errors.As(err, &pErr) {
+			return nil, util.UserVisibleError(codes.NotFound, "no suitable provider found, please enroll a provider")
+		}
 		return nil, providerError(err)
 	}
 
--- a/internal/providers/store.go
+++ b/internal/providers/store.go
@@ -56,6 +56,24 @@ type ProviderStore interface {
 	) ([]db.Provider, error)
 }
 
+// ErrProviderNotFoundBy is an error type which is returned when a provider is not found
+type ErrProviderNotFoundBy struct {
+	Name  string
+	Trait string
+}
+
+func (e ErrProviderNotFoundBy) Error() string {
+	msgs := []string{}
+	if e.Name != "" {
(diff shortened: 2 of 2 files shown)


jhrozek:
@dmjb hey, you have an eye for code cleanliness. I don't love that we are converting one UserVisibleError to another. Do you think it would be better to instead return a specific error value from `filteredResultNotFoundError` instead of a `UserVisibleError`? (More generally I wonder if we want to keep `UserVisibleError` only in handlers)

JAORMX (dismissed):

internal/controlplane/handlers_repositories.go
```diff
@@ -324,6 +324,10 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
 	providerName := in.GetContext().GetProvider()
 	provs, err := s.providerStore.GetByNameAndTrait(ctx, projectID, providerName, db.ProviderTypeRepoLister)
 	if err != nil {
+		var userErr *util.NiceStatus
+		if errors.As(err, &userErr) {
+			return nil, util.UserVisibleError(codes.NotFound, "no suitable provider found, please enroll a provider")
```
JAORMX, line 329: If this is already a nice status, why not return that?

jhrozek: Because we call:
 - ListRemoteRepositoriesFromProvider
 - GetByNameAndTrait
 - findProvider
 - filteredResultNotFoundError which returns a NiceStatus saying `provider not found with filters`. On the one hand this is a NiceStatus so it is suitable for rendering by clients, on the other hand it doesn't sound like an error that's very actionable by the user.

I wonder if we should rather not return a NiceStatus from a utility function or convert the NiceStatus from filteredResultNotFoundError to a bespoke `error` that would be then converted to a NiceStatus in the handlers.

JAORMX: I think we shouldn't. We should return relevant internal errors and only do gPRC status errors on the handler functions.

dmjb:
> (More generally I wonder if we want to keep `UserVisibleError` only in handlers)

I agree with this. The fact that the ProviderStore returns a user visible error is because I was minimizing the amount of changes which needed to be made to that code (since it was largely copied from elsewhere in the code). I feel like we should return custom errors, and then map them to user visible errors in the controlplane.

dmjb (approved):
Approved. Whether or not you want to refactor the provider store to return a custom error in this PR is up to you.

rdimitrov (dismissed):

jhrozek:
> Approved. Whether or not you want to refactor the provider store to return a custom error in this PR is up to you.

yeah, let's do this right, we've been piling too much debt lately it feels and my to-clean-up list is not getting shorter.

I'll submit a new version

jhrozek:
> > Approved. Whether or not you want to refactor the provider store to return a custom error in this PR is up to you.
> 
> yeah, let's do this right, we've been piling too much debt lately it feels and my to-clean-up list is not getting shorter.
> 
> I'll submit a new version

done. WDYT @dmjb ?

internal/controlplane/handlers_repositories.go
```diff
@@ -324,6 +327,9 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
 	providerName := in.GetContext().GetProvider()
 	provs, err := s.providerStore.GetByNameAndTrait(ctx, projectID, providerName, db.ProviderTypeRepoLister)
 	if err != nil {
+		if _, ok := err.(providers.ErrProviderNotFoundBy); ok {
```
JAORMX, line 330: I think you wanna use `errors.As`

jhrozek: good call, thanks

JAORMX (approved):

Issue #2971, opened 2024-04-05
Support full profile updates with PATCH


jhrozek:
At the moment PATCH only supports toggling remediate and alerts on profiles. We should support full profile updates.

jhrozek:
@kd @davolokh 👀

PR #2990, opened 2024-04-08
Allow full profile updates through the PATCH handler

jhrozek:
# Summary

We used to use a bespoke PATCH handler to update profiles but the PATCH
method only allowed changing alert and remediate. We can make the code
simpler and allow more updates by modifying the existing profile using
the UpdateMask and then just relaying the update to the existing
UpdateProfile method of the Profile service.

Fixes: #2971

## Change Type

# Testing

There are unit tests but you can also test with curl:

Adding a rule to a profile:
```
curl -XPATCH -H "Authorization: Bearer $MINDER_BEARER_TOKEN"
'http://localhost:8080/api/v1/profile/ce746798-c30c-46f3-8a12-0654747b49ec?context.project=1900c9f6-ef0f-43e4-8af7-108beb72de0e&context.provider=github'
-d '{"name":"dependabot-enabled-profile", "labels":[], "buildEnvironment":[], "repository":[{"type":"dependabot_configured","params":null,"def":{"apply_if_file":"requirements.txt","package_ecosystem":"pypi","schedule_interval":"weekly"},"name":"dependabot_configured"},{"type":"dependabot_configured","params":null,"def":{"apply_if_file":"go.mod","package_ecosystem":"gomod","schedule_interval":"weekly"},"name":"dependabot_configured_go"}], "pullRequest":[], "remediate ":"on", "alert":"off", "type":"", "version":"", "displayName":"gah" }
```

Remove the rule back:
```
curl -XPATCH -H "Authorization: Bearer $MINDER_BEARER_TOKEN"
'http://localhost:8080/api/v1/profile/ce746798-c30c-46f3-8a12-0654747b49ec?context.project=1900c9f6-ef0f-43e4-8af7-108beb72de0e&context.provider=github'
-d '{"name":"dependabot-enabled-profile", "labels":[], "buildEnvironment":[], "repository":[{"type":"dependabot_configured","params":null,"def":{"apply_if_file":"requirements.txt","package_ecosystem":"pypi","schedule_interval":"weekly"},"name":"dependabot_configured"}], "pullRequest":[], "remediate":"on", "alert":"off", "type":"", "version":"", "displayName":"gah"}}'
```

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_profile.go
+++ b/internal/controlplane/handlers_profile.go
@@ -31,10 +31,8 @@ import (
 	"github.com/stacklok/minder/internal/db"
 	"github.com/stacklok/minder/internal/engine"
 	"github.com/stacklok/minder/internal/engine/entities"
-	"github.com/stacklok/minder/internal/events"
 	"github.com/stacklok/minder/internal/logger"
 	prof "github.com/stacklok/minder/internal/profiles"
-	"github.com/stacklok/minder/internal/reconcilers"
 	"github.com/stacklok/minder/internal/util"
 	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 )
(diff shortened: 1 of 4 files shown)


internal/controlplane/handlers_profile.go
```diff
@@ -547,71 +549,63 @@ func (s *Server) PatchProfile(ctx context.Context, ppr *minderv1.PatchProfileReq
...
-	})
+	oldProfilePb, err := getProfilePBFromDB(ctx, profileID, entityCtx, s.store)
 	if err != nil {
-		if errors.Is(err, sql.ErrNoRows) {
-			return nil, util.UserVisibleError(codes.NotFound, "profile not found")
-		}
 		return nil, status.Errorf(codes.Internal, "failed to get profile: %s", err)
 	}
 
-	// TEMPORARY HACK: Since we do not need to support this patch logic in the
```
dmjb, line 563: We still need this check.

jhrozek: I can add a test but wouldn't the update code just error out later?

internal/controlplane/handlers_profile.go
```diff
@@ -547,71 +549,63 @@ func (s *Server) PatchProfile(ctx context.Context, ppr *minderv1.PatchProfileReq
...
+	if updateMask == nil {
+		return
 	}
 
-	// Update top-level profile db object
-	_, err = s.store.UpdateProfile(ctx, params)
-	if err != nil {
-		return nil, util.UserVisibleError(codes.Internal, "error updating profile: %v", err)
-	}
+	for _, attr := range updateMask.Paths {
```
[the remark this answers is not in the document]
dmjb: Subscription ID is one which we do not want to allow to be changed.

dmjb (changes requested):
Leaving this as "Request changes" until the question around editing bundle profiles is settled.

jhrozek:
> After talking to @jhrozek offline, I understand why that check can be deleted. I would like to see a test added for the bundled subscription ID case though.

I added a test and moved a bunch of the patch handler to profile service.

btw this PR is NOT important to merge at this moment

internal/profiles/service.go
```diff
@@ -276,6 +291,63 @@ func (p *profileService) UpdateProfile(
 	return profile, nil
 }
 
+func (p *profileService) PatchProfile(
+	ctx context.Context,
+	projectID uuid.UUID,
+	profileID uuid.UUID,
+	subscriptionID uuid.UUID,
```
evankanderson, line 298: Do we want to allow `PATCH` with subscriptions?  If not, could we drop this argument?

evankanderson: (It looks like we pass `uuid.Nil` as the argument to `UpdateProfile` from the other gRPC endpoint, so it looks like we the PATCH-on-Subscription codepath is effectively dead...)

internal/profiles/service.go
```diff
@@ -276,6 +291,63 @@ func (p *profileService) UpdateProfile(
...
+
+func patchProfilePb(oldProfilePb, patchPb *minderv1.Profile, updateMask *fieldmaskpb.FieldMask) {
+	// if there is no update mask, return the old profile. The grpc-rest gateway always sets the update mask
+	if updateMask == nil {
+		return
+	}
+
+	for _, attr := range updateMask.Paths {
+		oldReflect := oldProfilePb.ProtoReflect()
+		patchReflect := patchPb.ProtoReflect()
```
evankanderson, line 321: Do you want to do these inside the `for` loop?

internal/profiles/service.go
```diff
@@ -276,6 +291,63 @@ func (p *profileService) UpdateProfile(
...
+func (p *profileService) PatchProfile(
+	ctx context.Context,
+	projectID uuid.UUID,
+	profileID uuid.UUID,
+	subscriptionID uuid.UUID,
+	patch *minderv1.Profile,
+	updateMask *fieldmaskpb.FieldMask,
+	qtx db.Querier,
+) (*minderv1.Profile, error) {
+	oldProfilePb, err := getProfilePBFromDB(ctx, profileID, projectID, qtx)
```
evankanderson, line 303: Do you want to call `old` "base" here?  It's slightly confusing, because we go through the steps and then pass `old` to the Update method.

internal/profiles/service.go
```diff
@@ -276,6 +291,63 @@ func (p *profileService) UpdateProfile(
...
+		return nil, fmt.Errorf("failed to get profile: %w", err)
+	}
+
+	patchProfilePb(oldProfilePb, patch, updateMask)
+
+	return p.UpdateProfile(ctx, projectID, subscriptionID, oldProfilePb, qtx)
+}
+
+func patchProfilePb(oldProfilePb, patchPb *minderv1.Profile, updateMask *fieldmaskpb.FieldMask) {
+	// if there is no update mask, return the old profile. The grpc-rest gateway always sets the update mask
```
evankanderson, line 314: This is confusing, as it says "return", but there's no return value.  Maybe you mean "an empty field mask means apply no changes"?

evankanderson (approved):
Only nits; I _like_ the new code much better!  Thanks!

internal/profiles/service.go
jhrozek: probably not, you're right. We'll likely to only update subscription profiles through a replace, not patch.

internal/profiles/service.go
jhrozek: thanks, fixed

internal/profiles/service.go
jhrozek: thanks, renamed

jhrozek:
@evankanderson thanks for the review, I fixed your comments in separate patches for easier review.

PR #3032, opened 2024-04-10
Default to github-app in provider enroll

### LINKED ISSUE #2898: Make GitHub App the default provider in the CLI
When running `minder provider enroll`, the default provider should be the new `github-app` provider, instead of the legacy `github` provider.
We should change this once all the work for https://github.com/stacklok/epics/issues/154 is done.

We should also remove the messaging "You are about to enroll repositories from your personal account." and the confirmation request, since the installation owner is determined by the user in the browser, and not by the CLI flags.

dmjb:
Fixes #2898

Some additional changes have been made as well:

1) Make the flag to specify the class of provider specific to the enroll
   command. The existing top-level provider flag is left intact, but
   marked as deprecated.
2) The confirmation box is only shown for the legacy Github app workflow
   as requested in the ticket.
3) The timeout duration is listed in the CLI output when the browser is
   opened.
4) Small changes to message verbiage here and there.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/cmd/cli/app/provider/provider.go
+++ b/cmd/cli/app/provider/provider.go
@@ -20,7 +20,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/stacklok/minder/cmd/cli/app"
-	ghclient "github.com/stacklok/minder/internal/providers/github/oauth"
 	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 )
 
@@ -37,9 +36,9 @@ var ProviderCmd = &cobra.Command{
 func init() {
 	app.RootCmd.AddCommand(ProviderCmd)
 	// Flags for all subcommands
-	// TODO: remove the provider flag from here and add it only to the subcommands that need it
-	ProviderCmd.PersistentFlags().StringP("provider", "p", ghclient.Github, "Name of the provider, i.e. github")
 	ProviderCmd.PersistentFlags().StringP("project", "j", "", "ID of the project")
+	// TODO: get rid of this
+	ProviderCmd.PersistentFlags().StringP("provider", "p", "", "DEPRECATED - use `class` flag of `enroll` instead")
 }
 
 func getImplementsAsStrings(p *minderv1.Provider) []string {
--- a/cmd/cli/app/provider/provider_enroll.go
+++ b/cmd/cli/app/provider/provider_enroll.go
@@ -45,7 +45,7 @@ const (
 	legacyGitHubProvider = minderv1.ProviderClass_PROVIDER_CLASS_GITHUB
 
 	// githubAppProvider is the name used to identify the new GitHub App provider class
-	githubAppProivder = minderv1.ProviderClass_PROVIDER_CLASS_GITHUB_APP
+	githubAppProvider = minderv1.ProviderClass_PROVIDER_CLASS_GITHUB_APP
 )
 
 var enrollCmd = &cobra.Command{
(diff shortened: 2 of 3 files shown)


cmd/cli/app/provider/provider.go
```diff
@@ -37,9 +36,9 @@ var ProviderCmd = &cobra.Command{
 func init() {
 	app.RootCmd.AddCommand(ProviderCmd)
 	// Flags for all subcommands
-	// TODO: remove the provider flag from here and add it only to the subcommands that need it
-	ProviderCmd.PersistentFlags().StringP("provider", "p", ghclient.Github, "Name of the provider, i.e. github")
 	ProviderCmd.PersistentFlags().StringP("project", "j", "", "ID of the project")
+	// TODO: get rid of this
+	ProviderCmd.PersistentFlags().StringP("provider", "p", "", "DEPRECATED - use `class` flag of `enroll` instead")
```
eleftherias, line 41: Since this was only useful for the `provider enroll` command, should it be moved there? Otherwise other commands like `provider list` see the deprecation message

cmd/cli/app/provider/provider_enroll.go
```diff
@@ -275,8 +283,10 @@ func init() {
 	ProviderCmd.AddCommand(enrollCmd)
 	// Flags
 	enrollCmd.Flags().StringP("token", "t", "", "Personal Access Token (PAT) to use for enrollment")
```
eleftherias, line 285: The token is also for legacy GitHub only

internal/providers/github/oauth/oauth.go
```diff
@@ -34,8 +34,12 @@ import (
 	provifv1 "github.com/stacklok/minder/pkg/providers/v1"
 )
 
-// Github is the string that represents the GitHubOAuth provider
-const Github = "github"
+const (
```
eleftherias, line 37: We do have a `ProviderClass` enum in `minder.proto`. Could we reuse that, instead of re-defining the options here?

cmd/cli/app/provider/provider.go
dmjb: I left it that way for the sake of backwards compatibility. I also think that we should consider changing the name of the parameter: in some commands elsewhere in the CLI, `provider` might mean the name of the provider, here it means the class/type

cmd/cli/app/provider/provider_enroll.go
JAORMX: for now. Other providers (e.g. DockerHub) will make use of this.

dmjb: The problem is that with the current approach, this command will end up with every possible optional flag for every possible provider. In the Minder CLI proposal document, I have suggested that we move to a structure with subcommands per provider, e.g.:

```
minder provider enroll github-app [flags]
minder provider enroll github [flags]
minder provider enroll oci [flags]
```

However, since we need to maintain backwards compatibility, I am not making this change at this point.

Ria's feedback does raise the issue of whether we should avoid calling certain parts of the code if we are creating a `github-app` provider

eleftherias: Eventually, it would be nice if we could use the `AuthorizationFlows` for each provider to create a selection box for the user. For example the legacy `github` provider support the OAuth flow and the user input flow, so when they run `minder provider enroll --class github` we could show:
```
Select your authorization flow:
[] provider the token myself
[] use OAuth 2.0 authorization code flow to generate a token 
```

Then we remove `--token` from the parameters of `enroll` and new providers just have to define their accepted authorization flows.

cmd/cli/app/provider/provider_enroll.go
```diff
@@ -43,6 +43,9 @@ const (
 
 	// legacyGitHubProvider is the legacy GitHub OAuth provider class
 	legacyGitHubProvider = minderv1.ProviderClass_PROVIDER_CLASS_GITHUB
+
+	// githubAppProvider is the name used to identify the new GitHub App provider class
+	githubAppProivder = minderv1.ProviderClass_PROVIDER_CLASS_GITHUB_APP
```
eleftherias, line 48: typo

eleftherias (approved):

PR #3034, opened 2024-04-10
Don't crash if there's no package list client

jhrozek:
# Summary

If a github App provider is enrolled but no fallback token configured through `GitHubAppConfig.FallbackClient` then the github config structure's `packageListingClient` will be `nil`. While this only affects the initial reconcile of the artifacts on repo registration and not retrieving the artifacts through webhook notifications, minder shouldn't crash.

The patch also checks the result for null because on error the result might be null but the err handler was dereferencing result unconditionally.

## Change Type

# Testing

enroll with a github app without the fallback client

# Review Checklist:

### DIFF
--- a/internal/providers/github/common.go
+++ b/internal/providers/github/common.go
@@ -59,6 +59,8 @@ var (
 	ErrNotFound = errors.New("not found")
 	// ErrBranchNotFound denotes if the branch was not found
 	ErrBranchNotFound = errors.New("branch not found")
+	// ErrNoPackageListingClient denotes if there is no package listing client available
+	ErrNoPackageListingClient = errors.New("no package listing client available")
 )
 
 // GitHub is the struct that contains the shared GitHub client operations
@@ -171,6 +173,11 @@ func (c *GitHub) ListPackagesByRepository(
 	// create a slice to hold the containers
 	var allContainers []*github.Package
 
+	if c.packageListingClient == nil {
+		zerolog.Ctx(ctx).Error().Msg("No client available for listing packages")
+		return allContainers, ErrNoPackageListingClient
+	}
+
 	type listPackagesRespWrapper struct {
 		artifacts []*github.Package
 		resp      *github.Response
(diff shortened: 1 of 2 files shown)


internal/providers/github/common.go
```diff
@@ -176,6 +181,10 @@ func (c *GitHub) ListPackagesByRepository(
 		var resp *github.Response
 		var err error
 
+		if c.packageListingClient == nil {
```
puerco, line 184: I think the condition you're introducing on line 170 makes this code unreachable.

internal/providers/github/common.go
```diff
@@ -167,6 +167,11 @@ func (c *GitHub) ListPackagesByRepository(
 	// create a slice to hold the containers
 	var allContainers []*github.Package
 
+	if c.packageListingClient == nil {
+		zerolog.Ctx(ctx).Error().Msg("No client available for listing packages")
+		return allContainers, nil
```
puerco, line 172: Should this return an error instead?

internal/providers/github/common.go
```diff
@@ -167,6 +167,11 @@ func (c *GitHub) ListPackagesByRepository(
 	// create a slice to hold the containers
 	var allContainers []*github.Package
 
+	if c.packageListingClient == nil {
```
JAORMX, line 170: @jhrozek for my education, in what case would there be no package listing client?

internal/providers/github/common.go
jhrozek: thanks, I started fixing the code where I saw the crash happening and then didn't remove this condition.

internal/providers/github/common.go
jhrozek: good point, I changed the code to return an error code from the GH function, but then catch and ignore the error with a log message on the only place where it's used.

internal/providers/github/common.go
jhrozek: > @jhrozek for my education, in what case would there be no package listing client?

I added an explanation to the PR description.

eleftherias (approved):

JAORMX (approved):

Issue #2948, opened 2024-04-04
Handle user deletion gracefully when the user created the organization


ethomson:
### Describe the issue

If you create an organization (top-level project), then it will be created with your GitHub username. You can then invite somebody to that organization, then delete your user account (minder auth delete). Your project will continue to exist, still named as your GitHub username. You now cannot even log in again, since doing so would create a new organization with your GitHub username, but one already exists.

### To Reproduce

1. Create a new user account (and thus a new project).
   `minder auth login`
   At this point, you will have a new Minder organization (top-level project) that is named after your GitHub username.
2. Invite someone else to your project
   `minder project role grant ...`
3. Delete your account
   `minder auth delete`

Now you can no longer create a new account (`minder auth login`). Creating your new account would try to create a new project with your GitHub username as the org name. However, that already exists (you orphaned it in step 2). You can no longer use Minder.

evankanderson:
There is a workaround with the GitHub App flow if you start from https://github.com/apps/minder-by-stacklok (or whatever your app name is) to register the organizations in Minder before you create your account.  At that point, CreateUser will create projects for each pending App Install rather than the project named based on the user.

evankanderson:
I'm also going to check in a small fix to remove the project when the last `admin` member is removed, rather than the last user, since a `viewer` or `editor` on the project can't recover full use of the project once there are no admins.

evankanderson:
Looking at the code further, it seems like we actually remove the auth before the projects, which could lead to a second problem if the database transaction fails.  We probably need to rethink the flow here to be less all-at-once and more sequential.

PR #3082, opened 2024-04-14
Do not delete user from OpenFGA in a DB transaction to avoid issues with minder auth delete

jhrozek:
# Summary

NOTE: This patch touches a bit delicate part of minder. Please review carefully - I was not super acquainted with the code earlier and I wouldn't be suprised if I missed something.

We've been occasionally hitting issues where `minder auth delete` would
delete the user entry but leave the project around. Recovering from this
state then requires database write permissions to manually clean up the
projects.

At least in the case that I reproduced by accident the flow went like
this:
- user issues auth delete
- the transaction starts
- the user projects are returned and contain a non-empty list
- the user entry is deleted from OpenFGA
- the user entry is staged for deletion from the DB as part of the
  transaction
- one of the projects are attempted to be deleted, but the deletion
  fails. This error rolls back the transaction and deleting the user.
- the end-user gets an error message, thinks 'a fluke' and runs auth
  delete again
- the new run starts, calls list project for user from OpenFGA which
  returns an empty list
- the user is deleted from the db and the transaction is commited
- the project are left behind because openFGA returned none for the user
  because as far as openFGA knows the user is gone
- the next auth login can't create the project for the user becasue it already exists

This patch reorganizes the `DeleteUser` flow is such a way that if the
all the database operations are wrapped in a transaction, but the
OpenFGA delete user call is kept outside; the list of projects before the
transaction and deleting the user entry after it.

The call to delete unmanaged projects is changed so that only the
projects that have other subjects than the one being deleted are
deleted.

## Change Type

# Testing

I reverted the patch that prevents deleting a project in case it still contains
artifacts as a way to easily break project deletion.

Then without the patch, deleting a project caused the projects to be essentially orphaned
and subsequent auth login calls were all failing.

# Review Checklist:

### DIFF
--- a/internal/controlplane/identity_events.go
+++ b/internal/controlplane/identity_events.go
@@ -109,36 +109,17 @@ func HandleEvents(
 }
 
 // DeleteUser deletes a user and all their associated data from the minder database
-func DeleteUser(ctx context.Context, store db.Store, authzClient authz.Client, providerService service.GitHubProviderService,
-	userId string) (retErr error) {
+func DeleteUser(
+	ctx context.Context,
+	store db.Store,
+	authzClient authz.Client,
+	providerService service.GitHubProviderService,
+	userId string,
+) error {
 	l := zerolog.Ctx(ctx).With().
 		Str("operation", "delete").
 		Str("subject", userId).
 		Logger()
-
-	tx, err := store.BeginTransaction()
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if retErr != nil {
-			if err := store.Rollback(tx); err != nil && !errors.Is(err, sql.ErrTxDone) {
-				l.Debug().Msgf("error rolling back transaction: %v", err)
-			}
-		}
-	}()
-	qtx := store.GetQuerierWithTransaction(tx)
-
-	var userDBID *int32
-
-	usr, err := qtx.GetUserBySubject(ctx, userId)
-	// If the user doesn't exist, we still want to clean up any associated data
-	if err != nil && !errors.Is(err, sql.ErrNoRows) {
-		return fmt.Errorf("error retrieving user %v", err)
-	} else if err == nil {
-		userDBID = &usr.ID
-		l = l.With().Int32("user_id", usr.ID).Logger()
-	}
 
 	// Fetching the projects for user before the deletion was made.
 	// This allows us to clean up the project from the database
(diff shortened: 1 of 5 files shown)


JAORMX (commented):
lgtm, did you manage to test this locally with something other than unit tests?

eleftherias:
@jhrozek Do you know why this situation happened:
> one of the projects are attempted to be deleted, but the deletion fails. 

Was it a fluke or something intentional that kept the project from being deleted?

jhrozek:
> @jhrozek Do you know why this situation happened:
> 
> > one of the projects are attempted to be deleted, but the deletion fails.
> 
> Was it a fluke or something intentional that kept the project from being deleted?

It was before f4807f133ce9db30f2cc084f0e494d588796ef22 was fixed, so I couldn't delete the project because it had artifacts attached to one of the repos.

jhrozek:
> lgtm, did you manage to test this locally with something other than unit tests?

I reverted the fix for f4807f133ce9db30f2cc084f0e494d588796ef22 and tried to delete the project while it had artifacts attached to the repos. I got consistent error that I can't remove the project, but I didn't end up in this half-state of no FGA assignments and no user entry, but projects left behind.

jhrozek:
> > @jhrozek Do you know why this situation happened:
> > > one of the projects are attempted to be deleted, but the deletion fails.
> > 
> > 
> > Was it a fluke or something intentional that kept the project from being deleted?
> 
> It was before [f4807f1](https://github.com/stacklok/minder/commit/f4807f133ce9db30f2cc084f0e494d588796ef22) was fixed, so I couldn't delete the project because it had artifacts attached to one of the repos.

but your question was valid - if we think that this change is not 100% or too risky, let's not merge it this week.

JAORMX (commented):
I still think this is an improvement and it makes sense. I think @rdimitrov said he'd test it, so I'll wait for that.

rdimitrov (dismissed):
tldr: I haven't found any issues 👍 

I've tested it eventually by using the approach mentioned by @jhrozek by reverting the artifacts fix. I've created sub-projects, confirmed that without this fix we end up in a weird state, then enabled the fix and did the same test where I had my main project and the sub-projects and tried deleting my account. Deletion failed because of the reverted artifact fix but it did not broke anything for me.

JAORMX (dismissed):
shipit

internal/projects/delete.go
```diff
@@ -42,15 +48,23 @@ func CleanUpUnmanagedProjects(
 		return fmt.Errorf("error getting role assignments for project %v", err)
 	}
 
-	if len(as) == 0 {
+	if !hasOtherRoleAssignmens(as, subject) {
```
evankanderson, line 51:
```suggestion
	if !hasOtherRoleAssignments(as, subject) {
```

internal/projects/delete.go
```diff
@@ -42,15 +48,23 @@ func CleanUpUnmanagedProjects(
...
 		if err := DeleteProject(ctx, proj, querier, authzClient, providerService, l); err != nil {
 			return fmt.Errorf("error deleting project %v", err)
 		}
+	} else {
+		l.Debug().Msg("project has other role assignments")
 	}
 	return nil
 }
 
+func hasOtherRoleAssignmens(as []*v1.RoleAssignment, subject string) bool {
```
evankanderson, line 62:
```suggestion
func hasOtherRoleAssignments(as []*v1.RoleAssignment, subject string) bool {
```

internal/controlplane/identity_events.go
```diff
@@ -109,71 +109,66 @@ func HandleEvents(
...
-	if err != nil && !errors.Is(err, sql.ErrNoRows) {
-		return fmt.Errorf("error retrieving user %v", err)
-	} else if err == nil {
-		userDBID = &usr.ID
-		l = l.With().Int32("user_id", usr.ID).Logger()
-	}
-
 	// Fetching the projects for user before the deletion was made.
 	// This allows us to clean up the project from the database
 	// if there are no more role assignments for the project.
```
evankanderson, line 126: I'm not sure this comment still makes sense -- we should probably update it.

Given that project IDs are UUIDs, it _seems_ reasonable to assert that we clean up the database first because any "dangling" user-to-project assignments in OpenFGA will be effectively harmless, because we should never end up reusing a project ID.  It seems worthwhile including a comment to that effect.

internal/controlplane/identity_events.go
```diff
@@ -109,71 +109,66 @@ func HandleEvents(
...
-		l.Debug().Msg("deleting user from database")
-		if err := qtx.DeleteUser(ctx, *userDBID); err != nil {
-			return fmt.Errorf("error deleting user %v", err)
+		// We only delete the user if it still exists in the database
+		if userDBID != nil {
+			l.Debug().Msg("deleting user from database")
+			if err := qtx.DeleteUser(ctx, *userDBID); err != nil {
+				return db.User{}, fmt.Errorf("error deleting user %v", err)
+			}
 		}
```
evankanderson, line 151: This is subtle -- we'll end up with userDBID being `nil` if neither the `if` nor the the `else if` match.

Can we use `if usr.IdentitySubject != ""` or `if !errors.Is(err, sql.ErrNoRows)` here?

```suggestion
		// We only delete the user if it still exists in the database
		if usr.IdentitySubject != "" {
			l = l.With().Int32("user_id", usr.ID).Logger()
			l.Debug().Msg("deleting user from database")
			if err := qtx.DeleteUser(ctx, usr.ID); err != nil {
				return db.User{}, fmt.Errorf("error deleting user %v", err)
			}
		}
```

internal/controlplane/identity_events.go
```diff
@@ -109,71 +109,66 @@ func HandleEvents(
...
-	for _, proj := range projs {
-		l.Debug().Str("project_id", proj.String()).Msg("cleaning up project")
-		if err := projects.CleanUpUnmanagedProjects(ctx, proj, qtx, authzClient, providerService, l); err != nil {
-			return fmt.Errorf("error deleting project %v", err)
+		for _, proj := range projs {
+			l.Debug().Str("project_id", proj.String()).Msg("cleaning up project")
+			if err := projects.CleanUpUnmanagedProjects(ctx, userId, proj, qtx, authzClient, providerService, l); err != nil {
+				return db.User{}, fmt.Errorf("error deleting project %v", err)
+			}
 		}
```
evankanderson, line 158: Q: Do we want to return an error here after the first Unmanaged project fails?  Would we want to delete the projects before the user?

A: It turns out the order doesn't matter, because it's happening in a transaction.  It would still feel slightly better to me to delete the user after the projects, but the real effect is minimal.

evankanderson (commented):
I have #3098 as well, but this looks generally correct, so let's merge it and then I'll merge #3098 after that.

evankanderson (dismissed):

evankanderson:
This fixes the other half of #2948 that I was worried about, more elegantly than I was imagining, because I still forget about database transactions...

evankanderson (approved):

PR #3083, opened 2024-04-15
Frizbee: Cache GitHub action entries

JAORMX:
# Summary

This uses frizbee's builtin cache to store the found github actions
in-between file modifications.

We always had cache support but simply were not using it. This should
reduce the number of calls we do to GitHub.

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/engine/actions/remediate/pull_request/types_actions_replace_tags.go
+++ b/internal/engine/actions/remediate/pull_request/types_actions_replace_tags.go
@@ -67,9 +67,10 @@ func newFrizbeeTagResolveModification(
 
 func (ftr *frizbeeTagResolveModification) createFsModEntries(ctx context.Context, _ interfaces.ActionsParams) error {
 	entries := []*fsEntry{}
+	cache := utils.NewRefCacher()
 
 	err := ghactions.TraverseGitHubActionWorkflows(ftr.fs, ".github/workflows", func(path string, wflow *yaml.Node) error {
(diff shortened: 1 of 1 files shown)


rdimitrov (approved):
Just so I understand the changes - when we start parsing the project tree we'll cache all action entries we find so no repetitive requests will be made for the same github action during the whole evaluation?

eleftherias (approved):
It looks like this cache gets overwritten on every remediation, so it's likely to be empty most of the time. Would it be more beneficial to create the cache further up the call chain so that it gets reused over multiple runs?

JAORMX:
> Just so I understand the changes - when we start parsing the project tree we'll cache all action entries we find so no repetitive requests will be made for the same github action during the whole evaluation?

So, before this change, we'd iterate each file and replace al GH actions. There is an inner cache in frizbee, but it gets replaced on each call to `ModifyReferencesInYAML`. Meaning, it only applied to one file.

After this PR, the cache is shared between files, so a single remediation round will share the cache. We could optimize further sharing the cache the whole policy run. But that's for another PR.

JAORMX:
> It looks like this cache gets overwritten on every remediation, so it's likely to be empty most of the time. Would it be more beneficial to create the cache further up the call chain so that it gets reused over multiple runs?

yes, I just didn't want to introduce that big changes to the remediator just yet. If we think it's worth it, I can revisit this in a further PR. For now, this is a local repo improvement. Though, this remediation would most likely only be called once per policy execution.

PR #3088, opened 2024-04-15
Webhook handler chain

dmjb:
# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


eleftherias:
I don't think the chain of responsibility pattern makes sense in this case. My reasoning is:
1) Our handlers have arbitrary ordering. Usually with this type of pattern, the handlers are in a priority order, so that the highest priority handler gets executed first. In our case, we don't care which handler goes first.
2) The shape of the incoming request is known beforehand, so we could go directly to the right handler. With this pattern we unnecessarily iterate through all the possible handlers.

With this approach debugging would also be harder, because we don't know which handler the request was intended for. For example, if we get the secret wrong for the GitHub event handler, we'll have a hard time tracing that it's the GitHub handler that's erroring. We'll simply see that none of the handlers could handle the request, but we won't know where the problem lies within the chain.

dmjb:
Closing since I have heard a number of arguments for using a different approach.

Issue #3071, opened 2024-04-12
Remove Provider name from EntityInfoWrapper


dmjb:
### Please describe the enhancement

Since we are moving away from provider name in a number of our DB entities, change the event system to use provider IDs instead of identifying them by a mix of name and project ID.

### Solution Proposal

1) Add the ProviderID field to the message format
2) Modify the consumer to use provider ID if present else, use the name
3) In a separate PR: populate the field on the producer
4) In a separate PR: get rid of the name field

PR #3093, opened 2024-04-15
Use provider ID instead of name when sending events

dmjb:
Relates to #3071

This PR changes any code which creates an EntityInfoWrapper to use Provider ID instead of name. The name is still left as a field in the event for backwards compatibility purposes, and the consumer can still use the name to create the provider instance. However, the name is no longer populated.

Some other refactoring was carried out, specifically moving code from the helper package to more appropriate packages.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/artifacts/db.go
+++ b/internal/artifacts/db.go
@@ -1,0 +1,81 @@
+// Copyright 2024 Stacklok, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package artifacts stores logic relating to the artifact entity type
+package artifacts
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/google/uuid"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
+	"github.com/stacklok/minder/internal/db"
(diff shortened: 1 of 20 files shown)


internal/engine/executor.go
```diff
@@ -202,10 +202,10 @@ func (e *Executor) prepAndEvalEntityEvent(ctx context.Context, inf *entities.Ent
 	// TODO: clean this up once we get rid of provider name
 	var provider *db.Provider
 	var err error
-	if inf.ProviderID == uuid.Nil {
-		provider, err = e.providerStore.GetByName(ctx, inf.ProjectID, inf.Provider)
-	} else {
+	if inf.ProviderID != uuid.Nil {
```
dmjb, line 205: Swapped this condition because it is now expected that the Provider ID path will the more commonly used one

internal/artifacts/db.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2024 Stacklok, Inc
```
dmjb, line 1: Previously part of the helpers package

internal/repositories/db.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2024 Stacklok, Inc
```
dmjb, line 1: Previously part of the helpers package

internal/eea/entities.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2024 Stacklok, Inc
```
dmjb, line 1: Previously part of the helpers package. Moved under eea, because only the eea code uses these functions.

evankanderson (commented):
What's the sequencing on this with the previous PR?  Do we need one pushed in production and clearing all the previous messages which only had name before we start this one?

dmjb:
> What's the sequencing on this with the previous PR? Do we need one pushed in production and clearing all the previous messages which only had name before we start this one?

My previous PR was part of yesterday evening's deploy to prod, so this PR is safe to merge.

rdimitrov (approved):

dmjb:
Did some local verification

PR #3131, opened 2024-04-18
Implement ProviderFactory and ProviderClassFactory

dmjb:
Relates to #2845

Introduce two new constructs for creating providers. The top-level ProviderFactory takes the name/project or ID of a provider and instantiates it. In order to instantiate it, it checks the class of the provider, and delegates it to a ProviderClassFactory, which is responsible for creating it. This split exists so that we can register new types of provider while minimizing changes to existing code.

This PR does not make use of the new interfaces. Instead, it just adds the implementations. This is to simplify review - the process of getting rid of the ProviderBuilder will involve a large number of small changes to many parts of code, and that will happen in another PR.

Some points worth noting:

1) This does not provide a ProviderClassFactory for REST or Git since we
   do not use standalone instances of those providers at this point in
   time. It should be trivial to write factories for those at the point
   in time when we need them.
2) Based on discussion with Ozz and Jakub, I have decided to assume that
   the provider class column of the providers table will not be null for
   valid providers. This approach uses the class to figure out which
   concrete type is used to instantiate the provider, and is orthogonal
   to the traits, which describe the set of interfaces which that
   provider can support. In a future PR, I will make the class column
   non-nullable.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/providers/github/factory/factory.go
```diff
@@ -0,0 +1,223 @@
+// Copyright 2024 Stacklok, Inc.
```
dmjb, line 1: The code here is mostly copy-paste from `ProviderBuilder`. The main difference is that the process of figuring out which type of credential is determined dynamically here, whereas `ProviderBuilder` requires it to be figured out when the `ProviderBuilder` struct is instantiated.

internal/providers/factory/factory.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+	}
+	return &providerFactory{
+		classFactories: classes,
+		store:          store,
+	}, nil
+}
+
+func (p *providerFactory) BuildFromID(ctx context.Context, providerID uuid.UUID) (v1.Provider, error) {
+	config, err := p.store.GetByID(ctx, providerID)
```
dmjb, line 84: I've taken to referring to `db.Provider` as the "provider configuration" which I feel is an accurate description of its intended purpose. In future, I may introduce a PR to rename the struct type (but not change the table, the rename can be done using sqlc's configuration). This avoids the confusion from having two types called "Provider" in the system which are very for very different things.

internal/providers/github/factory/factory.go
dmjb: Another note: I have decided to rely on existing test code to verify this logic once it is wired up in the application.

pkg/providers/v1/providers.go
```diff
@@ -133,3 +134,12 @@ func ParseAndValidate(rawConfig json.RawMessage, to any) error {
 
 	return nil
 }
+
+// As is a type-cast function for Providers
+func As[T Provider](provider Provider) (T, error) {
```
JAORMX, line 139: is this used anywhere?

JAORMX (commented):
this lgtm and makes sense.

pkg/providers/v1/providers.go
dmjb: It will be used in the next PR. It will be used as a helper function for casting the Provider to a specific trait

internal/providers/github/factory/factory.go
```diff
@@ -0,0 +1,223 @@
...
+	// This should be validated by the caller, but let's check anyway
+	if !slices.Contains(supportedClasses, class) {
+		return nil, fmt.Errorf("provider does not implement github")
+	}
+
+	if config.Version != v1.V1 {
+		return nil, fmt.Errorf("provider version not supported")
+	}
+
+	creds, err := g.getProviderCredentials(ctx, config)
```
eleftherias, line 92: There are some places where we don't use the DB credential and pass a different credential into the provider constructor.
I think it's valid to not use a provider in those cases, or pass the credential into the function call. 
I'm only calling it out here to be aware of it when we refactor that piece of code later.

eleftherias (approved):

Issue #3142, opened 2024-04-22
Add external ID to Minder entities


dmjb:
### Please describe the enhancement

Per recent discussion on providers, add an external ID to artifacts and repositories. This will be an opaque identifier which the providers will use to identify the entity in the third party API.

### Solution Proposal

Per discussion with Evan - store the ID information in a protobuf, serialize, and store in the DB.

evankanderson:
@dmjb -- I think we may be doing this somewhere else.  Do you want to link this in, or close it?

PR #3143, opened 2024-04-22
Add external IDs to entities

dmjb:
MARKING AS DRAFT: This needs even more thinking than I thought originally.

Relates to #3142

Per our design discussions, add an opaque identifier column to each entity type. This is intended to contain provider-specific information used to identify that entity in the provider's API. This will allow us to build generic interfaces around providers which do not need to know exactly what the provider needs as an ID for each entity type.

Alternatives considered:

1) Use a string. The problem with this approach is that since the string
   can be inspected, it creates the temptation to write code outside of
   the provider which uses the structure/contents of that string. This
   is meant to be an opaque identifier.

2) Use a UUID, which serves as a foreign key reference to data in the
   provider's database. I considered this for a while, but it creates a
   requirement for each provider to have a database. This may be a
   reasonable assumption for many cases, but I think we should leave the
   door open for stateless providers when carrying out simple
   integrations.

Note that this PR does not actually use the new external IDs, or migrate existing entities to have them. That will be implemented in future PRs.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -869,9 +881,19 @@ func reconcilePrWithDb(
 
 	switch prEvalInfo.Action {
 	case WebhookActionEventOpened, WebhookActionEventSynchronize:
+		externalID, err := proto.Marshal(&pb.GitHubPullRequestID{
```
dmjb, line 884: Ideally we would have a `PullRequestService` which encapsulates methods for creating these entities. I am avoiding this sort of refactoring until we have the provider/core code more cleanly split.

internal/reconcilers/repository.go
```diff
@@ -152,6 +153,16 @@ func (r *Reconciler) handleArtifactsReconcilerEvent(ctx context.Context, evt *Re
 	for _, artifact := range artifacts {
 		// store information if we do not have it
 		typeLower := strings.ToLower(artifact.GetPackageType())
+		externalID, err := proto.Marshal(&pb.GitHubArtifactID{
```
dmjb, line 156: Same comment as for pull requests: ideally we should factor this out into a dedicated struct/interface... but now is not the time.

database/query/pull_requests.sql
```diff
@@ -1,14 +1,9 @@
--- name: CreatePullRequest :one
```
dmjb, line 1: This query was only ever used in a test, and it could easily be replaced by the upsert operation without breaking the test.

database/migrations/000054_entity_external_id.up.sql
```diff
@@ -0,0 +1,22 @@
...
+-- limitations under the License.
+
+-- external_id is a serialization of a protobuf struct which contains
+-- provider-specific used to identify the entity
+-- e.g. a github repo may contain info on owner and repo name, or an OCI
+-- provider may contain the image name/tag and digest
+-- TODO: make these non-nullable once we populate the table
+ALTER TABLE repositories ADD COLUMN external_id BYTEA;
+ALTER TABLE artifacts ADD COLUMN external_id BYTEA;
+ALTER TABLE pull_requests ADD COLUMN external_id BYTEA;
```
JAORMX, line 22: should we also have a migration that fills in these from the `repo_id`?

dmjb: Going to do that in a separate PR.

proto/minder/v1/minder.proto
```diff
@@ -2143,3 +2143,33 @@ message Provider {
     // This is an output-only field. It may be: "set", "unset", "not_applicable".
     string credentials_state = 9;
 }
+
+// These are for persistence use only, and are not part of the API contract
+
+// GitHubRepoID is the external ID for a repository in GitHub
+message GitHubRepoID {
```
JAORMX, line 2150: I'm not sure I understand this part, is the idea then to serialize this and use it as the external ID? Thought we'd use the GH ID for this as it allows us to detect if there has been a repo change or takeover.

PR #3155, opened 2024-04-24
Initial implementation of IDP interface and Keycloak implementation

### LINKED ISSUE #3151: Introduce an interface for looking up IDP identities

evankanderson:
# Summary

Defines and implements an IDP interface and a Keycloak implementation of said interface which allows us to specify and look up users by their GitHub login aliases while recording their KeyCloak user IDs in the actual database.  Manually tested with the following after logging in both `evankanderson` and `stacklokdemo`:

```
minder project role grant -r admin -s stacklokdemo
```

The data written to OpenFGA uses the Keycloak user IDs as `user:<GUID>`, but the API reports the github names.

As part of this change, I've introduced the use of [OpenFeature](https://openfeature.dev/) with the [GO Feature Flag library in module configuration (i.e. reading from a file)](https://gofeatureflag.org/docs/go_module/configuration).  To activate, you'll need to add the following `flags-config.yaml` to your minder directory:

```yaml
idp_resolver:
  variations:
    UUIDonly: false
    NiceNames: true
  defaultRule:
    variation: NiceNames
```
(There are many more fine-grained options, but this turns _on_ the flag.  The file is mounted into the Docker container, and the server will notice updates within a minute or two, so you can toggle the feature on and off live.)

Fixes #3151 

## Change Type

# Testing

I've added unit tests with decent coverage for the various new code.  I've also put the feature behind a feature flag, because there's more work to be done to clean up and unify this interface.

# Review Checklist:


internal/controlplane/handlers_authz_test.go
```diff
@@ -250,6 +253,7 @@ func TestProjectAuthorizationInterceptor(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+		tc := tc
```
dmjb, line 256: There shouldn't be any need for this in Go 1.22

internal/controlplane/handlers_authz_test.go
```diff
@@ -157,6 +157,7 @@ func TestEntityContextProjectInterceptor(t *testing.T) {
...
```
dmjb, line 160: Same comment as below - curious if you ran into a problem that required this to be reinserted.

cmd/server/app/serve.go
```diff
@@ -103,6 +104,15 @@ var serveCmd = &cobra.Command{
 			return fmt.Errorf("unable to prepare authz client for run: %w", err)
 		}
 
+		kc, err := keycloak.NewKeyCloak("", cfg.Identity.Server)
```
dmjb, line 107: What is the significance of the `name` argument to `NewKeyCloak`, and why is it left blank here?

dmjb:
Changes LGTM aside from the small questions.

internal/controlplane/handlers_authz_test.go
evankanderson: Agreed, but I was getting a bunch of linter complaints.  I can [turn off the `loopclosure` analysis in VSCode](https://github.com/golang/vscode-go/issues/227), but I worry that it might catch other issues.  If you've got a good recipe for ignoring these in our go1.22 code and not elsewhere, let me know.  😁

cmd/server/app/serve.go
evankanderson: From the definition of IdentityProvider:
```
	// String returns the name of the identity provider.  This should be a short
	// one-word string suitable for presentation.  As a special case, a _single_
	// provider may use the empty string as its name to act as a default / fallback
	// provider.
```

We're using KeyCloak as the default provider.  At some point in the future, I'd like to be able to add various machine identities as providers which can be granted access to the system without needing to have accounts in our primary IDP.  (For example, an AWS role, a GCP service account, or a GitHub Action -- these might be stored as `user:gha/repo:octo-org/octo-repo:environment:prod` as the user key in OpenFGA, for example.)

I'm not ready to commit to implementing one of these machine identity providers yet, but I wanted to put the foundations in place while I was doing this refactoring.

Currently, we only have the one default identity provider, which is Keycloak.  This PR doesn't switch over all our Keycloak identity usage yet, only puts a backwards-compatible shim over the Keycloak + OpenFGA combination.

evankanderson:
This is ready for review.  Sorry it got a bit large, but hopefully it's not too painful to review.  Writing the tests already helped me catch a couple bugs that weren't in the common path, so yay tests!

JAORMX (commented):
I'm guessing we'll need a migration from our current role assignments to assignments using the new subjects, right?

evankanderson:
> I'm guessing we'll need a migration from our current role assignments to assignments using the new subjects, right?

No, that's what the `""` IDP is about -- by registering the current Keycloak as the default IDP, the `user:<Keycloak ID>` assignments in OpenFGA don't need to change.  In the future, we can add something like `aws/<account>:<role>` or what-not and it will be stored in OpenFGA as `user:aws/<underlying id>`, but the default IDP is stored without a prefix so that we don't need to migrate.

internal/flags/flags.go
```diff
@@ -0,0 +1,86 @@
...
+// Bool provides a simple wrapper around client.Boolean to normalize usage for Minder.
+func Bool(ctx context.Context, client *openfeature.Client, feature Experiment) bool {
+	ret := client.Boolean(ctx, string(feature), false, FromContext(ctx))
+	// TODO: capture in telemetry records
+	return ret
+}
+
+// ProviderFromFlags installs an OpenFeature Provider based on the flags config.
+// This curently only supports the GoFeatureFlag file-based provider.
+func ProviderFromFlags(ctx context.Context, cfg config.FlagsConfig) {
```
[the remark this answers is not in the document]
evankanderson: You mean e.g. `OpenFeatureProviderFromFlags`? Yeah, I could do that.

Unfortunate naming collision.  ¯\_(ツ)_/¯

JAORMX (approved):

JAORMX (approved):

JAORMX (approved):

PR #3159, opened 2024-04-24
Use `go generate` directives for mock generation

dmjb:
Instead of adding commands to the Makefile, use go generate directives in the files which contain the interfaces we want to mock. This approach has the following advantages:

1) Less context switching: you can declare the need for a mock in the
   code and not have to edit the makefiles.
2) The same directive can be reused everywhere - you no longer need to
   specify the filepath, interface name and the path to the generated
   mock.
3) If the interface is renamed or moved, the code generation will still
   work without modification.

I have left some commands in the makefile for generating mocks from code in `pkg`, and from autogenerated code. We could in theory get rid of these usages as well, but I did not feel there was much point at this time since they are edge cases.

You may note that there are a lot of additional lines of code: this is because I'm using the "source" mode of mockgen instead of the "reflect" mode. The source mode will generate mocks for all interfaces in the file, and not just the specified ones. However, it does so much quicker than the reflect mode - `make gen` now takes about half the time it did previously.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


JAORMX:
With this PR:
```
________________________________________________________
Executed in    9.18 secs    fish           external
   usr time   26.27 secs  395.00 micros   26.27 secs
   sys time    9.51 secs  168.00 micros    9.51 secs
```

Without this PR:
```
________________________________________________________
Executed in   35.45 secs    fish           external
   usr time  137.61 secs    0.00 micros  137.61 secs
   sys time   23.82 secs  476.00 micros   23.82 secs

```

JAORMX (approved):

PR #3202, opened 2024-04-29
Use ProviderManager in webhook handler

dmjb:
Relates to #2845

Switch to the use of the new ProviderManager interface in webhook handler to instantiate the Github provider.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -864,9 +846,8 @@ func getPullRequestInfoFromPayload(
 	}, nil
 }
 
-func reconcilePrWithDb(
+func (s *Server) reconcilePrWithDb(
```
dmjb, line 849: Changed some functions to methods so we do not need to pass around the `db.Store` pointer.

internal/controlplane/handlers_githubwebhooks_test.go
```diff
@@ -246,27 +243,6 @@ func (s *UnitTestSuite) TestHandleWebHookRepository() {
 	projectID := uuid.New()
 	providerID := uuid.New()
 
-	// create a test oauth2 token
```
dmjb, line 249: No longer needed because the token querying is not triggered unless the provider is actually instantiated. The test case for the handler does not need the provider to be instantiated.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -641,33 +624,32 @@ func (s *Server) parseArtifactPublishedEvent(
...
 	ctx context.Context,
 	whPayload map[string]any,
 	msg *message.Message,
 	dbrepo db.Repository,
-	store db.Store,
-	prov *providers.ProviderBuilder,
+	providerID uuid.UUID,
 	action string,
 ) error {
-	// NOTE(jaosorior): this webhook is very specific to github
```
evankanderson, line 653: This comment is still true, we just think it's obvious, correct?

internal/controlplane/handlers_githubwebhooks.go
evankanderson: Long-term, we want these to be in the provider implementation and not the server, correct?

dmjb: Eventually, yes.

internal/controlplane/handlers_githubwebhooks.go
dmjb: Correct - but this particular codepath is so Github-specific that it's not really worth noting any more.

PR #3204, opened 2024-04-29
Delete docs/docs/how-to/using-minder-with-ghas.md

meganbruce:
This content is already in the Tutorials section for Minder SaaS, and I think that's the best place for it. I'm proposing that we delete this file in the Policy and Profile Management section, in favor of having it in Tutorials (so that we don't have the same content in two different places in the Minder docs).

## Change Type


jhrozek:
going to modify the PR opening comment to make the PR validator happy.

PR #3209, opened 2024-04-30
Add flag telling if repo from provider is already registered

### LINKED ISSUE #2627: Update the ListRemoteRepositoriesFromProvider response with a field saying which repo is registered and which not
The goal of this issue is to update the `ListRemoteRepositoriesFromProvider` response(ref. `UpstreamRepositoryRef`) with a field saying which repo is registered and which not.

cc: @peppescg

blkt:
# Summary

`ListRemoteRepositoriesFromProvider` gets a list of repositories from a provider, but used to ignore whether any of them were already registered.

This changes add boolean flag populated via a lookup on a map of already-registered repos.

Fixes #2627 (server side)

## Change Type

# Testing

I just updated an existing test.

# Review Checklist:


internal/controlplane/handlers_repositories.go
```diff
@@ -368,6 +368,25 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
 			continue
 		}
 
+		registeredRepos, err := s.store.ListRepositoriesByProjectID(ctx, db.ListRepositoriesByProjectIDParams{
+			ProjectID: projectID,
+			Provider:  getNameFilterParam(provider.Name),
+		})
+		if err != nil {
+			// Here we try to handle errors gracefully by
+			// simply not filtering.
```
blkt, line 377: I would love to hear opinions on this.

internal/controlplane/handlers_repositories_test.go
```diff
@@ -470,8 +616,12 @@ func createServer(ctrl *gomock.Controller, repoServiceSetup repoMockBuilder, git
...
 			}, nil).AnyTimes()
 		store.EXPECT().
-			GetProviderByID(gomock.Any(), gomock.Any()).
-			Return(provider, nil).AnyTimes()
+			ListRepositoriesByProjectID(gomock.Any(), gomock.Any()).
+			Return([]db.Repository{dbExistingRepo}, nil).AnyTimes()
+	}
+
+	if storeSetup != nil {
+		store = storeSetup(ctrl)
```
blkt, line 624: I added this brutal override to avoid messing up other tests, but I believe it would be worth refactoring lines 589-through-621 completely. Let me know what you think.

internal/controlplane/handlers_repositories.go
eleftherias: This seems like the right approach to me

internal/controlplane/handlers_repositories_test.go
eleftherias: Maybe the `RepositoryService` needs a `ListRepositories` method. Then we wouldn't have to mock the store here. What do you think @dmjb?

dmjb: I agree. @blkt Let me know if you want to pair on this.

internal/controlplane/handlers_repositories.go
dmjb: Concern: Image a customer calls this endpoint in a script and uses the registered/not registered information for some purpose, such as selecting new repos to register or something like this. If we have a transient failure, their script could end up treating some registered repos as unregistered, and this could cause problems.

I'm not convinced that swallowing this error is helpful, and it could cause problems for the client.

blkt: Yes, that's what I was concerned with as well, but as far as I can tell the only problem that can occur is that a Customer tries to `minder repo register` the same repo again, and at that point either it fails completely or it succeeds and returns a "repo already registered" message.

Besides, if a Customer uses this endpoint as last interaction with Minder before triggering his own internal processes (e.g. for some sort of reporting feature or so), then he is probably misusing it and we might need a separate endpoint just for such use cases.

It's not a big change either way and we might want to fail unconditionally to keep it simple for now, let me know what you folks prefer.

eleftherias: I don't feel strongly about this. I'm fine with propagating the error to the caller as well.

blkt: After discussing this, we decided to fail rather than hiding the error.

internal/controlplane/handlers_repositories_test.go
blkt: I replaced direct access to store with access to repository service and removed all unnecessary utilities from tests.
I did not add tests for the Service itself, but the code path is still tested by tests on handler.

eleftherias (approved):

PR #3226, opened 2024-05-02
Artifacts: Properly handle untagged images

### LINKED ISSUE #3225: Artifact: Ingesting images by digest always returns "latest"
### Describe the issue

When ingesting a container image by digest and no tags are defined, the ingester will skip the specified digest and ingest the last image returned by the the API (generally matching `latest`)

### To Reproduce

This entity should ingest `e3efa96133139eaf23335d6b8aace18c8db9555c026c02d1244ab61af986e181` but instead ingests `4a80bffe07d74bece13c3b954c6463aa25b36492bdfcfc0f674381e0b43f58c9` (`latest` at the time of this report)

```yaml
---
owner: jakubtestorg
name: good-repo-go
type: container
versions: 
  - sha: sha256:e3efa96133139eaf23335d6b8aace18c8db9555c026c02d1244ab61af986e181
  - tags: []
```

900bb165611a1b04ab769255ca7f2719fb2b268b

puerco:
# Summary

This PR fixes the `artifact.isSkippable` function to properly handle the ingestion of untagged images. It also adds a test to ensure it behaves as expected.

It also optimizes the selection of images to drop early any artifact candidates when hashes don't match, improving the performance of the filtering process considerably.

Fixes #3225

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/engine/ingester/artifact/artifact.go
+++ b/internal/engine/ingester/artifact/artifact.go
@@ -279,6 +279,20 @@ func getAndFilterArtifactVersions(
 			continue
 		}
 
+		match := false
+		for _, v := range artifact.Versions {
+			if v.Sha == *version.Name {
+				match = true
+				break
+			}
+		}
+		if !match {
+			zerolog.Ctx(ctx).Debug().Str("name", *version.Name).
+				Str("hashes", fmt.Sprintf("%+v", artifact.Versions)).
+				Msg("artifact versions do not match")
+			continue
+		}
+
 		// If the artifact version is applicable to this rule, add it to the list
 		zerolog.Ctx(ctx).Debug().Str("name", *version.Name).Strs("tags", tags).Msg("artifact version matched")
 		res = append(res, *version.Name)
@@ -309,10 +323,7 @@ func isSkippable(artifactType verifyif.ArtifactType, createdAt time.Time, opts m
 		}
 		tags, ok := opts["tags"].([]string)
 		if !ok {
-			return nil
-		} else if len(tags) == 0 {
-			// if the artifact has no tags, skip it
-			return fmt.Errorf("artifact has no tags")
+			tags = []string{}
 		}
 		// if the artifact has a .sig tag it's a signature, skip it
 		if verifier.GetSignatureTag(tags) != "" {
@@ -320,7 +331,7 @@ func isSkippable(artifactType verifyif.ArtifactType, createdAt time.Time, opts m
 		}
 		// if the artifact tags don't match the tag matcher, skip it
 		if !filter.MatchTag(tags...) {
-			return fmt.Errorf("artifact tags does not match")
+			return fmt.Errorf("artifact tags do not match (%+v)", tags)
 		}
 		return nil
 	default:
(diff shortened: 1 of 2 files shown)


dmjb (approved):

jhrozek:
Actually, let's not push this just yet, I think there might be a bug. I was seeing some weird behaviour with this branch which went away once I switched to main.

puerco:
Let me know which weird behavior and I can check it

jhrozek:
> Let me know which weird behavior and I can check it

OK, I know what happened, but I don't have a good answer on fixing it. So there's two ways we can ingest an artifact:
1) "directly", through either a webhook or the mindev command. This does pass on the Version, typically the one you ingest.
2) through a reconcile. This is normally used when you first create a profile and then you register the repo. The code is [here](https://github.com/stacklok/minder/blob/7fc9740fa5735c4707eaeed297ece69971c481a0/internal/reconcilers/repository.go#L183). In this case we send a nil Versions which means that the loop iterates over slice with len == which means the default matches=false is used.

Would it be reasonable to adjust the new code to only execute on len(artifact.Versions) > 0 ?

JAORMX:
> 2\. through a reconcile. This is normally used when you first create a profile and then you register the repo. The code is [here](https://github.com/stacklok/minder/blob/7fc9740fa5735c4707eaeed297ece69971c481a0/internal/reconcilers/repository.go#L183). In this case we send a nil Versions which means that the loop iterates over slice with len == which means the default matches=false is used.

I actually think the reconcile should be the one to iterate through the versions and pass a single event per version it reconciles. That way it's the same behavior as the webhook and we don't need to special case this. It also makes the artifact ingest logic simpler.

jhrozek:
> > 2. through a reconcile. This is normally used when you first create a profile and then you register the repo. The code is [here](https://github.com/stacklok/minder/blob/7fc9740fa5735c4707eaeed297ece69971c481a0/internal/reconcilers/repository.go#L183). In this case we send a nil Versions which means that the loop iterates over slice with len == which means the default matches=false is used.
> 
> I actually think the reconcile should be the one to iterate through the versions and pass a single event per version it reconciles. That way it's the same behavior as the webhook and we don't need to special case this. It also makes the artifact ingest logic simpler.

This would be a LOT of versions and events, most of which would be thrown away.

JAORMX:
> This would be a LOT of versions and events, most of which would be thrown away.

uhm... why not just try to get the status of the `latest` tag? Do we really want to iterate over everything? Trying to think of use-cases.

jhrozek:
> > This would be a LOT of versions and events, most of which would be thrown away.
> 
> uhm... why not just try to get the status of the `latest` tag? Do we really want to iterate over everything? Trying to think of use-cases.

I'm not sure I understand what the proposal is but IIRC the GH API didn't let you iterate over tags, you had to list everything and then filter yourself. Maybe this is different with the OCI API?

JAORMX:
> > > This would be a LOT of versions and events, most of which would be thrown away.
> > 
> > 
> > uhm... why not just try to get the status of the `latest` tag? Do we really want to iterate over everything? Trying to think of use-cases.
> 
> I'm not sure I understand what the proposal is but IIRC the GH API didn't let you iterate over tags, you had to list everything and then filter yourself. Maybe this is different with the OCI API?

yeah, there are query parameters that allow you to order by creation date and limit the number of responses.

rdimitrov:
I remember we were unsure of the practicality of having untagged images along with some implementation details that made us discard them.

Let's wait merging this for a bit until we meet to chat more about it.

PR #3238, opened 2024-05-03
Artifact: Unify signer identity URIs

puerco:
# Summary

This commit modifies the signer identity function in the artifact code to handle URIs evenly regardless of where they came from.

We trim the ref and return the path only to allow minder to create rules based on it regardless of the repository.

Fixes for the smoke tests

## Change Type

# Testing

Epxande the function test to capture the full range of cases

# Review Checklist:

### DIFF
--- a/internal/engine/ingester/artifact/artifact.go
+++ b/internal/engine/ingester/artifact/artifact.go
@@ -205,9 +205,14 @@ func (i *Ingest) getVerificationResult(
 
 			// If we got verified provenance info for the artifact version, populate the rest of the verification result
 			if res.IsVerified {
+				siIdentity, err := signerIdentityFromCertificate(res.Signature.Certificate)
+				if err != nil {
+					zerolog.Ctx(ctx).Err(err).Msg("error parsing signer identity")
+				}
+
 				verResult.Repository = res.Signature.Certificate.SourceRepositoryURI
 				verResult.Branch = branchFromRef(res.Signature.Certificate.SourceRepositoryRef)
-				verResult.SignerIdentity = signerIdentityFromCertificate(res.Signature.Certificate)
+				verResult.SignerIdentity = siIdentity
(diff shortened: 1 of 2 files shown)


dmjb (approved):

internal/engine/ingester/artifact/artifact.go
```diff
@@ -336,21 +341,30 @@ func branchFromRef(ref string) string {
...
+		// when a user authenticated using an OIDC provider.
+		// Any other SAN types are returned verbatim
+		return c.SubjectAlternativeName.Value, nil
+	}
 
-	} else if c.SubjectAlternativeName.Value != "" {
-		// This is the use case where there is no build signer URI but there is a subject alternative name
-		// Usually this is the case when signing through an OIDC provider. The value is the signer's email identity.
-		return c.SubjectAlternativeName.Value
+	uString, _, _ := strings.Cut(builderURL, "@")
```
jhrozek, line 364: @puerco this is much better than what we had and I like that the code is generic, but in this case I wonder if making the code a little GH specific would go a long way towards better usability, because the returned path still contains the org and repo paths (`jakubtestorg/demo-repo-go/.github/workflows/build-image-signed-cosign-static.yml` in my case).

Do you think it would be too much of a hack to check if the extensions.Issuer is `https://token.actions.githubusercontent.com` and in that case strip also `GithubWorkflowRepository`? Is it possible to forge signatures with that data that we would successfully verify?

JAORMX: If we're validating the signature and it comes from a public (or GH) sigstore, it should be fine to make that assumption. That is, to check the issuer and get the relevant metadata from there. Public sigstore has a subset of IdPs they support and they do check the issuer before giving out a Fulcio certificate. This would be a different case if we'd be verifying with a private sigstore though.

PR #3262, opened 2024-05-07
Refactor engine code to use specific provider traits in more places

dmjb:
Relates to #2845

When evaluating a policy, we instantiate a ProviderBuilder and then pass it around to various parts of the code. Once the specific trait is known, we get the specific trait we need from the ProviderBuilder.

While replacing the ProviderBuilder, I notice that there are many parts of the code where we continue to pass around the ProviderBuilder even though we know the specific provider trait which will be needed. I changed various code to explicitly require the provider trait it needs instead of taking the ProviderBuilder and getting an instance of the desired type. As a result of this change, test setup is simplified, and this will make my PR to replace the ProviderBuilder smaller.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/engine/actions/alert/alert.go
+++ b/internal/engine/actions/alert/alert.go
@@ -43,7 +43,11 @@ func NewRuleAlert(rt *pb.RuleType, pbuild *providers.ProviderBuilder) (engif.Act
 		if alertCfg.GetSecurityAdvisory() == nil {
 			return nil, fmt.Errorf("alert engine missing security-advisory configuration")
 		}
-		return security_advisory.NewSecurityAdvisoryAlert(ActionType, rt.GetSeverity(), alertCfg.GetSecurityAdvisory(), pbuild)
+		client, err := pbuild.GetGitHub()
+		if err != nil {
+			return nil, fmt.Errorf("could not instantiate provider: %w", err)
+		}
+		return security_advisory.NewSecurityAdvisoryAlert(ActionType, rt.GetSeverity(), alertCfg.GetSecurityAdvisory(), client)
 	}
 
 	return nil, fmt.Errorf("unknown alert type: %s", alertCfg.GetType())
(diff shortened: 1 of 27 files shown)


internal/engine/actions/alert/alert.go
```diff
@@ -43,7 +43,11 @@ func NewRuleAlert(rt *pb.RuleType, pbuild *providers.ProviderBuilder) (engif.Act
 		if alertCfg.GetSecurityAdvisory() == nil {
 			return nil, fmt.Errorf("alert engine missing security-advisory configuration")
 		}
-		return security_advisory.NewSecurityAdvisoryAlert(ActionType, rt.GetSeverity(), alertCfg.GetSecurityAdvisory(), pbuild)
+		client, err := pbuild.GetGitHub()
```
JAORMX, line 46: Can we first check if the provider implements github and skip if it doesn't?

JAORMX (changes requested):

internal/engine/actions/remediate/pull_request/pull_request.go
```diff
@@ -91,7 +90,7 @@ type paramsPR struct {
 func NewPullRequestRemediate(
 	actionType interfaces.ActionType,
 	prCfg *pb.RuleType_Definition_Remediate_PullRequestRemediation,
-	pbuild *providers.ProviderBuilder,
+	ghCli provifv1.GitHub,
```
JAORMX, line 93: How will this look like once we generalize it for another provider, say GitLab?

dmjb: We will need a more generic trait at that point for pull requests.

internal/engine/actions/alert/alert.go
dmjb: I'm not sure I agree with this suggestion:

1. The code copies the existing behaviour: https://github.com/stacklok/minder/blob/main/internal/engine/actions/alert/security_advisory/security_advisory.go#L161
2. If other aspects of the configuration are incorrect, we also return an error.

dmjb: Note in my next PR, I will change the behaviour to print an explicit error if the provider is not Github

JAORMX: There is an explicit "error" that gets the engine to mark something as skipped. Anyway, it's fine to do it in a separate PR.

JAORMX (approved):

PR #3270, opened 2024-05-07
Remove ProviderBuilder from engine

dmjb:
Relates to #2845

Refactor engine to use ProviderManager. Various bits of refactoring along the way, including renaming some struct fields and variables for clarity.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/cmd/dev/app/rule_type/rttst.go
+++ b/cmd/dev/app/rule_type/rttst.go
@@ -17,7 +17,6 @@ package rule_type
 import (
 	"bytes"
 	"context"
-	"database/sql"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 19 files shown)


internal/engine/executor.go
```diff
@@ -298,9 +229,7 @@ func (e *Executor) evalEntityEvent(
 func (e *Executor) getEvaluator(
 	ctx context.Context,
 	inf *entities.EntityInfoWrapper,
-	ectx *EntityContext,
```
dmjb, line 301: `ectx` was passed in to provide the project ID and the provider name. Since we no longer need provider name, and `inf` has the project ID - get rid of it.

cmd/dev/app/rule_type/rttst.go
```diff
@@ -148,38 +149,30 @@ func testCmdRun(cmd *cobra.Command, _ []string) error {
...
-			GitHubApp: &serverconfig.GitHubAppConfig{
-				AppName: "test",
-			},
-		},
-		nil, // this is unused here
-	))
+		nil,
+		clients.NewGitHubClientFactory(telemetry.NewNoopMetrics()),
+		false,
+	)
```
JAORMX, line 168: This probably blocks https://github.com/stacklok/minder/pull/2983 from moving forward. What's the alternative? Seems like this locks us into a one-provider rule type engine for now which is not ideal.

JAORMX (approved):

cmd/dev/app/rule_type/rttst.go
dmjb: For posterity, we chatted about this on a call. I think we need to revisit this later - there are a bunch of things going on which may change what the ideal solution would be.

PR #3277, opened 2024-05-08
Improve Trusty integration

puerco:
# Summary

This PR surfaces the Trusty maliciousness data on the comment added by minder when analyzing dependencies. It also adds the required piping to add the data to the rest of the required low scored dependencies.

I've broken the Trusty evaluator to more scoped utility functions to make them more testable and added a few initial unit tests. It needs a little bit more mocking to write an integration test and the rest of the unit tests

I've simplified the comment template, we now have a single template instead of 3 and it is now pure markdown which is smaller. Here' is a screenshot of the output with malicious dependencies:

image

Demo PRs showing

- [A malicious dependency](https://github.com/puerco/demo-repo-python/pull/29)
- [A low score dependency](https://github.com/puerco/demo-repo-python/pull/27)
- [Both malicious and low score](https://github.com/puerco/demo-repo-python/pull/28)

## Change Type

# Testing

Added initial tests (up to 22% from 0 :) )

# Review Checklist:

### DIFF
--- a/internal/engine/eval/trusty/config.go
+++ b/internal/engine/eval/trusty/config.go
@@ -47,6 +47,14 @@ type ecosystemConfig struct {
 	// If `evaluate_score` is set to something else (e.g. `provenance`)
 	// then that score is used, which comes from the details field.
 	EvaluateScore string `json:"evaluate_score" mapstructure:"evaluate_score"`
+
+	// The provenance field contains the minimal provenance score
+	// to consider the origin of the package as trusted.
+	Provenance float64 `json:"provenance" mapstructure:"provenance"`
+
+	// Activity is the minimal activity score that minder needs to find to
+	// consider the package as trustworthy.
+	Activity float64 `json:"activity" mapstructure:"activity"`
 }
 
 // config is the configuration for the vulncheck evaluator
--- a/internal/engine/eval/trusty/config.go
+++ b/internal/engine/eval/trusty/config.go
@@ -60,16 +60,22 @@ func defaultConfig() *config {
 		Action: pr_actions.ActionSummary,
 		EcosystemConfig: []ecosystemConfig{
 			{
-				Name:  "npm",
-				Score: 5.0,
+				Name:       "npm",
+				Score:      5.0,
+				Provenance: 5.0,
+				Activity:   5.0,
 			},
 			{
-				Name:  "pypi",
-				Score: 5.0,
+				Name:       "pypi",
+				Score:      5.0,
+				Provenance: 5.0,
+				Activity:   5.0,
 			},
 			{
-				Name:  "go",
-				Score: 5.0,
+				Name:       "go",
+				Score:      5.0,
+				Provenance: 5.0,
+				Activity:   5.0,
 			},
 		},
 	}
--- a/internal/engine/eval/trusty/config.go
+++ b/internal/engine/eval/trusty/config.go
@@ -40,14 +40,6 @@ type ecosystemConfig struct {
 	// evaluated depends on the `evaluate_score` field.
 	Score float64 `json:"score" mapstructure:"score" validate:"required"`
 
-	// EvaluateScore tells the trusty executor which score to use
-	// for evaluation. This is useful when the trusty API returns.
-	// The default is the summary score. If `score` or an empty string, the
-	// summary score is used.
-	// If `evaluate_score` is set to something else (e.g. `provenance`)
-	// then that score is used, which comes from the details field.
-	EvaluateScore string `json:"evaluate_score" mapstructure:"evaluate_score"`
-
 	// The provenance field contains the minimal provenance score
 	// to consider the origin of the package as trusted.
 	Provenance float64 `json:"provenance" mapstructure:"provenance"`
@@ -116,31 +108,4 @@ func (c *config) getEcosystemConfig(ecosystem pb.DepEcosystem) *ecosystemConfig 
 	}
 
 	return nil
-}
-
-func (ec *ecosystemConfig) getScoreSource() string {
-	if ec.EvaluateScore == DefaultScore || ec.EvaluateScore == SummaryScore {
-		return SummaryScore
-	}
-
-	return ec.EvaluateScore
-}
-
-func (ec *ecosystemConfig) getScore(inSummary ScoreSummary) (float64, error) {
-	if inSummary.Score != nil && (ec.EvaluateScore == DefaultScore || ec.EvaluateScore == SummaryScore) {
-		return *inSummary.Score, nil
-	}
-
-	// If the score is not the summary score, then it must be in the details
-	rawScore, ok := inSummary.Description[ec.EvaluateScore]
-	if !ok {
-		return 0, fmt.Errorf("score %s not found in details", ec.EvaluateScore)
-	}
-
-	s, ok := rawScore.(float64)
-	if !ok {
-		return 0, fmt.Errorf("score %s is not a float64", ec.EvaluateScore)
-	}
-
-	return s, nil
 }
--- a/internal/engine/eval/trusty/trusty.go
+++ b/internal/engine/eval/trusty/trusty.go
@@ -41,6 +41,7 @@ const (
 type Evaluator struct {
 	cli      provifv1.GitHub
 	endpoint string
+	client   *trustyClient
 }
 
 // NewTrustyEvaluator creates a new trusty evaluator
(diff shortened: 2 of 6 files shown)


yrobla (dismissed):

jhrozek (dismissed):
one minor comment, but looks good.

internal/engine/eval/trusty/trusty.go
```diff
@@ -59,119 +60,203 @@ func NewTrustyEvaluator(ctx context.Context, ghcli provifv1.GitHub) (*Evaluator,
...
+
+	prSummaryHandler, err := newSummaryPrHandler(prDependencies.Pr, e.cli, e.endpoint)
+	if err != nil {
+		return fmt.Errorf("failed to create summary handler: %w", err)
+	}
+
+	// Classify all dependencies, tracking all that are malicious or scored low
+	for _, dep := range prDependencies.Deps {
+		if err := classifyDependency(ctx, &logger, e.client, ruleConfig, prSummaryHandler, dep); err != nil {
+			return fmt.Errorf("classifying dependency: %w", err)
```
jhrozek, line 105: Not blocking, but do you think we should error out the whole evaluation here? I wonder if we could classify a dependency as something like unknown instead.

lukehinds:
There is more payload from trusty we could leverage here.

Provenance: sigstore or historical provenance. It might make sense to surface a threshold here for folks to set within the policy. This could be an float (I think?) between 1-10, or a simply bool cc: @therealnb , @yrobla 

We also have deprecated or achieved available as fields to report in the same way as malicious

internal/engine/eval/trusty/trusty.go
puerco: Since this logic is responsible for blocking PRs introducing malware into the codebase, I would rather build a more resilient client to make the requests more robust and indeed fail if we can't get the trusty score instead of just letting it through as an unknown. WDYT? 

The only two cases where this might fail are because of an error talking to trusty or a misconfiguration in the profile.

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

puerco:
OK, I've surfaced the rest of the malicious data and the score components on the PR comment. The rule evaluator will now honor minimal scores for provenance and activity in the rule configuration. 

image

stacklokbot (commented):

stacklokbot (commented):

jhrozek:
I really like the new format, but it seems like a high-scoring package is getting flagged now as well, see e.g. https://github.com/jakubtestorg/bad-python/pull/193

internal/engine/eval/trusty/config.go
```diff
@@ -40,14 +40,6 @@ type ecosystemConfig struct {
 	// evaluated depends on the `evaluate_score` field.
 	Score float64 `json:"score" mapstructure:"score" validate:"required"`
 
-	// EvaluateScore tells the trusty executor which score to use
-	// for evaluation. This is useful when the trusty API returns.
-	// The default is the summary score. If `score` or an empty string, the
-	// summary score is used.
-	// If `evaluate_score` is set to something else (e.g. `provenance`)
-	// then that score is used, which comes from the details field.
-	EvaluateScore string `json:"evaluate_score" mapstructure:"evaluate_score"`
```
jhrozek, line 49: Oh this is probably why the "good" packages are now flagged as low scoring? Does it mean that everyone who deployed this profile with the old ruletype (before https://github.com/stacklok/minder-rules-and-profiles/pull/111) would get all their deps flagged?

puerco: mmh interesting. Let me check why that one is getting flagged. The removal of `EvaluateScore` should not affect this one in particular as it has a high score and also a high provenance component.

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

puerco:
This is how its looking now. I will investigate Jakubs bugfind next but we can merge as it is and iterate on smaller PRs.

image

internal/engine/eval/trusty/config.go
puerco: I could not reproduce it ([see here](https://github.com/puerco/demo-repo-python/pull/75)) it is weird because the profile is the same and the previous `EvaluateScore` is not used anymore. I'll keep looking

jhrozek:
> This is how its looking now. I will investigate Jakubs bugfind next but we can merge as it is and iterate on smaller PRs.
> 

Let me try to reproduce again so I don't send you down the wrong path

jhrozek:
I think there are two issues: The evaluator has hardcoded default configuration which should include also the two new attributes and additionally it seems like the condition when evaluating the scores are reversed - shouldn't we check that the score of the package is higher than the config? Check out the diff below:

```
diff --git a/internal/engine/eval/trusty/config.go b/internal/engine/eval/trusty/config.go
index 8b673b494..534427641 100644
--- a/internal/engine/eval/trusty/config.go
+++ b/internal/engine/eval/trusty/config.go
@@ -60,16 +60,22 @@ func defaultConfig() *config {
 		Action: pr_actions.ActionSummary,
 		EcosystemConfig: []ecosystemConfig{
 			{
-				Name:  "npm",
-				Score: 5.0,
+				Name:       "npm",
+				Score:      5.0,
+				Provenance: 5.0,
+				Activity:   5.0,
 			},
 			{
-				Name:  "pypi",
-				Score: 5.0,
+				Name:       "pypi",
+				Score:      5.0,
+				Provenance: 5.0,
+				Activity:   5.0,
 			},
 			{
-				Name:  "go",
-				Score: 5.0,
+				Name:       "go",
+				Score:      5.0,
+				Provenance: 5.0,
+				Activity:   5.0,
 			},
 		},
 	}
diff --git a/internal/engine/eval/trusty/trusty.go b/internal/engine/eval/trusty/trusty.go
index 634fdb915..53b4dcf6c 100644
--- a/internal/engine/eval/trusty/trusty.go
+++ b/internal/engine/eval/trusty/trusty.go
@@ -253,14 +253,14 @@ func classifyDependency(
 		}
 	}
 
-	if ecoConfig.Score <= packageScore {
+	if ecoConfig.Score > packageScore {
 		reasons = append(reasons, TRUSTY_LOW_SCORE)
 	}
-	if ecoConfig.Provenance <= descr["provenance"].(float64) {
+	if ecoConfig.Provenance > descr["provenance"].(float64) {
 		reasons = append(reasons, TRUSTY_LOW_PROVENANCE)
 	}
 
-	if ecoConfig.Activity <= descr["activity"].(float64) {
+	if ecoConfig.Activity > descr["activity"].(float64) {
 		reasons = append(reasons, TRUSTY_LOW_ACTIVITY)
 	}
 	if len(reasons) > 0 {

```

stacklokbot (commented):

stacklokbot (commented):

evankanderson:
> it seems like the condition when evaluating the scores are reversed - shouldn't we check that the score of the package is higher than the config?

Should we have a test for this?  It feels like the sort of bug that we could easily accidentally introduce again.

jhrozek:
> > it seems like the condition when evaluating the scores are reversed - shouldn't we check that the score of the package is higher than the config?
> 
> Should we have a test for this? It feels like the sort of bug that we could easily accidentally introduce again.

We do have a smoke test, but alas, no way of running it locally yet. You are right that we should do better on the unit testing front - not having proper tests is mostly my fault, we do have reasonable tests for the OSV evaluator and the current Trusty evaluator code was meant to be a "quick hack" until we can generalize the PR dependency evaluators into common code and built both OSV and trusty evaluators atop them. We just haven't prioritized that work.

jhrozek (approved):
thank you for being patient with this long review!

JAORMX:
I think we can merge this now. Let's start adding tests in further PRs

puerco:
> Should we have a test for this? It feels like the sort of bug that we could easily accidentally introduce again.
@evankanderson as part of this improvement I've been splitting the logic into more atomic functions to make them testable and to be able to mock the whole trusty evaluator. I'll strat sending incremental PRs to add more tests

PR #3335, opened 2024-05-15
Allow KeyStore to be configured with multiple keys

### LINKED ISSUE #3304: Hide encryption key fetching behind an interface

Place encryption key retrieval behind an interface so that we can create multiple retrieval mechanisms, and select one based on configuration. For example, we may want a secret store which reads straight from a secrets manager instead of reading off disk.

### Solution Proposal

Create interface for key retrieval.

_No response_

### Additional context

_No response_

### Acceptance Criteria

_No response_

dmjb:
Fixes #3304

Add a new configuration structure for encryption keys. This allows a default key and algorithm to be specified, along with fallbacks. A backwards compatibility mechanism is provided which will allow the keystore to work with the existing configuration structure.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

internal/config/utils.go
```diff
@@ -297,3 +273,33 @@ func getDefaultValueForInt64(value string) (any, error) {
 	// Return the original error, not time.ParseDuration's error
 	return nil, err
 }
+
+func getDefaultValue(field reflect.StructField, value string, valueName string) any {
```
dmjb, line 277: Factored out into a function to avoid cyclomatic complexity limits

internal/config/utils.go
```diff
@@ -297,3 +273,33 @@ func getDefaultValueForInt64(value string) (any, error) {
...
+	case reflect.Int32, reflect.Int16, reflect.Int8, reflect.Int,
+		reflect.Uint64, reflect.Uint32, reflect.Uint16, reflect.Uint8, reflect.Uint:
+		defaultValue, err = strconv.Atoi(value)
+	case reflect.Float64:
+		defaultValue, err = strconv.ParseFloat(value, 64)
+	case reflect.Bool:
+		defaultValue, err = strconv.ParseBool(value)
+	case reflect.Slice:
+		defaultValue = nil
+	case reflect.Map:
```
dmjb, line 295: This is the new addition

internal/config/server/crypto.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// KeyStoreConfig specifies the type of keystore to use and its configuration
+type KeyStoreConfig struct {
+	Type string `mapstructure:"type"`
+	// is currently expected to match the structure of LocalFileKeyStoreConfig
+	Config map[string]any `mapstructure:"config"`
+}
+
+// DefaultCrypto defines the default cyrpto to be used for new data
```
evankanderson, line 31:
```suggestion
// DefaultCrypto defines the default crypto to be used for new data
```

internal/config/server/crypto.go
```diff
@@ -0,0 +1,48 @@
...
+	KeyID     string `mapstructure:"key_id"`
+	Algorithm string `mapstructure:"algorithm"`
+}
+
+// FallbackCrypto defines the optional list of keys and algorithms to fall
+// back to.
+// When rotating keys or algorithms, add the old ones here.
+type FallbackCrypto struct {
+	KeyIDs     []string `mapstructure:"key_ids"`
+	Algorithms []string `mapstructure:"algorithms"`
```
evankanderson, line 42: Should this be a `[]struct{KeyID string, Algo string}`?  I'm wondering what happens if the two arrays are not the same length.

internal/config/server/crypto.go
```diff
@@ -0,0 +1,48 @@
...
+// CryptoConfig is the configuration for the crypto engine
+type CryptoConfig struct {
+	KeyStore KeyStoreConfig `mapstructure:"key_store"`
+	Default  DefaultCrypto  `mapstructure:"default"`
+	Fallback FallbackCrypto `mapstructure:"fallback"`
+}
+
+// KeyStoreConfig specifies the type of keystore to use and its configuration
+type KeyStoreConfig struct {
+	Type string `mapstructure:"type"`
```
evankanderson, line 26: Do you want defaults here?

internal/crypto/keystores/keystore.go
```diff
@@ -31,32 +34,59 @@ type KeyStore interface {
...
 
 // NewKeyStoreFromConfig creates an instance of a KeyStore based on the
 // AuthConfig in Minder.
 // Since our only implementation is based on reading from the local disk, do
 // all key loading during construction of the struct.
-// TODO: allow support for multiple keys/algos
-func NewKeyStoreFromConfig(config *serverconfig.AuthConfig) (KeyStore, error) {
-	key, err := config.GetTokenKey()
+func NewKeyStoreFromConfig(config serverconfig.CryptoConfig) (KeyStore, error) {
+	// You might think that a nil value cannot be deserialized to a struct,
```
evankanderson, line 50: This is because the particular item is a map, and `nil` is the zero value for a map, right?  (Sorry, this comment was confusing -- maybe call out that `Config` is a map.)

internal/crypto/keystores/keystore.go
```diff
@@ -31,32 +34,59 @@ type KeyStore interface {
...
+		return nil, errors.New("keystore config is missing")
+	}
+
+	// TODO: support other methods in future
+	if config.KeyStore.Type != LocalKeyStore {
+		return nil, fmt.Errorf("unexpected keystore type: %s", config.KeyStore.Type)
+	}
+
+	var keystoreCfg serverconfig.LocalFileKeyStoreConfig
+	err := mapstructure.Decode(config.KeyStore.Config, &keystoreCfg)
```
evankanderson, line 62: This is because Config is a `map[string]any`, right?

(I'd like to minimize the total area of `mapstructure` in our code, though I'm willing to be convinced that I shouldn't)

evankanderson (approved):
It looks like we ended up changing the configuration interface, but we didn't change the encryption / decryption interfaces, correct?

A few small suggestions, but this overall makes sense to me.

stacklokbot (commented):

stacklokbot (commented):

internal/crypto/keystores/keystore.go
dmjb: Yes, it's because the exact type of `Config` is not known until the `Type` field is read.

Perhaps this could be done as part of config parsing, but I felt it would be better to do at the point where we make use of the config.

stacklokbot (commented):

stacklokbot (commented):

internal/config/server/crypto.go
dmjb: It is expected that they will not always be the same length. For example, if we rotate a key, but do not change the algorithm, then we will add a fallback key but not a corresponding algorithm.

The selection of key/algorithm during decryption is based on the data we store in the database (see the `EncryptedData` struct in `models.go`) and the selection of the algorithm during encryption is based on the values in the `DefaultCrypto` struct. The fallback values state which algorithms and keys should be loaded at startup to support old data.

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

internal/config/server/crypto.go
evankanderson: Ah, okay.  Works for me.

internal/crypto/keystores/keystore.go
evankanderson: As I said, I'm okay with this pattern -- it would be nice if we could 👍  or 👎  the config at parsing time, but that's already not the case.

evankanderson (approved):

PR #3346, opened 2024-05-16
Make github webhook easier to extend with new events for auto registration.

### LINKED ISSUE #3359: Refactor GitHub webhook handler to have a case for each event of interest
Currently, the GitHub webhook handler treats different event types uniformly and the "dispatch logic" for each event is spread over multiple routines. This makes it harder to extend the logic in case new events need to be handled.

Refactor the handler so that every event type has its own routine implementing the necessary business logic.

blkt:
# Summary

The idea is to make it easier to extend github webhook with events that are tied to more than one repository, namely "installation" and "installation_repositories", which are necessary for repo auto registration.

Fixes #3359 

## Change Type

# Testing

Added several unit tests, plus I additionally tested it manually on a local environment.

# Review Checklist:


stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

stacklokbot (commented):

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -240,33 +267,124 @@ func (s *Server) HandleGitHubWebHook() http.HandlerFunc {
...
+			res, processingErr = s.processRepositoryVulnerabilityAlertEvent(ctx, event)
+		case *github.SecurityAdvisoryEvent:
+			res, processingErr = s.processSecurityAdvisoryEvent(ctx, event)
+		case *github.SecurityAndAnalysisEvent:
+			res, processingErr = s.processSecurityAndAnalysisEvent(ctx, event)
+		case *github.OrgBlockEvent,
+			*github.OrganizationEvent:
+			l.Info().Msgf("webhook events %s do not contain repo", wes.Typ)
+		case *github.PushEvent:
+			// TODO has GetRepo but the type is different :|
```
JAORMX, line 354: This needs to be handled so we can keep evaluating `git` ingester policies.

JAORMX (changes requested):

internal/controlplane/handlers_githubwebhooks.go
blkt: Resolved, I tested this here https://github.com/blkt/demo-repo-python/pull/2

blkt: Also
* [code](https://github.com/stacklok/minder/blob/a26fafdee87e9bbf06322fe9fad1cc61d7476cc9/internal/controlplane/handlers_githubwebhooks.go#L475)
* [tests](https://github.com/stacklok/minder/blob/a26fafdee87e9bbf06322fe9fad1cc61d7476cc9/internal/controlplane/handlers_githubwebhooks_test.go#L2573-L2642)

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -145,42 +334,64 @@ func (s *Server) HandleGitHubAppWebhook() http.HandlerFunc {
...
+
+			w.WriteHeader(http.StatusOK)
 			return
 		}
 
-		wes.Accepted = true
-		l.Info().Str("message-id", m.UUID).Msg("publishing event for execution")
+		if res != nil {
+			l.Info().Str("message-id", m.UUID).Msg("publishing event for execution")
+			if err := res.iiw.ToMessage(m); err != nil {
```
jhrozek, line 382: would it make sense to double-check that iiw is non-nil here in case we refactor the code later to handle more types?

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -145,42 +334,64 @@ func (s *Server) HandleGitHubAppWebhook() http.HandlerFunc {
...
 
-		ctx := r.Context()
 		l := zerolog.Ctx(ctx).With().
 			Str("webhook-event-type", m.Metadata[events.GithubWebhookEventTypeKey]).
 			Str("providertype", m.Metadata[events.ProviderTypeKey]).
 			Str("upstream-delivery-id", m.Metadata[events.ProviderDeliveryIdKey]).
 			Logger()
+		ctx = l.WithContext(ctx)
+
+		wes.Accepted = true
```
jhrozek, line 351: Question: Would it be safer to rely on how we already initialize `wes` here with `Accepted: false` and only set it to true when we know we accept the message instead of the other way around?

internal/controlplane/handlers_githubwebhooks.go
blkt: It might be the case that we want to return an interface and switch-case on its actual type.
Which way would you prefer?

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -237,82 +444,496 @@ func (s *Server) HandleGitHubWebHook() http.HandlerFunc {
...
+			"team_add":
+			res, processingErr = s.processRepositoryEvent(ctx, rawWBPayload)
+		case "package":
+			// This is an artifact-related event, and can
+			// only trigger a reconciliation.
+			res, processingErr = s.processPackageEvent(ctx, rawWBPayload)
+		case "pull_request":
+			res, processingErr = s.processPullRequestEvent(ctx, rawWBPayload)
+		// This event is not currently handled.
+		case "org_block",
```
jhrozek, line 494: I know we had them in the earlier list, but I can't tell why now. Since we don't handle them should we set accepted=false?

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -237,82 +444,496 @@ func (s *Server) HandleGitHubWebHook() http.HandlerFunc {
...
+			"organization":
+			l.Info().Msgf("webhook events %s do not contain repo", wes.Typ)
+		case "ping":
+			// For ping events, we do not set wes.Accepted
+			// to true because they're not relevant
+			// business events.
+			wes.Accepted = false
+			wes.Error = false
+			s.processPingEvent(ctx, rawWBPayload)
+		default:
```
jhrozek, line 504: should we set accepted=false here as well?

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -237,82 +444,496 @@ func (s *Server) HandleGitHubWebHook() http.HandlerFunc {
...
+			// webhook ID that doesn't correspond to the
+			// one we have stored in the DB.
+			return nil, newErrNotHandled("meta event with action %s not handled, hook ID %d does not match stored webhook ID %d",
+				event.GetAction(),
+				event.GetHookID(),
+				dbrepo.WebhookID.Int64,
+			)
+		}
+	}
+	// If we get this far it means we got a deleted event for a
```
jhrozek, line 706: this comment seems like it's no longer relevant (or should be moved to where we handle `webhookActionEventDeleted`

internal/controlplane/handlers_githubwebhooks.go
blkt: I agree on setting accepted=false, I'll fix that.

That said, I had a look back at github documentation and it looks like these events do contain a repository, but their semantics is somewhat unrelated to it
* `org_block` event occurs when organization owners or moderators block or unblock a non-member from collaborating on the organization's repositories
* `organization` event occurs when there is activity relating to an organization and its members

I think we should not process them, but I am ok adding them back.

internal/controlplane/handlers_githubwebhooks.go
blkt: I'll set `wes.Accepted` to true inside the switch case, looks clearer to me.

internal/controlplane/handlers_githubwebhooks.go
blkt: Removing it.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -237,82 +443,486 @@ func (s *Server) HandleGitHubWebHook() http.HandlerFunc {
...
+	if err := json.Unmarshal(payload, &event); err != nil {
+		return nil, err
+	}
+
+	if event.Action == nil {
+		return nil, errors.New("invalid event: action is nil")
+	}
+	if event.Package == nil || event.Repo == nil {
+		log.Printf("could not determine relevant entity for event. Skipping execution.")
+		return nil, nil // this is awkward
```
jhrozek, line 585: I agree, feel free to return e.g. `errNotHandled` or introduce a new error that would be handled in `handleParseError`

internal/controlplane/handlers_githubwebhooks.go
jhrozek: I don't really have a preference other than we don't panic :-)

internal/controlplane/handlers_githubwebhooks.go
jhrozek: I'm fine with removing. I don't remember why we added these events. I think in the distant past we handled all events and when we added the filtering we didn't want to filter too much. But I don't see a use for these events either.

jhrozek (commented):
Still reading the patchset, just want to submit some comments I had pending so the review keeps happening from both sides. I like the code so far!

jhrozek:
For the record, I like the code, I'm just doing some manual testing before finally acking.

jhrozek (approved):
Manual tests seem to work fine and the code looks good, too.

I think @JAORMX needs to take a look as well because currently he sets changes requested.

JAORMX (approved):

Issue #3380, opened 2024-05-20
Trusty Integration: Mark a flagged pull request as requesting changes


puerco:
When trusty catches problematic dependencies being introduced in a PR, it should have the capability to add a review requesting changes.

Issue #3379, opened 2024-05-20
Trusty Integration: Block PR on deprecated/malicious


puerco:
We should add to the trusty pr handler the capability to block pull requests when deprecated or malicious dependencies are found.

PR #3392, opened 2024-05-22
Trusty: Support blocking PRs through reviews

puerco:
# Summary

Minder now has the capability to request changes in PRs when it finds something odd based on Trusty dependency data.

This PR  also introduces a new setting in the trusty ruletype configuration to disable blocking when malicious deps are detected (all power to the users, but why would you want that?!).

Fixes https://github.com/stacklok/minder/issues/3380
Part of https://github.com/stacklok/minder/issues/3379

## Change Type

# Testing

image

# Review Checklist:

### DIFF
--- a/internal/engine/eval/trusty/config.go
+++ b/internal/engine/eval/trusty/config.go
@@ -47,9 +47,12 @@ type ecosystemConfig struct {
 	// Activity is the minimal activity score that minder needs to find to
 	// consider the package as trustworthy.
 	Activity float64 `json:"activity" mapstructure:"activity"`
+
+	// AllowMalicious disables blocking PRs introducing malicious dependencies
+	AllowMalicious bool `json:"allow_malicious" mapstructure:"allow_malicious"`
 }
 
-// config is the configuration for the vulncheck evaluator
+// config is the configuration for the trusty evaluator
 type config struct {
 	Action          pr_actions.Action `json:"action" mapstructure:"action" validate:"required"`
 	EcosystemConfig []ecosystemConfig `json:"ecosystem_config" mapstructure:"ecosystem_config" validate:"required"`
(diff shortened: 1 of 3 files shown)


dmjb (dismissed):

internal/engine/eval/trusty/config.go
```diff
@@ -49,15 +49,19 @@ type ecosystemConfig struct {
...
+	Block           map[string]bool   `json:"block" mapstructure:"block"`
 	EcosystemConfig []ecosystemConfig `json:"ecosystem_config" mapstructure:"ecosystem_config" validate:"required"`
 }
 
 func defaultConfig() *config {
 	return &config{
-		Action: pr_actions.ActionSummary,
+		Action: pr_actions.ActionReviewPr,
+		Block: map[string]bool{
+			"malicious": true,
```
rdimitrov, line 63: Isn't it safer if we invert this and don't rely on initialising the bool with true? 

Also I think we can even delete this option in the Review action. I mean turning it on kind of overlaps with the other actions and if someone doesn't want Minder to block the PR, they can use some of them, right? 

I think this is how the vulnerability/OSV rule works like, right?

rdimitrov (changes requested):
I don't have anything against, but I'll click the request changes in case someone decides to merge it before we resolve the comment 👍

internal/engine/eval/trusty/config.go
puerco: Ah so the idea is that this option is _for_ the review action. In addition to malicious, we will have another entry to block (or not) when deprecated packages are found. This will let users control when they want to block or just get feedback from minder separately for those two specific dependency types.

puerco: Also note that this is initializing missing map entries from the default settings [here](https://github.com/stacklok/minder/pull/3392/files#diff-4747ed96b4116a79071ea012528f62f247e7f143e6d8e7f276739816cb54d163R179-R181).

jhrozek: I'm actually fine with this config granularity @rdimitrov . The way I read it the action tells you the "what" to do with the PR and the "block" tells you which of the attributes to take into account when deciding whether to block. And I also think all of them should be on by default.

jhrozek (dismissed):
A test with a malicious package went fine. As discussed with @puerco it would be nice to display a list of alternatives to the user also when flagging a malicous package, but this can be done in a separate PR.

puerco:
In the rush to write this yesterday I noticed that the configuration flags I added didn't match those in the PRD. I rewrote the logic to rename the options and to make them configurable per ecosystem. @rdimitrov you'll be glad to see that the logic of the blocking flags is now flipped :)

rdimitrov (approved):

internal/engine/eval/trusty/config.go
rdimitrov: Ah, thanks for flipping the setting! 😃 Yeah, my motivation was to keep vulncheck and trusty behaviour similar, but I agree on the functionality of this + we'll be refactoring that part so we might as well make it available for all PR related Minder actions not just Trusty.

PR #3413, opened 2024-05-23
Add batch creation logic for the reminder service

Vyom-Yadav:
# Summary

Part - 2 #2262 

**Algorithm:**

1. Generate a random v4 `uuid` as starting point.
2. Iterate and pick repos until the end of the table is reached.
3. After end of table is reached, start iterating from the first row in the table.

Only the starting point is random; batch creation becomes sequential after one end of the table is reached. If after reaching the end of the table, the iteration point is selected randomly again, the coverage of all repos won't be guaranteed. 

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/database/query/profile_status.sql
+++ b/database/query/profile_status.sql
@@ -74,6 +74,16 @@ SELECT p.id, p.name, ps.profile_status, ps.last_updated FROM profile_status ps
 INNER JOIN profiles p ON p.id = ps.profile_id
 WHERE p.project_id = $1;
 
+-- ListOldestRuleEvaluationsByRepositoryId has casts in select statement as sqlc generates incorrect types.
+-- cast after MIN is required due to a known bug in sqlc: https://github.com/sqlc-dev/sqlc/issues/1965
+
+-- name: ListOldestRuleEvaluationsByRepositoryId :many
+SELECT re.repository_id::uuid AS repository_id, MIN(rde.last_updated)::timestamp AS oldest_last_updated
+FROM rule_evaluations re
+    INNER JOIN rule_details_eval rde ON re.id = rde.rule_eval_id
+WHERE re.repository_id = ANY (sqlc.arg('repository_ids')::uuid[])
+GROUP BY re.repository_id;
+
 -- DeleteRuleStatusesForProfileAndRuleType deletes a rule evaluation
 -- but locks the table before doing so.
 
--- a/database/query/repositories.sql
+++ b/database/query/repositories.sql
@@ -45,6 +45,25 @@ WHERE project_id = $1 AND webhook_id IS NOT NULL
     AND (lower(provider) = lower(sqlc.narg('provider')::text) OR sqlc.narg('provider')::text IS NULL)
 ORDER BY repo_name;
 
+-- name: ListRepositoriesAfterID :many
+SELECT *
+FROM repositories
+WHERE id > $1
+ORDER BY id
+LIMIT sqlc.arg('limit')::bigint;
+
+-- name: UpdateReminderLastSentById :exec
+UPDATE repositories
+SET reminder_last_sent = NOW()
+WHERE id = $1;
+
+-- name: RepositoryExistsAfterID :one
+SELECT EXISTS (
+  SELECT 1
+  FROM repositories
+  WHERE id > $1)
+AS exists;
+
 -- name: DeleteRepository :exec
 DELETE FROM repositories
 WHERE id = $1;
--- a/internal/reminder/reminder.go
+++ b/internal/reminder/reminder.go
@@ -16,12 +16,9 @@
 package reminder
 
 import (
-	"bytes"
 	"context"
-	"encoding/gob"
 	"errors"
 	"fmt"
-	"os"
 	"sync"
 	"time"
 
@@ -32,10 +29,6 @@ import (
 	reminderconfig "github.com/stacklok/minder/internal/config/reminder"
 	"github.com/stacklok/minder/internal/db"
 )
-
-func init() {
-	gob.Register(map[projectProviderPair]string{})
-}
 
 // Interface is an interface over the reminder service
 type Interface interface {
@@ -53,8 +46,7 @@ type reminder struct {
 	stop     chan struct{}
 	stopOnce sync.Once
 
-	projectListCursor string
-	repoListCursor    map[projectProviderPair]string
+	repositoryCursor uuid.UUID
 
 	ticker *time.Ticker
 
@@ -62,27 +54,18 @@ type reminder struct {
 	eventDBCloser  driverCloser
 }
 
-type projectProviderPair struct {
-	// Exported for gob
-
-	ProjectId uuid.UUID
-	Provider  string
-}
-
 // NewReminder creates a new reminder instance
 func NewReminder(ctx context.Context, store db.Store, config *reminderconfig.Config) (Interface, error) {
-	logger := zerolog.Ctx(ctx)
 	r := &reminder{
-		store:          store,
-		cfg:            config,
-		stop:           make(chan struct{}),
-		repoListCursor: make(map[projectProviderPair]string),
+		store: store,
+		cfg:   config,
+		stop:  make(chan struct{}),
 	}
-	err := r.restoreCursorState(ctx)
-	if err != nil {
-		// Non-fatal error, if we can't restore the cursor state, we'll start from scratch.
-		logger.Error().Err(err).Msg("error restoring cursor state")
-	}
+
+	// Set to a random UUID to start
+	r.repositoryCursor = uuid.New()
+	logger := zerolog.Ctx(ctx)
+	logger.Info().Msgf("initial repository cursor: %s", r.repositoryCursor)
 
 	pub, cl, err := r.setupSQLPublisher(ctx)
 	if err != nil {
@@ -149,70 +132,111 @@ func (r *reminder) Stop() {
 	})
 }
 
-// storeCursorState stores the cursor state to a file
-// Not thread-safe, should be called from a single goroutine
-func (r *reminder) storeCursorState(ctx context.Context) error {
+func (r *reminder) sendReminders(ctx context.Context) []error {
 	logger := zerolog.Ctx(ctx)
-	logger.Debug().Msg("storing cursor state")
 
-	var buf bytes.Buffer
-	enc := gob.NewEncoder(&buf)
-
-	data := map[string]interface{}{
-		"projectListCursor": r.projectListCursor,
-		"repoListCursor":    r.repoListCursor,
-	}
-	if err := enc.Encode(data); err != nil {
-		return err
-	}
-
-	return os.WriteFile(r.cfg.CursorFile, buf.Bytes(), 0600)
-}
-
-// restoreCursorState restores the cursor state from a file
-// Not thread-safe, should be called from a single goroutine
-func (r *reminder) restoreCursorState(ctx context.Context) error {
-	logger := zerolog.Ctx(ctx)
-	logger.Debug().Msg("restoring cursor state")
-
-	if _, err := os.Stat(r.cfg.CursorFile); os.IsNotExist(err) {
-		return nil
-	}
-
-	data, err := os.ReadFile(r.cfg.CursorFile)
+	// Fetch a batch of repositories
+	repos, err := r.getRepositoryBatch(ctx)
 	if err != nil {
-		return err
+		logger.Error().Err(err).Msg("unable to fetch repositories")
+		return []error{err}
 	}
 
-	buf := bytes.NewBuffer(data)
-	dec := gob.NewDecoder(buf)
+	logger.Info().Msgf("created repository batch of size: %d", len(repos))
 
-	cursorData := make(map[string]interface{})
-
-	if err := dec.Decode(&cursorData); err != nil {
-		return err
-	}
-
-	if val, ok := cursorData["projectListCursor"]; ok {
-		if v, ok := val.(string); ok {
-			r.projectListCursor = v
-		} else {
-			return fmt.Errorf("projectListCursor is not a string")
+	// Update the reminder_last_sent for each repository to export as metrics
+	for _, repo := range repos {
+		logger.Debug().Str("repo", repo.ID.String()).
+			Time("previously", repo.ReminderLastSent.Time).Msg("updating reminder_last_sent")
+		err := r.store.UpdateReminderLastSentById(ctx, repo.ID)
+		if err != nil {
+			logger.Error().Err(err).Str("repo", repo.ID.String()).Msg("unable to update reminder_last_sent")
+			return []error{err}
 		}
-	}
-
-	if val, ok := cursorData["repoListCursor"]; ok {
-		if v, ok := val.(map[projectProviderPair]string); ok {
-			r.repoListCursor = v
-		} else {
-			return fmt.Errorf("repoListCursor is not a map[projectProviderPair]string")
-		}
+		// TODO: Send the actual reminders
 	}
 
 	return nil
 }
 
-// TODO: Will be implemented in a separate PR
-func (_ *reminder) sendReminders(_ context.Context) []error {
-	return nil
+func (r *reminder) getRepositoryBatch(ctx context.Context) ([]db.Repository, error) {
+	logger := zerolog.Ctx(ctx)
+
+	logger.Debug().Msgf("fetching repositories after cursor: %s", r.repositoryCursor)
+	repos, err := r.store.ListRepositoriesAfterID(ctx, db.ListRepositoriesAfterIDParams{
+		ID:    r.repositoryCursor,
+		Limit: int64(r.cfg.RecurrenceConfig.BatchSize),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	eligibleRepos, err := r.getEligibleRepositories(ctx, repos)
+	if err != nil {
+		return nil, err
+	}
+	logger.Debug().Msgf("%d/%d repositories are eligible for reminders", len(eligibleRepos), len(repos))
+
+	r.updateRepositoryCursor(ctx, repos)
+
+	return eligibleRepos, nil
+}
+
+func (r *reminder) getEligibleRepositories(ctx context.Context, repos []db.Repository) ([]db.Repository, error) {
+	eligibleRepos := make([]db.Repository, 0, len(repos))
+
+	repoIds := make([]uuid.UUID, 0, len(repos))
+	for _, repo := range repos {
+		repoIds = append(repoIds, repo.ID)
+	}
+
+	oldestRuleEvals, err := r.store.ListOldestRuleEvaluationsByRepositoryId(ctx, repoIds)
+	if err != nil {
+		return nil, err
+	}
+
+	idToLastUpdatedMap := make(map[uuid.UUID]time.Time)
+	for _, oldestRuleEval := range oldestRuleEvals {
+		idToLastUpdatedMap[oldestRuleEval.RepositoryID] = oldestRuleEval.OldestLastUpdated
+	}
+
+	for _, repo := range repos {
+		if oldestRuleEval, ok := idToLastUpdatedMap[repo.ID]; ok &&
+			oldestRuleEval.Add(r.cfg.RecurrenceConfig.MinElapsed).Before(time.Now()) {
+			eligibleRepos = append(eligibleRepos, repo)
+		}
+	}
+
+	return eligibleRepos, nil
+}
+
+func (r *reminder) updateRepositoryCursor(ctx context.Context, repos []db.Repository) {
+	logger := zerolog.Ctx(ctx)
+
+	if len(repos) == 0 {
+		r.repositoryCursor = uuid.Nil
+	} else {
+		r.repositoryCursor = repos[len(repos)-1].ID
+		r.adjustCursorForEndOfList(ctx)
+	}
+
+	logger.Debug().Msgf("updated repository cursor to: %s", r.repositoryCursor)
+}
+
+func (r *reminder) adjustCursorForEndOfList(ctx context.Context) {
+	logger := zerolog.Ctx(ctx)
+	// Check if the cursor is the last element in the db
+	exists, err := r.store.RepositoryExistsAfterID(ctx, r.repositoryCursor)
+	if err != nil {
+		logger.Error().Err(err).Msgf("unable to check if repository exists after cursor: %s"+
+			", resetting cursor to zero uuid", r.repositoryCursor)
+		r.repositoryCursor = uuid.Nil
+		return
+	}
+
+	if !exists {
+		logger.Info().Msgf("cursor %s is at the end of the list, resetting cursor to zero uuid",
+			r.repositoryCursor)
+		r.repositoryCursor = uuid.Nil
+	}
 }
--- a/cmd/reminder/app/start.go
+++ b/cmd/reminder/app/start.go
@@ -49,11 +49,9 @@ func start(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("unable to read config: %w", err)
 	}
 
-	// Configuration is normalized to ensure that all values are valid
-	// and if they can't be fixed, an error is returned
-	_, err = cfg.Normalize(cmd)
+	err = cfg.Validate()
 	if err != nil {
-		return fmt.Errorf("unable to normalize config: %w", err)
+		return fmt.Errorf("error validating config: %w", err)
 	}
 
 	ctx = logger.FromFlags(cfg.LoggingConfig).WithContext(ctx)
--- a/config/reminder-config.yaml.example
+++ b/config/reminder-config.yaml.example
@@ -16,8 +16,6 @@
 recurrence:
   interval: "1h"
   batch_size: 100
-  max_per_project: 10
-  min_project_fetch_limit: 10
   min_elapsed: "1h"
 
 database:
@@ -27,8 +25,6 @@ database:
   dbpass: postgres
   dbname: minder
   sslmode: disable
-
-cursor_file: "/tmp/reminder-cursor"
 
 logging
   level: "debug"
--- a/database/migrations/000065_repo_reminder.down.sql
+++ b/database/migrations/000065_repo_reminder.down.sql
@@ -1,0 +1,15 @@
+-- Copyright 2024 Stacklok, Inc
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+ALTER TABLE repositories DROP COLUMN reminder_last_sent;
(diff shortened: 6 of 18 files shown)


database/migrations/000060_repo_reminder.up.sql
```diff
@@ -0,0 +1,17 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+ALTER TABLE repositories ADD COLUMN reminder_last_sent TIMESTAMP;
+
+CREATE EXTENSION IF NOT EXISTS tsm_system_rows;
```
Vyom-Yadav, line 17: I'm not sure if enabling extensions should be part of migration files.

sqlc.yaml
```diff
@@ -24,4 +24,10 @@ packages:
...
     emit_exact_table_names: false
-    emit_empty_slices: true
\ No newline at end of file
+    emit_empty_slices: true
+overrides:
+  - db_type: 'pg_catalog.interval'
+    go_type: 'github.com/jackc/pgtype.Interval'
+  - db_type: 'pg_catalog.interval'
+    go_type: 'github.com/jackc/pgtype.NullInterval'
+    nullable: true
```
Vyom-Yadav, line 33: See: https://github.com/sqlc-dev/sqlc/issues/429#issuecomment-607930265

rdimitrov (commented):
I missed if it was discussed but should we also have a way to disable reminder for certain repos by filtering repos out of the batch creation logic?

Vyom-Yadav:
> I missed if it was discussed but should we also have a way to disable reminder for certain repos by filtering repos out of the batch creation logic?

User configurable option? 

Why would someone like to do that? To preserve their rate limit? I don't think reminder should be exposed to end users; it's just background reconciliation to keep the system up to date.

blkt (commented):
Small suggestion as I keep looking at the code.

internal/reminder/reminder.go
```diff
@@ -149,70 +139,174 @@ func (r *reminder) Stop() {
...
+
+	// Only fetch additional repositories if we are under the limit
+	if len(repos) < r.cfg.RecurrenceConfig.BatchSize {
+		additionalRepos, err = r.getAdditionalRepos(ctx, repos)
+		if err != nil {
+			return nil, err
+		}
+
+		repos, intersectionPoint = r.mergeRepoBatch(repos, additionalRepos)
 	}
```
evankanderson, line 202: This feels complicated -- it feels like we should call `ListEligibleRepositoriesAfterID` and then use the last returned element of that query to set the cursor for the next run (checking `RepositoryExistsAfterID` and setting the cursor to the zero UUID if no more repos exist).

internal/reminder/reminder.go
```diff
@@ -149,70 +139,174 @@ func (r *reminder) Stop() {
...
+}
+
+func (_ *reminder) mergeRepoBatch(repos []db.Repository, additionalRepos []db.Repository) ([]db.Repository, int) {
+	var mergedBatch []db.Repository
+	mergedBatch = append(mergedBatch, repos...)
+	reposSet := sets.New[db.Repository](repos...)
+
+	// There may be an intersection between the two sets
+	// If there is an intersection, then we need to update the cursor to the last fetched
+	// non-common repository
```
evankanderson, line 256: I'm not sure I understand how this happens, unless there are only e.g. 4 repos eligible in the whole database.

internal/reminder/reminder.go
```diff
@@ -149,70 +139,174 @@ func (r *reminder) Stop() {
...
 	}
-	if err := enc.Encode(data); err != nil {
-		return err
+
+	logger.Debug().Msgf("fetched %d repositories initially", len(repos))
+
+	intersectionPoint := -1
+	var additionalRepos []db.Repository
+
+	// Only fetch additional repositories if we are under the limit
```
evankanderson, line 194: I don't think it's a requirement that we find _exactly_ `BatchSize` repos, just that we don't _exceed_ `BatchSize` repos in one pass.  We're trying to upper-bound the amount of background work added to the system.

I think this will get a lot simpler if we just do one pass:

```golang
repos, err := r.store.ListEligibleRepositoriesAfterID(ctx, {...})
if err != nil {
  return nil, err
}

// Don't actually ignore the error here...
if len(repos) == 0 || !RepositoryExistsAfterID(ctx, repos[len(repos)-1].ID) {
  r.repositoryCursor = uuid.UUID{}
} else {
  r.repositoryCursor = repos[len(repos)-1].ID
}

return repos, nil
```

internal/reminder/reminder.go
```diff
@@ -149,70 +139,174 @@ func (r *reminder) Stop() {
...
+	var mergedBatch []db.Repository
+	mergedBatch = append(mergedBatch, repos...)
+	reposSet := sets.New[db.Repository](repos...)
+
+	// There may be an intersection between the two sets
+	// If there is an intersection, then we need to update the cursor to the last fetched
+	// non-common repository
+	intersectionPoint := -1
+	for i, repo := range additionalRepos {
+		if reposSet.Has(repo) {
```
evankanderson, line 259: This is using `repo` as a hashable object, but we've done two different queries, so the repos might differ in some minutiae like `updated_at` and be included in the batch twice.  Given that we know that the two lists are sorted, we could also march through `additionalRepos` comparing less-than the first item in `repos` and find the intersection point without needing to use the set.

But, as mentioned, I think we can avoid a lot of this code if we're willing to not _exactly_ fill each reminder batch.

evankanderson (commented):
Good to see you back!  I'm going to encourage you to be a little less precise here in favor of ensuring there's an upper bound on the amount of work that we put into the system at any given time, which I think is the more critical piece for background operation.

database/migrations/000060_repo_reminder.up.sql
blkt: suggestion: I think we can get rid of query `GetRandomRepository` by calculating a random UUID on the application side and passing it straight to `ListEligibleRepositoriesAfterID`. Added benefits would be
* we have one less query
* one less step in the process
* the whole process becomes deterministic (and easier to test)

Vyom-Yadav: If we want to start at a random point, fetching a valid random repo is required. If we can start at any point, then sequential iteration would do the job.

If we don't fetch a valid repo / generate a random uuid, then we can potentially have a case where the generated uuid is out of range, and we have to either generate a uuid again or start from the beginning (which defeats the purpose of random generation)

It all comes down to whether we want to start from a valid random point or not.

database/query/repositories.sql
```diff
@@ -16,6 +16,10 @@ INSERT INTO repositories (
     provider_id
     ) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, sqlc.arg(default_branch), sqlc.arg(license), sqlc.arg(provider_id)) RETURNING *;
 
+-- name: GetRandomRepository :one
+SELECT * FROM repositories
+TABLESAMPLE SYSTEM_ROWS(1);
+
```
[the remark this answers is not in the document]
Vyom-Yadav: > "first" random uuid on the application side and then use that use that to start?

There is no way that I know that generates a random uuid in some range (I'd rather not play with that). If we generate on the application side, then we can end up with an out-of-range uuid, which leads to simple sequential iteration.

internal/reminder/reminder.go
```diff
@@ -53,36 +49,30 @@ type reminder struct {
...
-		cfg:            config,
-		stop:           make(chan struct{}),
-		repoListCursor: make(map[projectProviderPair]string),
+		store: store,
+		cfg:   config,
+		stop:  make(chan struct{}),
 	}
-	err := r.restoreCursorState(ctx)
+
+	randomRepo, err := r.store.GetRandomRepository(ctx)
```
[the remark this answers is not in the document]
Vyom-Yadav: It may generate an out-of-range uuid, which will result in sequential iteration. There is nothing wrong with sequential iteration, but I coded it in a way in which we start from a random valid uuid (theoretically speaking, it is possible to get the first uuid as a random uuid from the db which will result in sequential iteration)

This point of generating a random valid uuid may not be that important, so I'm willing to gen random uuid on application side if that's better (in terms of simplicity)

database/query/repositories.sql
```diff
@@ -45,6 +49,28 @@ WHERE project_id = $1 AND webhook_id IS NOT NULL
...
 
+-- name: ListEligibleRepositoriesAfterID :many
+SELECT r.* FROM repositories r
+  INNER JOIN rule_evaluations re ON re.repository_id = r.id
+  INNER JOIN rule_details_eval rde ON rde.rule_eval_id = re.id
+WHERE r.id > $1
+GROUP BY r.id
+HAVING MIN(rde.last_updated) + sqlc.arg('min_elapsed')::interval < NOW()
+ORDER BY r.id
+LIMIT sqlc.narg('limit')::bigint;
```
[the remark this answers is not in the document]
Vyom-Yadav: Agreed on the statement analysis. The current query (without the sub-select) guarantees that we get `limit` rows (if they are there). With a sub-select, selected rows might not be eligible i.e. `last_updated` was recent. 

Agreed with the point that this will be a single time-consuming query rather than a steady load form of query. To keep things and mocking simpler on the application side, I'd go with a sub-query rather than two different queries, i.e. selection and filtration.  

Just using `reminder_last_sent` isn't a good parameter IMO. We can have a case where a reminder was sent 24h ago but the repo was recently updated (edge based), so this would result in extra load on the server (reconciling the repo).

internal/reminder/reminder.go
```diff
@@ -53,36 +49,30 @@ type reminder struct {
...
 	if err != nil {
-		// Non-fatal error, if we can't restore the cursor state, we'll start from scratch.
-		logger.Error().Err(err).Msg("error restoring cursor state")
+		return nil, err
```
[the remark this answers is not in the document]
Vyom-Yadav: Yes, it should be:
```go
if err != nil && !errors.Is(err, sql.ErrNoRows) {
  return nil, err
}
```

sqlc.yaml
Vyom-Yadav: Done.

internal/reminder/reminder.go
```diff
@@ -149,70 +139,174 @@ func (r *reminder) Stop() {
...
+	}
+
+	logger.Info().Msgf("created repository batch of size: %d", len(repos))
 
-	data := map[string]interface{}{
-		"projectListCursor": r.projectListCursor,
-		"repoListCursor":    r.repoListCursor,
+	// Update the reminder_last_sent for each repository to export as metrics
+	for _, repo := range repos {
+		logger.Debug().Msgf("updating reminder_last_sent for repository: %s", repo.ID)
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

internal/reminder/reminder.go
```diff
@@ -149,70 +139,174 @@ func (r *reminder) Stop() {
...
+		err := r.store.UpdateReminderLastSentById(ctx, repo.ID)
+		if err != nil {
+			logger.Error().Err(err).Msgf("unable to update reminder_last_sent for repository: %s", repo.ID)
+			return []error{err}
+		}
+	}
+
+	// TODO: Send the actual reminders
```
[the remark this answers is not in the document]
Vyom-Yadav: I moved the todo comment. This part will be completed in a follow up PR.

internal/reminder/reminder.go
```diff
@@ -149,70 +139,174 @@ func (r *reminder) Stop() {
...
+	}
+
+	logger.Debug().Msgf("fetching repositories after cursor: %s", r.repositoryCursor)
+	repos, err := r.store.ListEligibleRepositoriesAfterID(ctx, db.ListEligibleRepositoriesAfterIDParams{
+		ID:         r.repositoryCursor,
+		MinElapsed: duration,
+		Limit: sql.NullInt64{
+			Int64: int64(r.cfg.RecurrenceConfig.BatchSize),
+			Valid: true,
+		},
```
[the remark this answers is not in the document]
Vyom-Yadav: Bad copy paste, changed it.

internal/reminder/reminder.go
Vyom-Yadav: Ahh, I missed this, but now we fetch only once, so this isn't required.

Vyom-Yadav:
@evankanderson, I updated the PR to fetch only once, and `batch_size` will only be used as the upper limit. The only thing remaining is to decide the starting point random `uuid` generation.

pkg/api/protobuf/go/minder/v1/minder.pb.gw.go
```diff
@@ -4578,21 +4578,21 @@ func RegisterProvidersServiceHandlerServer(ctx context.Context, mux *runtime.Ser
 // RegisterHealthServiceHandlerFromEndpoint is same as RegisterHealthServiceHandler but
 // automatically dials to "endpoint" and closes the connection when "ctx" gets done.
 func RegisterHealthServiceHandlerFromEndpoint(ctx context.Context, mux *runtime.ServeMux, endpoint string, opts []grpc.DialOption) (err error) {
-	conn, err := grpc.DialContext(ctx, endpoint, opts...)
+	conn, err := grpc.NewClient(endpoint, opts...)
```
Vyom-Yadav, line 4581: Haven't verified the changes that came with `protoc-gen-go v1.34.1`

internal/reminder/reminder.go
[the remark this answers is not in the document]
evankanderson: I'm not sure what an out-of-range uuid is -- I'd meant `uuid.NewRandom()`, which generates a valid (v4) UUID.

So, there is one edge case here: we may generate a UUID which is _larger_ than any of the UUIDs in the database (at which point, we should start at the sequentially-first UUID).  I believe that we're using v4 (random) UUIDs, so we should end up with a rougly-even distribution of UUIDs choosing at random.  (If we were using time-sorted UUIDs, picking a random UUID would likely either start at the beginning or the end of the sequence.)

database/query/repositories.sql
```diff
@@ -45,6 +49,35 @@ WHERE project_id = $1 AND webhook_id IS NOT NULL
...
+-- name: ListEligibleRepositoriesAfterID :many
+SELECT r.*
+FROM (
+  SELECT *
+  FROM repositories
+  WHERE repositories.id > $1
+  ORDER BY id
+  LIMIT sqlc.arg('limit')::bigint
+) r
+JOIN LATERAL (
```
evankanderson, line 61: It looks like `LATERAL` converts this from a HashAggregate of a NestedLoop to a NestedLoop that runs an Aggregate inside:

```
EXPLAIN SELECT r.*
FROM (
  SELECT *
  FROM repositories
  ORDER BY id
  LIMIT 10
) r
JOIN LATERAL (
  SELECT MIN(rde.last_updated) AS min_last_updated
  FROM rule_evaluations re
  INNER JOIN rule_details_eval rde ON rde.rule_eval_id = re.id
  WHERE re.repository_id = r.id
) sub ON sub.min_last_updated + interval '1h' < NOW()
ORDER BY r.id;
                                                       QUERY PLAN                                                        
-------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=31.75..318.77 rows=10 width=338)
   ->  Limit  (cost=0.14..2.30 rows=10 width=338)
         ->  Index Scan using repositories_pkey on repositories  (cost=0.14..47.45 rows=220 width=338)
   ->  Aggregate  (cost=31.61..31.63 rows=1 width=8)
         Filter: ((min(rde.last_updated) + '01:00:00'::interval) < now())
         ->  Nested Loop  (cost=8.12..31.60 rows=5 width=8)
               ->  Bitmap Heap Scan on rule_evaluations re  (cost=7.97..15.08 rows=3 width=16)
                     Recheck Cond: (repository_id = repositories.id)
                     ->  Bitmap Index Scan on rule_evaluations_results_name_lower_idx  (cost=0.00..7.97 rows=3 width=0)
                           Index Cond: (repository_id = repositories.id)
               ->  Index Scan using idx_rule_detail_eval_ids on rule_details_eval rde  (cost=0.15..5.50 rows=1 width=24)
                     Index Cond: (rule_eval_id = re.id)
(12 rows)
```

vs

```
EXPLAIN SELECT r.id, r.provider, r.project_id, r.repo_owner, r.repo_name, r.repo_id, r.is_private, r.is_fork, r.webhook_id
FROM (
  SELECT *
  FROM repositories
  ORDER BY id
  LIMIT 10
) r
JOIN rule_evaluations AS re on re.repository_id = r.id
JOIN rule_details_eval rde ON rde.rule_eval_id = re.id
  WHERE re.repository_id = r.id
GROUP BY r.id, r.provider, r.project_id, r.repo_owner, r.repo_name, r.repo_id, r.is_private, r.is_fork, r.webhook_id
HAVING MIN(rde.last_updated) + interval '1h' < NOW()                                                                                                                             ORDER BY r.id;
                                                             QUERY PLAN                                                              
-------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=28.33..28.36 rows=13 width=146)
   Sort Key: r.id
   ->  HashAggregate  (cost=27.39..28.09 rows=13 width=146)
         Group Key: r.id, r.provider, r.project_id, r.repo_owner, r.repo_name, r.repo_id, r.is_private, r.is_fork, r.webhook_id
         Filter: ((min(rde.last_updated) + '01:00:00'::interval) < now())
         ->  Nested Loop  (cost=2.67..26.39 rows=40 width=154)
               ->  Hash Join  (cost=2.52..19.79 rows=26 width=162)
                     Hash Cond: (re.repository_id = r.id)
                     ->  Seq Scan on rule_evaluations re  (cost=0.00..15.10 rows=510 width=32)
                     ->  Hash  (cost=2.40..2.40 rows=10 width=146)
                           ->  Subquery Scan on r  (cost=0.14..2.40 rows=10 width=146)
                                 ->  Limit  (cost=0.14..2.30 rows=10 width=338)
                                       ->  Index Scan using repositories_pkey on repositories  (cost=0.14..47.45 rows=220 width=338)
               ->  Index Scan using idx_rule_detail_eval_ids on rule_details_eval rde  (cost=0.15..0.25 rows=1 width=24)
                     Index Cond: (rule_eval_id = re.id)
(15 rows)
```

While there are more steps in the non-lateral query plan, it seems to plan and execute substantially faster than the LATERAL version.  In an environment with more rows, this is the difference between `EXPLAIN ANALYZE` taking 30ms and 2ms of execution time.

internal/reminder/reminder.go
```diff
@@ -149,70 +137,77 @@ func (r *reminder) Stop() {
...
+		return []error{err}
 	}
-	if err := enc.Encode(data); err != nil {
-		return err
+
+	logger.Info().Msgf("created repository batch of size: %d", len(repos))
+
+	// Update the reminder_last_sent for each repository to export as metrics
+	for _, repo := range repos {
+		logger.Debug().Msgf("updating reminder_last_sent for repository: %s", repo.ID)
```
evankanderson, line 154: Zerolog supports structured logging, which allows us to later easily query and extract (for example) log lines that reference a specific repository id:

```suggestion
		logger.Debug().Str("repo", repo.ID.String()).Time("previously", repo.ReminderLastSent.Time).
			Msg("updating reminder_last_sent")
```

internal/reminder/reminder.go
```diff
@@ -149,70 +137,77 @@ func (r *reminder) Stop() {
...
+		logger.Debug().Msgf("previous reminder_last_sent: %s", repo.ReminderLastSent.Time)
+		err := r.store.UpdateReminderLastSentById(ctx, repo.ID)
+		if err != nil {
+			logger.Error().Err(err).Msgf("unable to update reminder_last_sent for repository: %s", repo.ID)
```
evankanderson, line 158:
```suggestion
			logger.Error().Err(err).Str("repo", repo.ID.String()).Msg("unable to update reminder_last_sent")
```

internal/reminder/reminder.go
```diff
@@ -149,70 +137,77 @@ func (r *reminder) Stop() {
...
+	logger := zerolog.Ctx(ctx)
 
-	if val, ok := cursorData["projectListCursor"]; ok {
-		if v, ok := val.(string); ok {
-			r.projectListCursor = v
-		} else {
-			return fmt.Errorf("projectListCursor is not a string")
-		}
+	if len(repos) == 0 {
+		r.repositoryCursor = uuid.Nil
```
evankanderson, line 189: Since we're filtering the repos we get back from the batch in the database, it's possible that we read e.g. 50 repos that _all_ got filtered.  I think you want to check `RepositoryExistsAfterID` here.

internal/reminder/reminder.go
```diff
@@ -149,70 +137,77 @@ func (r *reminder) Stop() {
...
-// TODO: Will be implemented in a separate PR
-func (_ *reminder) sendReminders(_ context.Context) []error {
-	return nil
+func (r *reminder) adjustCursorForEndOfList(ctx context.Context) {
+	logger := zerolog.Ctx(ctx)
+	// Check if the cursor is the last element in the db
+	exists, err := r.store.RepositoryExistsAfterID(ctx, r.repositoryCursor)
+	if err != nil {
+		logger.Error().Err(err).Msgf("unable to check if repository exists after cursor: %s", r.repositoryCursor)
+		logger.Info().Msg("resetting cursor to zero uuid")
```
evankanderson, line 204: I'd tend to put this as one message in the logs, rather than two separate messages.  Especially when searching structured logs (say, for the last 48 hours across 8 or 10 servers), it can be hard to "scroll" forward or backward to the next log line from a particular server.

evankanderson (commented):
This is looking pretty close -- just two concerns:

1. It looks like `LATERAL` is a lot slower than just a standard inner join.  /shrug
2. I think your logic for when to loop back the beginning isn't quite right.

database/query/profile_status.sql
```diff
@@ -74,6 +74,17 @@ SELECT p.id, p.name, ps.profile_status, ps.last_updated FROM profile_status ps
...
+-- ListOldestRuleEvaluationsByRepositoryId has casts in select statement as sqlc generates incorrect types.
+-- Though repository_id doesn't have non null constraint, but it always has a value in the database.
+-- cast after MIN is required due to a known bug in sqlc: https://github.com/sqlc-dev/sqlc/issues/1965
+
+-- name: ListOldestRuleEvaluationsByRepositoryId :many
+SELECT re.repository_id::uuid AS repository_id, MIN(rde.last_updated)::timestamp AS oldest_last_updated
+FROM rule_evaluations re
+    INNER JOIN rule_details_eval rde ON re.id = rde.rule_eval_id
+WHERE re.repository_id = ANY (sqlc.arg('repository_ids')::uuid[])
+GROUP BY re.repository_id;
```
Vyom-Yadav, line 86: Can `re.repository_id` be null? I was under the assumption that minder needs a repo to function on.

internal/reminder/reminder.go
Vyom-Yadav: Done

internal/reminder/reminder.go
Vyom-Yadav: Done

internal/reminder/reminder.go
Vyom-Yadav: I missed this branch 😞 . I updated the logic. Hopefully I haven't missed any logical branches this time.

Vyom-Yadav:
@evankanderson To address the problem of updating the cursor properly, I have split the fetching and filtering queries. Now, repositories would be fetched unconditionally (to update the cursor) and later filtered on the application side.

internal/reminder/reminder.go
```diff
@@ -149,70 +132,111 @@ func (r *reminder) Stop() {
...
-		} else {
-			return fmt.Errorf("repoListCursor is not a map[projectProviderPair]string")
+	idToLastUpdatedMap := make(map[uuid.UUID]time.Time)
+	for _, oldestRuleEval := range oldestRuleEvals {
+		idToLastUpdatedMap[oldestRuleEval.RepositoryID] = oldestRuleEval.OldestLastUpdated
+	}
+
+	for _, repo := range repos {
+		if oldestRuleEval, ok := idToLastUpdatedMap[repo.ID]; ok &&
+			oldestRuleEval.Add(r.cfg.RecurrenceConfig.MinElapsed).Before(time.Now()) {
```
evankanderson, line 205: Why do this two-step loop here?  It feels like either passing in the cutoff to the query (i.e. `SELECT ... WHERE rde.last_updated < $1 AND re.repository_id = ANY(...) GROUP BY re.repository_id` or the equivalent `HAVING`), or looping through the results once should be sufficient:

```golang
cutoff := time.Now().Sub(r.cfg.RecurrenceConfig.MinElapsed)
for _, evalTime := range oldestRuleEvals {
  if evalTime.OldestLastUpdated.Before(cutoff) {
    eligibleRepos = append(eligibleRepos, repo)
  }
}
```

evankanderson (commented):
This is getting really close!

I'd actually missed the "query might return zero repos but we still need to advance the cursor" issue -- nice catch!

internal/reminder/reminder.go
Vyom-Yadav: `ListOldestRuleEvaluationsByRepositoryId` only queries `rule_evaluations` and `rule_details_eval`, so it returns a slice of:
```go
type ListOldestRuleEvaluationsByRepositoryIdRow struct {
	RepositoryID      uuid.UUID `json:"repository_id"`
	OldestLastUpdated time.Time `json:"oldest_last_updated"`
}
```

We have to iterate over repos as the repo object isn't returned.

internal/reminder/reminder.go
```diff
@@ -149,70 +132,111 @@ func (r *reminder) Stop() {
...
+	logger.Info().Msgf("created repository batch of size: %d", len(repos))
+
+	// Update the reminder_last_sent for each repository to export as metrics
+	for _, repo := range repos {
+		logger.Debug().Str("repo", repo.ID.String()).
+			Time("previously", repo.ReminderLastSent.Time).Msg("updating reminder_last_sent")
+		err := r.store.UpdateReminderLastSentById(ctx, repo.ID)
+		if err != nil {
+			logger.Error().Err(err).Str("repo", repo.ID.String()).Msg("unable to update reminder_last_sent")
+			return []error{err}
```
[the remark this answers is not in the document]
Vyom-Yadav: This function will largely change in the next PR i.e. connecting minder and reminder. A slice of errors is returned as we can get errors while creating messages, sending them, etc. Again, we can discuss this in the next PR. (See `sendReminders` function in the [old PR](https://github.com/stacklok/minder/pull/2475/files#diff-88a5d577ad710aa677287ede4909acba02f5ef8b7c7c6c1b6e9ac6bf039e7d39))

database/query/profile_status.sql
Vyom-Yadav: Added index on rule_evaluations(repository_id). The index will be created in concurrent mode as I thought blocking `rule_evaluations` writes isn't a good option.

internal/reminder/reminder.go
evankanderson: Ah, that seems like it's worth a comment about how we're doing a bunch of transforms of types to fit into the sqlc-generated code (rather than for some other reason, like performance or thread-safety).

Also, a slight preference for:

```golang
cutoff := time.Now().Sub(r.cfg.RecurrenceConfig.MinElapsed)
for _, repo := range repos {
  if t, ok := idToLastUpdate[repo.ID]; ok && t.Before(cutoff) {
    ....
  }
}
```

This has two benefits:

1. You can fit the condition on one line (shortening the time var that lives for a single line, removing `Map` from the name of the map, and doing the date math on a separate line).
2. You only do the date math _once_, and only fetch `time.Now()` once, rather than using a slightly different time for each check.

evankanderson (approved):

evankanderson:
I'm going to merge and then send a PR for the cleanup in `getEligibleRepositories` to avoid needing to do another go-round.  😁

PR #3447, opened 2024-05-28
Webhook handler now processes installation_repositories events

### LINKED ISSUE #3267: Add logic for handling added repositories to the GitHub App webhook handler
On adding a new repository we would receive a webhook message that a new repository was added. The message looks like this:

```
{
    "action": "added",
    "installation": {
        "id": 49896345,
        "account": {
            "login": "jakubtestorg",
            "id": 142222776,
            "node_id": "O_kgDOCHoluA",
            "avatar_url": "https://avatars.githubusercontent.com/u/142222776?v=4",
            "gravatar_id": "",
            "url": "https://api.github.com/users/jakubtestorg",
	     ...
        },
  ...
        "app_id": 864067,
        "app_slug": "jakub-s-local-minder-test",
        },
        "events": [],
        "created_at": "2024-04-22T14:16:52.000+02:00",
        "updated_at": "2024-04-22T14:16:53.000+02:00",
    },
    "repository_selection": "all",
    "repositories_added": [
        {
            "id": 790254972,
            "node_id": "R_kgDOLxpVfA",
            "name": "newrepo2",
            "full_name": "jakubtestorg/newrepo2",
            "private": false
        }
    ],
```

After receiving such request we need to:
- Check that we know the installation ID and infer the project ID
- Route the repositories from repositories_added or repositories_deleted to the appropriate handler. For deletion we already have handleEntityDeleteEvent, we need to add a similar handleEntityAddedEvent
- The handler would create or remove the repository record in the database 
- In case the repository was added, also create the webhook. Once the repository is created, Minder issues an event to reconcile it with the profiles in that project.

blkt:
# Summary

Such events are triggered when a new repository is added to or removed
from the list of repositories accessible to the app.

Fixes #3267 

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_githubwebhooks.go
+++ b/internal/controlplane/handlers_githubwebhooks.go
@@ -187,6 +187,7 @@ func (r *repoEvent) GetHookID() int64 {
 
 type repo struct {
 	ID       *int64  `json:"id,omitempty"`
+	Name     *string `json:"name,omitempty"`
 	FullName *string `json:"full_name,omitempty"`
 	HTMLURL  *string `json:"html_url,omitempty"`
 	Private  *bool   `json:"private,omitempty"`
@@ -197,6 +198,13 @@ func (r *repo) GetID() int64 {
 		return *r.ID
 	}
 	return 0
+}
+
+func (r *repo) GetName() string {
+	if r.Name != nil {
+		return *r.Name
+	}
+	return ""
 }
 
 func (r *repo) GetFullName() string {
@@ -218,6 +226,17 @@ func (r *repo) GetPrivate() bool {
 		return *r.Private
 	}
 	return false
+}
+
+func (r *repo) GetOwner() string {
+	if r.FullName != nil {
+		parts := strings.SplitN(*r.FullName, "/", 2)
+		// It is ok to always return the first item since it
+		// defaults to empty string in case the string has no
+		// separators.
+		return parts[0]
+	}
+	return ""
 }
 
 // pullRequestEvent are events related to pull requests issued around
@@ -282,6 +301,48 @@ func (i *installationEvent) GetAction() string {
 }
 
 func (i *installationEvent) GetInstallation() *installation {
+	return i.Installation
+}
+
+// installationRepositoriesEvent are events occurring when there is
+// activity relating to which repositories a GitHub App installation
+// can access.
+type installationRepositoriesEvent struct {
+	Action              *string       `json:"action,omitempty"`
+	RepositoriesAdded   []*repo       `json:"repositories_added,omitempty"`
+	RepositoriesRemoved []*repo       `json:"repositories_removed,omitempty"`
+	RepositorySelection *string       `json:"repository_selection,omitempty"`
+	Sender              *user         `json:"sender,omitempty"`
+	Installation        *installation `json:"installation,omitempty"`
+}
+
+func (i *installationRepositoriesEvent) GetAction() string {
+	if i.Action != nil {
+		return *i.Action
+	}
+	return ""
+}
+
+func (i *installationRepositoriesEvent) GetRepositoriesAdded() []*repo {
+	return i.RepositoriesAdded
+}
+
+func (i *installationRepositoriesEvent) GetRepositoriesRemoved() []*repo {
+	return i.RepositoriesRemoved
+}
+
+func (i *installationRepositoriesEvent) GetRepositorySelection() string {
+	if i.RepositorySelection != nil {
+		return *i.RepositorySelection
+	}
+	return ""
+}
+
+func (i *installationRepositoriesEvent) GetSender() *user {
+	return i.Sender
+}
+
+func (i *installationRepositoriesEvent) GetInstallation() *installation {
 	return i.Installation
 }
 
@@ -352,7 +413,7 @@ func (s *Server) HandleGitHubAppWebhook() http.HandlerFunc {
 			Logger()
 		ctx = l.WithContext(ctx)
 
-		var res *processingResult
+		var results []*processingResult
 		var processingErr error
 
 		switch github.WebHookType(r) {
@@ -364,7 +425,10 @@ func (s *Server) HandleGitHubAppWebhook() http.HandlerFunc {
 			s.processPingEvent(ctx, rawWBPayload)
 		case "installation":
 			wes.Accepted = true
-			res, processingErr = s.processInstallationAppEvent(ctx, rawWBPayload)
+			results, processingErr = s.processInstallationAppEvent(ctx, rawWBPayload)
+		case "installation_repositories":
+			wes.Accepted = true
+			results, processingErr = s.processInstallationRepositoriesAppEvent(ctx, rawWBPayload)
 		default:
 			l.Info().Msgf("webhook event %s not handled", wes.Typ)
 		}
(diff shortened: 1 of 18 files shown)


internal/reconcilers/entity_add.go
```diff
@@ -0,0 +1,79 @@
...
+
+	repoID, _, _ := inf.GetEntityDBIDs()
+
+	// Telemetry logging
+	logger.BusinessRecord(ctx).ProviderID = inf.ProviderID
+	logger.BusinessRecord(ctx).Project = inf.ProjectID
+	switch inf.Type {
+	case pb.Entity_ENTITY_REPOSITORIES:
+		l.Info().Str("repo_id", repoID.UUID.String()).Msg("handling entity add event")
+		// Upsert entry in the DB.
```
jhrozek, line 52: the rest of the WIP looks good, but why call the store directly here and not the `RepositoryService.CreateRepository` method? (I know the code is commented out, I'm mostly pointing that method out to make sure you know about it)

blkt: Sorry for missing the comment. I was already intentioned to switch to `RepositoryService` but needed to figure out some other stuff before doing that, namely I have to get a `db.Provider` object as well.

This is a bit annoying, as the sole purpose of passing a `db.Provider` to `CreateRepository` is to get its `ID` and its `Name`. I believe this could be made a bit simpler by just passing those two parameters.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -904,12 +978,145 @@ func (_ *Server) processInstallationAppEvent(
...
+			return nil, err
+		}
+
+		results = append(results, res)
+	}
+
+	// We might want to ignore this case since we can only delete
+	// repositories there were previously registered, which would
+	// be deleted by means of "meta" and "repository" events as
+	// well.
```
blkt, line 1047: This is interesting: we run into a race condition if, for a given repository that was removed from the org, we get a `meta` before the `installation_repositories` event, or vice versa. The first deletion succeeds, while the second fails.

```
error deleting repository from DB: error retrieving repository: sql: no rows in result set
```

This is not a problem right now, as it is handled correctly by logging some errors, but we do retry the deletion a handful of times, which might cause issues at scale.

internal/reconcilers/messages/messages.go
```diff
@@ -0,0 +1,50 @@
...
+
+	evtStr, err := json.Marshal(evt)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling init event: %w", err)
+	}
+
+	msg := message.NewMessage(uuid.New().String(), evtStr)
+	msg.Metadata.Set("provider_id", providerID.String())
+	return msg, nil
+}
```
blkt, line 50: This was moved here to solve a circular dependency between `internal/reconcilers/reconcilers.go` and `internal/service/service.go`

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -904,12 +978,145 @@ func (_ *Server) processInstallationAppEvent(
...
+	if event.GetInstallation() == nil {
+		return nil, errors.New("invalid event: installation is nil")
+	}
+	if event.GetInstallation().GetID() == 0 {
+		return nil, errors.New("invalid installation: id is 0")
+	}
+
+	installationID := event.GetInstallation().GetID()
+	installation, err := s.store.GetInstallationIDByAppID(ctx, installationID)
+	if errors.Is(err, sql.ErrNoRows) {
```
jhrozek, line 1013: Do we special-case this error somewhere?
I also wonder, does an unclaimed installation have an installation ID? (I think yes, but worth testing)

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -904,12 +978,145 @@ func (_ *Server) processInstallationAppEvent(
...
+		return nil, fmt.Errorf("no installation found for id %d", installationID)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("could not determine provider id: %v", err)
```
jhrozek, line 1017: the error message seems copy-pasted

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -904,12 +978,145 @@ func (_ *Server) processInstallationAppEvent(
...
+	}
+	if !installation.ProviderID.Valid {
+		return nil, errors.New("invalid provider id")
+	}
+	if !installation.ProjectID.Valid {
+		return nil, errors.New("invalid project id")
+	}
```
jhrozek, line 1024: we should also test how these behave for an unclaimed install - I /think/ it should be OK as we create this special project (as we saw during pair-session on Friday), but worth testing.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -904,12 +978,145 @@ func (_ *Server) processInstallationAppEvent(
...
+
+	results := make([]*processingResult, 0, len(event.GetRepositoriesAdded())+len(event.GetRepositoriesRemoved()))
+	for _, repo := range event.GetRepositoriesAdded() {
+		// caveat: we're accessing the database once for every
+		// repository, which might be inefficient at scale.
+		res, err := s.repositoryAdded(
+			ctx,
+			repo,
+			event.GetAction(),
+			events.TopicQueueReconcileEntityAdd,
```
jhrozek, line 1034: can the topic ever be something else (asking if it needs to be a parameter)

internal/controlplane/handlers_githubwebhooks.go
jhrozek: is the `meta` event for the webhook/installation and the `installation_repositories` for the repos only? Could we process one or the other? I think in either case they could be solved by a single db lookup, no? (if we are to add and the repo is already there, then skip, if we are to delete and the repo is already deleted, skip).

Either way, nice that you caught this issue already.

jhrozek (commented):
Nice work. Left some comments inline, but overall this PR is shaping quite nicely!

internal/controlplane/handlers_githubwebhooks.go
blkt: The slight difference was meant to figure out in which case we end up by just looking at logs, as go does not report line numbers when logging.

internal/controlplane/handlers_githubwebhooks.go
blkt: I'm not following, what is an unclaimed installation?

internal/controlplane/handlers_githubwebhooks.go
blkt: You're right, this can only be a `events.TopicQueueReconcileEntityAdd`.
Initially the routine was just one, this is a leftover.

internal/controlplane/handlers_githubwebhooks.go
blkt: I have to test what happens if we revoke access to a repository _without_ deleting it from GitHub.

I believe that we don't get the `"meta"` event in such a case, i.e. we have to handle `"removed"` action for `"installation_repositories"` events as well.

I'll test it manually and reply here.

internal/controlplane/handlers_repositories_test.go
```diff
@@ -37,6 +37,7 @@ import (
 	mockmanager "github.com/stacklok/minder/internal/providers/manager/mock"
 	ghrepo "github.com/stacklok/minder/internal/repositories/github"
 	mockghrepo "github.com/stacklok/minder/internal/repositories/github/mock"
+	rf "github.com/stacklok/minder/internal/repositories/github/mock/fixtures"
```
blkt, line 40: The diff for this file should just be refactoring, I moved helper routines to this package.

internal/controlplane/handlers_githubwebhooks_test.go
```diff
@@ -3512,12 +3518,168 @@ func (s *UnitTestSuite) TestHandleGitHubAppWebHook() {
...
 				},
 			},
-			mockStoreFunc: newMockStore(),
+			mockStoreFunc: df.NewMockStore(),
 			topic:         installations.ProviderInstallationTopic,
 			statusCode:    http.StatusOK,
 			queued:        nil,
 		},
 
+		// installation repositories events
```
blkt, line 3527: Most of the diff for this file is refactoring, relevant tests are here.

blkt (commented):
Here we go! 🚀

internal/controlplane/handlers_githubwebhooks.go
jhrozek: The one you create without logging to minder first, e.g by going to `https://github.com/apps/minder-instance/installations/new`

internal/controlplane/handlers_githubwebhooks.go
jhrozek: thanks

jhrozek (approved):
This is great. I kicked the coverage tests because they seem to be failing while the "normal" unit test run succeeded, so it might be a fluke.

Issue #3266, opened 2024-05-07
Extend the GitHub App configuration with an autoRegistration object


jhrozek:
To enable configuring auto-registration of entities (repositories for a start) we need to extend the configuration of the GitHub App provider which is currently quite bare-bones:
```
message GitHubAppProviderConfig {
 // Endpoint is the GitHub API endpoint. If using the public GitHub API, Endpoint can be left blank.
 string endpoint = 1;
}

// and then in Go code
func ParseV1Config(rawCfg json.RawMessage) (*minderv1.GitHubAppProviderConfig, error) {
```

to let the user configure the auto registration configuration:
```
message AutoRegistration {
	 // There is no difference between intentionally unset and undefined so for the "let's not auto-register anything" case we'd just let the repeated Entity empty
	repeated Entity enabled = 1;
}

message GitHubAppProviderConfig {
 // Endpoint is the GitHub API endpoint. If using the public GitHub API, Endpoint can be left blank.
 string endpoint = 1;
}

message ProviderConfig {
 	AutoRegistration auto_registration = 1;
	google.protobuf.Struct config = 2;
}

func ParseV1Config(rawCfg json.RawMessage) (*minderv1.ProviderConfig, error) {
type wrapper struct {
   GitHub *minderv1.GitHubProviderConfig `json:"github" yaml:"github" mapstructure:"github" validate:"required"`
}
```

PR #3449, opened 2024-05-28
Extend the GitHub App configuration with an autoRegistration object

jhrozek:
- **Add the protobuf struct for generic provider options**
- **Parse out ProviderConfig as well when reading GH App configuration**

# Summary

Adds the protobuf struct for generic provider options

In order to actually set the auto-registration of entities, we need to
be able to configure those options.

This patch adds two structures: AutoRegistration which contains a list
of entities to auto-register and ProviderConfig which contains the
AutoRegistration.

The idea behind this is that any provider would be able to use the
ProviderConfig attributes regardless of the provider type. So as an
example, both would be valid:

```
{ "auto_registration": { "enabled": ["repository"] }, "github": { "endpoint": "https://api.github.com" }}
```
and:
```
{ "auto_registration": { "enabled": ["repository"] }, "github-app": { "endpoint": "https://api.github.com" }}
```

and all would require just embedding the ProviderConfig structure into
the provider-specific parser.

Fixes: #3266

Parse out ProviderConfig as well when reading GH App configuration

Amends the ParseV1AppConfig to also include parsing the
minderv1.GitHubAppProviderConfig structure.

The parsed-out object is not used yet, but it's parsed and validated and the
new code is tested.

Related: #3266

## Change Type

# Testing

make test for now

# Review Checklist:


internal/providers/github/clients/app_test.go
```diff
@@ -39,6 +39,78 @@ import (
...
+		{
+			name:   "valid app config",
+			config: json.RawMessage(`{ "github-app": { "endpoint": "https://api.github.com" } }`),
+			ghEvalFn: func(ghConfig *minderv1.GitHubAppProviderConfig) {
+				assert.Equal(t, "https://api.github.com", ghConfig.Endpoint)
+			},
+			provEvalFn: func(providerConfig *minderv1.ProviderConfig) {
+				assert.Nil(t, providerConfig)
+			},
+		},
```
blkt, line 61: I don't know if this was intended, but these functions capture `t` from the outer scope, making them report failures on `TestParseV1AppConfig` rather than e.g. `TestParseV1AppConfig/valid_app_and_provider_config`.

In case this was not the intended behaviour, they would have to receive it as an argument.

```suggestion
	scenarios := []struct {
		name       string
		config     json.RawMessage
		error      string
		ghEvalFn   func(*testing.T, *minderv1.GitHubAppProviderConfig)
		provEvalFn func(*testing.T, *minderv1.ProviderConfig)
	}{
		{
			name:   "valid app config",
			config: json.RawMessage(`{ "github-app": { "endpoint": "https://api.github.com" } }`),
			ghEvalFn: func(t *testing.T, ghConfig *minderv1.GitHubAppProviderConfig) {
				assert.Equal(t, "https://api.github.com", ghConfig.Endpoint)
			},
			provEvalFn: func(t *testing.T, providerConfig *minderv1.ProviderConfig) {
				assert.Nil(t, providerConfig)
			},
		},
```

blkt (approved):

internal/providers/github/clients/app_test.go
jhrozek: not intended, thank you. that was a late-evening refactoring gone wrong.

jhrozek:
@blkt I accepted your suggestion, can you re-review?

blkt (approved):

proto/minder/v1/minder.proto
```diff
@@ -1289,6 +1289,20 @@ enum Entity {
...
+message AutoRegistration {
+    // enabled is the list of entities that are enabled for auto-registration.
+    repeated string enabled = 1;
+}
+
+// ProviderConfig contains the generic configuration for a provider.
+message ProviderConfig {
+    // auto_registration is the configuration for auto-registering entities.
+    AutoRegistration auto_registration = 1;
+}
```
evankanderson, line 1304: I feel like we need some sort of documentation about how these messages line up (or don't) with other fields in the API.  I'm assuming these are serialized as Struct / JSON in the `config` field of `Provider`?

evankanderson (commented):
One minor note, sorry I didn't get to this earlier.

I'm assuming that some portion of the parameters like `GitHubProviderConfig`, `GitHubAppProviderConfig`, `DockerHubProviderConfig`, etc are not fully hooked up at the moment, but we should think about what the documentation looks like once they are hooked up -- I'm guessing that we'd want some sort of page which documents the supported fields for e.g. the GitHubApp provider.

proto/minder/v1/minder.proto
jhrozek: Yes, this is the next part I'm working on at the moment, I hope to get a PR out soon.

jhrozek:
> One minor note, sorry I didn't get to this earlier.
> 
> I'm assuming that some portion of the parameters like `GitHubProviderConfig`, `GitHubAppProviderConfig`, `DockerHubProviderConfig`, etc are not fully hooked up at the moment, but we should think about what the documentation looks like once they are hooked up -- I'm guessing that we'd want some sort of page which documents the supported fields for e.g. the GitHubApp provider.

Thanks for going back and taking a look! I filed https://github.com/stacklok/minder/issues/3458 to ensure we don't forget about documentation. Hopefully how the configuration is updated will be more clear once I send the next PR.

PR #3451, opened 2024-05-29
User API: Add missing project display name and description

### LINKED ISSUE #3178: Users endpoint does not return `displayName` for projects
### Describe the issue

I noticed that in the /api/v1/user endpoint, the projects are returned with the same type as they are returned elsewhere (in the generated sdk it's called v1Project ), but the returned value here seems to be missing the displayName attribute. We actually want to display this value to the user. 

### To Reproduce

Should be enough to call the endpoint and observe the output. The user will need to have at least one project of course.

_No response_

puerco:
# Summary

This PR populates the missing descriptions and display names of projects included in the user API.

Fixes #3178

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_user.go
+++ b/internal/controlplane/handlers_user.go
@@ -199,11 +199,22 @@ func (s *Server) getUserDependencies(ctx context.Context, user db.User) ([]*pb.P
 			return nil, err
 		}
 
+		// Try to parse the project metadata to complete the response fields
+		pDisplay := pinfo.Name
+		pDescr := ""
+		meta, err := projects.ParseMetadata(&pinfo)
+		if err == nil {
+			pDisplay = meta.Public.DisplayName
+			pDescr = meta.Public.Description
+		}
+
 		projectsPB = append(projectsPB, &pb.Project{
-			ProjectId: proj.String(),
-			Name:      pinfo.Name,
-			CreatedAt: timestamppb.New(pinfo.CreatedAt),
-			UpdatedAt: timestamppb.New(pinfo.UpdatedAt),
+			ProjectId:   proj.String(),
+			Name:        pinfo.Name,
+			CreatedAt:   timestamppb.New(pinfo.CreatedAt),
+			UpdatedAt:   timestamppb.New(pinfo.UpdatedAt),
+			DisplayName: pDisplay,
+			Description: pDescr,
 		})
 	}
 


internal/controlplane/handlers_user.go
```diff
@@ -199,11 +200,21 @@ func (s *Server) getUserDependencies(ctx context.Context, user db.User) ([]*pb.P
 			return nil, err
 		}
 
+		// Try to parse the project metadata to complete the response fields
+		meta := &projects.Metadata{}
```
eleftherias, line 204: You could use `projects.ParseMetadata` here, instead of unmarshalling it yourself https://github.com/stacklok/minder/blob/main/internal/projects/meta.go#L66

puerco: Thanks for the tip @eleftherias ! Done :+1:

internal/controlplane/handlers_user.go
```diff
@@ -199,11 +199,21 @@ func (s *Server) getUserDependencies(ctx context.Context, user db.User) ([]*pb.P
...
-			ProjectId: proj.String(),
-			Name:      pinfo.Name,
-			CreatedAt: timestamppb.New(pinfo.CreatedAt),
-			UpdatedAt: timestamppb.New(pinfo.UpdatedAt),
+			ProjectId:   proj.String(),
+			Name:        pinfo.Name,
+			CreatedAt:   timestamppb.New(pinfo.CreatedAt),
+			UpdatedAt:   timestamppb.New(pinfo.UpdatedAt),
+			DisplayName: pDisplay,
+			Description: pDescr,
```
rdimitrov, line 216: @kantord - hey, in case the parsing above fails for some reason is it okay with you to have the default variable values here, i.e. empty strings?

kantord: you mean specifically for description? I think for description, empty string is a reasonable fallback value, but for displayName I think that using the name as a fallback value would make more sense

kantord: also, I'm not very informed about how the BE codebase works, but as I understand the fallback logic happens at insertion time, meaning this code itself does not address existing data in the database. I guess those could be fixed using a database migration or a manual query?

evankanderson: I'm pretty sure this code is happening on the way _out_ of the database when answering the query, not on insertion.  (That's both what I would expect, and matches the pattern in the code where we're constructing a protocol buffer for output.)  If we change 203 to:

```golang
pDisplay := pinfo.Name
pDescr := ""
```

before calling `ParseMetadata`, that should be sufficient to fill in the defaults when the metadata doesn't exist in the database, without needing a backfill.

puerco: Thanks for the suggestion @evankanderson , I've changed the initial values as suggested :)

rdimitrov (approved):

PR #3459, opened 2024-05-30
Define rule instance table for Minder

dmjb:
Rule instances are currently stored as serialized protobuf structures in the database, grouped by entity type. A separate join table exists to track the relationship between a profile-entity and the rule types which are used. This arrangement leads to some rather awkward code and database queries making the profile-related code in Minder more complex than it should be.

This PR adds in a definition of a dedicated rule instance table. Subsequent PRs will migrate data over to it and then change the codebase to query this table instead of the current tables.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/database/migrations/000060_rule_instances.up.sql
+++ b/database/migrations/000060_rule_instances.up.sql
@@ -28,6 +28,11 @@ CREATE TABLE IF NOT EXISTS rule_instances(
     UNIQUE (profile_id, entity_type, name)
 );
 
+-- this reflects likely access patterns for this table - retrieving all rule
+-- instances for a given profile, and all rules by profile and entity type
+CREATE INDEX idx_rule_instances_profile ON rule_instances(profile_id);
+CREATE INDEX idx_rule_instances_profile_entity ON rule_instances(profile_id, entity_type);
+
 -- this will be used for migration purposes
 ALTER TABLE entity_profiles ADD COLUMN migrated BOOL DEFAULT FALSE NOT NULL;
 
--- a/database/migrations/000060_rule_instances.up.sql
+++ b/database/migrations/000060_rule_instances.up.sql
@@ -24,8 +24,7 @@ CREATE TABLE IF NOT EXISTS rule_instances(
     params       JSONB NOT NULL,
     created_at   TIMESTAMP NOT NULL DEFAULT NOW(),
     updated_at   TIMESTAMP NOT NULL DEFAULT NOW(),
-    -- equivalent to constraint on entity_profile_rules
-    UNIQUE (profile_id, entity_type, rule_type_id),
+    -- equivalent to constraints enforced by rule validation
     UNIQUE (profile_id, entity_type, name)
 );
 
--- a/database/migrations/000060_rule_instances.up.sql
+++ b/database/migrations/000060_rule_instances.up.sql
@@ -1,0 +1,35 @@
+-- Copyright 2024 Stacklok, Inc
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS rule_instances(
+    id           UUID NOT NULL DEFAULT gen_random_uuid() PRIMARY KEY,
+    profile_id   UUID NOT NULL REFERENCES profiles(id) ON DELETE CASCADE,
+    rule_type_id UUID NOT NULL REFERENCES rule_type(id),
+    name         TEXT NOT NULL,
+    entity_type  entities NOT NULL,
+    def          JSONB NOT NULL,
+    params       JSONB NOT NULL,
+    created_at   TIMESTAMP NOT NULL DEFAULT NOW(),
+    updated_at   TIMESTAMP NOT NULL DEFAULT NOW(),
+    -- equivalent to constraint on entity_profile_rules
+    UNIQUE (profile_id, entity_type, rule_type_id),
+    UNIQUE (profile_id, entity_type, name)
+);
+
+-- this will be used for migration purposes
+ALTER TABLE entity_profiles ADD COLUMN migrated BOOL DEFAULT FALSE NOT NULL;
+
+COMMIT;
--- a/database/migrations/000060_rule_instances.up.sql
+++ b/database/migrations/000060_rule_instances.up.sql
@@ -20,8 +20,8 @@ CREATE TABLE IF NOT EXISTS rule_instances(
     rule_type_id UUID NOT NULL REFERENCES rule_type(id),
     name         TEXT NOT NULL,
     entity_type  entities NOT NULL,
-    def          JSONB NOT NULL,
-    params       JSONB NOT NULL,
+    def          JSONB, -- stores the values needed by the rule type's `def`
+    params       JSONB, -- stores the values needed by the rule type's `params`
     created_at   TIMESTAMP NOT NULL DEFAULT NOW(),
     updated_at   TIMESTAMP NOT NULL DEFAULT NOW(),
     -- equivalent to constraints enforced by rule validation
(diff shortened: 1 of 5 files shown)


database/migrations/000060_rule_instances.up.sql
```diff
@@ -0,0 +1,35 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS rule_instances(
```
dmjb, line 17: This table will need some indices defined. That will show up in a future PR.

database/query/profiles.sql
```diff
@@ -22,15 +22,25 @@ WHERE id = $1 AND project_id = $2 RETURNING *;
...
     entity,
     profile_id,
-    contextual_rules) VALUES ($1, $2, sqlc.arg(contextual_rules)::jsonb) RETURNING *;
+    contextual_rules,
+    migrated
+) VALUES (
+    $1,
+    $2,
+    sqlc.arg(contextual_rules)::jsonb,
+    FALSE
```
dmjb, line 31: This is redundant given that the default value is FALSE for this column, but I want to set it explicitly to remind myself to change it in the next PR.

database/migrations/000060_rule_instances.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    profile_id   UUID NOT NULL REFERENCES profiles(id) ON DELETE CASCADE,
+    rule_type_id UUID NOT NULL REFERENCES rule_type(id),
+    name         TEXT NOT NULL,
+    entity_type  entities NOT NULL,
+    def          JSONB NOT NULL,
+    params       JSONB NOT NULL,
+    created_at   TIMESTAMP NOT NULL DEFAULT NOW(),
+    updated_at   TIMESTAMP NOT NULL DEFAULT NOW(),
+    -- equivalent to constraints enforced by rule validation
+    UNIQUE (profile_id, entity_type, name)
```
evankanderson, line 28: Ah - a rule can apply to more than one entity type, and we allow the same name if it's applying to a different type of entity?  (Is that what we want?)

database/migrations/000060_rule_instances.up.sql
```diff
@@ -0,0 +1,39 @@
...
+);
+
+-- this reflects likely access patterns for this table - retrieving all rule
+-- instances for a given profile, and all rules by profile and entity type
+CREATE INDEX idx_rule_instances_profile ON rule_instances(profile_id);
+CREATE INDEX idx_rule_instances_profile_entity ON rule_instances(profile_id, entity_type);
+
+-- this will be used for migration purposes
+ALTER TABLE entity_profiles ADD COLUMN migrated BOOL DEFAULT FALSE NOT NULL;
```
evankanderson, line 37: Ah, so we're going to use this to mark whether to look in `rule_instances` or in the column data for `profiles` for the rule definitions?  SGTM, that will allow us to migrate without too many dual-writes.

evankanderson (dismissed):

database/migrations/000060_rule_instances.up.sql
```diff
@@ -0,0 +1,39 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS rule_instances(
```
JAORMX, line 17: Why not reuse `entity_profile_rules` that already contains the link between an `entity_profile` (which links to a `profile`) and a `rule_type_id`

pkg/api/protobuf/go/minder/v1/minder.pb.go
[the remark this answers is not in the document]
dmjb: I think so, this has been moved to another branch.

database/migrations/000060_rule_instances.up.sql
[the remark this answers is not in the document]
dmjb: "Rule instances" is the term I've heard used most often within the team, so I think I'll stick to that for now.

dmjb: The issue is that we have a whole bunch of complexity throughout the code which results from storing each rule instance in a JSONB array instead of having a table of the individual rule instances.

database/migrations/000060_rule_instances.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    id           UUID NOT NULL DEFAULT gen_random_uuid() PRIMARY KEY,
+    profile_id   UUID NOT NULL REFERENCES profiles(id) ON DELETE CASCADE,
+    rule_type_id UUID NOT NULL REFERENCES rule_type(id),
+    name         TEXT NOT NULL,
```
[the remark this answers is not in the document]
dmjb: It is case sensitive in the existing code, and so I preserved that behaviour. I think it will make the migration harder if we introduce that change up front.

database/migrations/000060_rule_instances.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    entity_type  entities NOT NULL,
+    def          JSONB NOT NULL,
+    params       JSONB NOT NULL,
```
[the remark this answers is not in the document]
dmjb: Yeah, they are parameters to the def and params blocks in a rule type... will add a comment to that effect.

database/migrations/000060_rule_instances.up.sql
JAORMX: Yes, the JSON storage issue is understandable. However, that is not the table I'm talking about. We already have a table that is meant for rule instances, where we store the link between profiles and rule types. And that is `entity_profie_rules`. we could leverage that.

database/migrations/000060_rule_instances.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    created_at   TIMESTAMP NOT NULL DEFAULT NOW(),
+    updated_at   TIMESTAMP NOT NULL DEFAULT NOW(),
```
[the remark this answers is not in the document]
dmjb: We currently store created/updated at in the entity_profiles table, so I decided to carry it forward here.

When someone PUTs a profile, the rule instances in the updated profile are upserted, and the updated time is updated on conflict... that behaviour will be carried forward in the queries for this table.

database/migrations/000060_rule_instances.up.sql
dmjb: The combination of entity type + rule name is unique for a given profile - you can reuse the same rule name across different entity type blocks in the YAML. This reflects the existing behaviour/validation rules for rule names.

database/migrations/000060_rule_instances.up.sql
JAORMX: @dmjb this is the table I'm talking about https://github.com/stacklok/minder/blob/main/internal/db/models.go#L467

For each profile in an entity's context we store the instantiation of a rule type. This is done to prevent deleting rule types when they're instantiated by profile. Instead of creating a new table, why not use what we already have and enhance it with the `def` and `params` information?

dmjb: A profile may contain more than one instance of a rule type. In the new table, (profile_id, entity_type, rule_type_id) is not unique - I had tried to create this constraint first and then discovered during testing that this is an incorrect assumption.

However, in entity_profile_rules, (profile_id, rule_type_id) is unique. That implies that N rows in rule_instances match to one row in entity_profile_rules. Since I want to dual write to my new structure and the old structure until we migrate over, I don't think it's a good idea to change the uniqueness constraint on entity_profile_rules so that we can reuse that table.

database/migrations/000060_rule_instances.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    -- equivalent to constraints enforced by rule validation
```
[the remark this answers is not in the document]
dmjb: I believe there is at least one query in my next PR which queries by these three columns. I also need it for an upsert query.

database/migrations/000060_rule_instances.up.sql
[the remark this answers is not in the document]
evankanderson: My question here was:

Are there cases where the `created_at` or `updated_at` fields on different rule_instances with the same `profile_id` will have different values?

It sounds like the answer is "yes" if I start with a Profile with 2 rules, then add a third rule later.

evankanderson (approved):

PR #3463, opened 2024-05-30
Events of type "repository" now avoid checking hook id.

### LINKED ISSUE #3461: GitHub webhook events of type "repository" action "deleted" are not handled properly
### Describe the issue

Previous code handling repository deletion went through [this](https://github.com/stacklok/minder/blob/04db03c6784f80a3d47976e5dd443e3a8da4f03e/internal/controlplane/handlers_githubwebhooks.go#L531) line which evaluated to `false` and thus did not care about the hook id.

In its [current form](https://github.com/stacklok/minder/blob/main/internal/controlplane/handlers_githubwebhooks.go#L684), the check on the hook id is always performed (because we share the struct used to deserialize the JSON payload), while it does not make sense for events of type `repository`, as they do not contain any information about the hook (see [here](https://docs.github.com/en/webhooks/webhook-events-and-payloads#repository)).

Ensure that [this check](https://github.com/stacklok/minder/blob/main/internal/controlplane/handlers_githubwebhooks.go#L684) is only performed on events of type `meta`.

### To Reproduce

1. register a repo
2. delete repo from github
3. this log line should occur
```
webhook event not handled (webhook event not handled: meta event with action deleted not handled, hook ID 0 does not match stored webhook ID 481299324)
```

_No response_

blkt:
# Summary

Events of type "repository" having "action" delete must trigger the deletion of registered repositories. Given we use the same deserialization struct for both "repository" and "meta" events, we erroneously checked the received hook id against the one stored in the database, thus consistently failing.

Added fix, test case, and improved logs.

Fixes #3461 

## Change Type

# Testing

* reproduced issue locally
* added test case

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_githubwebhooks.go
+++ b/internal/controlplane/handlers_githubwebhooks.go
@@ -345,6 +345,9 @@ func (s *Server) HandleGitHubAppWebhook() http.HandlerFunc {
 			Str("webhook-event-type", m.Metadata[events.GithubWebhookEventTypeKey]).
 			Str("providertype", m.Metadata[events.ProviderTypeKey]).
 			Str("upstream-delivery-id", m.Metadata[events.ProviderDeliveryIdKey]).
+			// This is added for consistency with logs in
+			// Entity Execution Aggregator
+			Str("event", m.UUID).
 			Logger()
 		ctx = l.WithContext(ctx)
 
@@ -380,14 +383,19 @@ func (s *Server) HandleGitHubAppWebhook() http.HandlerFunc {
 			l.Info().Str("message-id", m.UUID).Msg("publishing event for execution")
 			if err := res.iiw.ToMessage(m); err != nil {
 				wes.Error = true
-				log.Printf("Error creating event: %v", err)
+				l.Error().Err(err).Msg("Error creating event")
 				w.WriteHeader(http.StatusInternalServerError)
 				return
 			}
 
+			// This ensures that loggers on downstream
+			// processors have all log attributes
+			// available.
+			m.SetContext(ctx)
+
 			if err := s.evt.Publish(res.topic, m); err != nil {
 				wes.Error = true
-				log.Printf("Error publishing message: %v", err)
+				l.Error().Err(err).Msg("Error publishing message")
 				w.WriteHeader(http.StatusInternalServerError)
 				return
 			}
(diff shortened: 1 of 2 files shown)


internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -681,7 +697,7 @@ func (s *Server) processRelevantRepositoryEvent(
 	if dbrepo.WebhookID.Valid {
 		// Check if the payload webhook ID matches the one we
 		// have stored in the DB for this repository
-		if event.GetHookID() != dbrepo.WebhookID.Int64 {
+		if event.GetHookID() != 0 && event.GetHookID() != dbrepo.WebhookID.Int64 {
```
eleftherias, line 700: question: Does this fix work because the hook ID for repository events is always 0?

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -681,7 +697,7 @@ func (s *Server) processRelevantRepositoryEvent(
 	if dbrepo.WebhookID.Valid {
 		// Check if the payload webhook ID matches the one we
```
eleftherias, line 698: question: Is the check for `event.GetHookID() != 0` better suited to be in this `if` statement?

blkt: You're right, I'll move it up.

internal/controlplane/handlers_githubwebhooks.go
blkt: Yes, that's exactly the point.
The issue was not there before because we dynamically looked up `hook_id` using `jq`, while now we're using a shared struct to deserialize the messages and this makes it always get the default value.

Simply checking for it being some value other than `0` is sufficient and cannot cause bugs because it is not a valid value, and `go-github` handles deserialization in the same way as we do here, so we should be on the safe side.

evankanderson (approved):

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -678,7 +694,7 @@ func (s *Server) processRelevantRepositoryEvent(
 	}
 
 	// This only makes sense for "meta" event type
-	if dbrepo.WebhookID.Valid {
+	if event.GetHookID() != 0 && dbrepo.WebhookID.Valid {
 		// Check if the payload webhook ID matches the one we
 		// have stored in the DB for this repository
 		if event.GetHookID() != dbrepo.WebhookID.Int64 {
```
evankanderson, line 700: Does the comment on line 701-703 still make sense?

evankanderson (commented):
Oops, one more comment

internal/controlplane/handlers_githubwebhooks.go
blkt: I think it does.

This routine is called only on `meta` and `repository` events, and this fix ensures that we check the hook id only when available, which is the case solely for the `meta` event.

The comment per sé is still valid, but can be removed if it's not useful.
I lean towards leaving it there for posterity, maybe rewriting it at the top of the `if`, but let me know.

eleftherias (approved):

internal/controlplane/handlers_githubwebhooks.go
eleftherias: I agree that it's valid and useful

Issue #3315, opened 2024-05-13
Migrate encrypted data in database to new encryption structure


dmjb:
### Please describe the enhancement

See design doc in epic

### Solution Proposal

In migration tool:

1. Change decrypt code to read from new fields then old fields
1. Begin writing new secrets to new structure only
1. Get each secret
1. Decrypt
1. Re-salt and re-encrypt using new structure
1. Validate that migration has worked and drop old column

PR #3467, opened 2024-05-30
Remove algorithm from crypto config

dmjb:
Relates to #3315

Previously, it was possible to set a default and fallback algorithm in the crypto config. This seemed like a good idea at the time but I came to realize the following:

1) Unlike the key, the algorithm is not an external piece of data - it
   corresponds to something implemented in the codebase. We cannot
   introduce new algorithms without changing the code.
2) After pairing on a key rotation with Ozz, I realized that the
   relationship between the key ID and the algorithm in the config is
   not really intuitive, and is likely to cause confusion which could
   result in us breaking our ability to decrypt secrets.

This PR removes the algorithm from the config and defines the default vs fallback in the code. As part of this change, the new GCM algorithm becomes the default for everyone, and this required changing some tests to use real encryption keys (the old CFB algorithm would accept arbitrary strings as keys, the GCM cipher expects a base64-encoded 32-byte key).

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


jhrozek (approved):
I think not letting users set the algorithms is a good idea in general.

JAORMX (approved):

Issue #3265, opened 2024-05-07
Add a mechanism to patch provider configuration post enrollment


jhrozek:
This is a follow-up to issue #3263. In addition to storing the provider configuration on provider creation we'll also want to patch the provider configuration after the fact.  We might want to implement something like:

```
message ProviderPatch {
	// to allow generic configuration changes. 
	google.protobuf.Struct config = 1;
}

message PatchProviderRequest {
	Context context = 1;
	ProviderPatch = 2;

	google.protobuf.FieldMask update_mask = 3;
}

message PatchProviderResponse {
	Provider provider = 1;
}

rpc PatchProvider(PatchProviderRequest) returns (PatchProviderResponse) {
	option (google.api.http) {
		patch: "/api/v1/provider"
		body: "patch"
	}
	option (rpc_options) = {
		target_resource = TARGET_RESOURCE_PROJECT;
		relation = RELATION_PROVIDER_UPDATE;
	}
}
```

and expose that through the CLI. To implement the patch handler we'll need to:
- Extend the ProviderManager interface with a new PatchProvider method, e.g PatchProvider(ctx context.Context, uuid uuid.UUID, patch minderv1.ProviderPatch) (minderv1.Provider, error)
- Extend the ProviderClassManager interface with a new Patch method
- The  ProviderManager.PatchProvider would get the correct class manager and call its Patch
- The per-provider Patch implementation is where the generic protobuf struct is unmarshalled into the correct Go struct and the configuration adjusted

jhrozek:
Co-assigning @blkt who will be working on this issue to get it out of draft and merged as I'm away tomorrow and Friday.

PR #3468, opened 2024-05-30
Updating provider config

jhrozek:
# Summary

- **Protobuf changes for providers update** - just adds the protobufs, nothing else

- **Extend the ProviderStore with an Update method, so far only allows updating config** - we used to have a database-level Update, but not exposed through the provider store interface

- **Add a PatchProvider server handler, currently just updating the config** - this is the main part of the PR. Adds a new method PatchProviderConfig to the `ProviderService` interface.
This method uses the mergo library to merge the provided config with the currently stored config and then update the provider record in the DB.

- **Treat the provider config as an override, keep the defaults in the code** -   Since storing the whole config in the database is problematic for a number of reasons, like hard patching or worse ability to change defaults, let's change the way we treat the provider configuration in the database. Instead of storing the full config, we treat the provider config as an override. To that end, we store a structure with defaults in the code and make all the fields in the protobuf messages optional which causes the resulting Go structures to have pointers to fields instead of direct attributes.

We store the config override verbatim, after just having unmarshalled and marshalled back to get rid of any extra keys. Then we merge with the default config on retrieving the provider configuration from the database.

Fixes: #3265 
Fixes: #3511 

## Change Type

# Testing

```
curl -XPATCH -H "Authorization: Bearer $MINDER_BEARER_TOKEN" 'http://localhost:8080/api/v1/providers?context.project=d2f78a70-96eb-45fa-868d-d48391508e4d&context.provider=github-app-jakubtestorg' -d '{ "config": { "auto_registration": { "entities": { "repository": {"enabled": true} } }}}'
```

# Review Checklist:


jhrozek:
@kd @peppescg would you mind adding your opinion on the API?

@evankanderson you have the most experience with protobuf, could you check the patches to the proto file and the mergo usage? Is there a nicer way?

Thank you all for taking a look.

proto/minder/v1/minder.proto
```diff
@@ -2098,6 +2110,22 @@ message ListProviderClassesResponse {
...
+message ProviderPatch {
+    // to allow generic configuration changes.
+    google.protobuf.Struct config = 1;
+}
+
+message PatchProviderRequest {
+    Context context = 1;
+    ProviderPatch patch = 2;
+
+    google.protobuf.FieldMask update_mask = 3;
```
evankanderson, line 2122: I don't recall exactly how good the library support is for `update_mask`.  It's real handy as a client, but make sure it won't get you into too much trouble in the implementation.

pkg/api/openapi/minder/v1/minder.swagger.json
```diff
@@ -1418,6 +1418,50 @@
...
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/v1PatchProviderResponse"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "patch",
```
evankanderson, line 1434: I'm assuming this name doesn't show up anywhere.

pkg/api/openapi/minder/v1/minder.swagger.json
```diff
@@ -1418,6 +1418,50 @@
...
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/v1ProviderPatch"
```
evankanderson, line 1438: This feels a little odd in comparison with `v1CreateProviderRequest`, which has fields:

* `context`
* `provider`
  * `name`
  * `class`
  * `project`
  * `version`
  * `implements`
  * `config`
  * `authFlows`
  * `parameters`
  * `credentialsState`

It feels like you might want to make the body of the message be a `Provider`, and then disallow updates to anything other than the `config` field at the moment.

internal/controlplane/handlers_providers.go
```diff
@@ -230,6 +230,45 @@ func (s *Server) DeleteProviderByID(
...
+) (*minderv1.PatchProviderResponse, error) {
+	entityCtx := engine.EntityFromContext(ctx)
+	projectID := entityCtx.Project.ID
+	providerName := entityCtx.Provider.Name
+
+	if providerName == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "provider name is required")
+	}
+
+	err := s.providerManager.PatchProviderConfig(ctx, providerName, projectID, req.GetPatch().GetConfig().AsMap())
```
evankanderson, line 246: Should this be getting a provider instance and then calling `PatchConfig` on the provider instance?  I'm okay with having this on providerManager, I just don't have a feel for when we create an instance and when we don't.

proto/minder/v1/minder.proto
jhrozek: I actually don't use it now in the implementation. We might want to reconsider later, but yeah, it seems like a lot of trouble for very questionable gain (Unless the FE devs disgree, going to ping them now for API review)

pkg/api/openapi/minder/v1/minder.swagger.json
jhrozek: It does not, e.g. one calls the endpoint as:
```
 curl -XPATCH -H "Authorization: Bearer $MINDER_BEARER_TOKEN" 'http://localhost:8080/api/v1/providers?context.project=bcf5bc68-aa81-4f59-96ae-7dceb5c3663f&context.provider=github-app-jakubtestorg' -d '{ "config": { "auto_registration": { "enabled": ["repository"] }} }'
```
in other words, the body already contains the provider attributes.

pkg/api/openapi/minder/v1/minder.swagger.json
jhrozek: thanks, I didn't want to expose all the provider attributes at first, but you're right that if we don't make use of them then it doesn't matter and forward compatibility is better.

internal/controlplane/handlers_providers.go
jhrozek: yeah, me neither. Getting a provider instance and then patching it is pretty much what the implementation of PatchProviderConfig does now. 

@dmjb do you have an opinion since you created the `ProviderManager` interface?

evankanderson:
(let me know if you want further reviews)

blkt (approved):

blkt (approved):

Issue #3485, opened 2024-05-31
Migrate rule instances to new table structure


dmjb:
### Please describe the enhancement

Rule instances are currently stored as a list of JSON objects which makes querying challenging. Migrate rules over to a dedicated table structure.

### Solution Proposal

1. Define new table
2. Write rule instances to old and new tables
3. Migrate old rule instances to new table
4. Change business logic to use new tables exclusively

evankanderson:
@dmjb -- did we do this?

evankanderson:
I think we did this.

PR #3486, opened 2024-05-31
Dual write rule instances to new and old tables

dmjb:
Relates to #3485

Any time a profile is created or updated, write the rule instances to both the new and old rule instance tables. Mark the entries in the old tables as migrated so that we skip over them when we run the migration process.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


database/query/profiles.sql
```diff
@@ -28,7 +28,7 @@ INSERT INTO entity_profiles (
     $1,
     $2,
     sqlc.arg(contextual_rules)::jsonb,
-    FALSE
+    TRUE
```
dmjb, line 31: At this point, any new or updated profiles will have rules in the new rule instance table, and thus can be considered migrated.

internal/profiles/service.go
```diff
@@ -707,3 +743,58 @@ func deleteRuleStatusesForProfile(
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to serialize rule params: %w", err)
+		}
+
+		newInstance := db.UpsertRuleInstanceParams{
+			ProfileID:  profileID,
+			RuleTypeID: entityRuleTuple.RuleID,
+			Name:       rule.Name,
+			EntityType: entityType,
+			Def: pqtype.NullRawMessage{
```
jhrozek, line 781: The PR looks good, the only question I have is how would we later instantiate the rule params and defs - right now the code seems to presume that they are non-nil:
```
result, err = r.ingester.Ingest(ctx, inf.Entity, params.GetRule().Params.AsMap())
```
so we'll have to check that we don't instantiate them as `nil` in this case.

jhrozek (approved):
Added one comment but overall looks good.

rdimitrov (approved):

database/migrations/000061_rule_instance_not_null_def.down.sql
```diff
@@ -0,0 +1,20 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+ALTER TABLE rule_instances ALTER COLUMN def DROP NOT NULL;
```
jhrozek, line 17: Since we haven't written anything to the columns, this is safe to do.

internal/profiles/service.go
```diff
@@ -773,22 +772,14 @@ func upsertRuleInstances(
...
-			},
-			Params: pqtype.NullRawMessage{
-				RawMessage: params,
-				Valid:      params != nil,
-			},
-		}
-
-		id, err := qtx.UpsertRuleInstance(ctx, newInstance)
+			Def:        def,
+			Params:     params,
```
jhrozek, line 781: I assume that if there's no params defined then json.Marshall returns `{}`? I wasn't sure about that since the previous code had an explicit `params != nil` check. But I guess we'd want to either store `{}` or make sure to on retrieval convert whatever value is stored to `{}` .

jhrozek (approved):

PR #3506, opened 2024-06-03
Add ContextV2 protobuf structure

JAORMX:
# Summary

This takes a `project_id` by default. It also provides the functionality
to prefer a `ContextV2` from a request if it's present and available.

## Change Type

# Testing

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -1401,6 +1401,15 @@ message Context {
     optional string retired_organization = 2;
 }
 
+// ContextV2 defines the context in which a rule is evaluated.
+message ContextV2 {
+    // project is the project ID
+    string project_id = 1;
+
+    // name of the provider
+    optional string provider = 2;
```
evankanderson, line 1410: Let's make this use the default behavior, rather than wrapping the string in a message so we can distinguish between `""` and field-not-present.

JAORMX: Can you clarify?

evankanderson: When you say `optional string` in proto, it's the same as:

```
message DistinguishBetweenNullAndEmpty {
  string foo = 1;
}
```

In general, proto3 recommends using the empty default value for "unset" -- we'd only need `optional` here if we wanted to support a Provider named `""`.  Otherwise, many tools (like grpc-gateway) will behave better if we just say `string provider = 2`.

evankanderson (approved):

PR #3552, opened 2024-06-06
Add validation of iss claim parameter

### LINKED ISSUE #3542: Validate `iss` in OpenID JWTs
### Describe the issue

Per https://openid.net/specs/openid-connect-basic-1_0.html#IDTokenValidation:

> The Issuer Identifier for the OpenID Provider (which is typically obtained during Discovery) MUST exactly match the value of the `iss` (issuer) Claim.

In this case, we're not implementing discovery, and using a server-configured JWKS URL, but we should still validate the `iss` URL.

Note that this will require changing the `issuerUrl` (which may currently be an internal URL, like `http://keycloak:8080` in local development) in configuration to match the externally-configured Keycloak issuer URl (`http://localhost:8081` in our local docker-compose configuration).

### To Reproduce

N/A - generating a token signed by the jwks URL but not generated by Keycloak would be a lot of work.

_No response_

evankanderson:
# Summary

Per https://openid.net/specs/openid-connect-basic-1_0.html#IDTokenValidation:

> The Issuer Identifier for the OpenID Provider (which is typically obtained during Discovery) MUST exactly match the value of the iss (issuer) Claim.

In this case, we're not implementing discovery, and using a server-configured JWKS URL, but we should still validate the iss URL.

Fixes #3542

## Change Type

# Testing

Unit tests and manual login test, which revealed that:

1. We have an `issuer_url` parameter, which is actually a hostname, not the `iss` parameter.  In particular, it doesn't include the `/realms/stacklok` part, which is instead part of the URLs we build in Minder.
1. If we set `issuer_url` to `http://localhost:8081` for development (to match what's in the JWT), Minder in docker-compose is unable to reach Keycloak, because each is in a different network namespace.
1. We have a similar issue in production, where we don't expose the `/admin` API endpoints on the public internet.

All of this is pointed me to splitting the identity into two bits -- a keycloak server URL for the admin API endpoints, and an openID issuer URL.

# Review Checklist:


config/server-config.yaml.example
```diff
@@ -46,7 +46,8 @@ database:
 
 identity:
   server:
-    issuer_url: http://keycloak:8080 # Use http://localhost:8081 instead for running minder outside of docker compose
+    issuer_url: http://localhost:8081
+    issuer_claim: http://localhost:8081/realms/stacklok
```
rdimitrov, line 50: We still want to have the default be the all-in-docker setup?

cmd/server/app/serve.go
```diff
@@ -87,11 +87,19 @@ var serveCmd = &cobra.Command{
 		}
 
 		// Identity
-		jwksUrl, err := cfg.Identity.Server.Path("realms/stacklok/protocol/openid-connect/certs")
+		// TODO: cfg.Identity.Server.IssuerUrl _should_ be a URL to an issuer that has an
```
eleftherias, line 90: I'm not sure this comment is accurate. Maybe it's better to rename the `issuerUrl` to `hostnameUrl`. We intentionally use the hostname for deriving the user deletion URL as well, which is also Keycloak specific.

evankanderson: Ugh... so, in an ideal world, the URL that Keycloak includes in the `iss` claim would be reachable by the `minder` server and then we could use the standard OpenID JWKS discovery.

Unfortunately, the current docker-compose setup produces an `iss` of `http://localhost:8081/realms/stacklok`, but the `minder` container can only reach Keycloak via `keycloak:8080`.  This isn't a problem in staging or production, where Minder _can_ reach the external interface, and we just _happen_ to fetch some keys from an entirely different URL and use them to validate the JWT.

Let me think on this, and see if I can come up with a better way to handle this.

And yes, we should rename `issuer_url` to `identity_host` or `identity_url`, but I didn't want to gratuitously break existing configuration (including our own).

That also reminds me that I need to set these fields in staging and production _before_ we push this code there...

config/server-config.yaml.example
evankanderson: I can set this whichever way -- we override the config value with a flag in docker-compose for the all-in-docker setup.

dmjb (approved):

Issue #3268, opened 2024-05-07
Add a ReconcileEntityRegistration method to the Providers API and a reconcile repository implementation to the GitHub App


jhrozek:
In order to trigger repository reconciliation to either cover the case where we miss a webhook or after the initial enablement of the auto-enrollment we need a way to tell the provider to reconcile the auto-registered repositories with the upstream repositories.

To do that, we'd extend the providers API with an ReconcileEntityRegistration method exposed as an RPC. The Github app provider would implement it to reconcile repositories.

The RPC might look like this:
```
rpc ReconcileEntityRegistration (ReconcileEntityRegistrationRequest) returns (ReconcileEntityRegistrationResponse) {
        option (google.api.http) = {
            post: "/api/v1/provider/{provider}/register_all"
            additional_bindings {
                post: "/api/v1/provider/register_all",
                body: "*"
            }
            body: "*"
        };

        option (rpc_options) = {
            target_resource: TARGET_RESOURCE_PROJECT
	     // TODO: ensure that we either have per-entity APIs or we have a relation that allows any entity
            relation: RELATION_REPO_CREATE
        };
    }

message ReconcileEntityRegistrationRequest {
    Context context = 1;
    Entity entity = 2; // reconcile this entity type only
}

message RegisterRepositoryResponse {
	// empty by design
}

```

jhrozek:
Inside the handler I think we could do the following:
 - extract the code from `ListRepositoriesFromProvider` inside the `	for providerName, provider := range provs` to a reusable function in the server to get the list of repositories that are not registered yet.
 - for every repo that is not registered, create a message using `s.repositoryAdded()`
 - dispatch the message as we do for the auto-registration messages with `HandleGitHubAppWebhook`.

I think this would be OK for a start. We need to then have a way to signal to the reconcile handler to stop if a token is below certain threshold, but I suggest we file a separate ticket for that.

@blkt do you agree with the above?

blkt:
I agree, we should have a single code path for registration hooked into multiple entry points (i.e. `minder repo register`, automatic registration, etc...).

> We need to then have a way to signal to the reconcile handler to stop if a token is below certain threshold

I studied this subject a bit in the past, in particular I'd love to apply feedback control to this. Definitely worth another issue!

jhrozek:
> > We need to then have a way to signal to the reconcile handler to stop if a token is below certain threshold
> 
[...]

PR #3562, opened 2024-06-08
Reconcile entity registration

teodor-yanev:
# Summary

Fixes https://github.com/stacklok/minder/issues/3268

## Change Type

# Testing

Calling the method via the CLI. (note: please ignore the "nil" as it was part of the print testing)
For simplicity of the testing, I added this as an additional operation as part of the "register" flow. I'm mentioning this in case the screenshot seems confusing at first.

image

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -1054,6 +1065,16 @@ message ListRepositoriesResponse {
     string cursor = 2;
 }
 
+message ReconcileEntityRegistrationRequest {
+    Context context = 1;
+    string entity = 2;
+
+    string provider = 3 [deprecated=true];
```
jhrozek, line 1072: let's not add deprecated fields into new messages.

proto/minder/v1/minder.proto
```diff
@@ -342,6 +342,17 @@ service RepositoryService {
             relation: RELATION_REPO_CREATE
         };
     }
+    rpc ReconcileEntityRegistration(ReconcileEntityRegistrationRequest) returns (ReconcileEntityRegistrationResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/provider/{provider}/register_all"
```
jhrozek, line 347: I must have made a copy-paste in the original ticket. What I put in the additional bindings is what I think we should use as the REST API path, so `post: "/api/v1/provider/register_all"`. Which provider are we reconciling should be read from the context.

internal/controlplane/handlers_repositories.go
```diff
@@ -385,6 +349,56 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
+		zerolog.Ctx(ctx).Error().Err(err).Msg("cannot list repositories for provider")
+		return nil, err
+	}
+
+	registeredRepos, err := s.repos.ListRepositories(
+		ctx,
+		projectID,
+		providerName,
+	)
+	if err != nil {
```
jhrozek, line 383: I know this was in the original code, but could we also `zerolog` the error message?

internal/controlplane/handlers_entities.go
```diff
@@ -0,0 +1,132 @@
...
+		// Explicitly fetch the provider here as we need its ID for posting the event.
+		pvr, err := s.providerStore.GetByName(ctx, projectID, providerName)
+		if err != nil {
+			errorProvs = append(errorProvs, providerName)
+			continue
+		}
+
+		repos, err := s.fetchRepositoriesForProvider(ctx, projectID, providerName, provider)
+		if err != nil {
+			errorProvs = append(errorProvs, providerName)
```
jhrozek, line 72: since the handler only errors out if all providers fail could we `zerolog` the error message? (I guess there would realistically be only one provider though)

internal/controlplane/handlers_entities.go
```diff
@@ -0,0 +1,132 @@
...
+			if err != nil {
+				l.Error().Err(err).
+					Int64("repoID", repo.RepoId).
+					Str("providerName", providerName).
+					Msg("error creating registration entity message")
+				// This message will not be sent, but we can continue with the rest.
+				continue
+			}
+
+			if s.publishEntityMessage(&l, msg) != nil {
```
jhrozek, line 91: since we are checking for an error here could we write `if err := s.publishEntityMessage(&l, msg); err != nil {`
(we're probably not logging the err from this call underneath)

internal/controlplane/handlers_entities.go
```diff
@@ -0,0 +1,132 @@
+// Copyright 2023 Stacklok, Inc
```
jhrozek, line 1: it's 2024 :-P

jhrozek (commented):
looking good, a couple of nits here and there, but the main thing is to fix the protobuf to not include the provider (sorry, my fault for writing it wrong in the issue description)

internal/controlplane/handlers_entities.go
teodor-yanev: Ah 🤦 , my bad, thanks

proto/minder/v1/minder.proto
teodor-yanev: I honestly thought this would be more appropriate as well. Thanks.

proto/minder/v1/minder.proto
```diff
@@ -342,6 +342,17 @@ service RepositoryService {
             relation: RELATION_REPO_CREATE
         };
     }
+    rpc ReconcileEntityRegistration(ReconcileEntityRegistrationRequest) returns (ReconcileEntityRegistrationResponse) {
```
jhrozek, line 345: One more thing, because this RPC can register any types, shouldn't it be part of the provider service? (See also its http path..)

proto/minder/v1/minder.proto
```diff
@@ -1054,6 +1065,14 @@ message ListRepositoriesResponse {
     string cursor = 2;
 }
 
+message ReconcileEntityRegistrationRequest {
+    Context context = 1;
+    string entity = 2;
```
jhrozek, line 1070: We currently don't handle this part of the request at all. I'm thinking we should validate it and only allow repositories, but let's do that in a follow up.

proto/minder/v1/minder.proto
blkt: I agree, it's better placed under provider service.

internal/controlplane/handlers_entities.go
blkt: Line 75 already ships the error with the log as a structured field, do you mean something more detailed than that?

blkt (commented):
First batch of minor comments. Thanks for the great work @teodor-yanev!

internal/controlplane/handlers_repositories.go
```diff
@@ -385,6 +349,56 @@ func (s *Server) ListRemoteRepositoriesFromProvider(
...
 
+// fetchRepositoriesForProvider fetches repositories for a given provider
+//
+// Returns a list of repositories that with an up-to-date status of whether they are registered
+func (s *Server) fetchRepositoriesForProvider(
+	ctx context.Context,
+	projectID uuid.UUID,
+	providerName string,
+	provider v1.Provider,
+) ([]*pb.UpstreamRepositoryRef, error) {
```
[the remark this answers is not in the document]
jhrozek: I agree it would be nice but let's do it in a separate PR to not delay this one longer than needed.

proto/minder/v1/minder.proto
teodor-yanev: Noted

proto/minder/v1/minder.proto
teodor-yanev: I thought about this at the beginning as well but then I noticed that the provider service contains RPCs explicitly related to CRUD operations for the _provider_ entity itself. Indeed, the repository service is not a suitable place.
Still, I've moved this to the Provider service as it's the closest logically related one.

proto/minder/v1/minder.proto
teodor-yanev: Actually, I've addressed it in one of the commits here.

internal/controlplane/handlers_repositories.go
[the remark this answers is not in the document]
teodor-yanev: Yeah, I agree with both of the statements.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
teodor-yanev: Removed now, this was left from when I had the bindings URL.

internal/controlplane/handlers_entities.go
```diff
@@ -0,0 +1,143 @@
...
+
+	entityCtx := engine.EntityFromContext(ctx)
+	projectID := entityCtx.Project.ID
+
+	logger.BusinessRecord(ctx).Project = projectID
+
+	// Todo: We don't support other entities yet. This should be updated when we do.
+	entityType := in.GetEntity()
+	if pb.EntityFromString(entityType) != pb.Entity_ENTITY_REPOSITORIES {
+		return nil, util.UserVisibleError(codes.Internal, "entity type %s not supported", entityType)
```
jhrozek, line 55: One small change please, this should not return `codes.Internal`, but `codes.InvalidArgument` instead. Mainly so that we get the proper HTTP error code through the gateway (now we get a 500)

teodor-yanev: Yep, the copy/pasta betrayed me, thanks! 
Fixed

jhrozek (approved):

PR #3570, opened 2024-06-10
Add project_id column to rule_instances

dmjb:
This provides a more elegant solution to the `ON CASCADE DELETE` problem with the `rule_instance` table. Tested locally.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


database/migrations/000065_rule_instance_project_id.up.sql
```diff
@@ -0,0 +1,29 @@
...
+ALTER TABLE rule_instances ADD COLUMN project_id UUID REFERENCES projects(id) ON DELETE CASCADE;
+
+-- populate by joining on profiles table
+UPDATE rule_instances AS ri
+SET project_id = pf.project_id
+FROM profiles AS pf
+WHERE ri.profile_id = pf.id;
+
+-- now we can add the not null constraint
+ALTER TABLE rule_instances ALTER COLUMN project_id SET NOT NULL;
```
evankanderson, line 27: Won't this cause current-minder to fail after the DB migration until the new deployment occurs?

database/migrations/000065_rule_instance_project_id.up.sql
```diff
@@ -0,0 +1,29 @@
...
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+-- add column, leave as not null until we populate the column
+ALTER TABLE rule_instances ADD COLUMN project_id UUID REFERENCES projects(id) ON DELETE CASCADE;
```
evankanderson, line 18: Do we want or need an index on project_id here?

evankanderson (commented):
Overall, I like this simplification.  Maybe we're okay at the moment with causing minder failures during rollout due to adding NOT NULL columns to the schema, but I wanted to check. 

If you want to split it up, I would add the column with no constraints and the new queries, and then do a second migration PR that does the backfill and adds the constraint.

database/migrations/000065_rule_instance_project_id.up.sql
dmjb: Good point, will remove this.

database/migrations/000065_rule_instance_project_id.up.sql
dmjb: At this point in time, we don't use the new column in the queries since the tuple of (name, profile_id, entity_type) is still unique. As I add more queries, I will add more indices based on common query patterns.

rdimitrov (approved):

PR #3630, opened 2024-06-16
Connect reminder service to minder server to dispatch reminders

Vyom-Yadav:
# Summary

Issue #2262 (Task - I)

Adds logic for sending reminders to the minder server. Minder server processes these reminders to trigger a `TopicQueueReconcileRepoInit` event to reconcile a repository. 

## Change Type

# Testing

Tested locally (no unit tests)

# Review Checklist:

cc @evankanderson


internal/reminder/reminder.go
```diff
@@ -144,19 +147,64 @@ func (r *reminder) sendReminders(ctx context.Context) []error {
...
+			errorSlice = append(errorSlice, fmt.Errorf("error publishing messages: %w", err))
+		} else {
+			logger.Info().Msgf("sent %d reminders", len(messages))
+
+			// Commit the transaction i.e update reminder_last_sent
+			// only if the messages were sent successfully
+			if err = tx.Commit(); err != nil {
+				logger.Error().Err(err).Msg("unable to commit transaction")
+				errorSlice = append(errorSlice, err)
+			}
```
Vyom-Yadav, line 203: `UpdateReminderLastSentById` queries are part of a transaction and are only committed if the messages are published successfully. Still, the transaction could fail, which would result in reminders being sent, but the `reminder_last_sent` field won't be updated (better than false positive i.e. `reminder_last_sent` was updated but reminders weren't sent)

internal/reminder/reminder.go
```diff
@@ -144,19 +147,64 @@ func (r *reminder) sendReminders(ctx context.Context) []error {
...
+
+	defer tx.Rollback()
+
+	qtx := r.store.GetQuerierWithTransaction(tx)
+
+	// TODO: Collect Metrics
+	// Potential metrics:
+	// - Gauge: Number of reminders in the current batch
+	// - UpDownCounter: Average reminders sent per batch
+	// - Histogram: reminder_last_sent time distribution
```
Vyom-Yadav, line 166: How do we want to export the metrics? And some thoughts on adding these metrics?

jhrozek: Hmm, I wonder what metrics would be useful as a minder operator.

About the histogram, I think a histogram that shows me how out of date (how much before a cutoff) a repo was when it was selected would be useful.

I think the metrics would be interesting to fine-tune the algorithm, did you see some other value in the metrics?

Since reminder is a separate service, then I guess implementation-wise it should just have its own metrics server that could be scraped separately.

jhrozek (commented):
The code looks quite good to me. I left replies to your comments, neither of them is blocking. It wasn't clear to me if you wanted to tackle the metrics as part of this PR, but I normally prefer smaller PRs.

internal/reminder/reminder.go
```diff
@@ -144,19 +147,64 @@ func (r *reminder) sendReminders(ctx context.Context) []error {
...
+		}
+
+		messages = append(messages, repoReconcilerMessage)
+	}
+
+	if len(messages) != 0 {
+		err = r.eventPublisher.Publish(events.TopicQueueRepoReminder, messages...)
+		if err != nil {
+			errorSlice = append(errorSlice, fmt.Errorf("error publishing messages: %w", err))
+		} else {
```
evankanderson, line 195: Rather than an `if/else` here, generally prefer to use early-return, particularly since this seems like a terminal case.

More generally, I'm not convinced that there's a large benefit to batching the Publish messages, rather than sending each in the previous part of the loop.

internal/reminder/reminder.go
evankanderson: It seems like the `reminder_last_sent` field is mostly diagnostic, because we're actually ticking forward on the (in-memory) repository cursor, rather than using `reminder_last_sent` to actually dictate when repos are reconsidered.  I think this is correct, but that suggests to me that strict update correctness is less necessary.

(I'm not arguing against keeping this data, but it seems like the actual mechanism that prevents re-evaluation is when the reminder is _received_ and a rule evaluation is complete, which could be minutes after the reminder is sent on a particularly bad day.)

internal/reminder/messages/messages.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/go-playground/validator/v10"
+	"github.com/google/uuid"
+)
+
+// RepoReminderEvent is an event that is published by the reminder service to trigger repo reconciliation
+type RepoReminderEvent struct {
+	// Project is the project that the event is relevant to
+	Project uuid.UUID `json:"project"`
+	// RepositoryID is id of the repository to be reconciled
+	RepositoryID int64 `json:"repository" validate:"gte=0"`
```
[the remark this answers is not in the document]
evankanderson: Looking further, it looks like we use the GitHub-matching `repo_id` in NewRepoReconcilerMessage already, so that argues towards using it here as well.

internal/reminderprocessor/reminder_processor.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating repo reconcile event: %w", err)
+	}
+
+	// This is a non-fatal error, so we'll just log it and continue with the next ones
+	if err := rp.evt.Publish(events.TopicQueueReconcileRepoInit, repoReconcileMsg); err != nil {
+		log.Printf("error publishing reconciler event: %v", err)
+	}
+	return nil
+}
```
evankanderson, line 63: It feels like this is simply forwarding messages from one queue to another; why not simply sending the `TopicQueueReconcileRepoInit` message directly from `reminder.go` and skip this extra intermediary?  (There might be a good reason, but we should encapsulate that with a comment if there is such a reason.)

evankanderson (commented):
It looks like the current architecture re-uses the existing events configuration for Minder, since it's assuming that the `repo.reminder.event` events are in a Watermill instance which Minder is configured to see.  I think that's fine, but I didn't recall for sure where we'd landed on that earlier.  This setup is 👍  for me from a simplicity PoV.

internal/reminder/messages/messages.go
```diff
@@ -0,0 +1,67 @@
...
+	// ProviderID is the provider of the repository
+	ProviderID uuid.UUID `json:"provider"`
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

internal/reminder/messages/messages.go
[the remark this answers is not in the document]
Vyom-Yadav: Yupp, I used `RepositoryID` to represent GH Id, as Id collided with internal primary key.

internal/reminder/messages/messages.go
```diff
@@ -0,0 +1,67 @@
...
+func RepoReminderEventFromMessage(msg *message.Message) (*RepoReminderEvent, error) {
+	var evt RepoReminderEvent
+	if err := json.Unmarshal(msg.Payload, &evt); err != nil {
+		return nil, fmt.Errorf("error unmarshalling payload: %w", err)
+	}
+
+	validate := validator.New()
+	if err := validate.Struct(&evt); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
Vyom-Yadav: Yeah, no such benefit, removed it.

internal/reminder/reminder.go
Vyom-Yadav: I'll add monitoring in a sep PR.

internal/reminder/reminder.go
```diff
@@ -144,19 +147,64 @@ func (r *reminder) sendReminders(ctx context.Context) []error {
...
+			errorSlice = append(errorSlice, err)
+			continue
+		}
+
+		logger.Debug().
+			Str("repo", repo.ID.String()).
+			Time("previously", repo.ReminderLastSent.Time).
+			Msg("updating reminder_last_sent")
+
+		err = qtx.UpdateReminderLastSentById(ctx, repo.ID)
```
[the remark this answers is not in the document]
Vyom-Yadav: Changed this.

internal/reminder/reminder.go
```diff
@@ -144,19 +147,64 @@ func (r *reminder) sendReminders(ctx context.Context) []error {
...
 		if err != nil {
 			logger.Error().Err(err).Str("repo", repo.ID.String()).Msg("unable to update reminder_last_sent")
-			return []error{err}
+			errorSlice = append(errorSlice, err)
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

internal/reminder/reminder.go
Vyom-Yadav: Done.

> I'm not convinced that there's a large benefit to batching the Publish messages, rather than sending each in the previous part of the loop.

Bulk updates might be better than connecting to the db for every update, I went with bulk updates with slight changes in logic.

internal/reminderprocessor/reminder_processor.go
Vyom-Yadav: I'm open to removing this processor and directly sending events. But I think initially we discussed adding a new topic for reminder to have some distinction (which, now when I think cannot be internally distinguished, we don't have any form of tracing for reconciliations started by reminder).

internal/reminder/reminder.go
```diff
@@ -132,28 +133,46 @@ func (r *reminder) Stop() {
...
+
+	// TODO: Collect Metrics
+	// Potential metrics:
+	// - Gauge: Number of reminders in the current batch
+	// - UpDownCounter: Average reminders sent per batch
+	// - Histogram: reminder_last_sent time distribution
+
+	err = r.store.UpdateReminderLastSentForRepositories(ctx, repoIds)
+	if err != nil {
+		return fmt.Errorf("reminders published but error updating last sent time: %w", err)
```
Vyom-Yadav, line 175: Bulk updates for all repos is fine, IMO. We don't want to have a `ms` difference. The only value of this metric is in minutes/hours to make sure reconciliation is happening at a regular interval and we can reconcile every repo in like 24 hours or so.

internal/reminder/reminder.go
```diff
@@ -241,3 +260,25 @@ func (r *reminder) adjustCursorForEndOfList(ctx context.Context) {
...
+	logger := zerolog.Ctx(ctx)
+
+	messages := make([]*message.Message, 0, len(repos))
+	for _, repo := range repos {
+		repoReconcileMessage, err := remindermessages.NewRepoReminderMessage(
+			repo.ProviderID, repo.RepoID, repo.ProjectID,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("error creating reminder message: %w", err)
+		}
```
Vyom-Yadav, line 274: We now stop if any JSON marshalling fails. I expect this to be deterministic, so if one is failing, others will fail too.

Vyom-Yadav:
Sorry for the delay on this. Somehow, all weekends got occupied with one thing or the other :)

evankanderson:
> Sorry for the delay on this. Somehow, all weekends got occupied with one thing or the other :)

No problem, I've been in-and-out the last few weeks anyway.  I'll try to take a look at this later today or tomorrow.

Vyom-Yadav:
@evankanderson not found (404), where are you 😆 

image

internal/reminder/reminder.go
evankanderson: What I was thinking of recording on each pass of collecting reminders would effectively be the following for each reminded row:

```
now() - reminder_last_sent
```

Which would _hopefully_ give a nice even distribution between `MinElapsed` and `MinElapsed + Interval` in normal operation.  Unfortunately, common values might be `MinElapsed = 24h` and `Interval = 5m`, which will lead to a histogram where everything lands in a very large bucket.  So, for practicality, I'd probably report this as:

```
now() - reminder_last_sent - MinElapsed
```

Which should give a range `0 .. Interval`.  If the algorithm starts to get behind or reminders "bunch", we'd see the tail of the histogram get bigger, and go beyond `Interval`.  Given the target of `Interval = 5m` (hypothetical, but seems reasonable), we'd probably want to measure in seconds with a starting bucket size of 10s, roughly-doubling buckets (or a custom 1..5..10 bucketer) and going out to at least 20 minutes, which would mean about 9 buckets (0, 10, 20, 40, 80, ... 1280).  I'm less concerned about the exact bucket sizing as being able to see whether most items are ending up within that `MinElapsed + Interval` window.

I'd store this metric in `UpdateReminderLastSentForRepositories`, so that the rest of the loop didn't need to care, and it was next to the handling of `reminder_last_sent`.

internal/reminderprocessor/reminder_processor.go
evankanderson: In #4075 , I'm starting to look at switch from Watermill+SQL to CloudEvents + NATS, which has a somewhat more formalized envelope structure.  I think we could probably have some sort of envelope property that distinguishes between the two, but put both messages onto the same topic -- the reader would be able to record which source the `init` event came from, but would generally handle "new" and "remind" the same.

evankanderson (approved):
Sorry for disappearing!  I had a week of vacation and a week of conferences, and when I got done, I had no idea what happened before that time.

This is looking pretty good, I'm happy to add it in as-is for now, and figure out the next steps in a subsequent PR.  I did add a couple comments, but I don't think they block merging this.

PR #3666, opened 2024-06-19
Implement ListInvitations and ResolveInvitation APIs for UserService

rdimitrov:
# Summary

The following PR implements ListInvitations and ResolveInvitation APIs for UserService.

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_user.go
+++ b/internal/controlplane/handlers_user.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"path"
 	"strconv"
+	"time"
 
 	gauth "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/auth"
 	"github.com/rs/zerolog"
@@ -29,7 +30,9 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/stacklok/minder/internal/auth"
+	"github.com/stacklok/minder/internal/authz"
 	"github.com/stacklok/minder/internal/db"
+	"github.com/stacklok/minder/internal/flags"
 	"github.com/stacklok/minder/internal/logger"
 	"github.com/stacklok/minder/internal/projects"
 	"github.com/stacklok/minder/internal/util"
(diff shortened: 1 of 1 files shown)


internal/controlplane/handlers_user.go
```diff
@@ -259,3 +262,132 @@ func (s *Server) GetUser(ctx context.Context, _ *pb.GetUserRequest) (*pb.GetUser
...
+	ret, err := s.store.GetInvitationByCode(ctx, req.Code)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, util.UserVisibleError(codes.NotFound, "invitation not found")
+		}
+		return nil, status.Errorf(codes.Internal, "failed to get invitation: %s", err)
+	}
+
+	// Check if the invitation matches the user email
+	if token.Email() != ret.Email {
```
rdimitrov, line 358: Implemented this per - https://github.com/stacklok/minder/blob/07f93e0c1dcb59f74b6f15beb24cf684161287bb/database/query/invitations.sql#L15

which ensures if the email of the logged-in user corresponds to the one of the invite. 

@evankanderson - I know we discussed this to be more loose but then I saw that you made that comment so I implemented it that way. I'll ping you to see which is the desired behavior we want to keep. 

For the reviewers, don't take this comment as blocking as we can remove this check easily.

dmjb (changes requested):
It doesn't look like we have any unit test coverage for these, and there's enough new logic here that I feel like it would benefit from testing.

I wonder if splitting the logic between the handler, and an `InviteService` struct/interface would simplify writing tests too.

dmjb (approved):
Approving to unblock FE. Let's get some tests for this written sooner rather than later :)

internal/controlplane/handlers_user.go
```diff
@@ -259,3 +262,132 @@ func (s *Server) GetUser(ctx context.Context, _ *pb.GetUserRequest) (*pb.GetUser
...
+		if err := s.authzClient.Write(ctx, token.Subject(), authzRole, ret.Project); err != nil {
+			return nil, status.Errorf(codes.Internal, "error writing role assignment: %v", err)
+		}
+		isAccepted = true
+	}
+
+	// Delete the invitation since its resolved
+	ret, err = s.store.DeleteInvitation(ctx, req.Code)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to delete invitation: %s", err)
```
evankanderson, line 385: What happens in this case -- the user has been added to the OpenFGA database; if they try calling `ResolveInvitation` again, will they create a second tuple?  Will OpenFGA error with "already exists"?

We may want to return a `util.UserVisibleError(codes.Internal, "permission granted, but unable to revoke invitation")` after logging the error internally here.

We may also want to check what the result of attempting to grant the same permission twice is in terms of OpenFGA errors, and account for that specially.

internal/controlplane/handlers_user.go
```diff
@@ -259,3 +262,132 @@ func (s *Server) GetUser(ctx context.Context, _ *pb.GetUserRequest) (*pb.GetUser
...
+	if token.Email() != ret.Email {
+		return nil, util.UserVisibleError(codes.PermissionDenied, "invitation does not match user")
+	}
+
+	// Check if the invitation is expired
+	if time.Now().After(ret.UpdatedAt.Add(7 * 24 * time.Hour)) {
+		return nil, util.UserVisibleError(codes.PermissionDenied, "invitation expired")
+	}
+
+	isAccepted := false
```
evankanderson, line 367: Do we need `isAccepted`, or can we just use `req.Accept` below?

evankanderson (commented):
Overall LGTM, but a few questions / comments for a followup (sorry, I'm half-out today for Juneteenth, and then half-out Friday taking the kids on a trip).

PR #3672, opened 2024-06-19
Update AssignRole, RemoveRole and add new UpdateRole handlers

rdimitrov:
# Summary

The following PR update AssignRole, RemoveRole and add new UpdateRole handlers

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -323,7 +323,7 @@ func (s *Server) AssignRole(ctx context.Context, req *minder.AssignRoleRequest) 
 		if errors.Is(err, sql.ErrNoRows) {
 			return nil, util.UserVisibleError(codes.InvalidArgument, "target project with ID %s not found", targetProject)
 		}
-		return nil, status.Errorf(codes.Internal, "error getting project: %v", err)
+		return nil, status.Errorf(codes.InvalidArgument, "error getting project: %v", err)
 	}
 
 	// Validate the subject and email - decide if it's an invitation or a role assignment
@@ -674,7 +674,7 @@ func (s *Server) UpdateRole(ctx context.Context, req *minder.UpdateRoleRequest) 
 		if a.Subject == identity.String() {
 			roleToDelete, err := authz.ParseRole(a.Role)
 			if err != nil {
-				return nil, util.UserVisibleError(codes.InvalidArgument, err.Error())
+				return nil, util.UserVisibleError(codes.Internal, err.Error())
 			}
 			if err := s.authzClient.Delete(ctx, identity.String(), roleToDelete, targetProject); err != nil {
 				return nil, status.Errorf(codes.Internal, "error deleting previous role assignment: %v", err)
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"database/sql"
 	"errors"
-	"regexp"
 	"time"
 
 	"github.com/google/uuid"
@@ -323,11 +322,11 @@ func (s *Server) AssignRole(ctx context.Context, req *minder.AssignRoleRequest) 
 		if errors.Is(err, sql.ErrNoRows) {
 			return nil, util.UserVisibleError(codes.InvalidArgument, "target project with ID %s not found", targetProject)
 		}
-		return nil, status.Errorf(codes.InvalidArgument, "error getting project: %v", err)
+		return nil, status.Errorf(codes.Internal, "error getting project: %v", err)
 	}
 
 	// Validate the subject and email - decide if it's an invitation or a role assignment
-	if sub == "" && email != "" && isEmail(email) {
+	if sub == "" && email != "" {
 		if flags.Bool(ctx, s.featureFlags, flags.UserManagement) {
 			return s.inviteUser(ctx, targetProject, authzRole, email)
 		}
@@ -513,7 +512,7 @@ func (s *Server) RemoveRole(ctx context.Context, req *minder.RemoveRoleRequest) 
 	}
 
 	// Validate the subject and email - decide if it's about removing an invitation or a role assignment
-	if sub == "" && email != "" && isEmail(email) {
+	if sub == "" && email != "" {
 		if flags.Bool(ctx, s.featureFlags, flags.UserManagement) {
 			return s.removeInvite(ctx, targetProject, authzRole, email)
 		}
@@ -602,14 +601,6 @@ func (s *Server) removeRole(
 			Project: &prj,
 		},
 	}, nil
-}
-
-// isEmail checks if the subject is an email address or not
-func isEmail(subject string) bool {
-	// Define the regular expression for validating an email address
-	const emailRegexPattern = `^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$`
-	emailRegex := regexp.MustCompile(emailRegexPattern)
-	return emailRegex.MatchString(subject)
 }
 
 // UpdateRole updates a role for a user on a project
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -18,9 +18,11 @@ import (
 	"context"
 	"database/sql"
 	"errors"
+	"regexp"
 	"time"
 
 	"github.com/google/uuid"
+	gauth "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/auth"
 	"github.com/rs/zerolog"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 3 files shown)


rdimitrov:
Issue for creating unit tests for these handlers - https://github.com/stacklok/minder/issues/3673

rdimitrov:
Obviously don't merge as I haven't updated the unit tests but do review the code if you decide so 👍

internal/controlplane/handlers_authz.go
```diff
@@ -292,32 +296,157 @@ func (s *Server) ListRoleAssignments(
...
+	if err != nil {
+		// If the project is not found, return an error
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, util.UserVisibleError(codes.InvalidArgument, "target project with ID %s not found", targetProject)
+		}
+		return nil, status.Errorf(codes.Internal, "error getting project: %v", err)
+	}
+
+	// Validate the subject and email - decide if it's an invitation or a role assignment
+	if sub == "" && email != "" && isEmail(email) {
```
puerco, line 330: if `isEmail()` returns false here I think we should return an internal error as it points to an internal data inconsistency already present in the database.

internal/controlplane/handlers_authz.go
```diff
@@ -292,32 +296,157 @@ func (s *Server) ListRoleAssignments(
...
+				return nil, status.Errorf(codes.Internal, "error creating invitation: %v", err)
+			}
+		} else {
+			// Some other error happened, return it
+			return nil, status.Errorf(codes.Internal, "error getting invitations: %v", err)
+		}
+	} else {
+		// If we didn't get an error, this means there's an existing invite.
+		// We should update its expiration and send the response.
+		userInvite, err = s.store.UpdateInvitation(ctx, existingInvite.Code)
```
puerco, line 391: Here we should check the invite expiration date before determining if we should update it. We want to avoid replaying an invite in case someone gets access to an old email for example

internal/controlplane/handlers_authz.go
```diff
@@ -355,16 +493,150 @@ func (s *Server) AssignRole(ctx context.Context, req *minder.AssignRoleRequest)
...
+			Project: &prj,
+		},
+	}, nil
+}
+
+// isEmail checks if the subject is an email address or not
+func isEmail(subject string) bool {
+	// Define the regular expression for validating an email address
+	const emailRegexPattern = `^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$`
+	emailRegex := regexp.MustCompile(emailRegexPattern)
```
puerco, line 611: Compile the email regexp just once to reuse it across calls

```suggestion

var emailRegexPattern *regexp.Regexp
// isEmail checks if the subject is an email address or not
func isEmail(subject string) bool {
	// Define the regular expression for validating an email address
	const emailRegexPattern = `^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$`
	if emailRegexPattern == nil {
	  emailRegex = regexp.MustCompile(emailRegexPattern)
	}
```

internal/controlplane/handlers_authz.go
```diff
@@ -392,20 +664,61 @@ func (s *Server) RemoveRole(ctx context.Context, req *minder.RemoveRoleRequest)
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting role assignments: %v", err)
+	}
 
-	if err := s.authzClient.Delete(ctx, identity.String(), authzrole, projectID); err != nil {
+	for _, a := range as {
+		if a.Subject == identity.String() {
+			roleToDelete, err := authz.ParseRole(a.Role)
+			if err != nil {
+				return nil, util.UserVisibleError(codes.InvalidArgument, err.Error())
```
puerco, line 677: Just as in the other, if somehow an un parseable role is being returned by the client this points to an internal inconsistency, not an argument from the client
```suggestion
				return nil, util.UserVisibleError(codes.Internal, err.Error())
```

internal/controlplane/handlers_authz.go
```diff
@@ -334,18 +463,27 @@ func (s *Server) AssignRole(ctx context.Context, req *minder.AssignRoleRequest)
...
-	entityCtx := engine.EntityFromContext(ctx)
-	projectID := entityCtx.Project.ID
+	// Check in case there's an existing role assignment for the user
+	as, err := s.authzClient.AssignmentsToProject(ctx, targetPrj)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting role assignments: %v", err)
+	}
+
+	for _, a := range as {
+		if a.Subject == identity.String() {
```
puerco, line 473: I think this should also be checking for the role? We should match user + role before calling out that the assignment already exists

internal/controlplane/handlers_authz.go
```diff
@@ -292,32 +296,157 @@ func (s *Server) ListRoleAssignments(
...
 	}
 
+	// Ensure the target project exists
+	_, err = s.store.GetProjectByID(ctx, targetProject)
+	if err != nil {
+		// If the project is not found, return an error
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, util.UserVisibleError(codes.InvalidArgument, "target project with ID %s not found", targetProject)
+		}
+		return nil, status.Errorf(codes.Internal, "error getting project: %v", err)
```
puerco, line 326: And here I think it is a case for `InvalidArgument` as it is querying using the input from the user :D 
```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "error getting project: %v", err)
```

internal/controlplane/handlers_authz.go
rdimitrov: The email in this case is received as an argument that is part of the request message, so I think it makes sense to validate it and return `codes.InvalidArgument` if it doesn't look alright

internal/controlplane/handlers_authz.go
rdimitrov: Let me give some context. So the lifespan of the invites should be 7 days for expiration, but we'll keep them for 30 days before deleting them entirely so an admin can update them once they see they have expired. This bumps their expires time only so previous send links continue to work and this is the intended behavior.

This flow happens when an administrator invites someone else again and there was an existing invitation (meaning it is less than 30 days old, otherwise it should have been deleted).

The scenarios are:
* If the invite expires <= 7 days old, they are fine and can accept.
* If the invite is 7 days < expires < 30 days, the invite will not work unless the admin went and updated it through this flow in the last 7 days
* If the invite is 30 < expires, it shouldn't be present in the db, that is for both the admin and the invitee.

In that sense, I don't think we need to check the expiration because this flow is called by the administrator explicitly and is about updating the expiration of an invite.

internal/controlplane/handlers_authz.go
rdimitrov: Good tip! Let me update this manually though as I think some of the env names in the suggestion are not correct.

internal/controlplane/handlers_authz.go
rdimitrov: Thanks! I agree on this since the respond comes from calling openfga 👍

internal/controlplane/handlers_authz.go
rdimitrov: It was decided that a user can have only one role within a project so in this case this should be enough 👍

internal/controlplane/handlers_authz.go
rdimitrov: We have a deal on this one! 😃

internal/controlplane/handlers_authz.go
```diff
@@ -355,16 +493,150 @@ func (s *Server) AssignRole(ctx context.Context, req *minder.AssignRoleRequest)
...
+	// Get all invitations for this email, project and role
+	inviteToRemove, err := s.store.GetInvitationByEmailAndProjectAndRole(ctx, db.GetInvitationByEmailAndProjectAndRoleParams{
+		Email:   email,
+		Project: targetPrj,
+		Role:    role.String(),
+	})
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, util.UserVisibleError(codes.NotFound, "no invitation found for this email, project and role")
+		}
```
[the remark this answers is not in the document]
rdimitrov: It is what we do for other handlers, such as - https://github.com/stacklok/minder/blob/d08b8f68f4a308c3c8c69f49fbdb81710596ddcb/internal/controlplane/handlers_profile.go#L95

Let's discuss if we should change this and we can do that in a follow-up for all similar use cases so it's consistent 👍

jhrozek:
@puerco not to roll over your review, but since the frontend team was waiting on the API to land, I did a review of the PR with @rdimitrov. If you feel like your comments need addressing let's bring them up as follow-up PRs

internal/controlplane/handlers_authz.go
jhrozek: @puerco the email regex was actually removed to let the e-mail sending service verify it, not minder (it's unlikely we'd ever get the regex right for all weird cases)

internal/controlplane/handlers_authz.go
jhrozek: but in this case it's some unexpected error, I think Internal is OK in this case? e.g. the database connection was lost and such

jhrozek (approved):

rdimitrov:
@jhrozek, @puerco - thanks for the reviews! 🙏

PR #3676, opened 2024-06-19
Added minder provider update CLI command.

### LINKED ISSUE #3509: Add a CLI to update provider config
After we merge PR #3468 we'll get the means to patch provider config via a call. But we also need to expose this functionality through the CLI so that the users could patch their configs.

blkt:
# Summary

Update command allows changing configuration for provider on a per-field basis. It operates on a read-modify-write fashion, retrieving current configuration from the backend for modification.

Modification is done by reflection by walking over the struct's JSON tags. Implementation is recursive, but the maximum depth is determined by the deepest field in the configuration struct.

Argument parsing is trivial and assumes that arguments are either of the form `field1.field2.field3` (for `--unset-attribute`) or `field1.field2.field3=value` (for `--set-attribute`). The right parser for `value` is determined once the correct struct field is found.

It is not currently possible to modify `"github-app"` or `"github"` fields.

Fixes #3509

## Change Type

# Testing

* unit tested
* locally tested
* tested against SAAS (staging)

# Review Checklist:

### DIFF
--- a/cmd/cli/app/provider/provider_update.go
+++ b/cmd/cli/app/provider/provider_update.go
@@ -1,0 +1,509 @@
+//
+// Copyright 2024 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package provider
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/structpb"
+
+	"github.com/stacklok/minder/internal/util/cli"
+	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+)
+
+var (
+	errMissingProviderName = errors.New("provider name flag is missing")
+	errMissingProject      = errors.New("project flag is missing")
+)
+
+var updateCmd = &cobra.Command{
+	Use:   "update",
+	Short: "Updates a provider's configuration",
+	Long: `The minder provider update command allows a user to update a provider's
+configuration after enrollement.`,
+	RunE: cli.GRPCClientWrapRunE(UpdateProviderCommand),
+}
+
+type parser func(string) (*reflect.Value, error)
+
+func parseBool(val string) (*reflect.Value, error) {
+	v, err := strconv.ParseBool(val)
+	if err != nil {
+		return nil, fmt.Errorf("expected bool, got %s", val)
+	}
+	res := reflect.ValueOf(v)
+	return &res, nil
+}
+
+func parseInt(size int) parser {
+	return func(val string) (*reflect.Value, error) {
+		v, err := strconv.ParseInt(val, 10, size)
+		if err != nil {
+			return nil, fmt.Errorf("expected integer, got %s", val)
+		}
+		res := reflect.ValueOf(v)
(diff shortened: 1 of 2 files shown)


cmd/cli/app/provider/provider_update.go
```diff
@@ -0,0 +1,509 @@
...
+func UpdateProviderCommand(
+	ctx context.Context,
+	cmd *cobra.Command,
+	_ []string,
+	conn *grpc.ClientConn,
+) error {
+	// TODO: get rid of provider flag, only use class
+	providerName := viper.GetString("name")
+	if providerName == "" {
+		return cli.MessageAndError(
```
jhrozek, line 131: I think you can use `updateCmd.MarkFlagRequired("name")` in  `init` and then cobra catches the missing flag for you.

cmd/cli/app/provider/provider_update.go
```diff
@@ -0,0 +1,509 @@
...
+		Context: &minderv1.Context{
+			Project: &project,
+		},
+		Name: providerName,
+	})
+	if err != nil {
+		return cli.MessageAndError("Failed to get provider", err)
+	}
+
+	provider := resp.GetProvider()
```
jhrozek, line 160: do you think it makes sense to add a non-nil check for provider? I know it shouldn't happen if we check the error, so maybe it's not needed, but my C programmer past compels me to check all pointers :-)

cmd/cli/app/provider/provider_update.go
```diff
@@ -0,0 +1,509 @@
...
+	if err != nil {
+		// TODO this is likely to be an internal error and
+		// should be mapped to a more suitable user-facing
+		// error.
+		return cli.MessageAndError(
+			"invalid config",
+			fmt.Errorf("error marshalling provider config: %w", err),
+		)
+	}
+	serde := &configStruct{}
```
jhrozek, line 171: I like the name :-)

cmd/cli/app/provider/provider_update.go
```diff
@@ -0,0 +1,509 @@
...
+			return errors.New("too long")
+		}
+		if attrValue != nil {
+			parserFunc, found := parserMap[config.Kind()]
+			if !found {
+				return fmt.Errorf("field has unexpected kind: %s", config.Kind())
+			}
+			v, err := parserFunc(*attrValue)
+			if err != nil {
+				return fmt.Errorf("expected bool, got %s", *attrValue)
```
jhrozek, line 356: the literal bool in the error message is likely a typo?

jhrozek (approved):
This is great. I left some minor comments inline but I'm approving nonetheless so you can go ahead and merge.

cmd/cli/app/provider/provider_update.go
```diff
@@ -0,0 +1,509 @@
...
+		reflect.String:  identity,       // identity
+	}
+)
+
+type configStruct struct {
+	*minderv1.ProviderConfig
+	//nolint:lll
+	GitHub *minderv1.GitHubProviderConfig `json:"github,omitempty" yaml:"github" mapstructure:"github" validate:"required"`
+	//nolint:lll
+	GitHubApp *minderv1.GitHubAppProviderConfig `json:"github_app,omitempty" yaml:"github_app" mapstructure:"github_app" validate:"required"`
```
jhrozek, line 116: Do you have any thoughts on how would we handle additional custom providers?

cmd/cli/app/provider/provider_update.go
blkt: Yes, thanks for catching it.

blkt: I changed it to this
```
$ minder provider update -j ... -n ... --set-attribute auto_registration.entities.repository.enabled=50
...
Message: invalid attribute
Details: auto_registration.entities.repository.enabled is not a valid attribute: expected bool, got 50
```

cmd/cli/app/provider/provider_update.go
blkt: I haven't looked too much into this, but I believe that custom providers should ship their configuration structure as some form of spec (e.g. JSON Spec) so that every client can perform validation and guarantee type safety when parsing/producing a config object. That would also make it possible to dynamically produce valid attribute names as help strings for example.

Moreover, configuration structure is a problem already, because it's untyped outside of Minder Server, which is why I had to duplicate that struct. I think for our main providers we should be already shipping a spec for the config, and allow (or mandate) custom providers to do the same in the future.

jhrozek (approved):

Issue #3556, opened 2024-06-07
Implement new evaluation history tracking logic


dmjb:
### Please describe the enhancement

As described in design doc in epic.

### Solution Proposal

As described in design doc in epic.

PR #3688, opened 2024-06-21
Define evaluation metrics

dmjb:
Add in some basic metrics to track rule/entity evaluations to the engine. Note that the metrics for alerts and remediations are not wired in yet since they depend on some changes which will be added in the next PR.

I have taken a different approach to handling noop metrics than is used in the codebase so far: instead of defining an interface and separate real vs noop implementations, I have decided to make use of the noop metrics handling built into otel: there is now a metrics meter factory with alternative implementation for noop vs "real" metrics. Any part of the code which needs to create metrics can use the factory without caring whether the metrics are actually exported or not.

Relates to: #3556

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/entities/utils.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/google/uuid"
+
+	"github.com/stacklok/minder/internal/db"
+)
+
+// EntityFromIDs takes the IDs of the three known entity types and
+// returns a single ID along with the type of the entity.
+// This assumes that exactly one of the IDs is not equal to uuid.Nil
+func EntityFromIDs(
```
jhrozek, line 29: This is smart, but I worry that the code might get fragile in the future. Since the caller already passes attributes of `interfaces.EvalStatusParams`, why not pass its `EntityType` attribute directly?

jhrozek (commented):
Looking good, one question inline.

internal/entities/utils.go
dmjb: The problem is that the way we refer to entities right now requires us to pass around a UUID field for each type of entity, so if another entity gets added we will need to add another parameter anyway.

I would love to come up with a more elegant way of managing entities in the codebase, but it's definitely a future problem.

jhrozek (approved):

PR #3689, opened 2024-06-21
Refactor UpdateRole and add display names to invite-related responses

rdimitrov:
# Summary

The following PR includes a few updates on the user management implementation. 

The changes that are introduced are:
* Moved the invite update logic to happen under the UpdateRole handler instead of being part of the AssignRole handler
* Added a few display names based on the FE feedback
* Reflected these changes on the CLI
* Refactored the part of create/deleting an invitation to ensure proper behaviour
* Added a protection so minder discards processing the invite if the user tries to accept an invite for a role he already has and another one where we delete his current role assignment and accept the new one ensuring he should have only one assignment per project.
* Fixed the missing invite code in ListRoleAssignments and a bug where the Sponsor field was empty
* Updated the `minder role grant list` command to also print current invitations along with current role assignments
* Removed the checks for `flags.IDPResolver`

## Change Type

# Testing

Locally.

# Review Checklist:

### DIFF
--- a/cmd/cli/app/project/role/role_grant_list.go
+++ b/cmd/cli/app/project/role/role_grant_list.go
@@ -18,7 +18,9 @@ package role
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -78,17 +80,30 @@ func GrantListCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn 
 		}
 		cmd.Println(out)
 	case app.Table:
-		t := initializeTableForGrantList()
+		t := initializeTableForGrantListRoleAssignments()
 		for _, r := range resp.RoleAssignments {
-			t.AddRow(r.Subject, r.Role)
+			t.AddRow(r.Subject, r.Role, *r.Project)
 		}
 		t.Render()
+		if len(resp.Invitations) > 0 {
+			t := initializeTableForGrantListInvitations()
+			for _, r := range resp.Invitations {
+				t.AddRow(r.Email, r.Role, r.SponsorDisplay, r.ExpiresAt.AsTime().Format(time.RFC3339), strconv.FormatBool(r.Expired), r.Code)
+			}
+			t.Render()
+		} else {
+			cmd.Println("No pending invitations found.")
+		}
 	}
 	return nil
 }
 
-func initializeTableForGrantList() table.Table {
-	return table.New(table.Simple, layouts.Default, []string{"Subject", "Role"})
+func initializeTableForGrantListRoleAssignments() table.Table {
+	return table.New(table.Simple, layouts.Default, []string{"Subject", "Role", "Project"})
+}
+
+func initializeTableForGrantListInvitations() table.Table {
+	return table.New(table.Simple, layouts.Default, []string{"Invitee", "Role", "Sponsor", "Expires At", "Expired", "Code"})
 }
 
 func init() {
--- a/cmd/cli/app/project/role/role_update.go
+++ b/cmd/cli/app/project/role/role_update.go
@@ -17,6 +17,7 @@ package role
 
 import (
 	"context"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -38,40 +39,64 @@ to a user (subject) on a particular project.`,
 func UpdateCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc.ClientConn) error {
 	client := minderv1.NewPermissionsServiceClient(conn)
 
-	sub := viper.GetString("sub")
 	r := viper.GetString("role")
 	project := viper.GetString("project")
+	sub := viper.GetString("sub")
+	email := viper.GetString("email")
 
 	// No longer print usage on returned error, since we've parsed our inputs
 	// See https://github.com/spf13/cobra/issues/340#issuecomment-374617413
 	cmd.SilenceUsage = true
 
-	ret, err := client.UpdateRole(ctx, &minderv1.UpdateRoleRequest{
+	req := &minderv1.UpdateRoleRequest{
 		Context: &minderv1.Context{
 			Project: &project,
 		},
 		Roles:   []string{r},
 		Subject: sub,
-	})
-	if err != nil {
-		return cli.MessageAndError("Error updating role", err)
+	}
+	failMsg := "Error updating role"
+	successMsg := "Updated role successfully."
+	if email != "" {
+		req.Email = email
+		failMsg = "Error updating an invite"
+		successMsg = "Invite updated successfully."
 	}
 
-	cmd.Println("Update role successfully.")
-	cmd.Printf(
-		"Subject \"%s\" is now assigned to role \"%s\" on project \"%s\"\n",
-		ret.RoleAssignments[0].Subject,
-		ret.RoleAssignments[0].Role,
-		*ret.RoleAssignments[0].Project,
-	)
+	ret, err := client.UpdateRole(ctx, req)
+	if err != nil {
+		return cli.MessageAndError(failMsg, err)
+	}
 
+	cmd.Println(successMsg)
+
+	if email != "" {
+		t := initializeTableForGrantListInvitations()
+		for _, r := range ret.Invitations {
+			expired := "No"
+			if r.Expired {
+				expired = "Yes"
+			}
+			t.AddRow(r.Email, r.Role, r.Sponsor, r.ExpiresAt.AsTime().Format(time.RFC3339), expired, r.Code)
+		}
+		t.Render()
+		return nil
+	}
+	// Otherwise, print the role assignments if it was about updating a role
+	t := initializeTableForGrantListRoleAssignments()
+	for _, r := range ret.RoleAssignments {
+		t.AddRow(r.Subject, r.Role, *r.Project)
+	}
+	t.Render()
 	return nil
 }
 
 func init() {
 	RoleCmd.AddCommand(updateCmd)
 
-	updateCmd.Flags().StringP("sub", "s", "", "subject to update role access for")
 	updateCmd.Flags().StringP("role", "r", "", "the role to update it to")
-	updateCmd.MarkFlagsRequiredTogether("sub", "role")
+	updateCmd.Flags().StringP("sub", "s", "", "subject to update role access for")
+	updateCmd.Flags().StringP("email", "e", "", "email to send invitation to")
+	updateCmd.MarkFlagsOneRequired("sub", "email")
+	updateCmd.MarkFlagsMutuallyExclusive("sub", "email")
 }
--- a/database/query/invitations.sql
+++ b/database/query/invitations.sql
@@ -4,7 +4,7 @@
 -- the invitee.
 
 -- name: ListInvitationsForProject :many
-SELECT user_invites.email, role, users.identity_subject, user_invites.created_at, user_invites.updated_at
+SELECT user_invites.email, role, users.identity_subject, user_invites.created_at, user_invites.updated_at, user_invites.code
 FROM user_invites
   JOIN users ON user_invites.sponsor = users.id
 WHERE project = $1;
@@ -14,15 +14,22 @@ WHERE project = $1;
 -- to allow them to accept invitations even if email delivery was not working.
 -- Note that this requires that the destination email address matches the email
 -- address of the logged in user in the external identity service / auth token.
+-- This clarification is related solely for user's ListInvitations calls and does
+-- not affect to resolving invitations intended for other mail addresses.
 
 -- name: GetInvitationsByEmail :many
-SELECT * FROM user_invites WHERE email = $1;
+SELECT user_invites.*, users.identity_subject
+FROM user_invites
+  JOIN users ON user_invites.sponsor = users.id
+WHERE email = $1;
 
--- GetInvitationByEmailAndProjectAndRole retrieves an invitation by email, project,
--- and role.
+-- GetInvitationsByEmailAndProject retrieves all invitations by email and project.
 
--- name: GetInvitationByEmailAndProjectAndRole :one
-SELECT * FROM user_invites WHERE email = $1 AND project = $2 AND role = $3;
+-- name: GetInvitationsByEmailAndProject :many
+SELECT user_invites.*, users.identity_subject
+FROM user_invites
+  JOIN users ON user_invites.sponsor = users.id
+WHERE email = $1 AND project = $2;
 
 -- GetInvitationByCode retrieves an invitation by its code. This is intended to
 -- be called by a user who has received an invitation email and is following the
@@ -30,7 +37,10 @@ SELECT * FROM user_invites WHERE email = $1 AND project = $2 AND role = $3;
 -- invitation.
 
 -- name: GetInvitationByCode :one
-SELECT * FROM user_invites WHERE code = $1;
+SELECT user_invites.*, users.identity_subject
+FROM user_invites
+  JOIN users ON user_invites.sponsor = users.id
+WHERE code = $1;
 
 -- CreateInvitation creates a new invitation. The code is a secret that is sent
 -- to the invitee, and the email is the address to which the invitation will be
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -358,6 +358,21 @@ func (s *Server) inviteUser(
 	}
 
 	// If there are no invitations for this email, great, we should create one
+
+	// Resolve the sponsor's identity and display name
+	identity, err := s.idClient.Resolve(ctx, currentUser.IdentitySubject)
+	if err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error resolving identity")
+		return nil, util.UserVisibleError(codes.NotFound, "could not find identity %q", currentUser.IdentitySubject)
+	}
+
+	// Resolve the target project's display name
+	prj, err := s.store.GetProjectByID(ctx, targetProject)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get target project: %s", err)
+	}
+
+	// Create the invitation
 	userInvite, err = s.store.CreateInvitation(ctx, db.CreateInvitationParams{
 		Code:    invite.GenerateCode(),
 		Email:   email,
@@ -369,20 +384,7 @@ func (s *Server) inviteUser(
 		return nil, status.Errorf(codes.Internal, "error creating invitation: %v", err)
 	}
 
-	// Resolve the sponsor's identity and display name
-	identity, err := s.idClient.Resolve(ctx, currentUser.IdentitySubject)
-	if err != nil {
-		zerolog.Ctx(ctx).Error().Err(err).Msg("error resolving identity")
-		return nil, util.UserVisibleError(codes.NotFound, "could not find identity %q", currentUser.IdentitySubject)
-	}
-
 	// TODO: Publish the event for sending the invitation email
-
-	// Resolve the project's display name
-	prj, err := s.store.GetProjectByID(ctx, userInvite.Project)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get project: %s", err)
-	}
 
 	// Send the invitation response
 	return &minder.AssignRoleResponse{
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -239,6 +239,8 @@ func (s *Server) ListRoleAssignments(
 		return nil, status.Errorf(codes.Internal, "error getting role assignments: %v", err)
 	}
 
+	// Resolve the display names for the subjects
+	mapIdToDisplay := make(map[string]string, len(as))
 	for i := range as {
 		identity, err := s.idClient.Resolve(ctx, as[i].Subject)
 		if err != nil {
@@ -246,23 +248,13 @@ func (s *Server) ListRoleAssignments(
 			zerolog.Ctx(ctx).Error().Err(err).Msg("error resolving identity")
 			continue
 		}
-		as[i].Subject = identity.String()
+		as[i].DisplayName = identity.Human()
+		if mapIdToDisplay[as[i].Subject] == "" {
+			mapIdToDisplay[as[i].Subject] = identity.Human()
+		}
 	}
 
 	if flags.Bool(ctx, s.featureFlags, flags.UserManagement) {
-		mapIdToDisplay := make(map[string]string, len(as))
-		for i := range as {
-			if mapIdToDisplay[as[i].Subject] == "" {
-				user, err := s.idClient.Resolve(ctx, as[i].Subject)
-				if err != nil {
-					// if we can't resolve the subject, report the raw ID value
-					zerolog.Ctx(ctx).Error().Err(err).Str("user", as[i].Subject).Msg("error resolving user")
-					continue
-				}
-				mapIdToDisplay[as[i].Subject] = user.Human()
-			}
-			as[i].DisplayName = mapIdToDisplay[as[i].Subject]
-		}
 		// Add invitations, which are only stored in the Minder DB
 		invites, err := s.store.ListInvitationsForProject(ctx, targetProject)
 		if err != nil {
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -18,10 +18,8 @@ import (
 	"context"
 	"database/sql"
 	"errors"
-	"time"
 
 	"github.com/google/uuid"
-	gauth "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/auth"
 	"github.com/rs/zerolog"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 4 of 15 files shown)


internal/controlplane/handlers_authz.go
```diff
@@ -241,31 +239,22 @@ func (s *Server) ListRoleAssignments(
 		return nil, status.Errorf(codes.Internal, "error getting role assignments: %v", err)
 	}
 
-	if flags.Bool(ctx, s.featureFlags, flags.IDPResolver) {
```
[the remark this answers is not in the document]
rdimitrov: We are deprecating it and making this the default. cc: @evankanderson

internal/controlplane/handlers_authz.go
```diff
@@ -344,85 +334,70 @@ func (s *Server) inviteUser(
...
-		UserID:    currentUser.IdentitySubject,
-		HumanName: currentUser.IdentitySubject,
-	}
-	if flags.Bool(ctx, s.featureFlags, flags.IDPResolver) {
-		identity, err = s.idClient.Resolve(ctx, currentUser.IdentitySubject)
-		if err != nil {
-			zerolog.Ctx(ctx).Error().Err(err).Msg("error resolving identity")
-			return nil, util.UserVisibleError(codes.NotFound, "could not find identity %q", currentUser.IdentitySubject)
-		}
+	identity, err := s.idClient.Resolve(ctx, currentUser.IdentitySubject)
```
jhrozek, line 373: We create an invitation then resolve the ID. What if resolving the ID fails for some reason? Should we instead wrap creating the invite and resolving the project ID into a transaction or just create the invite as the last step?

The concern is that if something fails and the invite is created, but the reply couldn't be sent, is it easy to delete this invite and start over?

internal/controlplane/handlers_authz.go
```diff
@@ -344,85 +334,70 @@ func (s *Server) inviteUser(
...
+	if err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error resolving identity")
+		return nil, util.UserVisibleError(codes.NotFound, "could not find identity %q", currentUser.IdentitySubject)
```
jhrozek, line 376: is it correct to return this error code and error message always?

internal/controlplane/handlers_authz.go
```diff
@@ -530,32 +495,70 @@ func (s *Server) removeInvite(
 	role authz.Role,
 	email string,
 ) (*minder.RemoveRoleResponse, error) {
-	prj := targetPrj.String()
-	// Get all invitations for this email, project and role
-	inviteToRemove, err := s.store.GetInvitationByEmailAndProjectAndRole(ctx, db.GetInvitationByEmailAndProjectAndRoleParams{
+	// Get all invitations for this email and project
+	invitesToRemove, err := s.store.GetInvitationsByEmailAndProject(ctx, db.GetInvitationsByEmailAndProjectParams{
```
jhrozek, line 499: Would it be possible to have an SQL query to find an invite by role, e-mail and project (or extend GetInvitationsByEmailAndProject to accept an optional role)? Then you wouldn't have to do the filtering yourself and could just check if you got 0 replies, 1 or more than 1 which would be an internal error.

internal/controlplane/handlers_authz.go
```diff
@@ -530,32 +495,70 @@ func (s *Server) removeInvite(
...
 		Email:   email,
 		Project: targetPrj,
-		Role:    role.String(),
```
jhrozek, line 538: oh this removed line seems like you were doing that before?

internal/controlplane/handlers_authz.go
rdimitrov: That was literally the line I removed :D

internal/controlplane/handlers_authz.go
rdimitrov: Exactly, my reasoning was we discussed whether a user can have more than 1 role within a project (not allowed for now), so I decided to allow for this in the future.

internal/controlplane/handlers_authz.go
```diff
@@ -530,32 +495,70 @@ func (s *Server) removeInvite(
 	role authz.Role,
 	email string,
 ) (*minder.RemoveRoleResponse, error) {
-	prj := targetPrj.String()
-	// Get all invitations for this email, project and role
-	inviteToRemove, err := s.store.GetInvitationByEmailAndProjectAndRole(ctx, db.GetInvitationByEmailAndProjectAndRoleParams{
+	// Get all invitations for this email and project
```
jhrozek, line 500: A bit of a side-note, but I wonder if extracting all those non-exported methods like `removeInvite`, `removeRole` etc into interfaces (RoleService and InviteService maybe?) might make testing the area of code easier and would encapsulate the code better to only rely on what they need and not the whole big Server structure. Not for this PR though.

internal/controlplane/handlers_authz.go
```diff
@@ -632,23 +625,120 @@ func (s *Server) UpdateRole(ctx context.Context, req *minder.UpdateRoleRequest)
...
+		return nil, util.UserVisibleError(codes.NotFound, "no invitations found for this email and project")
+	} else if len(existingInvites) > 1 {
+		return nil, status.Errorf(codes.Internal, "multiple invitations found for this email and project")
+	}
+
+	// At this point, there should be exactly 1 invitation. We should either update its expiration or
+	// discard it and create a new one
+	if existingInvites[0].Role != authzRole.String() {
+		// If there's an existing invite with a different role, we should delete it and create a new one
+		// Delete the existing invitation
```
jhrozek, line 676: Is this OK without a transaction?

internal/controlplane/handlers_authz.go
```diff
@@ -632,23 +625,120 @@ func (s *Server) UpdateRole(ctx context.Context, req *minder.UpdateRoleRequest)
...
+			return nil, status.Errorf(codes.Internal, "error updating invitation: %v", err)
 		}
 	}
+	// Resolve the project's display name
+	prj, err := s.store.GetProjectByID(ctx, userInvite.Project)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get project: %s", err)
+	}
+	// Resolve the sponsor's identity and display name
+	identity, err := s.idClient.Resolve(ctx, currentUser.IdentitySubject)
```
jhrozek, line 705: Should this be done earlier to avoid touching the DB if we can't resolve the identity (wrapping the db writes in a transaction would help, too)

internal/controlplane/handlers_user.go
```diff
@@ -272,53 +273,41 @@ func (s *Server) ListInvitations(ctx context.Context, _ *pb.ListInvitationsReque
...
-	token, err := s.jwt.ParseAndValidate(tokenString)
+	tokenEmail, err := auth.GetUserEmailFromContext(ctx)
 	if err != nil {
-		return nil, status.Errorf(codes.Unauthenticated, "failed to parse bearer token: %v", err)
+		return nil, status.Errorf(codes.Internal, "failed to get user email: %s", err)
 	}
 
 	// Get the list of invitations for the user
-	invites, err := s.store.GetInvitationsByEmail(ctx, token.Email())
+	invites, err := s.store.GetInvitationsByEmail(ctx, tokenEmail)
```
jhrozek, line 282: This works only for invites for the address the user has in their token right?

jhrozek: Is there a way to list the invites by the identity instead and look up the identity by e-mail?

jhrozek (commented):
A bunch of questions inline

internal/controlplane/handlers_authz.go
rdimitrov: Addressed it in 1d54fdcfdf96692ac991451fba69eb8ac2c4a382

internal/controlplane/handlers_user.go
rdimitrov: Should we expect that the same user identity can have multiple emails in their token (they get that from Keycloak right?) The idea was to check that for listing invitations, but still allow the user to accept/decline invitations by their code (without cross-checking the email)

internal/controlplane/handlers_authz.go
rdimitrov: Addressed it in 4eb2fe6674474acfb331ec1c1daeae83cad6d8a2

internal/controlplane/handlers_authz.go
rdimitrov: Addressed it in - 4eb2fe6674474acfb331ec1c1daeae83cad6d8a2

internal/controlplane/handlers_authz.go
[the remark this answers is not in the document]
jhrozek: thanks

internal/controlplane/handlers_authz.go
jhrozek: thanks!

internal/controlplane/handlers_authz.go
jhrozek: 👍🏻

internal/controlplane/handlers_authz.go
jhrozek: now you know that I add comments in a stream without first reading the whole PR ;-)

internal/controlplane/handlers_authz.go
jhrozek: 👍🏻

internal/controlplane/handlers_authz.go
```diff
@@ -567,29 +572,22 @@ func (s *Server) removeRole(
...
 	}
 
 	// Verify if user exists
-	if _, err := s.store.GetUserBySubject(ctx, identity.UserID); err != nil {
+	if _, err := s.store.GetUserBySubject(ctx, identity.String()); err != nil {
 		if errors.Is(err, sql.ErrNoRows) {
 			return nil, util.UserVisibleError(codes.NotFound, "User not found")
 		}
 		return nil, status.Errorf(codes.Internal, "error getting user: %v", err)
 	}
```
evankanderson, line 588: Why do we need this at all?

internal/controlplane/handlers_authz.go
```diff
@@ -567,29 +572,22 @@ func (s *Server) removeRole(
...
 
+	// Delete the role assignment
 	if err := s.authzClient.Delete(ctx, identity.String(), role, targetProject); err != nil {
 		return nil, status.Errorf(codes.Internal, "error writing role assignment: %v", err)
 	}
```
evankanderson, line 593: Do we want to allow a user to remove themselves from the project?

evankanderson: We probably also want to enforce the invariant that projects without parents have at least one admin after `RemoveRole`.

internal/controlplane/handlers_authz.go
```diff
@@ -632,23 +627,134 @@ func (s *Server) UpdateRole(ctx context.Context, req *minder.UpdateRoleRequest)
...
 
-	// We may be given a human-readable identifier which can vary over time. Resolve
-	// it to an IDP-specific stable identifier so that we can support subject renames.
-	identity := &auth.Identity{
-		Provider:  nil,
-		UserID:    sub,
-		HumanName: sub,
+	// Validate the subject and email - decide if it's about updating an invitation or a role assignment
+	if sub == "" && email != "" {
```
evankanderson, line 631: IMO, this feels like it should be "if specified user is not in current OpenFGA roles, then look for invitations..."

internal/controlplane/handlers_user.go
jhrozek: OK, maybe I'm missing what is the list invitations for. Is it supposed to be used by the user after they had logged in? Or by the UI?

I'm concerned about 2 things with respect to the invitations - since we would get the e-mail from keycloak we 1) can't rely on having the e-mail available I think or 2) it could be unrelated to the e-mail where invites are supposed to be sent (e.g. my e-mail address is available but it's my personal e-mail)

jhrozek: OK, but looking at the invite table, I don't think we have any other data to cross-reference the invite with. So this is more to be careful where/how we use that endpoint - it might not give us a complete list I think.

internal/controlplane/handlers_authz.go
evankanderson: But I'm okay leaving it for now.

internal/controlplane/handlers_user.go
rdimitrov: Yes, this is supposed to be called from a logged in user to check their pending invitations (both cli and ui). My assumption is we can rely on having the mail in the token since I don't think we have a use case where this is not true, i.e logging through GitHub does set the primary mail there. If the invite is sent to your company email and for whatever reason you want to accept it from your minder account using your personal mail you can still do that through the accept/decline flow but you are correct that it will not be shown via ListInvitations (so far this is the intended behaviour).

rdimitrov: Also invitations work solely through using the email(for the time being you cannot send an invite to a user using their minder account)

internal/controlplane/handlers_authz.go
```diff
@@ -241,31 +239,22 @@ func (s *Server) ListRoleAssignments(
...
-				zerolog.Ctx(ctx).Error().Err(err).Msg("error resolving identity")
-				continue
-			}
-			as[i].Subject = identity.Human()
+	// Resolve the display names for the subjects
+	mapIdToDisplay := make(map[string]string, len(as))
+	for i := range as {
+		identity, err := s.idClient.Resolve(ctx, as[i].Subject)
+		if err != nil {
+			// if we can't resolve the subject, report the raw ID value
```
jhrozek, line 247: this comment is no longer valid is it? If we can't resolve the subject, we would then just use the zero value (empty string). Or did you mean to first initialize the map with the raw ID value before trying to resolve it?

internal/controlplane/handlers_authz.go
```diff
@@ -530,32 +497,70 @@ func (s *Server) removeInvite(
...
+	sponsorUser, err := s.store.GetUserByID(ctx, ret.Sponsor)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get user: %s", err)
+	}
+
+	// Resolve the sponsor's identity and display name
+	identity, err := s.idClient.Resolve(ctx, sponsorUser.IdentitySubject)
+	if err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error resolving identity")
+		return nil, util.UserVisibleError(codes.NotFound, "could not find identity %q", sponsorUser.IdentitySubject)
```
jhrozek, line 548: Does this need to be a fatal error? (Ditto for getting the sponsorUser above) - could we just return some raw ID? Or do you think it's better to just return all well-formed as the UI expects it or fail?

jhrozek (approved):
Only two more small comments, but non-blocking.

internal/controlplane/handlers_authz.go
```diff
@@ -632,23 +627,134 @@ func (s *Server) UpdateRole(ctx context.Context, req *minder.UpdateRoleRequest)
...
+	}
+	return nil, util.UserVisibleError(codes.InvalidArgument, "one of subject or email must be specified")
+}
+
+func (s *Server) updateInvite(
+	ctx context.Context,
+	targetProject uuid.UUID,
+	authzRole authz.Role,
+	email string,
+) (*minder.UpdateRoleResponse, error) {
```
evankanderson, line 648: I wish this could be an `upsert`... I'm assuming that doesn't make sense given our RPCs, though, right?

internal/controlplane/handlers_authz.go
```diff
@@ -632,23 +627,134 @@ func (s *Server) UpdateRole(ctx context.Context, req *minder.UpdateRoleRequest)
...
+	// Get all invitations for this email and project
+	existingInvites, err := s.store.GetInvitationsByEmailAndProject(ctx, db.GetInvitationsByEmailAndProjectParams{
+		Email:   email,
+		Project: targetProject,
+	})
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting invitations: %v", err)
+	}
+
+	// Exit early if there are no or multiple existing invitations for this email and project
```
evankanderson, line 665: What's the recovery method if there are multiple existing invites?  Is this something a user can resolve on their own?

internal/controlplane/handlers_authz.go
```diff
@@ -632,23 +627,134 @@ func (s *Server) UpdateRole(ctx context.Context, req *minder.UpdateRoleRequest)
...
+
+	// Begin a transaction to ensure that the invitation is updated atomically
+	tx, err := s.store.BeginTransaction()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error starting transaction: %v", err)
+	}
+	defer s.store.Rollback(tx)
+
+	// At this point, there should be exactly 1 invitation. We should either update its expiration or
+	// discard it and create a new one
```
evankanderson, line 680: I'm confused about this logic, and when we extend an invite vs re-create it.  I *think* this code does:

- User A invites foo@email.com to be a Viewer
- 8 days pass (expires) and then User A invites foo@email.com as a Viewer again (*Update*)
- 2 days pass and User A updates the invite for foo@email.com to Editor (*Replace*)
- 5 days pass and User B updates the invite for foo@email.com to Editor (*Update*)
- 12 days pass (expires) and User C renews the invite for foo@email.com to Editor (*Update*)

Is that correct?  Why do we preserve the original invite for users B and C, but reset the creation date when A updates the role?

Do we need both a creation and and an update for the invitation?

internal/controlplane/handlers_authz.go
```diff
@@ -632,23 +627,134 @@ func (s *Server) UpdateRole(ctx context.Context, req *minder.UpdateRoleRequest)
...
+		return nil, util.UserVisibleError(codes.NotFound, "could not find identity %q", currentUser.IdentitySubject)
+	}
+
+	// Commit the transaction to persist the changes
+	if err := s.store.Commit(tx); err != nil {
+		return nil, status.Errorf(codes.Internal, "error committing transaction: %v", err)
+	}
+
+	return &minder.UpdateRoleResponse{
+		// Leaving the role assignment empty as it's an invitation
```
evankanderson, line 725: This comment reads strangely given `Invitations.Role`.  Do we need the comment at all?

evankanderson (approved):
I'm fine with addressing some/all of my feedback in another PR.  I am slightly concerned that our final user management will end up feeling complex to users because of the number of different states and transitions we support -- it would be nice if `AssignRole` could basically do something like the following:

```
if user in project.roles:
  # This is not "upsert", because the user already exists.
  project.update_role(user, role)
else:
  project.upsert_invite(user, role)
```

cmd/cli/app/project/role/role_update.go
```diff
@@ -38,40 +39,64 @@ to a user (subject) on a particular project.`,
 func UpdateCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc.ClientConn) error {
 	client := minderv1.NewPermissionsServiceClient(conn)
 
-	sub := viper.GetString("sub")
 	r := viper.GetString("role")
 	project := viper.GetString("project")
+	sub := viper.GetString("sub")
+	email := viper.GetString("email")
```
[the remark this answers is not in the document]
rdimitrov: Yes, there's the `updateCmd.MarkFlagsMutuallyExclusive("sub", "email")` which should cover this case

cmd/cli/app/project/role/role_grant_list.go
```diff
@@ -78,17 +80,30 @@ func GrantListCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn
...
 }
 
-func initializeTableForGrantList() table.Table {
-	return table.New(table.Simple, layouts.Default, []string{"Subject", "Role"})
+func initializeTableForGrantListRoleAssignments() table.Table {
+	return table.New(table.Simple, layouts.Default, []string{"Subject", "Role", "Project"})
+}
+
+func initializeTableForGrantListInvitations() table.Table {
+	return table.New(table.Simple, layouts.Default, []string{"Invitee", "Role", "Sponsor", "Expires At", "Expired", "Code"})
```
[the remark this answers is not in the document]
rdimitrov: Addressed in https://github.com/stacklok/minder/pull/3710

cmd/cli/app/project/role/role_update.go
```diff
@@ -38,40 +39,64 @@ to a user (subject) on a particular project.`,
...
-	cmd.Println("Update role successfully.")
-	cmd.Printf(
-		"Subject \"%s\" is now assigned to role \"%s\" on project \"%s\"\n",
-		ret.RoleAssignments[0].Subject,
-		ret.RoleAssignments[0].Role,
-		*ret.RoleAssignments[0].Project,
-	)
+	cmd.Println(successMsg)
 
+	if email != "" {
```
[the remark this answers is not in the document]
rdimitrov: Addressed in https://github.com/stacklok/minder/pull/3710

database/query/invitations.sql
```diff
@@ -4,7 +4,7 @@
 -- the invitee.
 
 -- name: ListInvitationsForProject :many
-SELECT user_invites.email, role, users.identity_subject, user_invites.created_at, user_invites.updated_at
+SELECT user_invites.email, role, users.identity_subject, user_invites.created_at, user_invites.updated_at, user_invites.code
```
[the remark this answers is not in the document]
rdimitrov: Addressed in https://github.com/stacklok/minder/pull/3710

internal/controlplane/handlers_authz.go
rdimitrov: Addressed in https://github.com/stacklok/minder/pull/3710

internal/controlplane/handlers_authz.go
```diff
@@ -278,10 +267,11 @@ func (s *Server) ListRoleAssignments(
 				Email:          i.Email,
 				Project:        targetProject.String(),
 				CreatedAt:      timestamppb.New(i.CreatedAt),
-				ExpiresAt:      timestamppb.New(i.UpdatedAt.Add(7 * 24 * time.Hour)),
-				Expired:        time.Now().After(i.UpdatedAt.Add(7 * 24 * time.Hour)),
+				ExpiresAt:      invite.GetExpireIn7Days(i.UpdatedAt),
+				Expired:        invite.IsExpired(i.UpdatedAt),
 				Sponsor:        i.IdentitySubject,
 				SponsorDisplay: mapIdToDisplay[i.IdentitySubject],
+				Code:           i.Code,
```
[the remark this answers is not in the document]
rdimitrov: Addressed in https://github.com/stacklok/minder/pull/3710

internal/controlplane/handlers_authz.go
```diff
@@ -304,7 +294,7 @@ func (s *Server) AssignRole(ctx context.Context, req *minder.AssignRoleRequest)
 	targetProject := entityCtx.Project.ID
 
 	// Ensure user is not updating their own role
-	err := s.isUserSelfUpdating(ctx, sub, email)
+	err := isUserSelfUpdating(ctx, sub, email)
```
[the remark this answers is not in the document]
rdimitrov: I've added this in https://github.com/stacklok/minder/pull/3709, so if you're the sponsor of the invite, you cannot resolve it which should cover this use case.

internal/controlplane/handlers_authz.go
```diff
@@ -325,7 +315,7 @@ func (s *Server) AssignRole(ctx context.Context, req *minder.AssignRoleRequest)
 		return nil, status.Errorf(codes.Internal, "error getting project: %v", err)
 	}
 
-	// Validate the subject and email - decide if it's an invitation or a role assignment
+	// Decide if it's an invitation or a role assignment
```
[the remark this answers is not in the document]
rdimitrov: My understanding was that if a user doesn't have a role in a project and an admin wants to invite them to the project, there are 2 ways to do that. 
* Send an invite via their email address. This works both for registered and unregistered Minder users. The role is updated once the invite is accepted.
* Grant the role directly using their uuid. This works only for registered users and doesn't have to go through the invite flow. The role is updated immediately.

Was this what we had in mind or adding someone to a project should **always** result in an invite and the affect should happen once the invite is resolved?

internal/controlplane/handlers_authz.go
rdimitrov: Addressed in https://github.com/stacklok/minder/pull/3710

internal/controlplane/handlers_authz.go
rdimitrov: Yes. if this happens the user can list all invitations and remove the ones that are conflicting. We do have protections against it when creating invites but still a valid point 👍

internal/controlplane/handlers_authz.go
```diff
@@ -434,28 +411,18 @@ func (s *Server) assignRole(
...
-		HumanName: subject,
-	}
-	if flags.Bool(ctx, s.featureFlags, flags.IDPResolver) {
-		identity, err = s.idClient.Resolve(ctx, subject)
-		if err != nil {
-			zerolog.Ctx(ctx).Error().Err(err).Msg("error resolving identity")
-			return nil, util.UserVisibleError(codes.NotFound, "could not find identity %q", subject)
-		}
+	// Resolve the subject to an identity
+	identity, err := s.idClient.Resolve(ctx, subject)
```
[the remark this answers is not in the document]
rdimitrov: Not sure as it might be considered similar but how about we move resolving the identity to happen after `GetUserBySubject`? This way we should try to resolve only if the caller knows the `identity_subject` of the invitee?

internal/controlplane/handlers_authz.go
rdimitrov: This was there prior to these changes, but my understanding is to fail early if we try to do something on a user it doesn't exist.

internal/controlplane/handlers_authz.go
rdimitrov: I'll ask @ethomson for the first one. And I agree on the 2nd one, I'll add it to the PR with the rest of the fixes 👍

rdimitrov: Ed confirmed on the 1st one 👍 I'll add a follow up.

internal/controlplane/handlers_authz.go
rdimitrov: Addressed in https://github.com/stacklok/minder/pull/3710

internal/controlplane/handlers_authz.go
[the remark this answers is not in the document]
rdimitrov: This was me not understanding the context. We should always use invitations to assign people to projects so this needs to be updated. Will do that in a follow-up 👍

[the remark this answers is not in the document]
evankanderson: To be fair, I think the context had changed over time from the first engineering implementation to designing the situation with product and realizing that we probably wanted positive user acceptance of the new permission (ergo, invites).

PR #3731, opened 2024-06-27
Add database table and methods for profile selectors

### LINKED ISSUE #3720: Add new database tables and the associated operations for profile selectors
We need to add the table and the basic CRUD operation for the table

jhrozek:
# Summary

- **Add the selectors table**
- **Add DB methods for selector management**

So far just adds the table and the methods and their tests. These will be used in later PRs.

Fixes #3720

## Change Type

# Testing

with the unit tests

# Review Checklist:


database/query/selectors.sql
```diff
@@ -0,0 +1,24 @@
...
+INSERT INTO profile_selectors (profile_id, entity, selector, comment)
+VALUES ($1, $2, $3, $4)
+RETURNING id, profile_id, entity, selector, comment;
+
+-- name: GetSelectorsByProfileID :many
+SELECT id, profile_id, entity, selector, comment
+FROM profile_selectors
+WHERE profile_id = $1;
+
+-- name: UpdateSelector :one
```
jhrozek, line 11: I really wanted to add an upsert, but since the only column to conflict on would be the ID, the upsert would be able to add a zero UUID. I tried working around that with a little more SQL code, like:
```
INSERT INTO profile_selectors (id, profile_id, entity, selector, comment)
VALUES (
   CASE WHEN $1::uuid = '00000000-0000-0000-0000-000000000000' THEN gen_random_uuid() ELSE $1::uuid END,
   $2::uuid, $3, $4, $5
)
ON CONFLICT (id) DO UPDATE
    SET profile_id = $2::uuid,
        entity = $3,
        selector = $4,
        comment = $5
RETURNING id, profile_id, entity, selector, comment;
```

but that breaks sqlc - it wasn't able to infer the type of $1 and generated parameters with a column named `Column_1` and type `interface{}`. Trying to use `sqlc.arg(id)` instead of `$1` didn't work either, because sqlc [apparently requires](https://github.com/sqlc-dev/sqlc/issues/3166) that all parameters be used in the query.

So I went with just an update now and will handle the upsert case in the code instead.

dmjb (approved):

database/query/selectors.sql
evankanderson: If `profile_id` is not unique, I'm not sure what the `upsert` would look like unless you always have the UUID already created (e.g. client-side generation).  `profile_id` is _not_ marked as unique at the moment, so I'm not sure that upsert is particularly useful -- you'll know whether you have an existing UUID or not, and can switch between the insert and update pretty easily.

evankanderson (commented):
LGTM.

PR #3735, opened 2024-06-27
Initial implementation of the invite email sending service

rdimitrov:
# Summary

The following PR is the initial implementation of the email sending service for invitations.

**Details:**
* Adds a new event topic to the message queue for sending emails
* Adds a common email client interface
* Adds a client implementation based on AWS SES
* Adds a config section in server-config

**What's left:** 
- [ ] Test it once there's a configured AWS SES service (Note: If it gets merged before setting up the things on the AWS side of course it will not work but it won't cause any crashes even if the flag is enabled)

Moving away from using a builtin template for the mail content has a separate issue and will be done in a follow-up.

## Change Type

# Testing

Example rendering of the template: 
image

# Review Checklist:


internal/email/email.go
```diff
@@ -0,0 +1,107 @@
...
+	Email          string `json:"email"`
+}
+
+// MailManager is the email manager
+type MailManager struct {
+	client Service
+}
+
+// NewMailManager creates a new mail manager
+func NewMailManager(client Service) *MailManager {
```
dmjb, line 54: I would call this something like `MailEventConsumer` or `MailEventHandler` to make its purpose more clear.

internal/controlplane/handlers_authz.go
```diff
@@ -394,7 +397,15 @@ func (s *Server) inviteUser(
 		return nil, status.Errorf(codes.Internal, "error creating invitation: %v", err)
 	}
 
-	// TODO: Publish the event for sending the invitation email
+	// Publish the event for sending the invitation email
```
dmjb, line 400: Non-blocking comment: I feel like we should move this invite logic into an `InviteService` type interface.

dmjb (changes requested):

internal/controlplane/handlers_authz.go
rdimitrov: thanks, this was already one of the follow-up AIs on my list 👍

internal/email/email.go
rdimitrov: I did not like this name too 😄 thanks! 👍

rdimitrov: Fixed 👍

internal/email/awsses.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ses"
+	"github.com/rs/zerolog"
+)
+
+const (
+	// awsUSEast1 is the AWS region
+	awsUSEast1 = "us-east-1"
```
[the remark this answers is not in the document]
rdimitrov: Addressed it

eleftherias (approved):

rdimitrov:
@dmjb - The comment about renaming this was addressed. Dismissing this so we can unblock merging the PR. Do get back to me in case you wanted me to address something once you're back from PTO and I'll do a follow-up 👍

internal/service/service.go
```diff
@@ -195,16 +197,25 @@ func AllInOneServerService(
...
 	}
-
 	evt.ConsumeEvents(rec)
 
+	// Register the installation manager to handle provider installation events
 	im := installations.NewInstallationManager(ghProviders)
 	evt.ConsumeEvents(im)
 
+	// Register the email manager to handle email invitations
+	mailClient, err := email.NewAWSSES(cfg.Email.AWSSES.Sender, cfg.Email.AWSSES.Region)
```
evankanderson, line 212: This feels like we should only do this block if `cfg.Email.AWSSES.Sender` is non-empty, but I guess we'll do that in the future to allow local development to not need AWS credentials set up.

evankanderson (commented):
LGTM, but adding a few comments after-the-fact, because I got to this late.

internal/email/template.go
```diff
@@ -0,0 +1,291 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package email
+
+//nolint:lll
+const (
+	// bodyHTML is the HTML body of the email
+	bodyHTML = `
```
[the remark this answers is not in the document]
rdimitrov: Yes, this is the initial approach and there'll be a follow-up where we can decide how to distribute the template.

internal/service/service.go
rdimitrov: Addressed in https://github.com/stacklok/minder/pull/3783

PR #3741, opened 2024-06-28
Cache RuleTypeEngine instances in Executor

dmjb:
This simplifies an upcoming PR, and also addresses a comment in the code about the need to cache rule type lookups. This PR changes the executor to load all relevant rule types before evaluation, and construct RuleTypeEngine instances for each. The RuleTypeEngine instances are stored in a cache and reused during rule evaluation, instead of the current approach which constructs a new RuleTypeEngine for each rule instance which is evaluated.

Certain aspects of this change (specifically, the need to query for the rule type ID) are a bit clunky, but will be simplified in an upcoming PR.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


internal/engine/rtengine/cache.go
```diff
@@ -0,0 +1,88 @@
...
+	provinfv1 "github.com/stacklok/minder/pkg/providers/v1"
+)
+
+// Cache contains a set of RuleTypeEngine instances
+type Cache interface {
+	GetRuleEngine(ruleTypeID uuid.UUID) (*RuleTypeEngine, error)
+}
+
+type ruleEngineCache struct {
+	engines map[uuid.UUID]*RuleTypeEngine
```
JAORMX, line 35: should this be a [syncmap](https://pkg.go.dev/golang.org/x/sync/syncmap) instead so this could be updated/read in parallel?

dmjb: Discussed this with @JAORMX offline: the current implementation populates all the rule type instances up front. As a result, we do not need to worry about concurrent read/write scenarios.

JAORMX (commented):
LGTM. Let's run some smoke tests before merging this. Wanna make sure we're in a good state before this.

internal/engine/interfaces/interface.go
```diff
@@ -131,7 +131,6 @@ type EvalStatusParams struct {
 	Result           *Result
 	Profile          *pb.Profile
 	Rule             *pb.Profile_Rule
-	RuleTypeName     string
```
dmjb, line 134: Populating the eval params struct with the rule type name was a pain, and it was only used in one place for evaluation telemetry. I decided instead to pass the rule type name via a different mechanism to the telemetry code.

JAORMX (approved):

PR #3772, opened 2024-07-03
Show the subject display name along with the ID in minder role

rdimitrov:
# Summary

The following PR includes the subject display name when listing or updating role assignments. Previously it was just the ID.

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/cmd/cli/app/project/role/role_grant_list.go
+++ b/cmd/cli/app/project/role/role_grant_list.go
@@ -81,7 +81,7 @@ func GrantListCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn 
 	case app.Table:
 		t := initializeTableForGrantListRoleAssignments()
 		for _, r := range resp.RoleAssignments {
-			t.AddRow(r.Subject, r.Role, *r.Project)
+			t.AddRow(fmt.Sprintf("%s / %s", r.DisplayName, r.Subject), r.Role, *r.Project)
 		}
 		t.Render()
 		if len(resp.Invitations) > 0 {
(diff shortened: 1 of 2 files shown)


cmd/cli/app/project/role/role_grant_list.go
```diff
@@ -81,7 +81,7 @@ func GrantListCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn
 	case app.Table:
 		t := initializeTableForGrantListRoleAssignments()
 		for _, r := range resp.RoleAssignments {
-			t.AddRow(r.Subject, r.Role, *r.Project)
+			t.AddRow(fmt.Sprintf("%s / %s", r.DisplayName, r.Subject), r.Role, *r.Project)
```
eleftherias, line 84: Do you think it's worth adding another column to the table, because the current column title is "Subject" and with this change it will contain both the subject and the display name.

rdimitrov: I did do it like that initially, but it stretched the table too much, so I thought to reuse what we do in the `auth` command (where projects are displayed in a similar pattern - `project_name / project_id`). I can make it in the other way if you think it would be better 👍

eleftherias: Maybe we include both in the same column, but change the title from "Subject" to "User"?

rdimitrov: Sounds good 👍 I'll push a commit 👍

rdimitrov: Done 👍

eleftherias (approved):

PR #3783, opened 2024-07-04
Consider the environment when generating the invite URLs

rdimitrov:
# Summary

The following PR:
* Adds a Minder base url field to the email server config section. This is used when generating the user invitation URL, so we can prefix it with the correct minder frontend, i.e. staging vs prod
* Adds the invite URL to the responses when an invite is created or updated
* Adds a noop email client in case there's no aws ses config present. This is helpful for local development where we don't expect having it. It will log the email content which was supposed to be sent and return nil
* Removed the wrapper and the interface since it felt obsolete. If someone wants to use another client, they can safely implement the `events.Consumer` interface and that should be enough. The message payload includes everything that is needed - email, subject, html and text bodies.
* Updates the minder base url to the example server config

## Change Type

# Testing

# Review Checklist:


internal/email/awsses/awsses.go
```diff
@@ -12,43 +12,37 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package email provides the email utilities for minder
-package email
+// Package awsses provides the email utilities for minder
+package awsses
```
lukehinds, line 16: I figure this is vendor agnostic and others could in time add for example a `gmail` `mailchimp` etc?

proto/minder/v1/minder.proto
```diff
@@ -2307,6 +2307,8 @@ message Invitation {
     // project_display is the display name of the project to which the user
     // is invited.
     string project_display = 10;
+    // inviteURL is the URL that can be used to accept the invitation.
+    string invite_url = 11;
```
jhrozek, line 2311: I haven't been following this feature too closely, but why is the invite URL configurable at all? Shouldn't this be something set by the server?

internal/email/awsses/awsses.go
rdimitrov: Yeah. initially I had a wrapper and this interface - 
```go
// Service is the email service interface
type Service interface {
	SendEmail(ctx context.Context, to, subject, bodyHTML, bodyText string) error
}
```
But it felt obsolete since the watermill event message payload now includes everything that is needed for another client to be hooked up. 

If we or someone else wants to add another email service provider, they have to implement the `Register` method (`events.Consumer` interface) and that would be enough.

internal/controlplane/handlers_authz.go
```diff
@@ -404,11 +407,32 @@ func (s *Server) inviteUser(
...
+	// Read the server config, so we can get the Minder base URL
+	cfg, err := config.ReadConfigFromViper[serverconfig.Config](viper.GetViper())
+	if err != nil {
+		return nil, fmt.Errorf("unable to read config: %w", err)
+	}
+
+	inviteURL := ""
+	if cfg.Email.MinderURLBase != "" {
+		// Create the invite URL
+		inviteURL = fmt.Sprintf("%s/join/%s", cfg.Email.MinderURLBase, userInvite.Code)
```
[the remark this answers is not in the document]
rdimitrov: of course, good idea 👍

proto/minder/v1/minder.proto
rdimitrov: I think both are true, because the invite URL is generated on the backend side, but it is based on where the FE server is. I decided to pass this from the environment through the server config. Is this what you meant?

internal/controlplane/handlers_authz.go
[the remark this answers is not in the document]
rdimitrov: Addressed it 👍 Also added the use of db transaction for both creating and updating an invite so we fail in case the base URL fails parsing.

proto/minder/v1/minder.proto
jhrozek: What I meant to ask is, is the invite URL something the user can affect or is it just given to the user in the reply? If it's just in the reply and the user can't change the invite base URL when they create the invite, then it's OK

rdimitrov: Ah, I see. No, the user has no control over the invite URL and how it's constructed 👍

jhrozek (approved):

PR #3784, opened 2024-07-04
Added ListEvaluationHistory RPC implementation.

### LINKED ISSUE #3746: Implement API endpoints for evaluation history
Handler for `/api/v1/history` has a stubbed implementation [here](https://github.com/stacklok/minder/blob/main/internal/controlplane/handlers_evalstatus.go#L34-L58).

Add the following
* service implementation under [`HistoryService`](https://github.com/stacklok/minder/blob/674252bea530231404895bc097cf74eeba8d03eb/internal/history/service.go#L34-L45)
* add necessary SQL statements
* add implementation of filtering and pagination

blkt:
# Summary

ListEvaluationHistory RPC returns a paginated list of evaluation events based on the given cursor and filter.

Routines for managing a filter are implemented as well, in a way that should be reusable within other endpoints supporting filters. This implementation only allows and-joined predicates. Such predicates allow only or-joined equality/inequality checks. Simply put, if a filter entry starts with the exclamation mark, it is added to the inequality check, while it is added to the equality check otherwise. Finally, timerange based filtering is supported.

Routines for managing a cursor are implemented as well, but are not intended to be generic. Cursors are tightly coupled with the underlying extraction logic and are harder to refactor, and the additional effort was not considered valuable at this time.

Fixes #3746

## Change Type

# Testing

* unit tests
* manual tests (locally)

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -2590,6 +2590,9 @@ message EvaluationHistoryStatus {
     // details contains optional details about the evaluation.
     // the structure and contents are rule type specific, and are subject to change.
     string details = 2;
+
+    // created_at is the timestamp of creation of this evaluation
+    google.protobuf.Timestamp evaluated_at = 3;
```
blkt, line 2595: This field was missing from the spec.
@dmjb is this the right place?

database/query/eval_history.sql
```diff
@@ -93,3 +93,71 @@ INSERT INTO alert_events(
...
+   AND (sqlc.slice(notProfileNames)::text[] IS NULL OR p.name != ANY(sqlc.slice(notProfileNames)::text[]))
+   AND (sqlc.slice(notRemediations)::remediation_status_types[] IS NULL OR re.status != ANY(sqlc.slice(notRemediations)::remediation_status_types[]))
+   AND (sqlc.slice(notAlerts)::alert_status_types[] IS NULL OR ae.status != ANY(sqlc.slice(notAlerts)::alert_status_types[]))
+   AND (sqlc.slice(notStatuses)::eval_status_types[] IS NULL OR s.status != ANY(sqlc.slice(notStatuses)::eval_status_types[]))
+   -- time range filter
+   AND (sqlc.narg(fromts)::timestamp without time zone IS NULL
+        OR sqlc.narg(tots)::timestamp without time zone IS NULL
+        OR s.most_recent_evaluation BETWEEN sqlc.narg(fromts) AND sqlc.narg(tots))
+ ORDER BY s.most_recent_evaluation DESC
+ LIMIT sqlc.arg(size)::integer;
```
eleftherias, line 163: This may have already been discussed, but have you considered using a SQL generator, like https://github.com/Masterminds/squirrel? I'm worried this raw SQL will become unmaintainable and fragile, especially as we add these types of filters to more endpoints.

blkt: I started off thinking it was not possible to implement this without an SQL generator, and decided to prove that.
It turned out to be possible instead, so I haven't looked any further.

It is worth noting that the simple filtering logic we decided to implement only allows where conditions of the form
```
(X = val1 OR X = val2) AND (Y = val3 OR Y = val4)
```
These conditions are always definable in static SQL, with an additional `column1 IS NOT NULL OR` at the beginning. Filtering and pagination as discussed in design docs do not allow for more complex filters, and in such cases general search should be implemented instead.

I was not aware of the existence of squirrel, while I originally considered goqu, I'll have a look at both. Besides, the query is fairly efficient and using sqlc we have all the structs and bindings autogenerated, which I don't think is the case using squirrel.

I understand the query generator approach may still be preferable, I'll test squirrel on top of this branch to see what's the impact.

proto/minder/v1/minder.proto
dmjb: Yes, that makes sense

internal/history/service.go
```diff
@@ -195,3 +204,352 @@ type ruleEntityParams struct {
...
+func convert[
+	T db.Entities |
+		db.RemediationStatusTypes |
+		db.AlertStatusTypes |
+		db.EvalStatusTypes,
+](
+	values []string,
+	mapf func(string) (T, error),
+) ([]T, error) {
+	converted := []T{}
```
dmjb, line 465: Nitpick: Use `converted := make([]T, len(values))`

internal/history/models.go
```diff
@@ -0,0 +1,583 @@
...
+type ListEvaluationCursor struct {
+	Time      time.Time
+	Direction Direction
+}
+
+var (
+	// DefaultCursor is a cursor starting from the beginning of
+	// the data set.
+	DefaultCursor = ListEvaluationCursor{
+		Time:      time.UnixMicro(999999999999999999).UTC(),
```
jhrozek, line 63: Why does a cursor at the beginning of data set have a very large value?

internal/history/models.go
```diff
@@ -0,0 +1,583 @@
...
+		Direction: Next,
+	}
+)
+
+// ParseListEvaluationCursor interprets an opaque payload and returns
+// a ListEvaluationCursor. The opaque paylaod is expected to be of one
+// of the following forms
+//
+//   - `"+1257894000000000"` meaning the next page
+//     of data starting from the given UUID excluded
```
jhrozek, line 73: the UUIDs in the comments seem to be a typo?

internal/history/models.go
```diff
@@ -0,0 +1,583 @@
...
+//
+//   - `"1257894000000000"` meaning the next page
+//     of data (default) starting from the given UUID excluded
+func ParseListEvaluationCursor(payload string) (*ListEvaluationCursor, error) {
+	decoded, err := base64.StdEncoding.DecodeString(payload)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", ErrMalformedCursor, err)
+	}
+
+	switch {
```
jhrozek, line 86: do we expect more cases in this switch-case? It seems like `+`, `-` and `default` only differ in the direction and whether they parse the first character or not..

internal/history/models.go
blkt: Because the first item in the returned list is the most recent one, so the default cursor start from "year 9999" rather than the usual Unix epoch.

internal/history/models.go
blkt: Thanks for catching this, I initially developed the cursor using UUIDs and then moved to timestamp.

internal/history/models.go
blkt: No we don't, and yes, you're correct they only differ in direction and first character.
Let me rewrite it a bit better.

proto/minder/v1/minder.proto
```diff
@@ -2559,70 +2559,73 @@ message ListEvaluationHistoryResponse {
...
-    // entity contains details of the entity which was evaluated.
-    EvaluationHistoryEntity entity = 1;
+    message Entity {
+        // id is the ID of the entity.
+        string id = 1;
 
-    // rule contains details of the rule which the entity was evaluated against.
-    EvaluationHistoryRule rule = 2;
+        // type is the entity type.
+        minder.v1.Entity type = 2;
```
evankanderson, line 2567: By naming the message `Entity`, you're conflicting with the existing top-level definition of `Entity` elsewhere in the file, which seems like it may _also_ make the generated code more difficult to work with.

proto/minder/v1/minder.proto
```diff
@@ -2559,70 +2559,73 @@ message ListEvaluationHistoryResponse {
...
 
-    // status contains the evaluation status.
-    EvaluationHistoryStatus status = 3;
+        // name is the entity name.
+        string name = 3;
+    }
```
evankanderson, line 2571: Two comments:

1. I've rarely been happy in the future with where a nested message has been placed -- I'd lean towards making this a top-level message.
2. Can you include comments about uniqueness of the various fields, and ability to use those fields in future calls?  (For example, I assume that `id` is globally unique, but is `name` globally unique, or only unique within the scope of a Provider?)

@eleftherias on provider-scoping entity names, as I know she was looking to remove that linkage (and I was advocating for keeping it).

proto/minder/v1/minder.proto
```diff
@@ -2559,70 +2559,73 @@ message ListEvaluationHistoryResponse {
...
-}
+        // type is the name of the rule type.
+        string type = 2;
 
-message EvaluationHistoryEntity {
-    // id is the ID of the entity.
-    string id = 1;
+        // profile is the name of the profile which contains the rule.
+        string profile = 3;
+    }
```
evankanderson, line 2582: It feels odd to have `type` fields in both the `Entity` and `Rule` messages that mean different things -- `Entity.type` is an enum (IIRC), while `Rule.type` is effectively the name of a template.  Can we call this `rule_type` to disambiguate?

evankanderson: (Also, the lack of `id` here means that `profile` + `name` is the distinguishing set of fields for future follow-up queries.)

proto/minder/v1/minder.proto
```diff
@@ -2559,70 +2559,73 @@ message ListEvaluationHistoryResponse {
...
+        // status is one of (success, error, failure, skipped)
+        // not using enums to mirror the behaviour of the existing API contracts.
+        string status = 1;
 
-    // name is the entity name.
-    string name = 3;
-}
+        // details contains optional details about the evaluation.
+        // the structure and contents are rule type specific, and are subject to change.
+        string details = 2;
```
evankanderson, line 2591: I'm not a big fan of having a top-level stringly-typed details that has "rule type specific" contents and structure.  That may be where we are at right now, but it seems like this could mean that one rule emits markdown, another plain-text, and a third XML fragments.

proto/minder/v1/minder.proto
```diff
@@ -2559,70 +2559,73 @@ message ListEvaluationHistoryResponse {
...
+        // status is one of (success, error, failure, skipped, not available)
+        // not using enums to mirror the behaviour of the existing API contracts.
+        string status = 1;
 
-    // profile is the name of the profile which contains the rule.
-    string profile = 3;
-}
+        // details contains optional details about the remediation.
+        // the structure and contents are remediation specific, and are subject to change.
+        string details = 2;
```
evankanderson, line 2604: Ditto on stringly-typed `details`.  Again, I'm willing to accept this as a _bridge_, but we should get to a more well-defined outcome.

proto/minder/v1/minder.proto
```diff
@@ -2559,70 +2559,73 @@ message ListEvaluationHistoryResponse {
...
-    // name is the name of the rule instance.
-    string name = 1;
+        // created_at is the timestamp of creation of this evaluation
+        google.protobuf.Timestamp evaluated_at = 3;
+    }
 
-    // type is the name of the rule type.
-    string type = 2;
+    message Remediation {
+        // status is one of (success, error, failure, skipped, not available)
```
evankanderson, line 2598: It seems like `Remediation` has one extra status (`not available`) compared with `Status`.  It feels odd to me that `Remediation` isn't part of the status and doesn't have a Timestamp -- can we have a Remediation without a `Status`, or a different time-scale?

proto/minder/v1/minder.proto
```diff
@@ -2559,70 +2559,73 @@ message ListEvaluationHistoryResponse {
...
+        string status = 1;
 
-    // details contains optional details about the evaluation.
-    // the structure and contents are rule type specific, and are subject to change.
-    string details = 2;
-}
+        // details contains optional details about the alert.
+        // the structure and contents are alert specific, and are subject to change.
+        string details = 2;
+    }
```
evankanderson, line 2615: The comments on `Remediation` also apply to `Alert`.  In fact, I'm wondering whether an `Alert` is a particular type of remediation.  @puerco @ethomson

proto/minder/v1/minder.proto
```diff
@@ -2559,70 +2559,73 @@ message ListEvaluationHistoryResponse {
...
+    EvaluationHistory.Status status = 3;
+
+    // alert contains details of the alerts for this evaluation.
+    EvaluationHistory.Alert alert = 4;
 
-    // details contains optional details about the alert.
-    // the structure and contents are alert specific, and are subject to change.
-    string details = 2;
+    // remediation contains details of the remediation for this evaluation.
+    EvaluationHistory.Remediation remediation = 5;
```
evankanderson, line 2630: Do we want a single `status` / `alert` / `remediation` for each entry?  It feels like we'll end up repeating a lot of the envelope (`rule`, `entity`) many times when querying history.  I'd sort of prefer to see something like this:

```proto
message HistoryResult {
  Timestamp start = 1;
  uint32 occurrences = 2;
  string status = 3;  // Include both rule eval and alert / remediate result
  // Temp, while we work out schema
  string status_detail = 4;
  string alert_detail = 5;
  string remediate_detail = 6;
}

message EntityRuleEvaluationHistory {
  EntityInfo entity = 1;
  RuleInfo rule = 2;
  repeated HistoryResult history = 3;
}

message ListEvaluationHistoryResult {
  Pagination pagination = 1;
  repeated HistoryResult = 2;
}
```

database/query/eval_history.sql
```diff
@@ -93,3 +93,71 @@ INSERT INTO alert_events(
...
+            WHEN ere.pull_request_id IS NOT NULL THEN 'pull_request'::entities
+            WHEN ere.artifact_id IS NOT NULL THEN 'artifact'::entities
+       END AS entity_type,
+       -- entity id
+       CASE WHEN ere.repository_id IS NOT NULL THEN r.id
+            WHEN ere.pull_request_id IS NOT NULL THEN pr.id
+            WHEN ere.artifact_id IS NOT NULL THEN a.id
+       END AS entity_id,
+       -- entity name
+       CASE WHEN ere.repository_id IS NOT NULL THEN r.repo_name
```
evankanderson, line 111: Shouldn't `entity_name` for repositories be `CONCAT(r.repo_owner, '/', r.repo_name)`?

database/query/eval_history.sql
```diff
@@ -93,3 +93,71 @@ INSERT INTO alert_events(
     $3,
     $4
 );
+
+-- name: ListEvaluationHistory :many
+SELECT s.id::uuid AS evaluation_id,
```
evankanderson, line 98: Can we include the `project_id` at some place in the query, to avoid issues like https://github.com/stacklok/minder/security/advisories/GHSA-q6h8-4j2v-pjg4 ?

database/query/eval_history.sql
```diff
@@ -93,3 +93,71 @@ INSERT INTO alert_events(
...
+            WHEN ere.artifact_id IS NOT NULL THEN 'artifact'::entities
+       END AS entity_type,
+       -- entity id
+       CASE WHEN ere.repository_id IS NOT NULL THEN r.id
+            WHEN ere.pull_request_id IS NOT NULL THEN pr.id
+            WHEN ere.artifact_id IS NOT NULL THEN a.id
+       END AS entity_id,
+       -- entity name
+       CASE WHEN ere.repository_id IS NOT NULL THEN r.repo_name
+            WHEN ere.pull_request_id IS NOT NULL THEN pr.pr_number::text
```
evankanderson, line 112: Again, shouldn't the name for PRs include the `repo_owner` and `repo_name` from the related `repositories` table?

evankanderson (commented):
A few more comments looking at the SQL query (which I'm very impressed by!).

proto/minder/v1/minder.proto
dmjb: I think that is reflective of the current state of the universe, but not necessarily how things work in future.

One distinction which comes to mind is that a remediation generally operates against an entity, whereas an alert may not (for example, creating a Jira ticket does not involve changing the state of an entity).

proto/minder/v1/minder.proto
dmjb: This was discussed during the design review - the proposal was to match the current state of Minder and work on improving remediations/alerts later.

proto/minder/v1/minder.proto
dmjb: Same comment as for line 2604

proto/minder/v1/minder.proto
dmjb: @evankanderson Are you referring to the rule type ID, or the rule instance ID?

proto/minder/v1/minder.proto
blkt: The naming conflict is only visible at this very line of the protobuf definitions, and generated go types are totally unaffected. The last commit I pushed to this PR should clarify the impact on go code.

The reasoning behind embedding messages was that they're only meant to be used in this context and nowhere else. I don't think we want to use them anywhere else in the API, so I don't understand the point of having them as top-level messages.

I will revert the last commit and avoid embedding as it was a breaking change anyway.

proto/minder/v1/minder.proto
dmjb: At this moment in time, they might. I am not so sure that they will remain the same as we evolve them in future.

proto/minder/v1/minder.proto
dmjb: Right now, we can't have a remediation that happens at a different time to an evaluation (within the margin of a few seconds or so). The engine also tracks the status of the evaluation and the remediation separately.

database/query/eval_history.sql
blkt: I'll probably move that in the upper layer, I'm not particularly fond of having this sort of things in queries, as one can introduce very subtle bugs by changing just one line.

That said, I think this issues should be solved on the write path, by writing something like an extra `calculated_name` column, so that provider-specific logic can run before any data reaches storage.

database/query/eval_history.sql
blkt: 👍, see comment on `repo_name` above.

proto/minder/v1/minder.proto
```diff
@@ -2605,6 +2605,9 @@ message EvaluationHistoryStatus {
     // details contains optional details about the evaluation.
     // the structure and contents are rule type specific, and are subject to change.
     string details = 2;
+
+    // created_at is the timestamp of creation of this evaluation
+    google.protobuf.Timestamp evaluated_at = 3;
```
dmjb, line 2610: I wonder if we should move this to the top-level `EvaluationHistory` message

blkt: I agree, it's better placed in the top-level message, but let's hear from others as well.

database/query/eval_history.sql
blkt: I think I solved this, could you verify?

proto/minder/v1/minder.proto
blkt: I am confused by this comment, both `ListEvaluationHistoryResult` and `EntityRuleEvaluationHistory` embed a `repeated HistoryResult`, that's a typo, right?

I don't know how to make progress on this, I'd be happy to discuss if we want to change message structure.

evankanderson: Oops, I meant for `ListEvaluationHistoryResult` to embed a repeated `EntityRuleEvaluationHistory`.  Updated.

proto/minder/v1/minder.proto
evankanderson: I think it would be a rule instance ID if you wanted to be able to do a unique follow-up query.  (i.e. looking for more history for items which don't have any history or have some suspicious changes over a shorter time window)

database/query/eval_history.sql
evankanderson: Yes, this looks fixed.  I just wanted to share that knowledge around.

proto/minder/v1/minder.proto
evankanderson: It turns out that we're trying to present something different (a log-history based view of history), so my suggestion doesn't really work.

evankanderson (approved):

Issue #3721, opened 2024-06-26
Extend Profile database calls with selector support


jhrozek:
Minder needs to be able to list profiles but also get the selectors at the same time. We need to extend the database calls to retrieve selectors as well.

PR #3786, opened 2024-07-04
Extend the Profile List database calls to include selectors

jhrozek:
# Summary

Extend the Profile List database calls to include selectors

Since the profile selectors need to be included when listing profiles, we need to extend the two profile list calls to list selectors as well.

Because the selectors are stored in a different table, but we still need to retrieve them in the same Go struct defined in models, we define a new SQL type in a migration that represents the selectors and tell sqlc to use it in sqlc.yaml.

This new Go struct needs a Scan method as well to convert raw SQL rows into structs as well.

Co-authored-by: Michelangelo Mori <mmori@stacklok.com>

Fixes: #3721

## Change Type

# Testing

There are unit tests and I tested the code as part of a larger branch, too.

# Review Checklist:

### DIFF
--- a/database/query/profiles.sql
+++ b/database/query/profiles.sql
@@ -65,9 +65,9 @@ SELECT * FROM profiles WHERE lower(name) = lower(sqlc.arg(name)) AND project_id 
 -- name: ListProfilesByProjectID :many
 WITH helper AS(
      SELECT pr.id as profid,
-            ARRAY_AGG(ROW(ps.id, ps.profile_id, ps.entity, ps.selector, ps.comment)::profile_selector) FILTER (WHERE ps.id IS NOT NULL) AS selectors
+            ARRAY_AGG(ROW(ps.id, ps.profile_id, ps.entity, ps.selector, ps.comment)::profile_selector) AS selectors
        FROM profiles pr
-       LEFT JOIN profile_selectors ps
+       JOIN profile_selectors ps
          ON pr.id = ps.profile_id
       WHERE pr.project_id = $1
       GROUP BY pr.id
@@ -78,15 +78,15 @@ SELECT
     helper.selectors::profile_selector[] AS profiles_with_selectors
 FROM profiles
 JOIN profiles_with_entity_profiles ON profiles.id = profiles_with_entity_profiles.profid
-JOIN helper ON profiles.id = helper.profid
+LEFT JOIN helper ON profiles.id = helper.profid
 WHERE profiles.project_id = $1;
 
 -- name: ListProfilesByProjectIDAndLabel :many
 WITH helper AS(
      SELECT pr.id as profid,
-     ARRAY_AGG(ROW(ps.id, ps.profile_id, ps.entity, ps.selector, ps.comment)::profile_selector) FILTER (WHERE ps.id IS NOT NULL) AS selectors
+     ARRAY_AGG(ROW(ps.id, ps.profile_id, ps.entity, ps.selector, ps.comment)::profile_selector) AS selectors
        FROM profiles pr
-       LEFT JOIN profile_selectors ps
+       JOIN profile_selectors ps
          ON pr.id = ps.profile_id
       WHERE pr.project_id = $1
       GROUP BY pr.id
@@ -96,7 +96,7 @@ SELECT sqlc.embed(profiles),
        helper.selectors::profile_selector[] AS profiles_with_selectors
 FROM profiles
 JOIN profiles_with_entity_profiles ON profiles.id = profiles_with_entity_profiles.profid
-JOIN helper ON profiles.id = helper.profid
+LEFT JOIN helper ON profiles.id = helper.profid
 WHERE profiles.project_id = $1
 AND (
     -- the most common case first, if the include_labels is empty, we list profiles with no labels
--- a/database/query/profiles.sql
+++ b/database/query/profiles.sql
@@ -63,11 +63,40 @@ SELECT * FROM profiles WHERE id = $1 AND project_id = $2 FOR UPDATE;
 SELECT * FROM profiles WHERE lower(name) = lower(sqlc.arg(name)) AND project_id = $1 FOR UPDATE;
 
 -- name: ListProfilesByProjectID :many
-SELECT sqlc.embed(profiles), sqlc.embed(profiles_with_entity_profiles) FROM profiles JOIN profiles_with_entity_profiles ON profiles.id = profiles_with_entity_profiles.profid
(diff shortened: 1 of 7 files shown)


internal/db/domain.go
```diff
@@ -15,12 +15,62 @@
...
+		return nil
+	}
+
+	// Convert the value to a string
+	bytes, ok := value.([]byte)
+	if !ok {
+		return fmt.Errorf("failed to scan SelectorInfo: %v", value)
+	}
+	str := string(bytes)
+	fmt.Println(str)
```
jhrozek, line 36: oops I'll remove this locally.

internal/db/models.go
```diff
@@ -564,6 +564,11 @@ type ProfilesWithEntityProfile struct {
 	Profid          uuid.UUID             `json:"profid"`
 }
 
+type ProfilesWithSelector struct {
+	Profid    uuid.UUID         `json:"profid"`
+	Selectors []ProfileSelector `json:"selectors"`
+}
```
jhrozek, line 570: This nice and simple struct is pretty much why I've been doing the dance with the override. I wasn't able to force sqlc to generate the proper type otherwise.

internal/db/profiles.sql.go
```diff
@@ -358,13 +358,20 @@ func (q *Queries) GetProfileForEntity(ctx context.Context, arg GetProfileForEnti
...
+FROM profiles
+JOIN profiles_with_entity_profiles ON profiles.id = profiles_with_entity_profiles.profid
+JOIN profiles_with_selectors ON profiles.id = profiles_with_selectors.profid
 WHERE profiles.project_id = $1
 `
 
 type ListProfilesByProjectIDRow struct {
 	Profile                   Profile                   `json:"profile"`
 	ProfilesWithEntityProfile ProfilesWithEntityProfile `json:"profiles_with_entity_profile"`
+	ProfilesWithSelector      ProfilesWithSelector      `json:"profiles_with_selector"`
```
jhrozek, line 374: And here's how the new struct appears in the List response, thanks to `sqlc.embed`.

jhrozek:
@blkt 👀

jhrozek:
@blkt thank you for the help on tidying up the models!

I did some follow-up cleanup of the code and I think it's ready for review now. I'm not sure if you can give the final ack given that you are the coauthor on the PR now.

sqlc.yaml
```diff
@@ -24,4 +24,12 @@ packages:
...
     emit_interface: true
     emit_exact_table_names: false
-    emit_empty_slices: true
\ No newline at end of file
+    emit_empty_slices: true
+    overrides:
+      - db_type: "uuid"
+        go_type:
+          import: "github.com/google/uuid"
+          type: "UUID"
```
blkt, line 32: **question:** why is this override necessary now while it wasn't before?

database/query/profiles.sql
```diff
@@ -63,11 +63,40 @@ SELECT * FROM profiles WHERE id = $1 AND project_id = $2 FOR UPDATE;
...
+      GROUP BY pr.id
+)
+SELECT
+    sqlc.embed(profiles),
+    sqlc.embed(profiles_with_entity_profiles),
+    helper.selectors::profile_selector[] AS profiles_with_selectors
+FROM profiles
+JOIN profiles_with_entity_profiles ON profiles.id = profiles_with_entity_profiles.profid
+JOIN helper ON profiles.id = helper.profid
 WHERE profiles.project_id = $1;
```
blkt, line 82: **suggestion (non-blocking):** I think you could get rid of the `FILTER` clause by moving the `LEFT JOIN` on the outer query. I expect it to work the same given the result is a slice.

```suggestion
WITH helper AS(
     SELECT pr.id as profid,
            ARRAY_AGG(ROW(ps.id, ps.profile_id, ps.entity, ps.selector, ps.comment)::profile_selector) AS selectors
       FROM profiles pr
       JOIN profile_selectors ps
         ON pr.id = ps.profile_id
      WHERE pr.project_id = $1
      GROUP BY pr.id
)
SELECT
    sqlc.embed(profiles),
    sqlc.embed(profiles_with_entity_profiles),
    helper.selectors::profile_selector[] AS profiles_with_selectors
FROM profiles
JOIN profiles_with_entity_profiles ON profiles.id = profiles_with_entity_profiles.profid
LEFT JOIN helper ON profiles.id = helper.profid
WHERE profiles.project_id = $1;
```

database/query/profiles.sql
```diff
@@ -63,11 +63,40 @@ SELECT * FROM profiles WHERE id = $1 AND project_id = $2 FOR UPDATE;
...
+         ON pr.id = ps.profile_id
+      WHERE pr.project_id = $1
+      GROUP BY pr.id
+)
+SELECT sqlc.embed(profiles),
+       sqlc.embed(profiles_with_entity_profiles),
+       helper.selectors::profile_selector[] AS profiles_with_selectors
+FROM profiles
+JOIN profiles_with_entity_profiles ON profiles.id = profiles_with_entity_profiles.profid
+JOIN helper ON profiles.id = helper.profid
```
blkt, line 99: **suggestion (non-blocking):** I think you could get rid of the `FILTER` clause by moving the `LEFT JOIN` on the outer query. I expect it to work the same given the result is a slice.

```suggestion
WITH helper AS(
     SELECT pr.id as profid,
     ARRAY_AGG(ROW(ps.id, ps.profile_id, ps.entity, ps.selector, ps.comment)::profile_selector) AS selectors
       FROM profiles pr
       JOIN profile_selectors ps
         ON pr.id = ps.profile_id
      WHERE pr.project_id = $1
      GROUP BY pr.id
)
SELECT sqlc.embed(profiles),
       sqlc.embed(profiles_with_entity_profiles),
       helper.selectors::profile_selector[] AS profiles_with_selectors
FROM profiles
JOIN profiles_with_entity_profiles ON profiles.id = profiles_with_entity_profiles.profid
LEFT JOIN helper ON profiles.id = helper.profid
```

blkt (dismissed):
LGTM! 🎉

database/query/profiles.sql
jhrozek: suggestion taken!

database/query/profiles.sql
jhrozek: suggestion taken!

sqlc.yaml
jhrozek: (longer reply: This was a copy-paste from sqlc docs which I didn't remove by accident)

blkt (approved):

PR #3797, opened 2024-07-05
Add the protobuf message and the selector API

### LINKED ISSUE #3757: Define the selector grammar and the associated protobuf messages
Defining the selector grammar and its behaviour is more of a discussion/design task. The result should pretty much be a protobuf message that we agree can be used in the CEL evaluator.

jhrozek:
# Summary

**Define the protobuf API for selectors** - Creates a new protobuf message SelectorEntity that represents a set of attributes for a typed entity to be used in profile selectors.

At the moment, onyl SelectorRepository and SelectorArtifact are used and implemented. Additionally, the key-value store is not implemented either as it requires a new set of provider interfaces to fill this store.

**Add the selector evalautor** - Adds a new module called selectors that initializes CEL environments and evaluates selector

Fixes #3757

## Change Type

# Testing

there are unit tests included. I also cherry-pick these from a larger branch that takes this code into account

# Review Checklist:


internal/engine/selectors/selectors.go
```diff
@@ -0,0 +1,372 @@
...
+// artifactEnvFactory is a factory for creating a CEL environment
+// for the SelectorArtifact type representing an artifact
+type artifactEnvFactory struct{}
+
+func (_ *artifactEnvFactory) createEnv() (*cel.Env, error) {
+	return newEnvForEntity(
+		"artifact",
+		&minderv1.SelectorEntity_SelectorArtifact{},
+		"minder.v1.SelectorEntity.SelectorArtifact")
+}
```
blkt, line 71: **nit:** I tend to prefer function types to interfaces when they have a single method, which is usually the case with factories.

(part 1)

```suggestion
// celEnvFactory is a function type for creating CEL environments
// for an entity. Each entity must implement this interface to be
// usable in selectors
type celEnvFactory func() (*cel.Env, error)

// genericEnvFactory is a factory for creating a CEL environment
// for the generic SelectorEntity type
func genericEnvFactory() (*cel.Env, error) {
	return newEnvForEntity(
		"entity",
		&minderv1.SelectorEntity{},
		"minder.v1.SelectorEntity")
}

// repoEnvFactory is a factory for creating a CEL environment
// for the SelectorRepository type representing a repository
func repoEnvFactory() (*cel.Env, error) {
	return newEnvForEntity(
		"repository",
		&minderv1.SelectorEntity_SelectorRepository{},
		"minder.v1.SelectorEntity.SelectorRepository")
}

// artifactEnvFactory is a factory for creating a CEL environment
// for the SelectorArtifact type representing an artifact
func artifactEnvFactory() (*cel.Env, error) {
	return newEnvForEntity(
		"artifact",
		&minderv1.SelectorEntity_SelectorArtifact{},
		"minder.v1.SelectorEntity.SelectorArtifact")
}
```

internal/engine/selectors/selectors.go
```diff
@@ -0,0 +1,372 @@
...
+	err  error
+}
+
+// NewEnv creates a new Env struct with the default factories for each entity type. The factories
+// are used on first access to create the CEL environments for each entity type.
+func NewEnv() (*Env, error) {
+	factoryMap := map[minderv1.Entity]celEnvFactory{
+		minderv1.Entity_ENTITY_UNSPECIFIED:  &genericEnvFactory{},
+		minderv1.Entity_ENTITY_REPOSITORIES: &repoEnvFactory{},
+		minderv1.Entity_ENTITY_ARTIFACTS:    &artifactEnvFactory{},
```
blkt, line 129: **nit:** I tend to prefer function types to interfaces when they have a single method, which is usually the case with factories.

(part 2)

```suggestion
		minderv1.Entity_ENTITY_UNSPECIFIED:  genericEnvFactory,
		minderv1.Entity_ENTITY_REPOSITORIES: repoEnvFactory,
		minderv1.Entity_ENTITY_ARTIFACTS:    artifactEnvFactory,
```

internal/engine/selectors/selectors.go
```diff
@@ -0,0 +1,372 @@
...
+// envForEntity gets the CEL environment for a given entity type. If the environment is not cached,
+// it creates it using the factory for that entity type.
+func (e *Env) envForEntity(entity minderv1.Entity) (*cel.Env, error) {
+	cache, ok := e.entityEnvs[entity]
+	if !ok {
+		return nil, fmt.Errorf("no cache found for entity %v", entity)
+	}
+
+	cache.once.Do(func() {
+		cache.env, cache.err = e.factories[entity].createEnv()
```
blkt, line 205: **nit:** I tend to prefer function types to interfaces when they have a single method, which is usually the case with factories.

(part 3)

```suggestion
		cache.env, cache.err = e.factories[entity]()
```

internal/engine/selectors/selectors.go
```diff
@@ -0,0 +1,372 @@
...
+		Name:       fmt.Sprintf("%s/%s", a.GetOwner(), a.GetName()),
+		Entity: &minderv1.SelectorEntity_Artifact{
+			Artifact: &minderv1.SelectorEntity_SelectorArtifact{
+				Name:         fmt.Sprintf("%s/%s", a.GetOwner(), a.GetName()),
+				ArtifactName: a.GetName(),
+				Owner:        a.GetOwner(),
+				Type:         a.GetType(),
+			},
+		},
+	}
```
blkt, line 258: **question:** it feels to me like this piece of code belongs to `EntityInfoWrapper` rather than to `Selector`, but I am not 100% sure.

I suppose that every entity managed by Minder engine will eventually need this sort of transformation, and adding a new entity would lead up to changes to Selector. If my reasoning holds, it is the `Entity` being aware of `Selector`, and not the other way around, as `Selector` is really not much more than a boolean predicate.

What's your thought?

internal/engine/selectors/selectors.go
jhrozek: I think you're right that it would make the code tidier. Another thing I was considering was to have a single factory that would return both the environment and the converter just to make it easier to extend the code to handle a new entity.

internal/engine/selectors/selectors.go
jhrozek: > **question:** it feels to me like this piece of code belongs to `EntityInfoWrapper` rather than to `Selector`, but I am not 100% sure.

I was wondering about this, too when working on the code. I don't have a good answer (as you can see below)

> 
> I suppose that every entity managed by Minder engine will eventually need this sort of transformation, and adding a new entity would lead up to changes to Selector. If my reasoning holds, it is the `Entity` being aware of `Selector`, and not the other way around, as `Selector` is really not much more than a boolean predicate.
> 

The selector does expose some entity-specific attributes now like `is_fork`, so there is some coupling between the entity and the selector now and we'd still need to change the selector protobuf message if we add a new entity.

The protobuf messages though are just a convenience in general we could use a `map[string]any` (and that's what we pass to the evaluator), it's just that the protobuf messages make it easier to check the expressions and the input. 

Would you say that having the EntityInfoWrapper code be able to return the `SelectorEntity` (or actually..even returning the `map[string]any`...) to be treated as an opaque blob by the selector module would be better?

The other question I was wondering about - we will have to call the provider to fill in the key-value attributes (which are not present in the protobuf message now) and format the name.

Is this better done from the selector code that knows the expressions? Maybe this is premature optimization, but I wondered if we could only call the provider methods lazily if the attributes are needed in the expression.

blkt: Hey @jhrozek, sorry for the late reply!

> Would you say that having the EntityInfoWrapper code be able to return the `SelectorEntity` (or actually..even returning the `map[string]any`...) to be treated as an opaque blob by the selector module would be better?

I think this is exactly the case. My reasoning is that selectors are expressed in a language that is specific to Minder's domain, while entities are Minder's view into someone else's domain. Entities should specify themselves "in Minder's lingo", and not the other way around.

This is also reflected by my previous comment about adding a new entity. Adding support for a new entity should only require minimal (or no) changes to the engine.

> The other question I was wondering about - we will have to call the provider to fill in the key-value attributes (which are not present in the protobuf message now) and format the name.
>
> Is this better done from the selector code that knows the expressions?

I'm not sure about this, but I would gather all the info up front and avoid getting information lazily for now.

proto/minder/v1/minder.proto
```diff
@@ -2626,3 +2626,42 @@ message EvaluationHistoryAlert {
...
+    Entity entity_type = 1;
+    // the name of the entity, same as the name in the entity message
+    string name = 2;
+
+    oneof entity {
+        SelectorRepository repository = 5;
+        SelectorArtifact artifact = 6;
+    }
+
+    message SelectorRepository {
```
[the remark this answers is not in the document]
evankanderson: Yeah, nesting the type definitions doesn't tend to lead to a lot of joy.

I've been thinking at some point that we might want to split `minder.proto` into base-level types, intermediate types, and then just RPC + request/response messages.  I'm not sure what violence that would do to e.g. generated python types, but I think we've got `buf` set to ignore those API change breakages at the moment.

internal/engine/selectors/selectors.go
evankanderson: It feels like the selector should be using the [Visitor Pattern](https://en.wikipedia.org/wiki/Visitor_pattern) at a naive first pass, but I could be convinced otherwise.

That feels like it would like exposing a few methods that take attribute paths, ala: `StringProperty(string path) *string`, `BoolProperty(string path) *bool`, `IntProperty(string path) *int`.  (I'm not sure whether we should have both `IntProperty` and `FloatProperty` -- it feels like most properties right now are integer, but something like a Trusty score would be a float (I'm not suggesting we make this a property of Entities, but trying to think of examples).

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
jhrozek: let's move that message to a private `proto` file for now as you suggested elsewhere. There's a bunch of other private messages that we defined earlier as interface between ingesters and evaluators, those can be moved as well.

internal/engine/selectors/selectors.go
jhrozek: Yes, getting the key-value items from the provider was the plan all along, but you're right that we don't actually have to keep the conversion in the selector code.

This one though:
```
// Either the provider or some common code provides short-names for names like 'github.is_fork' --> 'is_fork'.
```
But we won't allow shortnames for *all* the key-values, but only for those that have an explicit attribute in the message right?

In other words, for a  a repository we might be able to use the following in the grammar: `repository.is_fork = false` which happens to be set to the same value as `repository.provider_properties.github.is_fork` but we can't use `repository.maturity=stable` even if our provider sets `repository.provider_properties.github.maturity=stable`?

proto/minder/v1/minder.proto
```diff
@@ -2626,3 +2626,42 @@ message EvaluationHistoryAlert {
     // the structure and contents are alert specific, and are subject to change.
     string details = 2;
 }
+
+message SelectorEntity {
+    // one of repository, pull_request, artifact (see oneof entity)
+    Entity entity_type = 1;
+    // the name of the entity, same as the name in the entity message
+    string name = 2;
```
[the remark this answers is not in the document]
jhrozek: I wanted to add the provider name as a separate attribute alongside name when we add fetching of the key-value pairs from the provider. Since you [suggested](https://github.com/stacklok/minder/pull/3797#discussion_r1669002640) to just use full_name were you thinking to make the provider name part of the full name as well?

I like how unambiguous that would be, but if we don't provide a very easy way "just ignore this repo" especially from the UI then the grammar gets less useful to the user.

proto/minder/v1/minder.proto
```diff
@@ -2626,3 +2626,42 @@ message EvaluationHistoryAlert {
...
+
+        // is_fork is true if the repository is a fork, nil if "don't know" or rather
+        // not applicable to this provider
+        optional bool is_fork = 4;
+        // is_private is true if the repository is private, nil if "don't know" or rather
+        // not applicable to this provider
+        optional bool is_private = 5;
+    }
+
+    message SelectorArtifact {
```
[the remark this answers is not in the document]
jhrozek: Sorry, did you mean if we need the `is_private` attribute of the `SelectorRepository` message or if we should add `is_private` to the `SelectorArtifact` message?

proto/minder/v1/minder.proto
```diff
@@ -2626,3 +2626,42 @@ message EvaluationHistoryAlert {
     // the structure and contents are alert specific, and are subject to change.
     string details = 2;
 }
+
+message SelectorEntity {
```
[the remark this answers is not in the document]
jhrozek: let's do a separate proto file

internal/engine/selectors/selectors.go
```diff
@@ -0,0 +1,372 @@
...
+
+// newConverterFactory creates a new converterFactory with the default converters for each entity type
+func newConverterFactory() *converterFactory {
+	return &converterFactory{
+		converters: map[minderv1.Entity]entityInfoConverter{
+			minderv1.Entity_ENTITY_REPOSITORIES: &repositoryInfoConverter{},
+			minderv1.Entity_ENTITY_ARTIFACTS:    &artifactInfoConverter{},
+		},
+	}
+}
```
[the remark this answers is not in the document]
jhrozek: these are not database table types but our public API types.

But yes, let's move the conversion into the providers.

internal/engine/selectors/selectors.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	return conv.toSelectorEntity(eiw.Entity)
+}
+
+// Selection is an interface for selecting entities based on a profile
+type Selection interface {
+	Select(*minderv1.SelectorEntity) (bool, error)
+	SelectEiw(*entities.EntityInfoWrapper) (bool, error)
+}
```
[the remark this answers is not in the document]
jhrozek: we don't, let's do a provider interface that returns either a selector entity or just a map for a single `Select()` method.

internal/engine/selectors/selectors.go
evankanderson: I'm not sure how we want to handle the shortnames, to be honest.  I can see two paths:

1. The shortnames are a convention, and it's up to providers (possibly with library help) to maintain that convention.
2. The shortnames are a minder intrinsic (e.g. produce a shortname for each fullname that has a unique last segment).  In this case, I'm imagining that `repository.is_fork` could be `is_fork` unless you also have a `github.tags.is_fork` (i.e. an "is fork" repo tag in the organization).

Looking at the second example, it feels like maybe we should go with case 1, which would also make it easier to report on the general set of properties which might be present, though we'd some syntax for `github.tags.* => string` to indicate that the GitHub tags were strings.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
evankanderson: I think we might want one property for the `full_name`, and one property for the `provider`.  That would allow for full distinguishing, but would default to names that were recognizable to people without some sort of special prefix.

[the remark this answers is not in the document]
evankanderson: It might also simplify "apply this policy to all `billing/fooserver` images at all registries under Minder automation"

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
evankanderson: I meant that we might want to be able to distinguish between public and private artifacts, as well as public and private repos.  But this may tie in to the [key-values from Provider comment](https://github.com/stacklok/minder/pull/3797#discussion_r1669286088).

internal/engine/selectors/selectors.go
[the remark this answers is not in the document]
evankanderson: If the common usage is going to be with an EntityInfoWrapper, I could see that being the standard interface, but it felt like having both be part of the public interface might lead to complication or drift between the two.

jhrozek:
thanks for the reviews @evankanderson @blkt . I'm going to resolve the comments that I think are resolved (but feel free to unresolve if you don't think they are fully addressed) and I'm going to comments on parts of the code where I'm unsure that I'm going in the right direction.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
jhrozek: the messages were moved to a private proto file and I've un-embedded them. Marking as resolved to keep the review tidier.

internal/engine/selectors/selectors.go
jhrozek: suggestion taken, thank you

internal/engine/selectors/selectors.go
jhrozek: In this version, I moved the logic that constructs the `SelectorEntity` to providers. I don't think we're done yet and I think we want the providers to also grab the key-value pairs and keep them as a map, but I'd prefer to do that in a follow-up PR if that's OK.

The attributes that are in the messages now are "stable" as far as the messages go, meaning I'd like to document them for users as something they can always use in the selector grammar. In the follow-up, I'd like to work on reading provider-specific key-value pairs per provider. First, I'd like to support languages for repos and the key-value pairs that github support (can't remember the names, but it's really a generic k-v pairs). At the same time, I'd like to work on "duck-typing" the attributes so that users can type `repository.properties.foo=bar` instead of `repository.provider.github.foo=bar`. But I think that merits a separate discussion.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
jhrozek: I added a provider message to the messages is that what you were describing? There's name and provider class exposed for start.

proto/minder/v1/minder.proto
```diff
@@ -2626,3 +2626,42 @@ message EvaluationHistoryAlert {
...
+    }
+
+    message SelectorRepository {
+        // the full name of the repository, e.g. stacklok/minder
+        string name = 1;
+        // the name of the repository, e.g. minder
+        string repo_name = 2;
+        // the owner of the repository, e.g. stacklok
+        string owner = 3;
+
```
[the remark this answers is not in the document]
jhrozek: Only the full name is used now, formatted by the provider. I dropped the repo_name and owner as separate fields. I named the attribute just `name` because I just think that's more user-friendly but I'm not attached to `name` vs. `full_name`.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
jhrozek: added! I would like to tie the k-v map in a follow up, though.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
jhrozek: moved to a separate proto file.

internal/engine/selectors/selectors.go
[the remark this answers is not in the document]
jhrozek: let's discuss the new interface, I'll point out places where I was unsure.

internal/proto/internal.proto
```diff
@@ -67,4 +67,48 @@ message PrContents {
...
   repeated File files = 2;
-}
\ No newline at end of file
+}
+
+message SelectorProvider {
+  // the name of the provider, e.g. github-app-jakubtestorg
+  string name = 1;
+  // the class of the provider, e.g. github-app
+  string class = 2;
```
jhrozek, line 76: This message is new in this PR version. It's just a way to avoid repeating what attributes we want to expose for provider.

internal/engine/selectors/selectors.go
```diff
@@ -42,7 +40,7 @@ func genericEnvFactory() (*cel.Env, error) {
 	return newEnvForEntity(
 		"entity",
 		&internalpb.SelectorEntity{},
-		"minder.v1.SelectorEntity")
+		"internal.SelectorEntity")
```
jhrozek, line 43: oops sorry these should have been squashed into the first patch

internal/providers/selectors/selector_entity.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"google.golang.org/protobuf/proto"
+
+	internalpb "github.com/stacklok/minder/internal/proto"
+	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+	provifv1 "github.com/stacklok/minder/pkg/providers/v1"
+)
+
+// entityInfoConverter is an interface for converting an entity from an EntityInfoWrapper to a SelectorEntity
+type entityInfoConverter interface {
```
jhrozek, line 30: just noticed the interface is misnamed, it's no longer tied to `EntityInfoWrapper`, but just a pair an entity as a `proto.message` and entity type.

internal/providers/selectors/selector_entity.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	return &repositoryInfoConverter{
+		converter: converter,
+	}
+}
+
+func (rc *repositoryInfoConverter) toSelectorEntity(ctx context.Context, entity proto.Message) *internalpb.SelectorEntity {
+	if rc == nil {
+		return nil
```
jhrozek, line 51: maybe it would be nicer to return an error here instead of a `nil`?

internal/providers/selectors/selector_entity.go
```diff
@@ -0,0 +1,127 @@
...
+	converters map[minderv1.Entity]entityInfoConverter
+}
+
+// newConverterFactory creates a new converterFactory with the default converters for each entity type
+func newConverterFactory(provider provifv1.Provider) *converterFactory {
+	return &converterFactory{
+		converters: map[minderv1.Entity]entityInfoConverter{
+			minderv1.Entity_ENTITY_REPOSITORIES: newRepositoryInfoConverter(provider),
+			minderv1.Entity_ENTITY_ARTIFACTS:    newArtifactInfoConverter(provider),
+		},
```
jhrozek, line 101: @puerco this is one of the things I don't like here, sure the code doesn't use a straight-up switch-case but a factory which is somewhat nicer, but we still hardcode the entities we know about..
I don't know if this is a good direction.

jhrozek:
@dmjb I'd appreciate your thoughts here as well, you usually have a good direction on how to split the interfaces nicely.

internal/proto/internal.proto
```diff
@@ -67,4 +67,48 @@ message PrContents {
...
+  minder.v1.Entity entity_type = 1;
+  // the name of the entity, same as the name in the entity message
+  string name = 2;
+  SelectorProvider provider = 3;
+
+  oneof entity {
+    SelectorRepository repository = 4;
+    SelectorArtifact artifact = 5;
+  }
+}
```
evankanderson, line 114: This is an internal implementation detail that currently makes it easier to implement CEL, right?

I'm concerned over time that encoding all the properties into a protocol buffer in core Minder will make it harder for new providers to add properties that make sense, but we can cross that bridge when we get there.  (The alternative is that it's easy for providers to add new properties, but little uniformity across providers -- this pushes us on the higher-friction-but-more-commonality path, which might even be right.)

internal/proto/internal.proto
evankanderson: Why is this _both_ in `SelectorRepository`/`SelectorArtifact` _and_ in `SelectorEntity`?  It feels like both copies of the message should have the same contents, so maybe we should have a canonical place?

internal/proto/internal.proto
```diff
@@ -67,4 +67,48 @@ message PrContents {
...
+  string name = 1;
+  // the provider of the repository
+  SelectorProvider provider = 2;
+
+  // is_fork is true if the repository is a fork, nil if "don't know" or rather
+  // not applicable to this provider
+  optional bool is_fork = 3;
+  // is_private is true if the repository is private, nil if "don't know" or rather
+  // not applicable to this provider
+  optional bool is_private = 4;
```
evankanderson, line 90: Do we want to include the ["Custom properties" key/values](https://github.blog/changelog/2023-10-12-github-repository-custom-properties-beta/) that GitHub supports under a name like "tags", "labels", or "properties"?

internal/proto/internal.proto
```diff
@@ -67,4 +67,48 @@ message PrContents {
...
+
+  // the type of the artifact, e.g. "container"
+  string type = 4;
+}
+
+message SelectorEntity {
+  // one of repository, pull_request, artifact (see oneof entity)
+  minder.v1.Entity entity_type = 1;
+  // the name of the entity, same as the name in the entity message
+  string name = 2;
```
evankanderson, line 107: This is to allow for access as both `name` and `repository.name`?

internal/providers/github/selector_entity.go
```diff
@@ -0,0 +1,59 @@
...
+				IsFork:    proto.Bool(r.GetIsFork()),
+				IsPrivate: proto.Bool(r.GetIsPrivate()),
+			},
+		},
+	}
+}
+
+// ArtifactToSelectorEntity converts an Artifact to a SelectorEntity
+func (_ *GitHub) ArtifactToSelectorEntity(_ context.Context, a *minderv1.Artifact) *internalpb.SelectorEntity {
+	fullName := fmt.Sprintf("%s/%s", a.GetOwner(), a.GetName())
```
evankanderson, line 47: Should this include `ghcr.io/`?  (I'm not sure)

internal/proto/internal.proto
```diff
@@ -67,4 +67,48 @@ message PrContents {
...
+  // not applicable to this provider
+  optional bool is_fork = 3;
+  // is_private is true if the repository is private, nil if "don't know" or rather
+  // not applicable to this provider
+  optional bool is_private = 4;
+}
+
+message SelectorArtifact {
+  // the full name of the artifact, e.g. stacklok/minder-server
+  string name = 1;
```
evankanderson, line 95: Do we want a `server` component here, e.g. `ghcr.io` or `eu.gcr.io`?

internal/providers/selectors/selector_entity.go
evankanderson: I think it's okay to return `nil` if there's no helpful error to be had.  Right now, the only error cases are passing `nil` or failing the cast, which should both be pretty obvious what went wrong.

evankanderson (commented):
It would help me a bit to see (in a comment) what this looks like wired in.  I'm assuming that we'll have something like the following in `executor.EvalEntityEvent`:

```golang
	// entityProto is a SelectorEntity proto
	entityProto, err := provider.GetEntityObject(inf)
	// ...
	err = e.forProjectsInHierarchy(
		ctx, inf, func(ctx context.Context, profile *pb.Profile, hierarchy []uuid.UUID) error {
			selector := selectors.SelectorForProfile(profile)
			if !selector.Select(entityProto) {
				// record as skipped
				return nil  // or SkippedErr
			}
			// Get only these rules that are relevant for this entity type
			relevant, err := profiles.GetRulesForEntity(profile, inf.Type)
			if err != nil {
				return fmt.Errorf("error getting rules for entity: %w", err)
			}
```

But, this formulation suggests that the `Env` and `SelectionBuilder` interfaces shouldn't need to take a `minderv1.Entity` as a property, and should simply spit out a selector that works on generic `SelectorEntity` protobufs.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
blkt: **question (non-blocking):** is there a particular reason why we're not calculating an _exact_, possibly unambiguous name and storing it alongside the entity? It really feels like we should have for repositories something like a URI/purl alongside repo name, owner, and provider, something like `<provider>:<owner>:<name>`, e.g. `github-app-foo:stacklok:minder`. On top of that, we can define custom functions to ship with the CEL evaluator that operate on top of that.

internal/providers/selectors/selector_entity.go
blkt: We have a bunch of places where we switch over entity types. We might implement some sort of registration mechanism alla `init()`, something like this.
```
package custom

import (
        "github.com/stacklok/minder/internal/entities
)

init() {
        if err := entities.Register(customv1.Entity_MY_CUSTOM_ENTITY, &CustomEntity{}); err != nil {
                log.Fatalf("failed registering custom entity: %s", customv1.Entity_MY_CUSTOM_ENTITY)
        }
}

type CustomEntity struct {
}

func (c *CustomEntity) Converter(...) error {
        return nil
}
```
The nice thing about this is that this could be dynamically loaded as a [plugin](https://pkg.go.dev/plugin), or be otherwise loaded in the system.
Could this be a decent tool for this purpose?

internal/proto/internal.proto
jhrozek: yes, I wanted to do that in a follow-up but perhaps it makes sense to include the k-v property already to make the review easier in the sense that we'd see the whole message.

internal/proto/internal.proto
jhrozek: Maybe I should have spelled this out more prominently in the design doc. There's some examples in the [unit tests](https://github.com/jhrozek/minder/blob/35e207a748d5e1239c9e1cbee2c917e2476778cb/internal/engine/selectors/selectors_test.go#L150)

and here's a simple inline example. Given these selectors:
```
  - comment: This should not apply to bad-go
    entity: repository
    selector: repository.name != 'jakubtestorg/bad-go'
  - comment: Skip any entity that contains demo in the name
    selector: entity.name.contains('demo') == false
```
The first one applies to repositories only, in this case the selector uses `repository.name`, but it could use also `repository.provider.name/class`, `repository.is_fork` or `repository.is_private`.

The second one applies to any entity type. It can use only attributes from the `SelectorEntity` message which currently are `entity.name`, `entity.provider.name` and `entity.provider.class`. (Let's talk about adding the purl or global name separately).

In both the entity-generic message and the entity-specific message we'll also add the k-v map.

internal/providers/github/selector_entity.go
jhrozek: I would say that not the name attribute, but I agree with both you and @blkt (in [this comment](https://github.com/stacklok/minder/pull/3797#discussion_r1675647618)) that we should have a global name as well.

I think just based on what users would expect in the simplest case we could have:
 - `.name` - for repos this would be `orgname/reponame`
 - `.full_name` - this could be a clone name perhaps. For repos, `https://github.com/org/repo` for artifacts `registry_name/namespace/artifact_name`
 -  `.id` (help me come with better name :-)) - a minder-specific identifier. I liked @blkt's suggestion

btw note that as long as we expose the individual attributes the user can also combine them with logical operations in the CEL expression so we don't have to go overboard with providing too many one-off attributes.

internal/proto/internal.proto
jhrozek: yes, good point. I'm not sure if we store them for artifacts (or for repos for that matter, we do have a clone_url for repos), but I like the idea.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
jhrozek: We can do this as well and Evan suggested something similar. Good idea, thanks.

internal/proto/internal.proto
jhrozek: See [the example here](https://github.com/stacklok/minder/pull/3797#discussion_r1675709579). Since the individual selectors can either be targetting a specific entity or any entity (but what is being evaluated is always an specific entity), the canonical place is the entity-specific message (e.g. `SelectorRepository`). What's stored in the generic message (`SelectorEntity`) is just a copy of the common data.

Is that confusing as a concept or would it help to make it crisper in the code in the sense that the provider methods that fill in the selector entity struct just fill in the entity-specific message and the generic attributes are copied by a common function?

internal/proto/internal.proto
jhrozek: > This is an internal implementation detail that currently makes it easier to implement CEL, right?

yes

> 
> I'm concerned over time that encoding all the properties into a protocol buffer in core Minder will make it harder for new providers to add properties that make sense, but we can cross that bridge when we get there. (The alternative is that it's easy for providers to add new properties, but little uniformity across providers -- this pushes us on the higher-friction-but-more-commonality path, which might even be right.)

Right, I hear you. The properties that are expressed as attributes are only those that those entities will most commonly have (names, is_fork/is_private and those are optional bool, too). I expect any provider-specific attributes to be just key-value pairs in a map.

We could even go with just a map, but the nice thing about passing proto messages to CEL is that you CEL can check [if the attributes exist](https://github.com/jhrozek/minder/blob/35e207a748d5e1239c9e1cbee2c917e2476778cb/internal/engine/selectors/selectors_test.go#L265) or [warn about mismatch between entity types and the attribute](https://github.com/jhrozek/minder/blob/35e207a748d5e1239c9e1cbee2c917e2476778cb/internal/engine/selectors/selectors_test.go#L253). I think it would be nice to surface these to the user when they create the selectors.

We won't have those for a generic map.

jhrozek:
> It would help me a bit to see (in a comment) what this looks like wired in. I'm assuming that we'll have something like the following in `executor.EvalEntityEvent`:
> 
> ```go
> 	// entityProto is a SelectorEntity proto
> 	entityProto, err := provider.GetEntityObject(inf)
> 	// ...
> 	err = e.forProjectsInHierarchy(
> 		ctx, inf, func(ctx context.Context, profile *pb.Profile, hierarchy []uuid.UUID) error {
> 			selector := selectors.SelectorForProfile(profile)
> 			if !selector.Select(entityProto) {
> 				// record as skipped
> 				return nil  // or SkippedErr
> 			}
> 			// Get only these rules that are relevant for this entity type
> 			relevant, err := profiles.GetRulesForEntity(profile, inf.Type)
> 			if err != nil {
> 				return fmt.Errorf("error getting rules for entity: %w", err)
> 			}
> ```
> 
> But, this formulation suggests that the `Env` and `SelectionBuilder` interfaces shouldn't need to take a `minderv1.Entity` as a property, and should simply spit out a selector that works on generic `SelectorEntity` protobufs.

@evankanderson I keep [a branch](https://github.com/jhrozek/minder/tree/selectors) with patches that will be sent later. The wiring looks like this in executor:

```go
			// this could be a private function for now and we could just select an inf
			selectors, err := e.selBuilder.NewSelectionFromProfile(inf.Type, profile.Selection)
			if err != nil {
				return fmt.Errorf("error creating selectors: %w", err)
			}
			selected, err := selectors.Select(provsel.EntityToSelectorEntity(ctx, provider, inf.Type, inf.Entity))
			if err != nil {
				return fmt.Errorf("error selecting entity: %w", err)
			}
```

evankanderson:
Why does extracting the entity properties need to know the profile?  Naively, I'd expect that we could extract the entity properties once per evaluation, not once per profile in the evaluation.

jhrozek:
> Why does extracting the entity properties need to know the profile? Naively, I'd expect that we could extract the entity properties once per evaluation, not once per profile in the evaluation.

(not 100% sure I'm answering your question without more context)

it doesn't, we can call the extraction earlier. We should also only build the entity properties (the selector message) once we hit the first profile with non-zero selectors because extracting the entity properties might involve calling to the provider (e.g. to get the key-value pairs which we don't store in the database) which might be expensive.

internal/engine/selectors/selectors.go
jhrozek: This version adds properties to repositories. There's no fetching of them in the executor but I think we don't need to show how to update a map of string,string.

internal/engine/selectors/selectors.go
```diff
@@ -0,0 +1,370 @@
...
+
+// artifactEnvFactory is a factory for creating a CEL environment
+// for the SelectorArtifact type representing an artifact
+func artifactEnvFactory() (*cel.Env, error) {
+	return newEnvForEntity(
+		"artifact",
+		&internalpb.SelectorArtifact{},
+		"internal.SelectorArtifact")
+}
+
```
JAORMX, line 74: Do we need an env factory for pull requests?

internal/proto/internal.proto
jhrozek: I added the properties to the message, filling them in would be done in a follow-up

internal/engine/selectors/selectors.go
```diff
@@ -0,0 +1,370 @@
...
+	err  error
+}
+
+// NewEnv creates a new Env struct with the default factories for each entity type. The factories
+// are used on first access to create the CEL environments for each entity type.
+func NewEnv() (*Env, error) {
+	factoryMap := map[minderv1.Entity]celEnvFactory{
+		minderv1.Entity_ENTITY_UNSPECIFIED:  genericEnvFactory,
+		minderv1.Entity_ENTITY_REPOSITORIES: repoEnvFactory,
+		minderv1.Entity_ENTITY_ARTIFACTS:    artifactEnvFactory,
```
JAORMX, line 160: do we need pull requests here?

internal/engine/selectors/selectors.go
jhrozek: if you can come up with resonable selector message for the pull request, sure. I wasn't sure what to put there honestly. Maybe we could just have properties and nothing else?

internal/proto/internal.proto
```diff
@@ -67,4 +67,50 @@ message PrContents {
...
+  SelectorProvider provider = 2;
+
+  // is_fork is true if the repository is a fork, nil if "don't know" or rather
+  // not applicable to this provider
+  optional bool is_fork = 3;
+  // is_private is true if the repository is private, nil if "don't know" or rather
+  // not applicable to this provider
+  optional bool is_private = 4;
+
+  map<string, string> properties = 5;
```
JAORMX, line 92: should this be a google.protobuf.Struct instead?

internal/proto/internal.proto
JAORMX: We can add that iteratively in another PR.

internal/engine/selectors/selectors.go
JAORMX: Let's add a TODO comment for this. I think it's just fine if it's just properties.

internal/proto/internal.proto
jhrozek: yes!

I added a couple of tests that check property usage for other types than string/string, too

internal/engine/selectors/selectors.go
jhrozek: done

JAORMX: Where? :confused:

internal/engine/selectors/selectors.go
jhrozek: added a todo

internal/engine/selectors/selectors.go
jhrozek: sorry, only to the proto message definition. I added another one to the env factory, too

JAORMX (approved):

PR #3815, opened 2024-07-09
Fix project creation conflict during user creation

rdimitrov:
# Summary

The following PR fixes the conflict that may happen if the project we try to create during `CreateUser` already exists.

The proposed solution checks if there's such project with that name, if there's appends a random 4 char string and tries again until it finds a unique combination.

## Change Type

# Testing

# Review Checklist:


internal/controlplane/handlers_user.go
```diff
@@ -78,7 +80,13 @@ func (s *Server) CreateUser(ctx context.Context,
 		// Set up the default project for the user
 		baseName := subject
 		if token.PreferredUsername() != "" {
-			baseName = token.PreferredUsername()
+			// Check if `project_name_lower_idx` unique constraint was violated. This happens when
```
jhrozek, line 83: Is this an outdated comment? We check the other way around, to see if there's `sql.ErrNoRows` returned right?

internal/controlplane/handlers_user.go
```diff
@@ -483,3 +491,33 @@ func isUserSelfResolving(ctx context.Context, store db.Store, i db.GetInvitation
...
+	if err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(bytes)[:length], nil
+}
+
+// getUniqueProjectBaseName is used to generate a unique project name
+func getUniqueProjectBaseName(ctx context.Context, store db.Store, baseName string) (string, error) {
+	uniqueBaseName := baseName
+	for {
```
jhrozek, line 508: I know it's unlikely we hit a really infinite loop here, but ` for {}` always makes me nervous. Do you think we could add some upper limit?

jhrozek (approved):
Since this is a hack to get around an immediate bug, feel free to merge, but please also open an issue for the proper fix

internal/controlplane/handlers_user.go
rdimitrov: You're right, this was a leftover copy-pasta from the other place where we had this check 👍

internal/controlplane/handlers_user.go
rdimitrov: I'll add a 10 times retry mechanism. It should be more than enough 👍

rdimitrov:
@jhrozek - Opened an issue to track the proper way to fix this through refactoring - https://github.com/stacklok/minder/issues/3826

jhrozek (approved):

PR #3827, opened 2024-07-10
Use project as targeting key, propagate message context through entity evaluation

evankanderson:
# Summary

Update entity evaluation to propagate message context and include data for flag evaluation in the context.

Update flag evaluation context to store user in the same form as project and provider, and make the project ID the primary targeting key for providers that need it (not GoFeatureFlag).  Since no one is probably using the current fine-grained targeting, this should be okay.

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/engine/handler.go
+++ b/internal/engine/handler.go
@@ -104,11 +104,10 @@ func (e *ExecutorEventHandler) HandleEntityEvent(msg *message.Message) error {
 	// TODO: Make this timeout configurable
 	msgCtx := context.WithoutCancel(msg.Context())
 	msgCtx, shutdownCancel := context.WithCancel(msgCtx)
-	{
-		e.lock.Lock()
-		defer e.lock.Unlock()
-		e.cancels = append(e.cancels, &shutdownCancel)
-	}
+
+	e.lock.Lock()
+	e.cancels = append(e.cancels, &shutdownCancel)
+	e.lock.Unlock()
 
 	// Let's not share memory with the caller.  Note that this does not copy Context
 	msg = msg.Copy()
@@ -127,9 +126,17 @@ func (e *ExecutorEventHandler) HandleEntityEvent(msg *message.Message) error {
 
 		ctx, cancel := context.WithTimeout(msgCtx, DefaultExecutionTimeout)
 		defer cancel()
+		defer func() {
+			e.lock.Lock()
+			e.cancels = slices.DeleteFunc(e.cancels, func(cf *context.CancelFunc) bool {
+				return cf == &shutdownCancel
+			})
+			e.lock.Unlock()
+		}()
+
 		ctx = engcontext.WithEntityContext(ctx, &engcontext.EntityContext{
 			Project: engcontext.Project{ID: inf.ProjectID},
-			// TODO: extract Provider name from ProviderID
+			// TODO: extract Provider name from ProviderID?
 		})
 
 		ts := minderlogger.BusinessRecord(ctx)
@@ -175,11 +182,6 @@ func (e *ExecutorEventHandler) HandleEntityEvent(msg *message.Message) error {
 		if err := e.evt.Publish(events.TopicQueueEntityFlush, msg); err != nil {
 			logger.Err(err).Msg("error publishing flush event")
 		}
-		e.lock.Lock()
-		defer e.lock.Unlock()
-		e.cancels = slices.DeleteFunc(e.cancels, func(cf *context.CancelFunc) bool {
-			return cf == &shutdownCancel
-		})
 	}()
 
 	return nil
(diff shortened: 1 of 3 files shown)


internal/engine/handler.go
```diff
@@ -142,6 +175,11 @@ func (e *ExecutorEventHandler) HandleEntityEvent(msg *message.Message) error {
 		if err := e.evt.Publish(events.TopicQueueEntityFlush, msg); err != nil {
 			logger.Err(err).Msg("error publishing flush event")
 		}
+		e.lock.Lock()
+		defer e.lock.Unlock()
+		e.cancels = slices.DeleteFunc(e.cancels, func(cf *context.CancelFunc) bool {
+			return cf == &shutdownCancel
+		})
```
puerco, line 182: Since we're appending a new cancelation function on every call to the handler at line 110, should we call this logic to drop `shutdownCancel` from the `e.cancels` slice before the return on line 171 too?

internal/flags/flags.go
```diff
@@ -39,10 +39,12 @@ func fromContext(ctx context.Context) openfeature.EvaluationContext {
 	// Note: engine.EntityFromContext is best-effort, so these values may be zero.
 	ec := engcontext.EntityFromContext(ctx)
 	return openfeature.NewEvaluationContext(
-		jwt.GetUserSubjectFromContext(ctx),
+		ec.Project.ID.String(),
 		map[string]interface{}{
-			"project":  ec.Project.ID.String(),
+			"project": ec.Project.ID.String(),
+			// TODO: is this useful, given how provider names are used?
```
dmjb, line 45: IMHO we should consider removing it, but I don't have strong feelings about it right now.

internal/engine/handler.go
```diff
@@ -95,11 +124,15 @@ func (e *ExecutorEventHandler) HandleEntityEvent(msg *message.Message) error {
...
 			time.Sleep(ArtifactSignatureWaitPeriod)
 		}
-		// TODO: Make this timeout configurable
-		ctx, cancel := context.WithTimeout(e.terminationcontext, DefaultExecutionTimeout)
+
+		ctx, cancel := context.WithTimeout(msgCtx, DefaultExecutionTimeout)
 		defer cancel()
+		ctx = engcontext.WithEntityContext(ctx, &engcontext.EntityContext{
+			Project: engcontext.Project{ID: inf.ProjectID},
+			// TODO: extract Provider name from ProviderID
```
dmjb, line 132: IMHO, if we do decide to retain the provider as a feature flag attribute, we should consider using the provider ID instead of the name.

dmjb:
Aside from the issue which @puerco raised, this looks good to me.

internal/flags/flags.go
evankanderson: I think you're probably right.

internal/engine/handler.go
evankanderson: Alternatively, we should use the provider _class_, rather than a particular provider ID.  (i.e. "enable this feature for users with the foo provider")  It still feels like a stretch, though.

I think I'll just remove this.

internal/engine/handler.go
```diff
@@ -57,13 +61,23 @@ func NewExecutorEventHandler(
...
 	}
+	go func() {
+		<-ctx.Done()
+		eh.lock.Lock()
+		defer eh.lock.Unlock()
+
+		for _, cancel := range eh.cancels {
+			(*cancel)()
+		}
+	}()
```
blkt, line 78: If I got this right, you're still waiting for `Done()` on what was originally the `terminationcontext`, but keep track of all downstream contexts to cancel them as well, right?

If yes, was this source of some sort of bug?

internal/engine/handler.go
```diff
@@ -79,9 +93,24 @@ func (e *ExecutorEventHandler) Wait() {
...
+	// provide the parallelism.
+	// We _do_ still want to cancel on shutdown, however.
+	// TODO: Make this timeout configurable
+	msgCtx := context.WithoutCancel(msg.Context())
+	msgCtx, shutdownCancel := context.WithCancel(msgCtx)
+	{
+		e.lock.Lock()
+		defer e.lock.Unlock()
+		e.cancels = append(e.cancels, &shutdownCancel)
+	}
```
blkt, line 111: **praise:** TIL that you can introduce arbitrary lexical scopes in Go as well.

puerco: Same, when I saw it I actually ran a small code snip to make sure I was understanding :)

evankanderson: Ha, I avoided screwing this up by accident!  It turns out that the other lock interaction was inside the goroutine, so this one sort of speeds through and doesn't run into the other lock anyway.  But I switched things up to reduce the scope of the lock by explicitly calling unlock on the following line.

internal/engine/handler.go
puerco: Awesome, thank you :)

puerco (approved):

Issue #3829, opened 2024-07-10
Split messages that are only used to generate Go structs from minder.proto


jhrozek:
We have several messages defined in minder.proto that are only ever used to generate Go structs and use them with proto reflection.

We should split those out to a separate proto file.

We will add the messages for the Selectors to this new proto file as well.

PR #3830, opened 2024-07-10
Move messages only used to generate Go structs out of minder.proto

jhrozek:
# Summary

This was brought up in PR #3797. We already have several messages that only
exist to generate Go structures out of them and use them with reflection. When
we add the profile selector messages, we'll have even more and our proto file
is going to become messy.

Instead, let's split out a new proto file and use it to store all the messages
that are really just a way to express Go structs.

Fixes: #3829

## Change Type

# Testing

make test

# Review Checklist:


jhrozek:
Hold this - @dmjb had a good point that we should try getting rid of the structs first and foremost.

pkg/api/openapi/internal.swagger.json
blkt, line 1: **nit:** Is it possible to avoid generating this file?

blkt (dismissed):
Minor comment.

pkg/api/openapi/internal.swagger.json
jhrozek: I did not find a way. I did find a way to avoid generating the doc file (see the change to `buf.gen.yaml`), but not the swagger

jhrozek:
> Hold this - @dmjb had a good point that we should try getting rid of the structs first and foremost.

I think we can actually review and merge this PR. @dmjb mentioned he wanted to get rid of the structs which is a good idea, but we'll need the internal proto file to include the selector messages anyway (the CEL evaluator works much easier with protobuf messages).

It doesn't really matter whether we remove the messages that this PR moves to the internal proto file from the original proto file or the internal one and having the internal proto file created would enable us to move the selector messages there.

dmjb (approved):

blkt (approved):

PR #3998, opened 2024-07-25
Return structured errors from the selectors API

jhrozek:
# Summary

In order to tell the user what's wrong with a selector they are about to
take into use, let's introduce a structured way of returning errors to
the selector API.

The errors can be represented as JSON in order for clients to be able to
parse the error without knowing much about the type.

Related: #3725

## Change Type

# Testing

with the added unit tests and as part of a larger topic branch

# Review Checklist:

### DIFF
--- a/internal/engine/selectors/selectors.go
+++ b/internal/engine/selectors/selectors.go
@@ -41,6 +41,8 @@ var (
 	ErrResultUnknown = errors.New("result is unknown")
 	// ErrUnsupported is returned when the entity type is not supported
 	ErrUnsupported = errors.New("unsupported entity type")
+	// ErrSelectorCheck is returned if the selector fails to be checked for syntax errors
+	ErrSelectorCheck = errors.New("failed to check selector")
 )
 
 // ErrKind is a string for the kind of error that occurred
@@ -100,6 +102,7 @@ type ErrStructure struct {
 // ParseError is an error type for syntax errors in CEL expressions
 type ParseError struct {
 	ErrDetails
+	original error
 }
 
 // Error implements the error interface for ParseError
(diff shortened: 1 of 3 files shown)


internal/engine/selectors/selectors_test.go
```diff
@@ -438,6 +481,29 @@ func TestSelectSelectorEntity(t *testing.T) {
...
 
+func testErrUnmarshallableValue(t *testing.T, err error, expected *ErrStructure) {
+	t.Helper()
+
+	var ce *CheckError
+	var pe *ParseError
+	var jsonString string
+
+	if errors.As(err, &ce) {
+		jsonString = ce.Error()
```
jhrozek, line 492: This might look like an odd test because go code could very well just look at the contents of the struct, but the thinking was to include the JSON representation as HTTP body when manipulating selectors from handlers throws an error, then FE would be able to parse the JSON which would include the location of the failed CEL expression.

Is that an acceptable pattern or would we prefer to add an error message to the protobuf message?

internal/engine/selectors/selectors.go
```diff
@@ -38,8 +39,110 @@ var (
...
+		ErrDetails: errDetailsFromCelIssues(source, issues),
+	}
+}
+
+func newCheckError(source string, issues *cel.Issues) error {
+	return &CheckError{
+		ErrDetails: errDetailsFromCelIssues(source, issues),
+	}
+}
+
```
JAORMX, line 145: should these errors wrap a common one to ease error handling? https://pkg.go.dev/errors you could implement `Unwraps` for this.

jhrozek: that's a really good idea!

jhrozek: done. We could even use `errors.Join(ErrCheckFailed, ErrParse)` for parse and join with a check error for the check but I'm not sure if it's worth it.

JAORMX (approved):

PR #4004, opened 2024-07-25
Take selectors into use in executor

### LINKED ISSUE #3724: Take selectors into use when evaluating entities
The executor needs to take selectors into account when evaluating entities and skip evaluation if selectors de-select an entity.

jhrozek:
# Summary

- **Change the BulkGetProfilesByID DB query to return an embedded profile** - This will make it easier to extend what the query returns
- **Extend the BulkGetProfilesByID DB query with selectors** - This is to return the selectors at the same time as querying the profiles from the DB and be able to use them in the executor.
- **Extend the ProfileAggregate structure with selectors** - returns the selectors to the engine through the ProfileAggregate struct
- **Change the selector's NewSelectionFromProfile API to accept the structure from the models API** - This will make it easier to use the method in the executor
- **Integrate the selectors into the executor** - When an entity is not selected for a profile, let's save that into a  profile-global status override and use that for all the statuses instead of calling eval
- **Include which selector did unselect the entity** - In addition to returning a bool, let's also return which selector did shortcut the evaluation when returning false

Fixes: #3724
Fixes: #3725 

## Change Type

# Testing

manual

# Review Checklist:

### DIFF
--- a/database/query/profiles.sql
+++ b/database/query/profiles.sql
@@ -126,6 +126,6 @@ GROUP BY r.entity_type;
 SELECT COUNT(*) AS num_named_profiles FROM profiles WHERE lower(name) = lower(sqlc.arg(name));
 
 -- name: BulkGetProfilesByID :many
-SELECT *
+SELECT sqlc.embed(profiles)
 FROM profiles
 WHERE id = ANY(sqlc.arg(profile_ids)::UUID[]);
--- a/database/query/profiles.sql
+++ b/database/query/profiles.sql
@@ -126,6 +126,17 @@ GROUP BY r.entity_type;
 SELECT COUNT(*) AS num_named_profiles FROM profiles WHERE lower(name) = lower(sqlc.arg(name));
 
 -- name: BulkGetProfilesByID :many
-SELECT sqlc.embed(profiles)
+WITH helper AS(
+    SELECT pr.id as profid,
+           ARRAY_AGG(ROW(ps.id, ps.profile_id, ps.entity, ps.selector, ps.comment)::profile_selector) AS selectors
+    FROM profiles pr
+             JOIN profile_selectors ps
+                  ON pr.id = ps.profile_id
+    WHERE pr.id = ANY(sqlc.arg(profile_ids)::UUID[])
+    GROUP BY pr.id
+)
+SELECT sqlc.embed(profiles),
+       helper.selectors::profile_selector[] AS profiles_with_selectors
 FROM profiles
+LEFT JOIN helper ON profiles.id = helper.profid
 WHERE id = ANY(sqlc.arg(profile_ids)::UUID[]);
--- a/internal/engine/executor.go
+++ b/internal/engine/executor.go
@@ -136,8 +136,9 @@ func (e *executor) EvalEntityEvent(ctx context.Context, inf *entities.EntityInfo
 		return fmt.Errorf("error while retrieving profiles and rule instances: %w", err)
 	}
 
-	// For each profile, get the profile-override status. Then, if there is no profile-override status,
-	// evaluate each rule and store the outcome in the database or store the override status for all rules
+	// For each profile, get the profileEvalStatus first. Then, if the profileEvalStatus is nil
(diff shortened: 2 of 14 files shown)


database/query/profiles.sql
```diff
@@ -126,6 +126,17 @@ GROUP BY r.entity_type;
...
+    SELECT pr.id as profid,
+           ARRAY_AGG(ROW(ps.id, ps.profile_id, ps.entity, ps.selector, ps.comment)::profile_selector) AS selectors
+    FROM profiles pr
+             JOIN profile_selectors ps
+                  ON pr.id = ps.profile_id
+    WHERE pr.id = ANY(sqlc.arg(profile_ids)::UUID[])
+    GROUP BY pr.id
+)
+SELECT sqlc.embed(profiles),
+       helper.selectors::profile_selector[] AS profiles_with_selectors
```
jhrozek, line 139: If this is too hairy, we can instead just loop over the profiles and do a get for each profile

internal/engine/selectors/selectors.go
```diff
@@ -107,6 +107,7 @@ func newEnvForEntity(varName string, typ any, typName string) (*cel.Env, error)
 }
 
 type compiledSelector struct {
+	orig    string
```
jhrozek, line 110: there is ast.ToString but it doesn't guarantee that the result will be exactly the same, only that it will have the same meaning..

internal/engine/selectors/selectors.go
```diff
@@ -258,9 +260,9 @@ type EntitySelection struct {
 }
 
 // Select return true if the entity matches all the compiled expressions and false otherwise
-func (s *EntitySelection) Select(se *internalpb.SelectorEntity, userOpts ...SelectOption) (bool, error) {
+func (s *EntitySelection) Select(se *internalpb.SelectorEntity, userOpts ...SelectOption) (bool, string, error) {
```
jhrozek, line 263: I started wondering if it would be better to return a struct that would contain bool and string for now. Three return values from a function already starts to smell especially if the function is used in an interface.

database/query/profiles.sql
jhrozek: on the other hand, having everything in a single call is probably safer in the long run

internal/engine/selectors/selectors.go
JAORMX: I think having a dedicated struct would be cleaner in terms of returning it and using it. A bool and a string can get confusing. However, this is not blocking IMO and can be done later.

internal/engine/executor.go
```diff
@@ -131,10 +136,14 @@ func (e *executor) EvalEntityEvent(ctx context.Context, inf *entities.EntityInfo
 		return fmt.Errorf("error while retrieving profiles and rule instances: %w", err)
 	}
 
-	// For each profile, evaluate each rule and store the outcome in the database
+	// For each profile, get the profile-override status. Then, if there is no profile-override status,
+	// evaluate each rule and store the outcome in the database or store the override status for all rules
```
JAORMX, line 140: remind me what the "profile override status" is

JAORMX (dismissed):
logic LGTM

internal/engine/executor.go
```diff
@@ -131,10 +136,14 @@ func (e *executor) EvalEntityEvent(ctx context.Context, inf *entities.EntityInfo
...
 	for _, profile := range profileAggregates {
+
+		profileEvalStatus := e.profileEvalStatus(ctx, provider, inf, profile)
```
jhrozek, line 143: @JAORMX I called the variable actually `profileEvalStatus`, but "profile-override status" in the comment. I'll fix the naming since it's confusing.

JAORMX (approved):

internal/engine/selectors/selectors.go
evankanderson: Alternatively, you could use presence or absence of the `error` as part of the boolean (and a particular error interface to indicate "not selected"), but I'm happy with this interface.

PR #4033, opened 2024-07-29
Backfill new columns in `rule_evaluations`

dmjb:
# Summary

Backfill the `rule_instance_id` and `rule_entity_id` columns and make them non
nullable. Note that the backfilling of `rule_entity_id` should not change any
rows in production - any `rule_evaluation` rows with corresponding entries in
`evaluation_rule_entities` should already have a non null `rule_entity_id`
column.

As a side effect of setting up the `rule_instance_id` foreign key association
it is no longer necessary to explicitly delete old rule statuses when updating
profiles - instead they are deleted by the FK association. This simplifies the
profile update code significantly.

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/database/migrations/000083_rule_evaluations_ere_id_backfill.up.sql
+++ b/database/migrations/000083_rule_evaluations_ere_id_backfill.up.sql
@@ -1,0 +1,64 @@
+-- Copyright 2024 Stacklok, Inc
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+-- backfill rows which don't have an rule instance ID
+UPDATE rule_evaluations
+SET rule_instance_id = ri.id
+FROM rule_instances AS ri
+WHERE rule_evaluations.rule_instance_id IS NULL
+    AND rule_evaluations.rule_name = ri.name
+    AND rule_evaluations.rule_type_id = ri.rule_type_id;
+
+-- backfill rows which don't have an rule entity ID
+-- This process only matches up rule_evaluations which have a corresponding
+-- evaluation_rule_entities entry. Entities which were last evaluated before
+-- the introduction of the evaluation history tables will not have entries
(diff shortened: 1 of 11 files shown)


database/migrations/000083_rule_evaluations_ere_id_backfill.up.sql
```diff
@@ -0,0 +1,64 @@
...
+FROM evaluation_rule_entities AS ere
+WHERE rule_evaluations.rule_entity_id IS NULL
+  AND rule_evaluations.entity = 'repository'::entities
+  AND rule_evaluations.rule_instance_id = ere.rule_id
+  AND rule_evaluations.repository_id = ere.repository_id;
+
+-- make field mandatory
+ALTER TABLE rule_evaluations ALTER COLUMN rule_instance_id SET NOT NULL;
+-- note that rule_entity_id is still expected to contain nulls until we backfill
+-- evaluation state which predates the evaluation history tables.
```
evankanderson, line 62: Thanks for this note, I was just going to ask about that.

Should we separate the `UPDATE` and `ALTER TABLE` for `rule_instance_id` to be adjacent / at the end of the transaction?  I'm worried that we might end up taking out table-level locks and running into conflicts.

evankanderson (commented):
I like the reduction in complexity in profile service, though I'm slightly concerned that it didn't affect any test code in that directory, and it seems to have _reduced_ code coverage by deleting that code.  I'm guessing the removed code was tested elsewhere somehow (or maybe we just didn't test anything about those rows, and used a fake rather than a mock)?

dmjb:
> I like the reduction in complexity in profile service, though I'm slightly concerned that it didn't affect any test code in that directory, and it seems to have _reduced_ code coverage by deleting that code. I'm guessing the removed code was tested elsewhere somehow (or maybe we just didn't test anything about those rows, and used a fake rather than a mock)?

I had noticed this as well. The old tests do not seem to set up any old statuses, or make assertions about their deletion. They also use embedded postgres instead of mocks which is why the tests did not require changes.

I plan to write some more test for the profile service once I've removed the usage of the `entity_profiles` table from the profile service.

database/migrations/000083_rule_evaluations_ere_id_backfill.up.sql
dmjb: The update statements for `rule_entity_id` depend on the `rule_instance_id` column being populated.

dmjb: Also - due to the relatively large batch of changes which went out in the last prod promo (which included turning on eval history for everyone, and adding the rule_instance_id and rule_entity_id columns) there should not be any rows in production which need migration. These migrations should only affect staging (aside from making the `rule_instance_id` non nullable)

evankanderson (approved):

PR #4089, opened 2024-08-06
Change profile/rule status queries to use evaluation history table

dmjb:
Depends on: #4065 

This change allows most of the functionality which previously used the `rule_evaluations` (and related) tables to use the new evaluation history tables instead.

Change in API behaviour: `GetProfileStatusByName` previously returned repo information when querying Github pull requests and artifacts. This behaviour relied on a quirk of the old rule evaluation tables, and is no longer carried forward in the new implementation.

# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/cmd/dev/app/rule_type/rttst.go
+++ b/cmd/dev/app/rule_type/rttst.go
@@ -26,7 +26,6 @@ import (
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"github.com/sqlc-dev/pqtype"
 	"google.golang.org/protobuf/reflect/protoreflect"
 
 	serverconfig "github.com/stacklok/minder/internal/config/server"
@@ -132,15 +131,12 @@ func testCmdRun(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("error reading fragment from file: %w", err)
 	}
 
-	remediateStatus := db.NullRemediationStatusTypes{}
+	remediateStatus := db.RemediationStatusTypesSkipped
 	if rstatus.Value.String() != "" {
-		remediateStatus = db.NullRemediationStatusTypes{
-			RemediationStatusTypes: db.RemediationStatusTypes(rstatus.Value.String()),
-			Valid:                  true,
-		}
+		remediateStatus = db.RemediationStatusTypes(rstatus.Value.String())
 	}
 
-	remMetadata := pqtype.NullRawMessage{}
+	var remMetadata json.RawMessage = []byte("{}")
 	if rMetaPath.Value.String() != "" {
 		f, err := os.Open(filepath.Clean(rMetaPath.Value.String()))
 		if err != nil {
@@ -153,10 +149,7 @@ func testCmdRun(cmd *cobra.Command, _ []string) error {
 			return fmt.Errorf("error decoding json: %w", err)
 		}
 
-		remMetadata = pqtype.NullRawMessage{
-			RawMessage: jsonMetadata,
-			Valid:      true,
-		}
+		remMetadata = jsonMetadata
 	}
 
 	// Disable actions
@@ -245,8 +238,8 @@ func runEvaluationForRules(
 	inf *entities.EntityInfoWrapper,
 	provider provifv1.Provider,
 	entitySelectors selectors.Selection,
-	remediateStatus db.NullRemediationStatusTypes,
-	remMetadata pqtype.NullRawMessage,
+	remediateStatus db.RemediationStatusTypes,
+	remMetadata json.RawMessage,
 	frags []*minderv1.Profile_Rule,
 	actionEngine *actions.RuleActionsEngine,
 ) error {
--- a/database/query/profile_status.sql
+++ b/database/query/profile_status.sql
@@ -78,39 +78,40 @@ WHERE p.project_id = $1;
 -- cast after MIN is required due to a known bug in sqlc: https://github.com/sqlc-dev/sqlc/issues/1965
 
 -- name: ListOldestRuleEvaluationsByRepositoryId :many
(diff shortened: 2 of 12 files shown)


cmd/dev/app/rule_type/rttst.go
```diff
@@ -132,15 +131,12 @@ func testCmdRun(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("error reading fragment from file: %w", err)
 	}
 
-	remediateStatus := db.NullRemediationStatusTypes{}
+	remediateStatus := db.RemediationStatusTypesSkipped
```
dmjb, line 134: A recurring theme with many of the code changes in this PR is that the old queries returned a lot more null types. Due to the schema of the new tables, and the fact that evaluation history is now inserted transactionally, there are far fewer nullable fields.

internal/util/cli/table/simple/simple.go
```diff
@@ -117,7 +117,7 @@ func profileStatusLayout(table *tablewriter.Table) {
 func ruleEvaluationsLayout(table *tablewriter.Table) {
 	defaultLayout(table)
 	table.SetHeader([]string{
-		"Rule ID", "Rule Name", "Entity", "Status", "Remediation", "Entity Info"})
+		"Rule Type ID", "Rule Name", "Entity", "Status", "Remediation", "Entity Info"})
```
dmjb, line 120: Spotted this during local testing. "Rule ID" is confusing - the table is in fact showing a rule type ID here.

database/query/profile_status.sql
```diff
@@ -137,21 +138,23 @@ SELECT
     rt.id AS rule_type_id,
     rt.guidance as rule_type_guidance,
     rt.display_name as rule_type_display_name
-FROM rule_evaluations res
```
dmjb, line 140: In the evaluation history tables, we are guaranteed that there will always be one alert and remediation row per evaluation status, and at least one evaluation status and latest evaluation status per evaluation rule entity. This is because the data is inserted transactionally into the DB.

cmd/dev/app/rule_type/rttst.go
evankanderson: So "skipped" now means actually-skipped, or "didn't report a status"?

evankanderson: (I do like removing null values and replacing with sentinels where possible, I'm just wondering whether having `""` as a default/valid value would be a good way to handle this.)

database/query/profile_status.sql
```diff
@@ -137,21 +138,23 @@ SELECT
...
-         INNER JOIN rule_type rt ON rt.id = res.rule_type_id
-WHERE res.profile_id = $1 AND
+FROM latest_evaluation_statuses les
+         INNER JOIN evaluation_rule_entities ere ON ere.id = les.rule_entity_id
+         INNER JOIN eval_details ed ON ed.id = les.evaluation_history_id
+         INNER JOIN remediation_details rd ON rd.evaluation_id = les.evaluation_history_id
+         INNER JOIN alert_details ad ON ad.evaluation_id = les.evaluation_history_id
+         INNER JOIN rule_instances AS ri ON ri.id = ere.rule_id
+         INNER JOIN rule_type rt ON rt.id = ri.rule_type_id
+         LEFT JOIN repositories repo ON repo.id = ere.repository_id
```
evankanderson, line 148: I'm not thrilled about this pile of joins, but I don't have a substantially better solution at the moment.

database/query/profile_status.sql
```diff
@@ -125,10 +126,10 @@ SELECT
...
-    res.repository_id,
-    res.entity,
-    res.rule_name,
+    ed.id AS rule_evaluation_id,
+    ere.repository_id,
+    ere.entity_type,
+    ri.name AS rule_name,
     repo.repo_name,
     repo.repo_owner,
     repo.provider,
```
evankanderson, line 135: I agree that pulling this repo stuff out here is a bit weird (especially since it seems GitHub-specific!).  How much does it improve things if we drop these contents from the RPC?  (And, is anyone using them right now / when were they added?)

evankanderson: If we _do_ need them, we could do a second query in the RPC, which feels like a better place, to be honest.

evankanderson (approved):
A bunch of smaller comments, but nothing that feels obviously-wrong / blocking.

database/query/profile_status.sql
dmjb: I think we should get rid of them, but removing them will impact the UI (from what I understand, one of the endpoints which uses this query powers the "Alerts" page in the UI) so we need to co-ordinate this change.

cmd/dev/app/rule_type/rttst.go
dmjb: Keep in mind that this file implements the CLI command for testing profiles/rules. The use of "Skipped" as a sentinel value vs NULL is of little consequence since this command persist anything to the DB.

database/query/profile_status.sql
evankanderson: What do you think of collecting the repo and other entity details in a _separate_ database call outside of this query?

If the frontend is using that information (and it makes sense that they would), it seems a shame to make them make two RPCs instead of one.  At the same time, I don't want this query getting bigger if we can avoid it, so maybe doing some iteration in the Minder server is better than extending the query.

cmd/dev/app/rule_type/rttst.go
evankanderson: Oh, I completely forgot that this bit existed.  Nevermind!

JAORMX:
So, with this change, what would be the output of `minder profile status list ... -d` ?

Issue #2723, opened 2024-03-20
--owner flag value is not validated while enrolling a provider


rdimitrov:
We still get the `Provider enrolled successfully` message even if we pass a non-existing org value while enrolling a provider via the `--owner` flag.

psekar:
Hello @rdimitrov 
I would like to contribute to this bug fix. Below is the question that i would like to get clarified before sending the fix.

We could validate if the users access token can query the list of repos from the user provided owner (GH organization) during provider enrollment. Incase of the API failure, should we invoke delete provider workflow ?

Thanks for your guidance 👍

PR #4107, opened 2024-08-09
Add owner flag validation during provider enrollment

psekar:
# Summary

This PR addresses the issue #2723 . When enrolling a provider, the owner flag is not validated. Incase the owner flag is invalid or the token has no access to the owner/organization during provider enrollment, the repo registration will fail which is a later step in the process. The change includes checking for a valid GH organization (passed as owner flag) that matches the owner flag and ensure the token has access to the GH organization.

I created a type struct to make Git API calls prior to actual provider enrollment flow, also created an interface to be able to mock the tests. The intent of the existing types seems to be useful post provider enrollment so decided to create a new type.

Fixes #(related issue)
#2723 
## Change Type

# Testing

👉 Existing tests will validate the valid owner path.
👉 Added the unit tests to validate the invalid owner flag code path. 
👉  Provide a valid owner flag (a GH org) and provider enrollment is successful
👉  Provide an invalid owner flag and provider enrollment fails during the oauth flow with clear error message about the owner flag.

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_oauth.go
+++ b/internal/controlplane/handlers_oauth.go
@@ -291,6 +291,14 @@ func (s *Server) processOAuthCallback(ctx context.Context, w http.ResponseWriter
 		return fmt.Errorf("error encoding token: %w", err)
 	}
 
+	// Verify if the token user is a member of the ownerFilter (Organization)
+	if stateData.OwnerFilter.Valid {
+		err := s.validateOwnerFilter(ctx, token, stateData.OwnerFilter.String)
+		if err != nil {
+			return err
+		}
+	}
+
 	var errConfig providers.ErrProviderInvalidConfig
 
 	p, err := s.sessionService.CreateProviderFromSessionState(ctx, db.ProviderClass(provider), &encryptedToken, state)
@@ -321,6 +329,20 @@ func (s *Server) processOAuthCallback(ctx context.Context, w http.ResponseWriter
 		return fmt.Errorf("error writing OAuth success page: %w", err)
 	}
 
+	return nil
+}
+
+func (s *Server) validateOwnerFilter(ctx context.Context, token *oauth2.Token, ownerFilter string) error {
+	member, err := s.ghProviders.ValidateOrgMembershipForToken(ctx, token, ownerFilter)
+	if err != nil {
+		return fmt.Errorf("error checking organization membership: %w", err)
+	}
+
+	if !member {
+		return newHttpError(http.StatusForbidden, "Invalid OwnerFilter Provided").SetContents(
+			"You do not have access to the organization %s, specified in the owner filter."+
+				" Please ensure that you are a member of the GitHub organization and try again.", ownerFilter)
+	}
 	return nil
 }
 
--- a/internal/controlplane/handlers_oauth_test.go
+++ b/internal/controlplane/handlers_oauth_test.go
@@ -462,6 +462,18 @@ func TestProviderCallback(t *testing.T) {
 			Return([]db.Provider{}, nil)
 	}
 
+	withValidOrgMemberships := func(service *mockprovsvc.MockGitHubProviderService) {
+		service.EXPECT().ValidateOrgMembershipForToken(gomock.Any(), gomock.Any(), gomock.Any()).Return(true, nil)
+	}
+
+	withInvalidOrgMemberships := func(service *mockprovsvc.MockGitHubProviderService) {
+		service.EXPECT().ValidateOrgMembershipForToken(gomock.Any(), gomock.Any(), gomock.Any()).Return(false, nil)
+	}
+
+	orgMembershipWithFailure := func(service *mockprovsvc.MockGitHubProviderService) {
+		service.EXPECT().ValidateOrgMembershipForToken(gomock.Any(), gomock.Any(), gomock.Any()).Return(false, errors.New("failed to validate org membership"))
+	}
+
 	testCases := []struct {
 		name                       string
 		redirectUrl                string
@@ -471,6 +483,7 @@ func TestProviderCallback(t *testing.T) {
 		projectIDBySessionNumCalls int
 		err                        string
 		config                     []byte
+		buildStubs                 func(service *mockprovsvc.MockGitHubProviderService)
 	}{{
 		name:                       "Success",
 		redirectUrl:                "http://localhost:8080",
@@ -478,7 +491,8 @@ func TestProviderCallback(t *testing.T) {
 		storeMockSetup: func(store *mockdb.MockStore) {
 			withProviderSearch(store)
 		},
-		code: 307,
+		code:       307,
+		buildStubs: withValidOrgMemberships,
 	}, {
 		name:                       "Success with remote user",
 		redirectUrl:                "http://localhost:8080",
@@ -487,7 +501,8 @@ func TestProviderCallback(t *testing.T) {
 		storeMockSetup: func(store *mockdb.MockStore) {
 			withProviderSearch(store)
 		},
-		code: 307,
+		code:       307,
+		buildStubs: withValidOrgMemberships,
 	}, {
 		name:                       "Wrong remote userid",
 		remoteUser:                 sql.NullString{Valid: true, String: "1234"},
@@ -497,6 +512,9 @@ func TestProviderCallback(t *testing.T) {
 		},
 		code: 403,
 		err:  "The provided login token was associated with a different user.\n",
+		buildStubs: func(_ *mockprovsvc.MockGitHubProviderService) {
+			// this codepath fails before the ghProviderService is called
+		},
 	}, {
 		name:        "No existing provider",
 		redirectUrl: "http://localhost:8080",
@@ -508,6 +526,7 @@ func TestProviderCallback(t *testing.T) {
 		storeMockSetup: func(store *mockdb.MockStore) {
 			withProviderCreate(store)
 		},
+		buildStubs: withValidOrgMemberships,
 	}, {
 		name:                       "Config does not validate",
 		redirectUrl:                "http://localhost:8080",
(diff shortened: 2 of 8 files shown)


internal/controlplane/handlers_oauth.go
```diff
@@ -280,6 +280,22 @@ func (s *Server) processOAuthCallback(ctx context.Context, w http.ResponseWriter
 	if err != nil {
 		return fmt.Errorf("error encoding token: %w", err)
 	}
+ 	login, err := s.ghLite.GetLoginByToken(ctx, token);
+	if (err != nil) {
+		return fmt.Errorf("error getting login: %w", err)
+	}
+	// Verify if the token user is a member of the ownerFilter (Organization)
+	if (stateData.OwnerFilter.Valid) {
+		member, err := s.ghLite.IsMemberInOrganization(ctx, token, stateData.OwnerFilter.String, login)
```
eleftherias, line 289: What do you think about using the [List org memberships for authenticated users](https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#list-organization-memberships-for-the-authenticated-user) API instead? Then we would only need 1 API call instead of 2.

eleftherias (changes requested):
Thanks for the contribution @psekar!

This approach looks good to me! I've added a few comments inline with suggestions that will simplify the code. 
Feel free to reach out if you have any questions.

README.md
```diff
@@ -206,6 +206,12 @@ See the documentation on how to do that - [Docs](https://minder-docs.stacklok.de
 
 #### Run `minder-server`
 
+For local development, create the token_key_passphrase in .ssh folder. For production, these tokens can be served from a secure vault.
+
+```bash
+openssl rand -base64 32 > .ssh/token_key_passphrase
```
dmjb, line 212: This is typically carried out by using the `make bootstrap` command. I would suggest that you document the make command instead of the individual steps it carries out.

internal/providers/github/common.go
```diff
@@ -86,6 +86,11 @@ type GitHub struct {
 // Ensure that the GitHub client implements the GitHub interface
 var _ provifv1.GitHub = (*GitHub)(nil)
 
+type GithubLite struct {
```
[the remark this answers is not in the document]
psekar: yea, this makes sense. Thanks for pointing it out.

internal/controlplane/handlers_oauth.go
psekar: Good catch. Yes this API serves the purpose here.

README.md
psekar: Sure 👍

internal/controlplane/handlers_githubwebhooks_test.go
```diff
@@ -117,7 +117,7 @@ func (s *UnitTestSuite) TestHandleWebHookPing() {
 	defer ctrl.Finish()
 
 	mockStore := mockdb.NewMockStore(ctrl)
-	srv, evt := newDefaultServer(t, mockStore, nil)
+	srv, evt := newDefaultServer(t, mockStore, nil, true)
```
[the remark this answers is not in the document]
psekar: I am mocking the behavior in the test but i am ghProviderService as an argument so other tests can still get a non mocked version of GhProviderService. Let me know if this makes sense.

internal/controlplane/handlers_oauth_test.go
```diff
@@ -518,6 +527,18 @@ func TestProviderCallback(t *testing.T) {
...
+		buildStubs: withValidOrgMemberships,
+	}, {
+		name:                       "Success with invalid ownerfilter",
+		redirectUrl:                "http://localhost:8080",
+		projectIDBySessionNumCalls: 1,
+		storeMockSetup: func(store *mockdb.MockStore) {
+		},
+		code: 403,
+		err: "The ownerFilter value provided does not allow access for the logged in user\n",
+		buildStubs: func(service *mockprovsvc.MockGitHubProviderService) {
```
eleftherias, line 539: For consistency could this be a function `withInvalidOrgMemberships`

internal/controlplane/handlers_oauth_test.go
```diff
@@ -497,6 +504,7 @@ func TestProviderCallback(t *testing.T) {
 		},
 		code: 403,
 		err:  "The provided login token was associated with a different user.\n",
+		buildStubs: nil,
```
eleftherias, line 507: This should be a no-op function. Setting it to nil causes a nil pointer dereference error.

internal/controlplane/handlers_oauth_test.go
```diff
@@ -572,7 +593,9 @@ func TestProviderCallback(t *testing.T) {
 			}
 			tc.storeMockSetup(store)
 
-			s, _ := newDefaultServer(t, store, clientFactory)
+			mockGhProviderService := mockprovsvc.NewMockGitHubProviderService(ctrl)
+			tc.buildStubs(mockGhProviderService)
+			s, _ := newDefaultServer(t, store, clientFactory, mockGhProviderService)
```
eleftherias, line 598: Instead of changing the signature of `newDefaultServer` let's set `s.ghProviders = mockGhProviderService` after creating the server. That's what we do for `providerAuthManager`, `sessionService` etc.

eleftherias (changes requested):
Thanks for the updates @psekar! 

I've added a few follow-up comments.
You can also run `make lint` before committing to makes sure the code passes the lint rules.

internal/controlplane/handlers_oauth_test.go
psekar: I do see this pattern in other places too after your comment. I tried not to change the server behavior outside of server_test but then we create a new instance so it shouldnt matter. Thanks for pointing it out 👍

psekar:
> Thanks for the updates @psekar!
> 
> I've added a few follow-up comments. You can also run `make lint` before committing to makes sure the code passes the lint rules.

I am still very new to Go and your review has been super useful 🙏

internal/controlplane/handlers_oauth.go
```diff
@@ -290,6 +290,18 @@ func (s *Server) processOAuthCallback(ctx context.Context, w http.ResponseWriter
...
+	if stateData.OwnerFilter.Valid {
+		member, err := s.ghProviders.ValidateOrgMembershipForToken(ctx, token, stateData.OwnerFilter.String)
+		if err != nil {
+			return fmt.Errorf("error checking organization membership: %w", err)
+		}
+
+		if !member {
+			return newHttpError(http.StatusForbidden, "Invalid OwnerFilter Provided").SetContents(
+				"The ownerFilter value provided does not allow access for the logged in user")
+		}
```
eleftherias, line 303: You may have noticed that the linter is complaining that the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of this function is too high.

Essentially, this means this function has several code paths which makes it hard to reason about when reading through it.

We can fix this by extracting the block here into a separate `validateOwnerFilter` function:
```golang
if stateData.OwnerFilter.Valid {
	err := s.validateOwnerFilter(ctx, token, stateData)
	if err != nil {
		return err
	}
}
```

eleftherias (changes requested):
Thanks for the quick updates again @psekar!

I've left some final comments about cleanup / linting changes.

eleftherias (commented):
I also noticed that the mocks need to be updated. You can run `make gen` in the root directory before committing your code and that will update them.

internal/controlplane/handlers_oauth.go
psekar: I fixed the 2 lint issues that i overlooked in my last review. Thanks

internal/controlplane/handlers_oauth.go
```diff
@@ -290,6 +290,18 @@ func (s *Server) processOAuthCallback(ctx context.Context, w http.ResponseWriter
...
+	// Verify if the token user is a member of the ownerFilter (Organization)
+	if stateData.OwnerFilter.Valid {
+		member, err := s.ghProviders.ValidateOrgMembershipForToken(ctx, token, stateData.OwnerFilter.String)
+		if err != nil {
+			return fmt.Errorf("error checking organization membership: %w", err)
+		}
+
+		if !member {
+			return newHttpError(http.StatusForbidden, "Invalid OwnerFilter Provided").SetContents(
+				"The ownerFilter value provided does not allow access for the logged in user")
```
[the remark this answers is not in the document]
psekar: After the refactor, I noticed we will be returning the same error message incase of invalid org or any github API failures. Do you think we should handle the github API failure separately ?

psekar:
> [Coverage Status](https://coveralls.io/builds/69189792)
> 
> coverage: 53.872% (-0.03%) from 53.899% when pulling **[94472cb](https://github.com/stacklok/minder/commit/94472cb57df55af1bb0091a5f59563bb9a55c74b) on tinytrail:validate-owner-filter** into **[e438848](https://github.com/stacklok/minder/commit/e438848a25bacfedc538d5f604e3490ec48b1683) on stacklok:main**.

Will add more test coverage and send it back for review tomorrow.

internal/controlplane/handlers_oauth.go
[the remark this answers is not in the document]
eleftherias: Instead of returning the boolean from the helper function, we can just return the error. Then the caller can directly return the error as well. This way, we can keep the error messages for invalid org or GitHub API failures.
It's fairly common to do that in our codebase, for example:
https://github.com/stacklok/minder/blob/16421648e284c9f5be7c5261aaa8e5a94b61c70c/internal/controlplane/handlers_oauth.go#L345-L348

psekar:
> > [Coverage Status](https://coveralls.io/builds/69189792)
> > coverage: 53.872% (-0.03%) from 53.899% when pulling **[94472cb](https://github.com/stacklok/minder/commit/94472cb57df55af1bb0091a5f59563bb9a55c74b) on tinytrail:validate-owner-filter** into **[e438848](https://github.com/stacklok/minder/commit/e438848a25bacfedc538d5f604e3490ec48b1683) on stacklok:main**.
> 
> Will add more test coverage and send it back for review tomorrow.

Fixed the linter issues and added more test coverage now. @eleftherias Appreciate your review 👍

internal/controlplane/handlers_oauth.go
```diff
@@ -291,6 +291,19 @@ func (s *Server) processOAuthCallback(ctx context.Context, w http.ResponseWriter
...
+		member, err := s.validateOwnerFilter(ctx, token, stateData.OwnerFilter.String)
+		if (err != nil) {
+			return err
+		}
+		if !member {
+			return newHttpError(http.StatusForbidden, "Invalid OwnerFilter Provided").SetContents(
+				"You do not have access to the organization %s, specified in the owner filter."+
+					" Please ensure that you are a member of the GitHub organization and try again.", stateData.OwnerFilter.String)
+		}
+	}
```
eleftherias, line 305: We don't need to return the `member` boolean from the helper, we can simply return the error, and then this second `if` statement is not needed.

The helper would be:
```
func (s *Server) validateOwnerFilter(ctx context.Context, token *oauth2.Token, ownerFilter string) error {
	member, err := s.ghProviders.ValidateOrgMembershipForToken(ctx, token, ownerFilter)
	if err != nil {
		return fmt.Errorf("error checking organization membership: %w", err)
	}

	if !member {
		return newHttpError(http.StatusForbidden, "Invalid OwnerFilter Provided").SetContents(
			"You do not have access to the organization %s, specified in the owner filter."+
				" Please ensure that you are a member of the GitHub organization and try again.", ownerFilter)
	}
	return nil
}
```

eleftherias (changes requested):
Thank you for your patience on this @psekar!
The cyclomatic complexity of `processOAuthCallback` is still high, I've proposed a change inline that will fix it. After this change, we should be ready to merge.

internal/controlplane/handlers_oauth.go
psekar: I recall running `make lint` before i sent the change to upstream, but i might have missed running it after the last refactor. Thanks for catching this.

PR #4123, opened 2024-08-12
Add central entities table with properties

JAORMX:
# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


jhrozek:
One sqlc-generated file was forgotten, so I force-pushed to your branch to see if CI would be greener.

database/migrations/000090_entity_properties.up.sql
```diff
@@ -0,0 +1,51 @@
...
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS entities (
+	id          UUID NOT NULL DEFAULT gen_random_uuid() PRIMARY KEY,
+	entity_type entities NOT NULL, 
+	name        TEXT NOT NULL,
+	project_id  UUID NOT NULL REFERENCES project(id),
+	provider_id UUID NOT NULL REFERENCES provider(id),
+	created_at  TIMESTAMP NOT NULL DEFAULT NOW(),
+	most_recent_evaluation TIMESTAMP, -- last time a profile/rule was run against this entity
```
jhrozek, line 24: If we don't have a use for this column, should we add it already?

JAORMX: We can remove it, no biggie.

database/migrations/000090_entity_properties.up.sql
```diff
@@ -0,0 +1,51 @@
...
+-- in code as we need to.
+
+INSERT INTO entities (id, entity_type, name, project_id, provider_id, created_at)
+SELECT id, 'repository', name, project_id, provider_id, created_at FROM repositories;
+
+INSERT INTO entities (id, entity_type, name, project_id, provider_id, created_at)
+SELECT id, 'artifact', name, project_id, provider_id, created_at FROM artifacts;
+
+INSERT INTO entities (id, entity_type, name, project_id, provider_id, created_at)
+SELECT id, 'pull_request', name, project_id, provider_id, created_at FROM pull_requests;
```
jhrozek, line 51: I'm not sure this is enough. Our infra first migrates the DB while the old code is running and then upgrades the minder code, which creates a race where an entry might be added to the old tables but never to the new ones.

What if we create a SQL trigger that starts writing to the new table on create and update instead, commit that first so that the existing code writes to the new tables and then copy the existing data? (I also wonder about the size of the data and if we'd create a huge transaction, but that's a matter of checking our SaaS data size)

JAORMX: I think we should just choose a migration on the golang migration wrapper.

evankanderson: +1 to doing the migration in the golang wrapper in a later PR after the tables are created and we start writing to them from the application code (but before we assume they are there on the read side).

I worry about doing SQL bulk inserts / updates leading to database locks, though I acknowledge that we haven't had those problems yet.

jhrozek (approved):

PR #4151, opened 2024-08-14
Provider properties interface

jhrozek:
# Summary

This PR adds a new provider interface for fetching properties as well as its implementation for the github provider. The interface returns new Property types which are convenience types to allow returning Go types from the unstructured properties returned by the providers. Lastly, the PR adds constants for the currently used properties, either generic entity properties, generic repo properties or github-specific repo properties.

Closes: https://github.com/stacklok/minder/issues/4165

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/entities/properties/constants.go
+++ b/internal/entities/properties/constants.go
@@ -1,0 +1,34 @@
+//
+// Copyright 2024 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package properties
+
+// General entity keys
+const (
+	// PropertyName represents the name of the entity. The name is formatted by the provider
+	PropertyName = "name"
+	// PropertyUpstreamID represents the ID of the entity in the provider
+	PropertyUpstreamID = "upstream_id"
+)
+
+// Repository property keys
+const (
+	// RepoPropertyIsPrivate represents whether the repository is private
+	RepoPropertyIsPrivate = "is_private"
+	// RepoPropertyIsArchived represents whether the repository is archived
+	RepoPropertyIsArchived = "is_archived"
+	// RepoPropertyIsFork represents whether the repository is a fork
+	RepoPropertyIsFork = "is_fork"
+)
--- a/internal/entities/properties/properties.go
+++ b/internal/entities/properties/properties.go
@@ -1,0 +1,126 @@
+//
+// Copyright 2024 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package properties provides a simple way to access properties of an entity
+package properties
+
+import (
+	"fmt"
+
+	"github.com/puzpuzpuz/xsync/v3"
+)
+
+// Property is a struct that holds a value. It's just a wrapper around an interface{}
+// with typed getters and handling of an empty receiver
+type Property struct {
+	value any
+}
+
+// NewProperty creates a new Property with a given value
+func NewProperty(value any) *Property {
+	return &Property{value: value}
+}
+
+func propertyValueAs[T any](value any) (T, error) {
+	var zero T
+	val, ok := value.(T)
+	if !ok {
+		return zero, fmt.Errorf("value is not of type %T", zero)
+	}
+	return val, nil
+}
+
+// AsBool returns the boolean value, or an error if the value is not a boolean
+func (p *Property) AsBool() (bool, error) {
+	if p == nil {
+		return false, fmt.Errorf("property is nil")
+	}
+	return propertyValueAs[bool](p.value)
+}
+
+// GetBool returns the boolean value, or false if the value is not a boolean
+func (p *Property) GetBool() bool {
+	val, err := p.AsBool()
+	if err != nil {
+		return false
+	}
+	return val
+}
+
+// AsString returns the string value, or an error if the value is not a string
+func (p *Property) AsString() (string, error) {
+	if p == nil {
+		return "", fmt.Errorf("property is nil")
+	}
+	return propertyValueAs[string](p.value)
+}
+
+// GetString returns the string value, or an empty string if the value is not a string
+func (p *Property) GetString() string {
+	if p == nil {
+		return ""
+	}
+	val, err := p.AsString()
+	if err != nil {
+		return ""
+	}
+	return val
+}
+
+// AsInt64 returns the int64 value, or an error if the value is not an int64
+func (p *Property) AsInt64() (int64, error) {
+	if p == nil {
+		return 0, fmt.Errorf("property is nil")
+	}
+	return propertyValueAs[int64](p.value)
+}
+
+// GetInt64 returns the int64 value, or 0 if the value is not an int64
+func (p *Property) GetInt64() int64 {
+	if p == nil {
+		return 0
+	}
+	val, err := p.AsInt64()
+	if err != nil {
+		return 0
+	}
+	return val
+}
+
+// Properties struct that holds the properties map and provides access to Property values
+type Properties struct {
+	props *xsync.MapOf[string, Property]
+}
+
+// NewProperties Properties from a map
+func NewProperties(props map[string]any) *Properties {
+	propsMap := xsync.NewMapOf[string, Property](xsync.WithPresize(len(props)))
+
+	for key, value := range props {
+		propsMap.Store(key, Property{value: value})
+	}
+	return &Properties{
+		props: propsMap,
+	}
+}
+
+// GetProperty returns the Property for a given key or an empty one as a fallback
+func (p *Properties) GetProperty(key string) *Property {
+	prop, ok := p.props.Load(key)
+	if !ok {
+		return nil
+	}
+	return &prop
+}
--- a/internal/providers/github/properties.go
+++ b/internal/providers/github/properties.go
@@ -1,0 +1,171 @@
+//
+// Copyright 2024 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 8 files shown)


pkg/providers/v1/providers.go
```diff
@@ -186,6 +186,13 @@ type OCI interface {
 	GetAuthenticator() (authn.Authenticator, error)
 }
 
+type Properties interface {
+	Provider
+
+	FetchAllProperties(ctx context.Context, name string, entType minderv1.Entity) (map[string]any, error)
```
jhrozek, line 193: I'm not sure if "name" is a good name for the parameter and even if string will always be what the provider queries about. We can probably always pass a stringified ID, but IIRC some providers use a numerical ID in their upstream API anyway.

internal/providers/github/properties.go
```diff
@@ -0,0 +1,146 @@
...
+	for lang := range ghLangs {
+		langs = append(langs, lang)
+	}
+
+	return map[string]any{
+		"properties.languages": langs,
+	}, nil
+}
+
+func (c *GitHub) FetchProperty(ctx context.Context, name string, entType minderv1.Entity, key string) (any, error) {
```
jhrozek, line 110: The API is really a wrapper around upstream APIs, which means there's no caching. The newly introduced tables are global not table-per-provider, but there is a provider UUID available.

Evan suggested in the design document (and I agree) that it should be the provider who is resposible for caching. To make use of the SQL tables that we have and that are quite simple I was thinking about adding a service layer around this provider API that would:
 - fetch the properties from the DB, that would produce a slice of (property,timestamp) tuples
 - pass them to the provider API for cache evaluation through a new method of the provider API. This way the provider would be in control of the caching
 - if the provider replies with a cache miss, fetch the property using this method

internal/providers/github/properties.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+var repoPropertyDefinitions = []propertyOrigin{
+	{
+		keys:    []string{"is_private", "is_archived"},
+		wrapper: getRepoWrapper,
+	},
+	{
+		keys:    []string{"properties.languages"},
+		wrapper: getLanguagesWrapper,
```
jhrozek, line 48: Not thrilled about this struct, but we can refactor later as long as the interface works for its consumers.

internal/providers/github/properties.go
```diff
@@ -0,0 +1,146 @@
...
+		"is_private":  repo.GetPrivate(),
+		"is_archived": repo.GetArchived(),
+	}
+
+	return repoProps, nil
+}
+
+func getLanguagesWrapper(ctx context.Context, ghCli v1.GitHub, name string) (map[string]any, error) {
+	// TODO: this should be a provider interface, even if private
+	slice := strings.Split(name, "/")
```
jhrozek, line 90: I think we should have a provider name formatter API

internal/repositories/github/service.go
```diff
@@ -143,22 +143,41 @@ func (r *repositoryService) CreateRepository(
...
+		isArchived = false
 	}
 
 	// skip if this is an archived repo
-	if githubRepo.GetArchived() {
+	if isArchived {
 		return nil, ErrArchivedRepoForbidden
 	}
 
+	isPrivate, ok := repoProperties["is_private"].(bool)
```
jhrozek, line 166: It would be nice to have typed accessor methods

internal/repositories/github/service.go
```diff
@@ -440,3 +447,52 @@ func (r *repositoryService) persistRepository(
 
 	return outid, pbr, nil
 }
+
+func pbRepoFromProperties(
+	repoProperties *properties.Properties,
+	githubHook *github.Hook,
+	hookUUID string,
+) (*pb.Repository, error) {
+	name, err := repoProperties.GetProperty("name").AsString()
```
jhrozek, line 456: Ozz suggested that we should have named constants for the common properties and after having written this code I agree

internal/repositories/github/service.go
```diff
@@ -369,7 +387,7 @@ func (r *repositoryService) pushReconcilerEvent(pbRepo *pb.Repository, projectID
 // returns DB PK along with protobuf representation of a repo
 func (r *repositoryService) persistRepository(
 	ctx context.Context,
-	githubRepo *github.Repository,
+	repoProperties *properties.Properties,
 	githubHook *github.Hook,
```
jhrozek, line 391: we still need to get rid of the github hook if we want to make the code better reusable but most of the data we store into the pbRepository is not really used past registration..so we could probably get away with refactoring webhokManager to return a struct with a more minimal attribute set and just don't fill a lot of the attributes

jhrozek:
@JAORMX there are no tests, but manual testing works fine. I think the code is ready for review, but I won't take it out of draft until we have tests

internal/entities/properties/properties.go
```diff
@@ -0,0 +1,124 @@
...
+package properties
+
+import (
+	"fmt"
+)
+
+// Property is a struct that holds a value. It's just a wrapper around an interface{}
+// with typed getters and handling of an empty receiver
+type Property struct {
+	value any
```
jhrozek, line 26: btw I tried to use protobuf's Value here and protobuf's Struct for Properties but Value only has one numerical type which is limited to float64 so we'd have to work around storing int64 in full.

internal/entities/properties/properties.go
```diff
@@ -0,0 +1,124 @@
...
+// AsBool returns the boolean value, or an error if the value is not a boolean
+func (p *Property) AsBool() (bool, error) {
+	if p == nil {
+		return false, fmt.Errorf("property is nil")
+	}
+	return propertyValueAs[bool](p.value)
+}
+
+// GetBool returns the boolean value, or false if the value is not a boolean
+func (p *Property) GetBool() bool {
```
jhrozek, line 52: The idea behind having separate Get and As is that Get is a convenience function for when you're OK not having the data but just the default value for that type and As does return an error that you need to handle for critical values (repo name, repo id, ...)

internal/entities/properties/properties.go
```diff
@@ -0,0 +1,124 @@
...
+		propertyMap[key] = Property{value: value}
+	}
+	return &Properties{props: propertyMap}
+}
+
+// GetProperty returns the Property for a given key or an empty one as a fallback
+func (p *Properties) GetProperty(key string) *Property {
+	prop, ok := p.props[key]
+	if !ok {
+		return &Property{}
```
jhrozek, line 121: This is to allow `Properties.GetProperty("i_dont_exist").GetBool("is_something")` to work

JAORMX: I'm guessing this is a slipup and you meant something as:

```
priv := ps.GetProperty("is_private").GetBool()
```

since `GetBool` doesn't take arguments. I think this looks pretty good!

pkg/providers/v1/providers.go
```diff
@@ -143,6 +144,7 @@ type GitHub interface {
 	AddAuthToPushOptions(ctx context.Context, options *git.PushOptions) error
 	StartCheckRun(context.Context, string, string, *github.CreateCheckRunOptions) (*github.CheckRun, error)
 	UpdateCheckRun(context.Context, string, string, int64, *github.UpdateCheckRunOptions) (*github.CheckRun, error)
+	GetRepoLanguages(ctx context.Context, owner, repo string) (map[string]int, error)
```
JAORMX, line 147: Do we need to expose this? This can probably jus the an implementation detail of how GitHub populates properties.

internal/providers/github/properties.go
JAORMX: I agree. It's fine if it's a TODO for now.

internal/entities/properties/properties.go
```diff
@@ -0,0 +1,124 @@
...
+		return 0
+	}
+	val, err := p.AsInt64()
+	if err != nil {
+		return 0
+	}
+	return val
+}
+
+// Properties struct that holds the properties map and provides access to Property values
```
JAORMX, line 103: Let's document the assumption that this will have a short lifetime and only a single writer at a time. If we want to start parallelizing or even lazy loading, I'd suggest we switch to a syncmap instead of a raw map.

internal/repositories/github/service.go
JAORMX: The github webhook installation will be a trait by itself. But that will be generalized as part of another body of work. This is fine for now.

internal/entities/properties/constants.go
```diff
@@ -0,0 +1,31 @@
...
+// limitations under the License.
+
+package properties
+
+// Repository property keys
+const (
+	RepoPropertyName   = "name"
+	RepoPropertyOwner  = "owner"
+	RepoPropertyRepoID = "repo_id"
+	// TODO: is this GH-specific? If yes, it should be moved under a provider-specific property
```
JAORMX, line 23: Let's add the GitHub specific constants under GitHub. This should only have "standard" ones. e.g. we can keep "is_private", "is_archived", "is_fork".

internal/entities/properties/constants.go
```diff
@@ -0,0 +1,31 @@
...
+	RepoPropertyDeployURL     = "deploy_url"
+	RepoPropertyCloneURL      = "clone_url"
+	RepoPropertyIsPrivate     = "is_private"
+	RepoPropertyIsArchived    = "is_archived"
+	RepoPropertyIsFork        = "is_fork"
+	RepoPropertyDefaultBranch = "default_branch"
+	RepoPropertyLanguages     = "properties.languages"
```
JAORMX, line 30: I'm unsure about `properties.languages`. Yes, we need languages, I just wanna figure out what's a good way to express this property.

jhrozek: I removed them for now. I think we need to figure out a better way to express the nested properties anyway, this was too much of a hack

internal/entities/properties/constants.go
jhrozek: done, there's a separate patch now with just constants for generic entity that tracks with your PR #4178 then a bunch of provider-agnostic repo properties and then the rest of the properties have been moved under the provider

internal/entities/properties/properties.go
jhrozek: I ended up using a syncmap because this would hit us sooner or later. Good suggestion!

internal/entities/properties/properties.go
jhrozek: right!

internal/providers/github/properties.go
jhrozek: this is now tracked in issue #4166

internal/providers/github/properties.go
```diff
@@ -0,0 +1,167 @@
...
+
+	repo, err := ghCli.GetRepository(ctx, slice[0], slice[1])
+	if err != nil {
+		return nil, err
+	}
+
+	repoProps := map[string]any{
+		// general entity
+		properties.PropertyName:       fmt.Sprintf("%s/%s", repo.GetOwner().GetLogin(), repo.GetName()),
+		properties.PropertyUpstreamID: repo.GetID(),
```
JAORMX, line 105: I'm not sure about this, but I'd like to get your opinion: Should we serialize the upstream ID as a string? This would probably change from provider to provider. Some might use serial integers, while others might use a UUID or another sort of random identifier. What do you think? Maybe this is irrelevant for now, but I wanted to get your thoughts on this.

jhrozek: Ah this is actually a great idea, the generic ID should be a string and we should have a github-specific property as an int.

JAORMX (approved):
LGTM

PR #4178, opened 2024-08-16
Add general entity protobuf

JAORMX:
# Summary

This adds two new messages to the Minder public protobuf implementation.

The EntityInstance, which gives general information about an entity.

The upstream reference, which gives enough information for a provider to identify an entity in the upstream system. This will be especially useful when registering an entity in Minder.

## Change Type

# Testing

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -2671,3 +2671,22 @@ message EvaluationHistoryAlert {
...
+    // context is the context in which the entity is evaluated.
+    ContextV2 context = 2;
+
+    // name is the name of the entity.
+    string name = 3;
+
+    // type is the type of the entity.
+    // DISCUSSION: If we're aiming for a BYO entity type, we should probably
+    // have this be a string, and have the user provide the type.
+    Entity type = 4;
```
jhrozek, line 2688: I think this is a good goal. Since this is a public API, do you think it makes sense to have the attribute typed to a string from the start so that we don't have to deprecate or change the type later? OTOH until the type is used in a public API we might as well be a little looser with the compatibility.

proto/minder/v1/minder.proto
```diff
@@ -2671,3 +2671,22 @@ message EvaluationHistoryAlert {
...
+
+    // properties is a map of properties of the entity.
+    google.protobuf.Struct properties = 5;
```
jhrozek, line 2691: I'm going to test this out because using a `google.protobuf.Struct` for the properties which is a `map<string, Value>` and using a `Value` for individual properties would simplify even the provider property code a bit, but the reason I didn't use the `Value` type in the github provider interface was that the `Value` is defined as:
```go
message Value {
  // The kind of value.
  oneof kind {
    // Represents a null value.
    NullValue null_value = 1;
    // Represents a double value.
    double number_value = 2;
    // Represents a string value.
    string string_value = 3;
    // Represents a boolean value.
    bool bool_value = 4;
    // Represents a structured value.
    Struct struct_value = 5;
    // Represents a repeated `Value`.
    ListValue list_value = 6;
  }
}
```
so the only numerical type is `double` which translates to `int64` in Go, but most of the numerical attributes in the Github Go API are `int64`.

jhrozek: so trying with a stupid Go test:

```go
		const key1Value = int64(math.MaxInt64 - 1)
		props := map[string]any{
			"key1": key1Value,
		}
		ei := minderv1.EntityInstance{}

		var err error
		ei.Properties, err = structpb.NewStruct(props)
		require.NoError(t, err)
		key1 := ei.Properties.Fields["key1"]

		require.NotEqual(t, key1.GetNumberValue(), key1Value)
		i64val, err := strconv.ParseInt(key1.GetStringValue(), 10, 64)
		require.NoError(t, err)
		require.Equal(t, key1Value, i64val)
```

I expected that `GetNumberValue` would return a wrong value then also because Value is a `oneof` internally, we can't just grab the value as string.

@evankanderson do you maybe have any tips? The concern is that GH stores many attributes as `int64` and we might risk overflowing the `float64` range. We could of course store the large values as strings and retrieve as string and convert internally, but..eek.

jhrozek: This is no longer an issue we coded up our `Property` type to convert `int64` to a stringified form. If we need to have a simple way to get a `Value` out of a `Property` or `Struct` out of `Properties` we could just add a getter that does so.

proto/minder/v1/minder.proto
```diff
@@ -2683,3 +2683,41 @@ message EvaluationHistoryAlert {
     // the structure and contents are alert specific, and are subject to change.
     string details = 2;
 }
+
+message EntityInstance {
+    // id is the unique identifier of the entity.
+    string id = 1;
+
+    // context is the context in which the entity is evaluated.
+    ContextV2 context = 2;
```
jhrozek, line 2692: Question: how is our `ContextV2` support these days? I see V2 being used in one RPC, do our middlewares support V2 already?

JAORMX: They do https://github.com/stacklok/minder/blob/main/internal/controlplane/common.go

proto/minder/v1/minder.proto
```diff
@@ -2683,3 +2683,41 @@ message EvaluationHistoryAlert {
...
+    // have this be a string, and have the user provide the type.
+    Entity type = 4;
+
+    // properties is a map of properties of the entity.
+    google.protobuf.Struct properties = 5;
+}
+
+// UpstreamEntityRef providers enough information for the
+// provider to identify the entity in the upstream system.
+message UpstreamEntityRef {
```
jhrozek, line 2708: Do you think we should add a provider ID? I wonder if a common case for both `EntityInstance` and `UpstreamEntityRef` to instantiate its provider and then manipulate the data.

We could also add an internal struct that wraps these along with a provider ID and a project ID and maybe that's cleaner to not have those in the public API.

JAORMX: @jhrozek we include provider information in `ContextV2`, I'm unsure if we need a provider ID in this case.

proto/minder/v1/minder.proto
JAORMX: sounds good to me. So, should we keep it as is for now?

proto/minder/v1/minder.proto
jhrozek: sounds good!

proto/minder/v1/minder.proto
jhrozek: yes, I think for now it's safer to have a typed attribute and the migration seems easy.

jhrozek (approved):

proto/minder/v1/minder.proto
evankanderson: > @evankanderson do you maybe have any tips? The concern is that GH stores many attributes as int64 and we might risk overflowing the float64 range. We could of course store the large values as strings and retrieve as string and convert internally, but..eek.

Sorry, missed this earlier until the followup message.  `float64` has a _larger_ range than `int64`, but loses precision above about 2^53.  This follows from a reading of the JSON spec, where javascript does not have an `int` type, but many people have ignored that because they really need an `int` type.

It looks like when decoding JSON to protobuf, both string and integer types are supported: https://protobuf.dev/programming-guides/proto3/#json

PR #4181, opened 2024-08-16
Move properties fetcher methods into Provider

JAORMX:
# Summary

This effectively enforces that providers **must** implement properties fetching as part of their implementation.

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/providers/dockerhub/dockerhub.go
+++ b/internal/providers/dockerhub/dockerhub.go
@@ -27,6 +27,7 @@ import (
 	"golang.org/x/oauth2"
 
 	"github.com/stacklok/minder/internal/db"
+	"github.com/stacklok/minder/internal/entities/properties"
 	"github.com/stacklok/minder/internal/providers/oci"
 	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 	provifv1 "github.com/stacklok/minder/pkg/providers/v1"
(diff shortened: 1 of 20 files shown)


internal/providers/dockerhub/dockerhub.go
```diff
@@ -174,3 +175,15 @@ func (d *dockerHubImageLister) ListImages(ctx context.Context) ([]string, error)
 
 	return containers, nil
 }
+
+// FetchAllProperties implements the provider interface
+// TODO: Implement this
+func (_ *dockerHubImageLister) FetchAllProperties(_ context.Context, _ string, _ minderv1.Entity) (*properties.Properties, error) {
+	return nil, nil
```
jhrozek, line 182: todo for me: Handle `nil` receiver on `Properties` (`Property` already does)

internal/providers/http/http.go
```diff
@@ -36,9 +36,6 @@ type REST struct {
 	credential provifv1.RestCredential
 }
 
-// Ensure that REST implements the REST interface
-var _ provifv1.REST = (*REST)(nil)
```
jhrozek, line 40: why do we need to remove these?

pkg/providers/v1/providers.go
```diff
@@ -45,6 +45,11 @@ type Provider interface {
 	// CanImplement returns true/false depending on whether the Provider
 	// can implement the specified trait
 	CanImplement(trait minderv1.ProviderType) bool
+
+	// FetchAllProperties fetches all properties for the given entity
+	FetchAllProperties(ctx context.Context, name string, entType minderv1.Entity) (*properties.Properties, error)
+	// FetchProperty fetches a single property for the given entity
+	FetchProperty(ctx context.Context, name string, entType minderv1.Entity, key string) (*properties.Property, error)
```
jhrozek, line 52: I agree with moving these

internal/providers/http/http.go
JAORMX: so, REST and Git are not actually providers, but helper code that providers use to implement the relevant traits. I was thinking it would be good to start migrating away from these "trying" to implement the provider interface. This gets a little tricky, as we do treat them as providers in some test code, so I might just create a provider wrapper that uses these.

jhrozek (approved):

Issue #4167, opened 2024-08-16
Add reference to central entities table to EEA


JAORMX:
Now that we're replicating data into the central entities table, we should also start migrating the EEA to take that into use. This should simplify that logic significantly.

For now, this merely duplicates the ID and does not change the logic.

PR #4191, opened 2024-08-19
Add reference to central entities table in EEA

JAORMX:
# Summary

This adds a nullable reference to the central entities table in the EEA
tables. This will simplify the EEA as we won't need to check multiple
tables anymore. However, we can't just migrate to it in one go; we need
to first duplicate the data then do it.

Closes: https://github.com/stacklok/minder/issues/4167

## Change Type

# Testing

# Review Checklist:


internal/eea/eea.go
```diff
@@ -156,11 +168,12 @@ func (e *EEA) aggregate(msg *message.Message) (*message.Message, error) {
 		logger.Info().Msg("executor not ready to process event. Queuing in flush cache.")
 
 		_, err := e.querier.EnqueueFlush(ctx, db.EnqueueFlushParams{
-			Entity:        entities.EntityTypeToDB(inf.Type),
-			RepositoryID:  repoID,
-			ArtifactID:    artifactID,
-			PullRequestID: pullRequestID,
-			ProjectID:     projectID,
+			Entity:           entities.EntityTypeToDB(inf.Type),
```
dmjb, line 171: Do you plan to get rid of the type field in a future PR? (I know removing this requires a few steps to avoid breaking backwards compatibility)

internal/engine/entities/entity_event.go
```diff
@@ -347,6 +347,33 @@ func (eiw *EntityInfoWrapper) GetEntityDBIDs() (repoID uuid.NullUUID, artifactID
...
+
+	id, ok := eiw.getIDForEntityType(eiw.Type)
+	if ok {
+		return id, nil
+	}
+
+	return uuid.Nil, fmt.Errorf("no entity ID found")
+}
+
+func (eiw *EntityInfoWrapper) getIDForEntityType(t minderv1.Entity) (uuid.UUID, bool) {
```
jhrozek, line 364: I guess you could make the methods simpler by making `getIDForEntityType` accept a nil receiver and return uuid.Nil in that case as well as not return the second bool from `getIDForEntityType` but just uuid.Nil. But it's a matter of taste, both are fine.

jhrozek (approved):

internal/eea/eea.go
JAORMX: Right, type won't be needed in the future since the type won't matter. We'd be able to solely rely on the entity instance ID.

This PR merely duplicates info and is not making any logic moves yet.

PR #4248, opened 2024-08-22
Property service for saving properties to database

jhrozek:
# Summary

Adds a service that either returns cached properties for a very short amount of time or contacts the provider to fetch updated service.

The service also marshalls the properties to the database properly.

Fixes: #4171

## Change Type

# Testing

mostly unit tests so far

# Review Checklist:


internal/entities/properties/properties.go
```diff
@@ -20,17 +20,22 @@ import (
...
 
 	"github.com/puzpuzpuz/xsync/v3"
+	"google.golang.org/protobuf/types/known/structpb"
 )
 
 // Property is a struct that holds a value. It's just a wrapper around an interface{}
 // with typed getters and handling of an empty receiver
 type Property struct {
-	value any
+	value *structpb.Value
```
jhrozek, line 30: There are several reasons why I chose `structpb.Value` over just `any`:
 - `NewValue` contains pretty nice code that can e.g. serialize a `[]byte` slice base64 encoded
 - Returning the value as `any` actually does checks, for example for over/underflow for numerical values
 - there is another draft PR that proposes to use `structpb.Struct` in the general entity protobuf message which would make the value easily usable should we need to

internal/entities/properties/properties.go
```diff
@@ -141,3 +142,12 @@ func (p *Properties) GetProperty(key string) *Property {
 	}
 	return &prop
 }
+
+// Iterate implements the seq2 iterator so that the caller can call for key, prop := range Iterate()
+func (p *Properties) Iterate() iter.Seq2[string, *Property] {
```
jhrozek, line 147: This is 50% a convenience method and 50% me wanting to play with iterators :-P

internal/db/entity_helpers.go
```diff
@@ -0,0 +1,154 @@
...
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+const propValueV1 = "1"
```
JAORMX, line 27: We've been using `v*` in other places, do we want to do that here too?

internal/entities/properties/properties.go
```diff
@@ -47,56 +52,60 @@ func (p *Property) AsBool() (bool, error) {
...
 func (p *Property) AsInt64() (int64, error) {
 	if p == nil {
 		return 0, fmt.Errorf("property is nil")
 	}
-	return propertyValueAs[int64](p.value)
+	fval, err := propertyValueAs[float64](p.value.AsInterface())
+	if err != nil {
+		return 0, fmt.Errorf("value is not of type int64: %w", err)
+	}
+	return int64(fval), nil
```
jhrozek, line 92: This need more work because internally `structpb.Value` stores any numerical value as `float64`. In order to handle `int64` and `uint64` I will, when creating a property or properties, check if the leaf value is either of those and change the value to be an object containing our own type annotation and the stringified value. The reverse on getting the value.

internal/entities/properties/properties.go
JAORMX: I dig it!

internal/entities/properties/service/service.go
```diff
@@ -0,0 +1,267 @@
...
+const (
+	// propertiesCacheTimeout is the default timeout for the cache of properties
+	propertiesCacheTimeout = 60
+	// bypassCacheTimeout is a special value to bypass the cache timeout
+	// it is not exported from the package and should only be used for testing
+	bypassCacheTimeout = -1
+)
+
+// PropertiesService is the interface for the properties service
+type PropertiesService interface {
```
jhrozek, line 42: At first I wanted to have just the `Retrieve` methods that would save the entity as well when saving the properties. But - I don't know if we can generalize saving the entity, especially while we are still in migrating mode and are saving the entity with a defined UUID that we take from the older tables. So at least for now I split the methods into separate retrieve and save where the idea is that the caller would retrieve properties, save the entity and then save the properties.

internal/entities/properties/service/service.go
```diff
@@ -0,0 +1,267 @@
...
+	// RetrieveAllProperties fetches all properties for the given entity
+	RetrieveAllProperties(context.Context, v1.Provider, uuid.UUID, string, minderv1.Entity) (*properties.Properties, error)
+	// RetrieveProperty fetches a single property for the given entity
+	RetrieveProperty(context.Context, v1.Provider, uuid.UUID, string, minderv1.Entity, string) (*properties.Property, error)
+	// SaveAllProperties saves all properties for the given entity
+	SaveAllProperties(context.Context, uuid.UUID, *properties.Properties) error
+	// SaveProperty saves a single property for the given entity
+	SaveProperty(context.Context, uuid.UUID, string, *properties.Property) error
```
JAORMX, line 50: can we name the parameters? These function signatures get pretty hard to read without names

internal/entities/properties/service/service.go
```diff
@@ -0,0 +1,267 @@
...
+	propertiesCacheTimeout = 60
+	// bypassCacheTimeout is a special value to bypass the cache timeout
+	// it is not exported from the package and should only be used for testing
+	bypassCacheTimeout = -1
+)
+
+// PropertiesService is the interface for the properties service
+type PropertiesService interface {
+	// RetrieveAllProperties fetches all properties for the given entity
+	RetrieveAllProperties(context.Context, v1.Provider, uuid.UUID, string, minderv1.Entity) (*properties.Properties, error)
```
jhrozek, line 44: should we retrive the entity by name or upstream ID?

internal/entities/properties/service/service.go
```diff
@@ -0,0 +1,267 @@
...
+
+	"github.com/stacklok/minder/internal/db"
+	"github.com/stacklok/minder/internal/entities/properties"
+	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+	v1 "github.com/stacklok/minder/pkg/providers/v1"
+)
+
+const (
+	// propertiesCacheTimeout is the default timeout for the cache of properties
+	propertiesCacheTimeout = 60
```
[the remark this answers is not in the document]
JAORMX: should we use golang durations so we keep the context of it being seconds... or something else?

internal/entities/properties/service/service.go
JAORMX: should the `Retrieve*` methods _also_ receive properties? it could be a subset of properties that help the provider identify the entity. e.g. only the upstream ID.

internal/db/entity_helpers.go
jhrozek: good idea

internal/entities/properties/service/service.go
jhrozek: Interesting! What would the Retrieve method do with the properties? Iterate over them trying to find the identifier to fetch the entity by (we could maybe have cases for name and upstream ID) - but then I guess the method would only save the properties that would be retrieved from upstream and not the ones passed as parameters?

internal/entities/properties/service/service.go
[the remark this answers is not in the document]
jhrozek: right, the code uses a duration actually but you are right it's not visible from the const

internal/entities/properties/service/service.go
JAORMX: uhm... wait a minute.... If this is retrieving all the properties for an entity, wouldn't it suffice to use the actual entity ID? I got a bit confused with the case of the provider fetching properties for an entity Minder doesn't know about yet.

jhrozek: Maybe we're talking about different IDs?

The entity might not be cached yet which would generate the ID that is used in the database as the entity's primary key. But the caller must know what entity do they want to refresh - by "an attribute". I read your suggestion as to provide Properties that the code would then iterate over and pick the property that it would fetch the entity from DB by.

Now the code assumes you fetch an entity by name. What I read you suggesting was to pass properties:

```go
fetchBy, _ := properties.NewProperties(map[string]any{
  properties.PropertyName: "stacklok/minder",
}
```

or:
```go
fetchBy, _ := properties.NewProperties(map[string]any{
  properties.PropertyUpstreamID: "this-can-even-be-a-uuid",
}
```

and then the Retrieve method would do something like:
```go
for name, value := range fetchBy.Iterate() {
 if name == properties.PropertyName {
   db.FetchEntityByName()
 } else if name == properties.PropertyUpstreamID. {
   db.FetchEntityByUpstreamID()
 }
}
```

or did you mean something else?

jhrozek:
I fixed a couple of tests, but this is as far as I go on a Friday afternoon. Pieces of the PR have been submitted separately and apart from the tests, I think this is ready to go.

JAORMX (approved):

PR #4262, opened 2024-08-23
Project deletion metadata audit log

teodor-yanev:
# Summary

In order for us to be able to calculate customer churn, we need to keep track of the deleted projects/accounts on Minder and the relevant metadata from a business perspective.

Fixes https://app.zenhub.com/workspaces/minder-sprints-666857b34fcef725d2efba7c/issues/gh/stacklok/infra/1906

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/database/query/entitlements.sql
+++ b/database/query/entitlements.sql
@@ -5,3 +5,8 @@
 SELECT f.settings FROM entitlements e
 INNER JOIN features f ON f.name = e.feature
 WHERE e.project_id = sqlc.arg(project_id)::UUID AND e.feature = sqlc.arg(feature)::TEXT;
+
+-- name: GetEntitlementsByProjectID :many
+SELECT *
+FROM entitlements
+WHERE project_id = sqlc.arg(project_id)::UUID;
--- a/database/query/entitlements.sql
+++ b/database/query/entitlements.sql
@@ -6,7 +6,7 @@ SELECT f.settings FROM entitlements e
 INNER JOIN features f ON f.name = e.feature
 WHERE e.project_id = sqlc.arg(project_id)::UUID AND e.feature = sqlc.arg(feature)::TEXT;
 
--- name: GetEntitlementsByProjectID :many
-SELECT *
+-- name: GetEntitlementFeaturesByProjectID :many
+SELECT feature
 FROM entitlements
 WHERE project_id = sqlc.arg(project_id)::UUID;
--- a/internal/controlplane/handlers_projects.go
+++ b/internal/controlplane/handlers_projects.go
@@ -303,7 +303,7 @@ func (s *Server) DeleteProject(
 	if err := s.store.Commit(tx); err != nil {
 		return nil, status.Errorf(codes.Internal, "error committing transaction: %v", err)
 	}
-	
+
 	// Create the audit log after a successful deletion
 	var entitlementsFeatures []string
 	for _, e := range entitlements {
--- a/internal/controlplane/handlers_projects.go
+++ b/internal/controlplane/handlers_projects.go
@@ -22,7 +22,6 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/rs/zerolog/log"
-	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
@@ -30,7 +29,6 @@ import (
 	"github.com/stacklok/minder/internal/auth/jwt"
 	"github.com/stacklok/minder/internal/db"
 	"github.com/stacklok/minder/internal/engine/engcontext"
-	"github.com/stacklok/minder/internal/logger"
 	"github.com/stacklok/minder/internal/projects"
 	"github.com/stacklok/minder/internal/projects/features"
 	"github.com/stacklok/minder/internal/util"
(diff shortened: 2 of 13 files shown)


internal/logger/telemetry_store.go
```diff
@@ -64,6 +64,19 @@ type RuleEvalData struct {
 	// TODO: do we want to store params?
 }
 
+// ProjectMetadata can be used to store project metadata in the context.
+type ProjectMetadata struct {
```
evankanderson, line 68: Can we call this "ProjectTombstone" or "ProjectDeletion", so it's clear that this is specifically for the case when a project is deleted?

evankanderson (commented):
I think there are other paths which can delete projects, in particular at least https://github.com/stacklok/minder/blob/main/internal/projects/deleter.go#L70 and https://github.com/stacklok/minder/blob/main/internal/projects/deleter.go#L117.

There's a path via https://github.com/stacklok/minder/blob/main/internal/controlplane/identity_events.go#L81 where users who delete their Stacklok IDP accounts (e.g. in Keycloak) will have their resources automatically cleaned up.  That seems to call [the non-struct version of DeleteUser](https://github.com/stacklok/minder/blob/main/internal/controlplane/identity_events.go#L111), which is also called by the [DeleteUser grpc endpoint](https://github.com/stacklok/minder/blob/main/internal/controlplane/handlers_user.go#L163).

evankanderson (changes requested):
(Realized the previous comment should have been as "request changes")

internal/logger/telemetry_store.go
teodor-yanev: Thanks for catching that and my bad for the obvious miss!

I've added the log audit at the core function which seems to be called from all other paths.

internal/controlplane/handlers_projects.go
```diff
@@ -267,35 +265,6 @@ func (s *Server) DeleteProject(
...
-	g.Go(func() error {
-		var err error
-		reposCount, err = qtx.CountRepositoriesByProjectID(ctx, projectID)
-		return err
-	})
-	g.Go(func() error {
-		var err error
-		entitlements, err = qtx.GetEntitlementsByProjectID(ctx, projectID)
-		return err
-	})
```
JAORMX: If these calls aren't used anymore, can you clean these up from the database queries?

teodor-yanev: They are, in `internal/projects/deleter.go`

internal/projects/deleter.go
```diff
@@ -170,3 +179,44 @@ func hasOtherRoleAssignments(as []*v1.RoleAssignment, subject string) bool {
...
+		var err error
+		entitlements, err = qtx.GetEntitlementsByProjectID(ctx, projectID)
+		return err
+	})
+
+	if err := g.Wait(); err != nil {
+		return nil, fmt.Errorf("error getting project metadata: %w", err)
+	}
+
+	var entitlementsFeatures []string
```
evankanderson, line 211: Use `make` with a size argument to avoid needing extra resizes:

```suggestion
	entitlementsFeatures := make([]string, 0, len(entitlements))
```

internal/projects/deleter_test.go
```diff
@@ -51,6 +51,12 @@ func TestDeleteProjectOneProjectWithNoParents(t *testing.T) {
 		}, nil)
 	mockStore.EXPECT().ListProvidersByProjectID(gomock.Any(), []uuid.UUID{proj}).
 		Return([]db.Provider{}, nil)
+	mockStore.EXPECT().CountProfilesByProjectID(gomock.Any(), proj).
+		Return(int64(0), nil)
+	mockStore.EXPECT().CountRepositoriesByProjectID(gomock.Any(), proj).
+		Return(int64(0), nil)
+	mockStore.EXPECT().GetEntitlementsByProjectID(gomock.Any(), proj).
+		Return([]db.Entitlement{}, nil)
```
evankanderson, line 59: Can you add at least one test where you verify that these are getting added to the telemetry context?

evankanderson (commented):
If we care about this functionality, we should add a test for it.

database/query/entitlements.sql
```diff
@@ -4,4 +4,9 @@
...
 SELECT f.settings FROM entitlements e
 INNER JOIN features f ON f.name = e.feature
-WHERE e.project_id = sqlc.arg(project_id)::UUID AND e.feature = sqlc.arg(feature)::TEXT;
\ No newline at end of file
+WHERE e.project_id = sqlc.arg(project_id)::UUID AND e.feature = sqlc.arg(feature)::TEXT;
+
+-- name: GetEntitlementsByProjectID :many
+SELECT *
+FROM entitlements
+WHERE project_id = sqlc.arg(project_id)::UUID;
```
evankanderson, line 12: I'm guessing if this was `SELECT feature FROM entitlements` you still wouldn't get a `[]string` as the return type...

evankanderson (dismissed):
A couple very minor simplifications, but I think this can merge as-is.

database/query/entitlements.sql
teodor-yanev: It actually does. I did it this way to be more generic and reusable in the future but if you think it's better for this one to only return a slice of the feature names then I'm ok with it.

teodor-yanev: Done, thanks

internal/logger/telemetry_store.go
```diff
@@ -64,6 +64,35 @@ type RuleEvalData struct {
...
+		pt.RepositoriesCount == other.RepositoriesCount &&
+		len(pt.Entitlements) == len(other.Entitlements) &&
+		func() bool {
+			for i, v := range pt.Entitlements {
+				if v != other.Entitlements[i] {
+					return false
+				}
+			}
+			return true
+		}()
```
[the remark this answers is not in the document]
teodor-yanev: True -> updated

internal/logger/telemetry_store.go
```diff
@@ -64,6 +64,36 @@ type RuleEvalData struct {
...
+func (pt ProjectTombstone) Equals(other ProjectTombstone) bool {
+	if len(pt.Entitlements) != len(other.Entitlements) {
+		return false
+	}
+
+	for i, v := range pt.Entitlements {
+		if v != other.Entitlements[i] {
+			return false
+		}
+	}
```
evankanderson, line 90: Ooops, I just realized that there is [`slices.Equal`](https://pkg.go.dev/golang.org/x/exp/slices#Equal) that should do all this.

evankanderson (dismissed):

evankanderson (dismissed):

evankanderson (approved):

PR #4279, opened 2024-08-26
WIP: Remove per-entity tables

### LINKED ISSUE #4315: Use entity_instances table for entity information
With all the necessary information within the `entity_instances` table, we can now dispose of the per-entity tables and rely solely on the central one.

JAORMX:
# Summary

Fixes #4315

## Change Type

# Testing

# Review Checklist:


jhrozek:
About the failures in artifact.go - I wonder if the path of least resistance (even with a bit of manual coding that might be thrown away later) would be to substitute the database lookups like GetArtifact with wrappers around GetEntity that would for now fulfill the same contract, in case of GetArtifact still return a pb.Artifact.

database/migrations/000101_remove_entities_tables.up.sql
```diff
@@ -20,4 +20,72 @@ DROP TABLE IF EXISTS repositories;
 DROP TABLE IF EXISTS artifacts;
 DROP TABLE IF EXISTS pull_requests;
 
+CREATE VIEW repositories AS
```
jhrozek, line 23: I'm not as familiar with views and especially replacing tables with views, so maybe this is a dumb question - but how do indexes work?

Also comparing the columns, would we miss something by not having the reminder_last_sent and updated_at columns?

JAORMX: Well, we don´t really care as much about the `updated_at` since we care now about properties. I do wonder how `reminder_last_sent` would work... would that need to be then part of the central table? maybe it needs to be.

database/query/eval_history.sql
```diff
@@ -144,22 +130,10 @@ WHERE s.id = sqlc.arg(evaluation_id) AND j.id = sqlc.arg(project_id);
...
-       CAST(
-         CASE
-         WHEN ere.repository_id IS NOT NULL THEN r.id
-         WHEN ere.pull_request_id IS NOT NULL THEN pr.id
-         WHEN ere.artifact_id IS NOT NULL THEN a.id
-         END AS uuid
-       ) AS entity_id,
-        ere.entity_instance_id as new_entity_id,
+       ei.entity_type,
+        ere.entity_instance_id as entity_id,
```
blkt, line 134: **nit:** I'm sorry, I could not unsee it 😂 

```suggestion
       ei.entity_type,
       ere.entity_instance_id as entity_id,
```

database/query/eval_history.sql
```diff
@@ -144,22 +130,10 @@ WHERE s.id = sqlc.arg(evaluation_id) AND j.id = sqlc.arg(project_id);
...
        -- raw fields for entity names
-       r.repo_owner,
-       r.repo_name,
-       pr.pr_number,
-       a.artifact_name,
-       j.id as project_id,
+       ei.name as entity_name,
```
blkt, line 136: **issue:** this need to be used in the `WHERE` condition as well.

database/query/eval_history.sql
```diff
@@ -219,25 +191,11 @@ SELECT s.id::uuid AS evaluation_id,
...
-         WHEN ere.pull_request_id IS NOT NULL THEN ere.pull_request_id
-         WHEN ere.artifact_id IS NOT NULL THEN ere.artifact_id
-         END AS uuid
-       ) AS entity_id,
-       ere.entity_instance_id as new_entity_id
+       ei.entity_type,
+       ere.entity_instance_id as entity_id
   FROM evaluation_statuses s
        JOIN evaluation_rule_entities ere ON s.rule_entity_id = ere.id
+       LEFT JOIN entity_instances AS ei ON ei.id = ere.entity_instance_id
```
blkt, line 198: **nit:** IIRC there is an `ON DELETE CASCADE` clause in the foreign key of from `evaluation_rule_entities`, so this should be a plain `JOIN`.

Is there a particular reason why this is a `LEFT JOIN` instead? Was data migrated to the new table?

database/query/profile_status.sql
```diff
@@ -66,38 +66,29 @@ SELECT
...
 WHERE les.profile_id = $1 AND
     (
         CASE
-            WHEN sqlc.narg(entity_type)::entities = 'repository' AND ere.repository_id = sqlc.narg(entity_id)::UUID THEN true
-            WHEN sqlc.narg(entity_type)::entities  = 'artifact' AND ere.artifact_id = sqlc.narg(entity_id)::UUID THEN true
-            WHEN sqlc.narg(entity_type)::entities  = 'pull_request' AND ere.pull_request_id = sqlc.narg(entity_id)::UUID THEN true
+            WHEN ei.entity_id = sqlc.narg(entity_id)::UUID THEN true
             WHEN sqlc.narg(entity_id)::UUID IS NULL THEN true
             ELSE false
             END
```
blkt, line 94: **nit:** very, very minor thing, but I would take the chance to rewrite this into the following
```sql
WHERE les.profile_id = $1
  AND (sqlc.narg(entity_id)::UUID IS NULL OR sqlc.narg(entity_id)::UUID = ei.entity_id)
```

blkt (commented):
Awesome work!

PR #4299, opened 2024-08-28
Take the property service into use in webhook and repository service

jhrozek:
# Summary

The PR moves us in the direction of using the generic entity tables for
repositories. On receiving a webhook, the repository entity along with the
properties are refreshed if needed (with caching provided by the property
service) and returns the entity and its properties. If this is a good
approach, we can also use it for repo reconciliation and profile
reconciliation. When sending the event, still the Repository protobuf message
is sent, but we could as well modify the entityInfoWrapper code to just accept
a message with the entity and properties in a more generic way.

The refresh is handled by a new method of the github repository service - we
could even use a more generic way and not code up a per-property, but we use
the repo-specific layer to migrate the "legacy" table entries to properties as
they are refreshed.

Because we store some attributes along with the repos that are not fetched
from upstream (webhook-related), we treat them as operational attributes and
carry them over during updates. What the operational attributes are is handled
by the provider without minder-core knowing or caring what they are.

Fixes: #4179 

## Change Type

# Testing

unit test + manual testing

# Review Checklist:


database/query/entities.sql
```diff
@@ -99,4 +99,13 @@ WHERE entity_id = $1;
 
 -- name: DeleteAllPropertiesForEntity :exec
 DELETE FROM properties
-WHERE entity_id = $1;
\ No newline at end of file
+WHERE entity_id = $1;
+
+-- name: GetTypedEntitiesByProperty :many
```
jhrozek, line 104: This is not useful in high performance code like the webhook updates because there's no way to index the properties well.

I can see two ways here:
 1) Add the upstream ID as a string value to the properties table
 2) Don't even query by upstream ID from minder-core, but always by name. I'm not sure if all providers will give us this option.

internal/entities/properties/properties.go
```diff
@@ -243,3 +243,48 @@ func (p *Properties) Iterate() iter.Seq2[string, *Property] {
...
+		}
+		return true
+	})
+
+	return &Properties{
+		props: propsMap,
+	}
+}
+
+func (p *Properties) Merge(other *Properties) *Properties {
```
jhrozek, line 267: Probably should have named this MergedCopy or merge in place. It would be easy to change one way or the other.

internal/entities/models/models.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"github.com/google/uuid"
+	"github.com/stacklok/minder/internal/engine/entities"
+
+	"github.com/stacklok/minder/internal/db"
+	"github.com/stacklok/minder/internal/entities/properties"
+	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+)
+
+type EntityInstance struct {
```
jhrozek, line 27: This is just a convenience struct to avoid returning and carrying around both db.EntityInstance and properties when handling an entity with properties. We might want to use the newly introduced protobuf message instead.

internal/providers/github/properties/fetcher.go
```diff
@@ -0,0 +1,97 @@
...
+type PropertyFetcher interface {
+	WrapperForProperty(propertyKey string) PropertyWrapper
+	AllPropertyWrappers() []PropertyWrapper
+	OperationalProperties() []string
+	GetName(props *properties.Properties) (string, error)
+}
+
+type PropertyFetcherFactory interface {
+	EntityPropertyFetcher(entType minderv1.Entity) PropertyFetcher
+}
```
jhrozek, line 37: These interfaces are not supposed to be used outside github, but I found it easier to decouple fetching of the properties and the operational attributes into objects that sit in the provider rather than having separate hardcoded lists. It will make handling of more properties easier and makes the provider code oblivious to what properties are operational and how to fetch properties. All that would be needed to add a new entity is to extend the PropertyFetcherFactory for new entities and implement the PropertyFetcher for a new entity.

pkg/providers/v1/providers.go
```diff
@@ -56,6 +56,8 @@ type Provider interface {
 	// The name is used to identify the entity within minder and is how
 	// it will be stored in the database.
 	GetEntityName(entType minderv1.Entity, props *properties.Properties) (string, error)
+	// OperationalProperties returns the operational properties for an entity
+	OperationalProperties(entType minderv1.Entity) []string
```
jhrozek, line 60: Question: Can we come up with a way to expose the operational properties without adding a new method to the generic Provider interface? It feels like the interface is getting too big already.

internal/repositories/github/service.go
```diff
@@ -325,6 +338,145 @@ func (r *repositoryService) DeleteByName(
 	return r.deleteRepository(ctx, client, &repo)
 }
 
+func (r *repositoryService) RefreshRepositoryByUpstreamID(
+	ctx context.Context,
+	upstreamRepoID int64,
```
jhrozek, line 343: We might not need this method at all once we finish migrating, but it provides a lazy migration from the legacy tables.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -889,19 +897,23 @@ func (s *Server) processRepositoryEvent(
...
+	}
+
 	eiw := entities.NewEntityInfoWrapper().
 		WithActionEvent(event.GetAction()).
-		WithProjectID(dbrepo.ProjectID).
-		WithProviderID(dbrepo.ProviderID).
+		WithProjectID(repoEnt.Entity.ProjectID).
+		WithProviderID(repoEnt.Entity.ProviderID).
 		WithRepository(pbRepo).
-		WithRepositoryID(dbrepo.ID)
```
jhrozek, line 904: I wonder if we could/should move away from using the type-specific protobuf messages? On the other hand, they are used by the ingestors and evaluators ...

database/query/entities.sql
JAORMX: We actually can have a partial index (using a GIN index) that would help here. e.g.

```
CREATE INDEX ON properties((value->>'upstream_id'));
```

or something of the sort.

internal/entities/models/models.go
JAORMX: I agree that we should not use the database structures but should instead have either a middle-struct (as you did here) or even the protobuf. I leave that to your discretion.

internal/providers/github/properties/fetcher.go
JAORMX: If it's not meant to be used outside of GitHub, can you add that as a comment and/or prepend the word `GitHub` in front?

internal/controlplane/handlers_githubwebhooks.go
JAORMX: yeah... I would REALLY love to get rid of them. It'll be some work :cry:

pkg/providers/v1/providers.go
JAORMX: Why do we need to expose or filter them?

jhrozek:
@JAORMX I think the only thing missing now is to fix up all the repeated webhook tests..

database/migrations/000102_properties_upstream_id_index.down.sql
```diff
@@ -0,0 +1,19 @@
+-- Copyright 2023 Stacklok, Inc
```
JAORMX, line 1: 2024

database/migrations/000102_properties_upstream_id_index.up.sql
```diff
@@ -0,0 +1,23 @@
...
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+BEGIN;
+
+-- Create index on properties for upstream_id
+-- this is not a GIN index, but in local testing postgres seems to be able to use it if the query
+-- contains these two attributes
```
JAORMX, line 19: Huh, I really thought we needed a GIN index

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -1189,12 +1207,12 @@ func (_ *Server) repositoryAdded(
 func (s *Server) fetchRepo(
 	ctx context.Context,
 	repo *repo,
-) (*db.Repository, error) {
+) (*models.EntityWithProperties, error) {
 	l := zerolog.Ctx(ctx)
 
-	dbrepo, err := s.store.GetRepositoryByRepoID(ctx, repo.GetID())
+	repoEnt, err := s.repos.RefreshRepositoryByUpstreamID(ctx, repo.GetID())
```
JAORMX, line 1213: perhaps not for this PR. But one TODO time we should add is figuring out what to do if we get multiple repositories out of this call and handling that. e.g. say we have several provider classes, they might have an upstream ID that clashes, we need to be able to handle that.

jhrozek: Good call, we need to search by provider, too

database/migrations/000102_properties_upstream_id_index.down.sql
jhrozek: sorry, I missed that. I will fix the year in another PR, I need to add better test coverage for some parts anyway

internal/providers/github/properties/fetcher.go
jhrozek: I added a prefix as to not stutter

pkg/providers/v1/providers.go
jhrozek: We need to filter them because at the same time we need to be able to remove properties that were deleted upstream but we need to retain the operational attributes like hook ID.
Anyway, we discussed that offline and agreed to move the attributes handling to the provider completely so that minder-core is oblivious to their existence.

JAORMX (approved):

PR #4338, opened 2024-09-02
Handle pull requests as general entities

jhrozek:
# Summary

- **Slight generalization of the GH property fetching** - The implementation of FetchProperty and FetchAllProperties for GitHub was not generic enough as it required that the properties contain the `name` key. For some properties, like pull request, the provider must fetch by other keys, so let's just pass the properties to the wrapper and let the wrapper error out if the get-by-properties don't contain the necessary data for fetching the property.
   Also moves the propertyOrigin struct to the generic fetcher.go source to make it clearer that it's to be reused by other entities. Finally the `AllPropertyWrappers` and `WrapperForProperty` methods are now implemented by a common structure that the per-entity structures just embed to reduce code duplication.
- **Add a pull request entity module to GH provider** - This will enable us to handle the pull request entity instances and their properties.
- **Handle pull requests as entities with properties in the github webhook**

Fixes: #4171

## Change Type

# Testing

manually + unit tests

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_githubwebhooks.go
+++ b/internal/controlplane/handlers_githubwebhooks.go
@@ -269,9 +269,17 @@ func (p *pullRequestEvent) GetPullRequest() *pullRequest {
 }
 
 type pullRequest struct {
+	ID     *int64  `json:"id,omitempty"`
 	URL    *string `json:"url,omitempty"`
 	Number *int64  `json:"number,omitempty"`
 	User   *user   `json:"user,omitempty"`
+}
+
+func (p *pullRequest) GetID() int64 {
+	if p.ID != nil {
+		return *p.ID
+	}
+	return 0
 }
 
 func (p *pullRequest) GetURL() string {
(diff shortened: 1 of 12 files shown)


internal/providers/github/properties/fetcher.go
```diff
@@ -27,14 +26,20 @@ import (
...
 
 // GhPropertyWrapper is a function that fetches a property from the GitHub API
-type GhPropertyWrapper func(ctx context.Context, ghCli *go_github.Client, name string) (map[string]any, error)
+type GhPropertyWrapper func(
+	ctx context.Context, ghCli *go_github.Client, lookupProperties *properties.Properties,
+) (map[string]any, error)
 
 // GhPropertyFetcher is an interface for fetching properties from the GitHub API
 type GhPropertyFetcher interface {
+	ghPropertyFetcherBase
```
jhrozek, line 35: I guess we wouldn't have to split the interface and could just make the per-entity structures inherit the base structure to satisfy the GhPropertyFetcher interface..?

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -967,44 +975,45 @@ func (s *Server) processPullRequestEvent(
...
+		ghprop.PullPropertyAuthorID:   event.GetPullRequest().GetUser().GetID(),
+		ghprop.PullPropertyAction:     event.GetAction(),
+	})
 	if err != nil {
-		l.Error().Err(err).Msg("error instantiating provider")
-		return nil, err
+		return nil, fmt.Errorf("error creating pull request properties: %w", err)
 	}
 
-	cli, err := provifv1.As[provifv1.GitHub](provider)
```
jhrozek, line 976: We just treat the provider as a generic provider here, not as github, except creating the properties to look up by and then calling the GH-specific function to format the fetched properties to pb.PullRequest

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -1547,25 +1562,38 @@ func (s *Server) reconcilePrWithDb(
...
+				Name:      pullName,
 				ProjectID: repoEnt.Entity.ProjectID,
 			})
-			if err != nil {
+			if err != nil && !errors.Is(err, sql.ErrNoRows) {
+				return nil, err
+			}
+
+			// temporary migration code to get rid of PRs with the old name
+			err = t.DeleteEntityByName(ctx, db.DeleteEntityByNameParams{
```
jhrozek, line 1590: Note that this PR /changes the name of the PRs/. We don't have to do that in this PR if we don't want to, but I since the old code just used repoName/pullNumber I wanted to address the case of two repos with the same name coming from different providers and yielding the same entity name.

JAORMX (approved):

PR #4342, opened 2024-09-02
Decouple legacy entity tables from results queries

### LINKED ISSUE #4316: Rely on central entities table for results
This will remove the `JOIN`s to the repositories table in favor of using the central entities table.

JAORMX:
# Summary

This is another step into moving away from the central entity tables and
into the new per-entity tables.

Fixes #4316

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_evalstatus.go
+++ b/internal/controlplane/handlers_evalstatus.go
@@ -200,7 +200,7 @@ func (s *Server) ListEvaluationHistory(
 }
 
 func fromEvaluationHistoryRows(
-	rows []history.OneEvalHistoryAndEntity,
+	rows []*history.OneEvalHistoryAndEntity,
 ) ([]*minderv1.EvaluationHistory, error) {
 	res := make([]*minderv1.EvaluationHistory, len(rows))
 
@@ -560,7 +560,7 @@ func filterProfileLists(
 func (s *Server) buildRuleEvaluationStatusFromDBEvaluation(
 	ctx context.Context,
 	profile *db.ListProfilesByProjectIDAndLabelRow, eval db.ListRuleEvaluationsByProfileIdRow,
-	efp entmodels.EntityWithProperties,
+	efp *entmodels.EntityWithProperties,
 ) (*minderv1.RuleEvaluationStatus, error) {
 	guidance := ""
 	// Only return the rule type guidance text when there is a problem
@@ -650,7 +650,7 @@ func (s *Server) buildRuleEvaluationStatusFromDBEvaluation(
 	}, nil
 }
 
-func buildEntityFromEvaluation(efp entmodels.EntityWithProperties) *minderv1.EntityTypedId {
+func buildEntityFromEvaluation(efp *entmodels.EntityWithProperties) *minderv1.EntityTypedId {
 	ent := &minderv1.EntityTypedId{
 		Type: efp.Entity.Type,
 	}
@@ -688,7 +688,8 @@ func buildProfileStatus(
 // buildEvalResultAlertFromLRERow build the evaluation result alert from a
 // database row.
 func buildEvalResultAlertFromLRERow(
-	eval *db.ListRuleEvaluationsByProfileIdRow, ent entmodels.EntityWithProperties) *minderv1.EvalResultAlert {
+	eval *db.ListRuleEvaluationsByProfileIdRow, ent *entmodels.EntityWithProperties,
+) *minderv1.EvalResultAlert {
 	era := &minderv1.EvalResultAlert{
 		Status:      string(eval.AlertStatus),
 		LastUpdated: timestamppb.New(eval.AlertLastUpdated),
--- a/internal/controlplane/handlers_evalstatus.go
+++ b/internal/controlplane/handlers_evalstatus.go
@@ -31,10 +31,14 @@ import (
 
 	"github.com/stacklok/minder/internal/db"
 	"github.com/stacklok/minder/internal/engine/engcontext"
+	entmodels "github.com/stacklok/minder/internal/entities/models"
+	"github.com/stacklok/minder/internal/entities/properties"
 	"github.com/stacklok/minder/internal/history"
+	ghprop "github.com/stacklok/minder/internal/providers/github/properties"
 	"github.com/stacklok/minder/internal/ruletypes"
 	"github.com/stacklok/minder/internal/util"
 	minderv1 "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+	provifv1 "github.com/stacklok/minder/pkg/providers/v1"
 )
 
 const (
@@ -77,25 +81,13 @@ func (s *Server) GetEvaluationHistory(
 		return nil, status.Error(codes.Internal, evalErrMsg)
 	}
 
-	entityName, err := getEntityName(
-		eval.EntityType,
-		eval.RepoOwner,
-		eval.RepoName,
-		eval.PrNumber,
-		eval.ArtifactName,
-	)
-	if err != nil {
-		zerolog.Ctx(ctx).Error().Err(err).Msg(evalErrMsg)
-		return nil, status.Error(codes.Internal, evalErrMsg)
-	}
-
 	pbEval := &minderv1.EvaluationHistory{
 		Id:          eval.EvaluationID.String(),
 		EvaluatedAt: timestamppb.New(eval.EvaluatedAt),
 		Entity: &minderv1.EvaluationHistoryEntity{
 			Id:   eval.EntityID.String(),
 			Type: dbEntityToEntity(eval.EntityType),
-			Name: entityName,
+			Name: eval.EntityName,
 		},
 		Rule: &minderv1.EvaluationHistoryRule{
 			Name:     eval.RuleName,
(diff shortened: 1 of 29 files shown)


internal/controlplane/handlers_evalstatus.go
```diff
@@ -720,53 +699,3 @@ func dbSeverityToSeverity(dbSev db.Severity) (*minderv1.Severity, error) {
...
-		return fmt.Sprintf("%s/%s#%d",
-			repoOwner.String,
-			repoName.String,
-			prNumber.Int64,
-		), nil
-	case db.EntitiesArtifact:
-		if !artifactName.Valid {
-			return "", errors.New("artifact_name is missing")
-		}
-		return artifactName.String, nil
```
jhrozek, line 754: Note that we'll be using a differently formatted name for artifacts, will that be a problem? Otherwise it's fine to use the artifact's entity name here.

internal/entities/properties/service/service.go
```diff
@@ -79,7 +79,7 @@ func WithEntityTimeout(timeout time.Duration) propertiesServiceOption {
 
 // NewPropertiesService creates a new properties service
 func NewPropertiesService(
-	store db.Store,
+	store db.ExtendQuerier,
```
jhrozek, line 82: This is correct, but reading this change made me realize that I should have exposed the ExtendQuerier as a parameter to all the property service interface methods so that we can always pass a transaction

internal/history/service.go
```diff
@@ -223,8 +205,56 @@ func (_ *evaluationHistoryService) ListEvaluationHistory(
...
+		if err != nil {
+			return nil, fmt.Errorf("error getting entity: %w", err)
+		}
+
+		p, err := ehs.providerManager.InstantiateFromID(ctx, ent.ProviderID)
+		if err != nil {
+			return nil, fmt.Errorf("error instantiating provider: %w", err)
+		}
+
+		propClient, err := provifv1.As[provifv1.Provider](p)
```
jhrozek, line 225: I don't think you need to cast to Provider here, what Instantiate returns is already a Provider, right?

jhrozek (commented):
Great work, this is not easy code to modify. About the issues with the code referencing repoName and repoOwner separately, it looks like we're either constructing the repo slug again anyway or just returning them in a generic `map[string]string` through the `entityInfo` variable, so it should be possible to use the name straight away hopefully?

internal/controlplane/handlers_evalstatus.go
JAORMX: I rather start using the new naming scheme already. Nothing is using artifacts just yet, right? We might want to double check that I don't break the smoke tests.

internal/entities/properties/service/service.go
JAORMX: I'm not sure that's needed. We can pass a transaction that's already begun, and handle the rollback and commit outside of the property service.

internal/controlplane/handlers_profile.go
```diff
@@ -273,41 +276,34 @@ func getProfilePBFromDB(
...
-			l.Err(err).Msg("error getting artifact by ID")
-			return entityInfo
-		}
-		entityInfo["artifact_id"] = artifact.ID.String()
-		entityInfo["artifact_name"] = artifact.ArtifactName
-		entityInfo["artifact_type"] = artifact.ArtifactType
-		entityInfo["provider"] = artifact.ProviderName
+	// entityInfo["artifact_id"] = artifact.ID.String()
+	// entityInfo["artifact_name"] = artifact.ArtifactName
+	// entityInfo["artifact_type"] = artifact.ArtifactType
```
jhrozek, line 296: I guess these could now be commented out?

internal/controlplane/handlers_profile.go
```diff
@@ -273,41 +276,34 @@ func getProfilePBFromDB(
...
+
+	entityInfo["provider"] = rs.Provider
+	entityInfo["entity_type"] = efp.Entity.Type.ToString()
+	entityInfo["entity_id"] = rs.EntityID.String()
+
+	// temporary: These will be replaced by entity_id
```
jhrozek, line 302: as well as these could be simplified?

internal/controlplane/handlers_profile.go
```diff
@@ -513,19 +524,17 @@ func (s *Server) getRuleEvalStatus(
...
 			repoPath = fmt.Sprintf("%s/%s", st.EntityInfo["repo_owner"], st.EntityInfo["repo_name"])
 		} else if dbRuleEvalStat.EntityType == db.EntitiesArtifact {
-			repoDetails, err := s.store.GetRepoPathFromArtifactID(ctx, dbRuleEvalStat.EntityID)
-			if err != nil {
-				l.Err(err).Msg("error getting repo details from db")
-				return st
+			artRepoOwner := efp.Properties.GetProperty(ghprop.ArtifactPropertyRepoOwner).GetString()
+			artRepoName := efp.Properties.GetProperty(ghprop.ArtifactPropertyRepoName).GetString()
+			if artRepoOwner != "" && artRepoName != "" {
+				repoPath = fmt.Sprintf("%s/%s", artRepoOwner, artRepoName)
```
jhrozek, line 530: Is this name ever going to be used in anything but presentation? I wonder if we should make the RepoOwner and RepoName properties not tied to github provider (and move them into contants.go like we have properties.RepoPropertyIsPrivate). Then we could use this code for non-github artifacts and maybe even make the code more generic by constructing new properties and calling provider.GetEntityName(REPOSITORY, theNewProperties)

internal/db/eval_history_test.go
```diff
@@ -70,8 +70,6 @@ func TestListEvaluationHistoryFilters(t *testing.T) {
 				require.Equal(t, es1, row.EvaluationID)
 				require.Equal(t, EntitiesRepository, row.EntityType)
 				require.Equal(t, repo1.ID, row.EntityID)
-				require.Equal(t, repo1.RepoOwner, row.RepoOwner.String)
-				require.Equal(t, repo1.RepoName, row.RepoName.String)
```
jhrozek, line 74: why did we have to remove these?

internal/history/service.go
```diff
@@ -48,19 +50,40 @@ type EvaluationHistoryService interface {
...
+}
+
 // NewEvaluationHistoryService creates a new instance of EvaluationHistoryService
-func NewEvaluationHistoryService() EvaluationHistoryService {
-	return &evaluationHistoryService{}
+func NewEvaluationHistoryService(providerManager manager.ProviderManager, opts ...options) EvaluationHistoryService {
+	ehs := &evaluationHistoryService{
+		providerManager: providerManager,
+		propServiceBuilder: func(qtx db.ExtendQuerier) propertiessvc.PropertiesService {
+			return propertiessvc.NewPropertiesService(qtx)
```
jhrozek, line 73: I think it would be OK to just require the PropertiesService interface as a parameter but the options don't hurt..

internal/history/service.go
```diff
@@ -223,8 +214,28 @@ func (_ *evaluationHistoryService) ListEvaluationHistory(
...
+	propsvc := ehs.propServiceBuilder(qtx)
+
+	data := make([]OneEvalHistoryAndEntity, 0, len(rows))
+	for _, row := range rows {
+		efp, err := propsvc.EntityForProperties(ctx, row.EntityID, row.ProjectID, ehs.providerManager, qtx)
+		if err != nil {
+			return nil, fmt.Errorf("error fetching entity for properties: %w", err)
+		}
+
+		err = propsvc.RetrieveAllPropertiesForEntity(ctx, efp)
```
jhrozek, line 226: this is potentially a lot of calls isn't it? Do we actually need up-to-date data here, couldn't we just fetch the entities and properties from cache regardless of whether it's expired or not?

internal/entities/properties/service/service.go
```diff
@@ -156,9 +164,29 @@ func (ps *propertiesService) RetrieveAllProperties(
 	return refreshedProps, nil
 }
 
+// RetrieveAllPropertiesForEntity fetches a single property for the given an entity
+// for properties model. Note that properties will be updated in place.
+func (ps *propertiesService) RetrieveAllPropertiesForEntity(
+	ctx context.Context, efp *models.EntityForProperties,
```
jhrozek, line 170: This is fine, but do you think that in a follow up we could change RetrieveAllProperties to accept EntityForProperties or just an Entity that alrady has projectID, providerID and type and do something like:

newProps, err := ps.RetrieveAllProperties(ctx, prov, entity, lookupProps)
efp := NewEntityFromPropertiesWithInstance(entity, newProps)

internal/entities/models/models.go
```diff
@@ -52,3 +53,46 @@ func NewEntityWithProperties(dbEntity db.EntityInstance, props *properties.Prope
...
+		Properties: props,
+	}
+}
+
+// EntityForProperties gives us the necessary fields to fetch properties
+type EntityForProperties struct {
+	*EntityWithProperties
+
+	// Provider is the provider for the entity
+	Provider provifv1.Provider
```
jhrozek, line 70: Why keep track of the provider and not just use the providerID in the EntityWithProperties? I wonder if the cleanest way would be to just pass the providerID to the propertyService, let the propertyService structure contain a providerManager and always instantiate the provider...

jhrozek (commented):
Thanks for the work!

The code reads easier now that it doesn't handle the entity type separately. I put some comments inline, they are mostly to make sure that the original design of the PropertyService wasn't bad to start with and that we're not adding hacks to address the shortcomings instead.

The biggest question I have is though - why do we call the Refreshes in this service at all and not just rely on the database properties? Wouldn't we cause a lot of upstream traffic this way?

internal/db/eval_history_test.go
JAORMX: Because else we'd need to return the repo owner via a JOIN in the ListEvaluationHistory call. And the intent was to remove that. We should not rely on any entity-specific info.

internal/controlplane/handlers_profile.go
JAORMX: Unfortunately, our output is heavily tied to GitHub, this is actually used to get the alert which won't exist in other providers. I can't wait to deprecate that.

JAORMX: Answering your question, I still think it's a github-specific attribute.

internal/controlplane/handlers_profile.go
JAORMX: I think we'll have a set of properties we output and have a general way of displaying them, so this will be replaced with a provider-specific call IMO.

internal/controlplane/handlers_profile.go
JAORMX: I had missed this, thanks!

internal/history/service.go
JAORMX: It was done for testing only.

internal/history/service.go
JAORMX: We are at a weird point where we don't have everything in cache just yet. The idea is to fetch from cache if it's possible, and persist it if it isn't there. Perhaps I should instead add "options" to this call and rely on the cache even if the data is marked as stale. wdyt?

internal/entities/models/models.go
JAORMX: Good observation. I was just trying to make it easier to instantiate everything. But this could actually be moved to the `RetrieveAllPropertiesForEntity` so we would not need the new wrapper. Let me refactor this.

JAORMX:
> The biggest question I have is though - why do we call the Refreshes in this service at all and not just rely on the database properties? Wouldn't we cause a lot of upstream traffic this way?

I added a comment about that above. I think we can deal with that by adding options to the retrieve calls.

internal/controlplane/handlers_profile.go
```diff
@@ -273,41 +276,38 @@ func getProfilePBFromDB(
...
+
+	entityInfo["provider"] = rs.Provider
+	entityInfo["entity_type"] = efp.Entity.Type.ToString()
+	entityInfo["entity_id"] = rs.EntityID.String()
+
+	// temporary: These will be replaced by entity_id
+	if rs.EntityType == db.EntitiesRepository {
+		entityInfo["repository_id"] = efp.Entity.ID.String()
+	} else if rs.EntityType == db.EntitiesArtifact {
+		entityInfo["artifact_id"] = efp.Entity.ID.String()
```
jhrozek, line 310: I think we could just use the upstream ID as a string here to simplify the logic (OK in a follow-up)

JAORMX: Do we use these for anything?

jhrozek (approved):

PR #4363, opened 2024-09-04
Use rule display name as profile rule name default

### LINKED ISSUE #4251: Use displayName in detailed profile view CLI

eleftherias:
If a profile does not have a custom name set for the rule, then the rule display name should be used as a default.

# Summary

If a profile does not have a custom name set for the rule, then the rule type display name should be used as a default.
If neither of the above exist, then the rule type name should be used.

Fixes #4251

#### Profile doesn't have custom rule name overrides. Rule types all have display names.

Screenshot 2024-09-04 at 15 36 38

#### Profile has some custom rule name overrides. Rule types all have display names.

Screenshot 2024-09-04 at 15 36 22

#### Profile has some custom rule name overrides. Rule types don't have display names.

Screenshot 2024-09-04 at 15 43 01

#### Profile doesn't have custom rule name overrides. Rule types don't have display names.

Screenshot 2024-09-04 at 15 49 35

## Change Type

# Testing

# Review Checklist:


internal/profiles/service.go
```diff
@@ -185,6 +182,9 @@ func (p *profileService) CreateProfile(
 	profile.Remediate = ptr.Ptr(string(newProfile.Remediate.ActionType))
 	profile.Alert = ptr.Ptr(string(newProfile.Alert.ActionType))
 
+	// Adds default rule names, if not present
+	PopulateRuleNames(profile, rulesInProf)
+
```
evankanderson, line 187: Moving this down here means that we're now storing a `NULL` for the rule name if it's not set.  Do we also need to call `PopulateRuleNames` on the Get / List output?  (Alternatively, moving this above line 155 should fix this problem.)

evankanderson: It looks like we didn't move profile listing into the `profiles` service, and we just call some methods like `MergeDatabaseListIntoProfiles` directly from `handlers_profile.go` without any support for mocking.  (And yes, that means the code there isn't tested...)

evankanderson (commented):
Can you check that this produces expected results when you go a get of the profile after creating it?

internal/profiles/service.go
eleftherias: It's intentionally moved to the end of this function to be near the other calls that populate the result.

> Moving this down here means that we're now storing a NULL for the rule name if it's not set

The display name will always be set, because it's defaulted to the rule type name.

> Do we also need to call PopulateRuleNames on the Get / List output?

No, the correct name is already stored in the DB during the create / update and fetched during the get / list.

internal/profiles/service.go
```diff
@@ -564,7 +564,7 @@ func upsertRuleInstances(
 	updatedIDs := make([]uuid.UUID, len(newRules))
 	for i, rule := range newRules {
 		// TODO: Clean up this logic once we no longer have to support the old tables.
-		entityRuleTuple, ok := rulesInProf[RuleTypeAndNamePair{
+		ruleIDAndName, ok := rulesInProf[RuleTypeAndNamePair{
 			RuleType: rule.Type,
 			RuleName: rule.Name,
```
evankanderson, line 569: I'll believe you, but what I'm seeing is that:

1. `CreateProfile` calls `PopulateRuleNames` on `profile` on `service.go:129`, passing a _pointer_ (in/out value).
2. `PopulateRuleNames` iterates over Repository, BuildEnvironment, PullRequest, Artifact slices (which are `[]*pb.Profile_Rule`), calling a function which sets the `r.Name` parameter on the pointed-to object.
3. `createProfileRulesForEntity` builds a map from the Repository, Artifact, etc slices of `[]*pb.Profile_Rule` (all 8 sub-fields, different from PopulateRuleNames), and then calls `createProfileRulesForEntity` on each of the lists (`service.go:168`).  These lists will have had `r.Name` populated in (2).
4. `createProfileRulesForEntity` will call `upsertRuleInstances` at `service.go:375`, passing in the slice from (3) AND marshalling the `[]*pb.Profile_Rule` into JSON (using `encoding/json` and not `protojson`!) and storing them in `CreateProfileForEntity` at `service.go:393`.
5. `upsertRuleInstances` iterates over the `[]*pb.Profile_Rule` slice from (3),  looking up `(r.Type, r.Name)` in `entityRuleTuple` and then calling `qtx.UpsertRuleInstance` in the database with `r.Name` on `service.go:590`.

I'm being a bit stubborn because I'm concern about the database stores in (4) and (6) -- I'd prefer not to do another cleanup like `000017`.

evankanderson (dismissed):

internal/profiles/service.go
eleftherias: Thanks, I see what you mean now.
I initially thought we should not return the defaulted rule name, when getting / listing profiles because the output should conform to the user input, i.e if the user didn't set a rule `name` override in the profile then they shouldn't see a `name` when fetching the profile.
But, I can see that's not how it used to work previously. Instead we did always return the defaulted name, even if it did not come from the profile itself. I'll update the code to continue to behave this way.

internal/profiles/service.go
```diff
@@ -126,7 +126,10 @@ func (p *profileService) CreateProfile(
 	}
 
 	// Adds default rule names, if not present
-	PopulateRuleNames(profile)
+	PopulateRuleNames(profile, rulesInProf)
```
evankanderson, line 129: GIven that `PopulateRuleNames` already uses `profile` as an in-out parameter, would it make sense to do the same for `rulesInProf` (i.e. update the map in `PopulateRuleNames` -- it looks like we're always calling `getUpdatedRuleMapping` and replacing the name afterwards; it feels like we could skip a second for loop by doing so)

evankanderson (approved):

evankanderson (approved):

PR #4377, opened 2024-09-05
Take properties into use executor to fill selectors' properties

jhrozek:
# Summary

Adds a lookup of an entity with properties to the evaluation engine so
that the resulting selectorEntity is augmented with properties and
passed to the selector engine. Changes the conversion interfaces to
selector entities to only use the EntityWithProperties model.

As a result, the user can use properties in selectors such as:
```
- description: This should not apply copyleft repos
	entity: repository
	selector: repository.properties['github/license'] != 'GPL'
```
And in future allow us to extend selectors based on any other properties
we add.

Changes the ruletype test command to not use an EntityInfoWrapper as
intput, but just properties. An example of a repo entity with properties
is:
```
name: jakubtestorg/bad-python
is_private: false
is_fork: false
is_archived: false
upstream_id: "706108924"
github/repo_id:
	minder.internal.type: int64
	minder.internal.value: "706108924"
github/repo_name: bad-python
github/repo_owner: jakubtestorg
github/default_branch: master
foo: bar
```

Fixes: #4173

- **GetEntityByID only accepts an ID since it's unique** - We can simplify the GetEntityByID call to not accept providerID since the call searches on the globally-unique and indexed entity ID.
- **Add GetEntityWithPropertiesByID to our entity model package** - Adds a convenience function to search an entity by ID with all its properties. Also moves two utility functions form the property service here.
- **Add ToProtoStruct to properties** - This utility function will allow us to use properties both in selectors that use the structupb.Struct and in the general entity protobuf message that we added recently.
- **Add an option to NewProperties to skip checking the internal prefix** - This is so unit tests can set int64 properties.

## Change Type

# Testing

mostly manually. I need to really go and add some selectors.

# Review Checklist:


internal/providers/gitlab/selector_entity.go
```diff
@@ -1,44 +0,0 @@
-//
-// Copyright 2024 Stacklok, Inc.
```
jhrozek, line 2: In a follow-up, I'm going to remove the provider-specific interface and just move the repo,PR and artifact conversion to provider-generic code because with the addition of properties, we don't really need anything provider-specific.

JAORMX (approved):

internal/entities/models/models.go
```diff
@@ -61,6 +64,54 @@ func NewEntityWithPropertiesFromInstance(entity EntityInstance, props *propertie
 	}
 }
 
+// GetEntityWithPropertiesByID retrieves an EntityWithProperties instance by its ID.
+func GetEntityWithPropertiesByID(ctx context.Context, q db.Querier, id uuid.UUID) (*EntityWithProperties, error) {
```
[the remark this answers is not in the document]
jhrozek: actually you are right, the properties service already has a Querier interface and we already also have the pattern of passing an ExtendQuerier to run the method in a transaction. Feel free to move the method there (or I can do it)

internal/entities/properties/properties.go
```diff
@@ -202,12 +202,32 @@ type Properties struct {
...
+// newPropertiesOption is a function that configures NewProperties
+type newPropertiesOption func(*newPropertiesConfig)
+
+type newPropertiesConfig struct {
+	skipPrefixCheck bool
+}
+
+// WithSkipPrefixCheckTestOnly returns an option to skip checking the prefix
+// This should only be used for testing purposes
+func WithSkipPrefixCheckTestOnly() newPropertiesOption {
```
JAORMX, line 214: If this is for testing purposes only, let's move it to the test file itself. It should be fine as it's part of the same package.

JAORMX (approved):

Issue #4420, opened 2024-09-10
Remove unused `GetEntitiesByType` SQL statement


blkt:
[This statement](https://github.com/stacklok/minder/blob/44d1523d784f4684439c85f8863013416025204a/database/query/entities.sql#L72-L77) is not used and should be removed from the codebase.

JAORMX:
This might not be needed with https://github.com/stacklok/minder/pull/4429

Issue #4331, opened 2024-08-30
Add entity registration to GitLab provider


JAORMX:
### Please describe the enhancement

-

### Solution Proposal

-

PR #4429, opened 2024-09-10
Use central entity tables to list remote repos

JAORMX:
# Summary

This is part of generalizing entity registration, as other providers
will need to work with this mechanism. As part of the migration, this
moved the logic to use our new entity model.

Related-To: https://github.com/stacklok/minder/issues/4331
Closes: https://github.com/stacklok/minder/issues/4420

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/database/query/entities.sql
+++ b/database/query/entities.sql
@@ -74,7 +74,9 @@ LIMIT 1;
 
 -- name: GetEntitiesByType :many
 SELECT * FROM entity_instances
-WHERE entity_instances.entity_type = $1 AND entity_instances.project_id = ANY(sqlc.arg(projects)::uuid[]);
+WHERE entity_instances.entity_type = $1
+    AND entity_instances.provider_id = sqlc.arg(provider_id)
+    AND entity_instances.project_id = ANY(sqlc.arg(projects)::uuid[]);
 
 -- GetEntitiesByProvider retrieves all entities of a given provider.
 -- this is how one would get all repositories, artifacts, etc. for a given provider.
--- a/internal/providers/manager/manager.go
+++ b/internal/providers/manager/manager.go
@@ -33,6 +33,12 @@ import (
 
 //go:generate go run go.uber.org/mock/mockgen -package mock_$GOPACKAGE -destination=./mock/$GOFILE -source=./$GOFILE
 
+// NameProviderTuple is a tuple of a provider name and the provider instance
+type NameProviderTuple struct {
+	Name     string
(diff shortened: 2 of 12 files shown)


database/query/entities.sql
```diff
@@ -74,7 +74,9 @@ LIMIT 1;
 
 -- name: GetEntitiesByType :many
 SELECT * FROM entity_instances
-WHERE entity_instances.entity_type = $1 AND entity_instances.project_id = ANY(sqlc.arg(projects)::uuid[]);
+WHERE entity_instances.entity_type = $1
+    AND entity_instances.provider_id = sqlc.arg(provider_id)
+    AND entity_instances.project_id = ANY(sqlc.arg(projects)::uuid[]);
```
jhrozek, line 79: 🚀

internal/providers/manager/manager.go
```diff
@@ -33,6 +33,12 @@ import (
 
 //go:generate go run go.uber.org/mock/mockgen -package mock_$GOPACKAGE -destination=./mock/$GOFILE -source=./$GOFILE
 
+// NameProviderTuple is a tuple of a provider name and the provider instance
+type NameProviderTuple struct {
+	Name     string
+	Provider v1.Provider
+}
```
jhrozek, line 40: This is a clear improvement. I was actually wondering about changing the generic provider interface to always have to implement `GetName` and `GetID`. But that's neither here or now.

internal/repositories/service.go
```diff
@@ -228,18 +228,28 @@ func (r *repositoryService) CreateRepository(
...
+		ProviderID: providerID,
+		Projects:   []uuid.UUID{projectID},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error fetching repositories: %w", err)
+	}
+
+	ents := make([]*models.EntityWithProperties, 0, len(repoEnts))
+	for _, ent := range repoEnts {
+		ewp, err := r.propSvc.EntityWithProperties(ctx, ent.ID, nil)
```
jhrozek, line 244: should this be a transaction so that handlers are not interrupted in case the repo goes away?

jhrozek (approved):
approved with two suggestions

internal/providers/manager/manager.go
JAORMX: That would be really handy, let's do it in a subsequent PR.

rdimitrov (approved):

PR #4430, opened 2024-09-10
Implement getting entity name for gitlab provider

JAORMX:
# Summary

This adds a simple implementation of getting the name for an entity
based on given properties. This also introduces the first properties for
the GitLab provider, which are the group and project.

Related-To: https://github.com/stacklok/minder/issues/4331

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/providers/gitlab/properties.go
+++ b/internal/providers/gitlab/properties.go
@@ -1,0 +1,83 @@
+//
+// Copyright 2024 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gitlab
+
+import (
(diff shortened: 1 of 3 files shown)


internal/entities/properties/properties.go
```diff
@@ -210,6 +210,15 @@ type newPropertiesConfig struct {
 	skipPrefixCheck bool
 }
 
+// MustNewProperties creates Properties from a map or panics
+func MustNewProperties(props map[string]any, opts ...newPropertiesOption) *Properties {
```
jhrozek, line 214: not against adding this to the properties module, but would it suffice to have the method in the tests?

internal/providers/gitlab/properties.go
```diff
@@ -0,0 +1,88 @@
...
+		return formatRepoName(groupName, projectName), nil
+	}
+
+	return "", fmt.Errorf("entity type %s not supported", entityType)
+}
+
+func getStringProp(props *properties.Properties, key string) (string, error) {
+	prop := props.GetProperty(key)
+	if prop == nil {
+		return "", fmt.Errorf("property %s not found", key)
```
jhrozek, line 75: the debug messages might not be as precise but AsString doesn't crash with a nil receiver, but already return an error, so maybe it would be enough to just return `props.GetProperty().AsString()`

jhrozek (approved):
approved with two nits/suggestions

internal/entities/properties/properties.go
JAORMX: I'm not able to import from a `_test.go` file. I moved this where it's used.

jhrozek (approved):

PR #4432, opened 2024-09-10
Fix retrieving properties by upstream ID and name

### LINKED ISSUE #4431: refreshing properties by both upstream ID and name doesn't work
We recently enhanced the property service to include lookup by ID in the cache, which is great, but we didn't handle the case where the lookup properties included the upstream ID but the entity properties in the database did not. In that case we should fall back to looking the property by name.

jhrozek:
# Summary

- **Don't fail if lookupProperties include upstream ID but the entity does not** - We recently enhanced the property service to include lookup by ID in the cache, which is great, but we didn't handle the case where the lookup properties included the upstream ID but the entity properties in the database did not. In that case we should fall back to looking the property by name.
- **Don't return provifv1.ErrEntityNotFound from the property service** - It was irritating to have to handle two return codes. The caller really doesn't care why was the entity not found, just that it was not found.  Let's conver the provider NotFound into service Not found so that the callers have one less error to handle.

Fixes #4431

# Testing

I added a unit test. I need to do better testing, I'm just running out of time right now. I'll continue in the evening.

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_evalstatus.go
+++ b/internal/controlplane/handlers_evalstatus.go
@@ -33,6 +33,7 @@ import (
 	"github.com/stacklok/minder/internal/engine/engcontext"
 	entmodels "github.com/stacklok/minder/internal/entities/models"
 	"github.com/stacklok/minder/internal/entities/properties"
+	propSvc "github.com/stacklok/minder/internal/entities/properties/service"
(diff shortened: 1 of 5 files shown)


JAORMX (approved):

jhrozek:
I did some manual testing and the patch seems to indeed help.

Issue #4393, opened 2024-09-08
Need to handle transactions when refreshing properties in historical evaluations better


jhrozek:
We refresh the properties in the historical evaluation endpoint (which btw should just be temporary to convert from the old tables to the new general tables) and missed the fact that the handler starts a transaction and then always rolls back - that was probably a way to keep consistence.

This means that the transactions that refresh properties are rolled back and on next call fetched again. Oops.

As a side-effect we should make sure that we handle transactions in the properties API well in general.

PR #4434, opened 2024-09-10
Handle calling the property service in transaction better

jhrozek:
# Summary

- **Pass a transaction handle to RetrieveAllProperties and RetrieveAllPropertiesForEntity** - These interface methods wouldn't take transaction into account.
- **Call an explicit commit in ListEvaluationHistory handler** - We refresh the properties in the historical evaluation endpoint (which btw should just be temporary to convert from the old tables to the new general tables) and missed the fact that the handler starts a transaction and then always rolls back - that was probably a way to keep consistence.  This means that the transactions that refresh properties are rolled back and on next call fetched again. This hurts performance quite a bit on repeated listHistory calls.

Fixes: #4393

## Change Type

# Testing

by calling the list history endpoint with expired entities

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_evalstatus.go
+++ b/internal/controlplane/handlers_evalstatus.go
@@ -192,6 +192,11 @@ func (s *Server) ListEvaluationHistory(
 		return nil, err
 	}
 
+	if err := s.store.Commit(tx); err != nil {
+		// non: fatal - this handler just refreshes properties but doesn't write any new data
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error committing transaction")
+	}
+
 	// return data set to client
 	resp := &minderv1.ListEvaluationHistoryResponse{}
 	if len(data) == 0 {
--- a/internal/controlplane/handlers_evalstatus.go
+++ b/internal/controlplane/handlers_evalstatus.go
@@ -592,7 +592,7 @@ func (s *Server) buildRuleEvaluationStatusFromDBEvaluation(
 			Msg("error converting severity will use defaults")
 	}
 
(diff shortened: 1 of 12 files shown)


internal/controlplane/handlers_evalstatus.go
```diff
@@ -192,6 +192,10 @@ func (s *Server) ListEvaluationHistory(
 		return nil, err
 	}
 
+	if err := s.store.Commit(tx); err != nil {
+		return nil, status.Errorf(codes.Internal, "error committing transaction: %v", err)
+	}
```
JAORMX, line 197: Do we care about this error though? This is a read operation after all.

JAORMX (dismissed):

internal/controlplane/handlers_evalstatus.go
jhrozek: good point, resolved

JAORMX: Now comes my second question... if the commit failed, do we need an explicit rollback?

jhrozek: we do a rollback via defer. Would an explicit rollback make much of a difference?

JAORMX: ah, you're right. I missed that. It's probably fine then.

JAORMX (dismissed):

jhrozek:
rebased on current main

jhrozek:
and a lint fix

JAORMX (approved):

PR #4459, opened 2024-09-12
Add "options" to properties service calls

JAORMX:
# Summary

Instead of passing the db store and other optional values as
explicit paramters, this introduces options to the properties service.
This allows us to start building queries in a more dynamic way.

The querier is passed always now as an option.

Disregarding stale data was introduced as an option, and it was also
used in the places it's needed, that is, in GET calls.

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/entities/properties/service/options.go
+++ b/internal/entities/properties/service/options.go
@@ -1,0 +1,113 @@
+//
+// Copyright 2024 Stacklok, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


internal/entities/properties/service/options.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	return psco.storeOrTransaction
+}
+
+// ReadOptions is a struct that contains the options for a read service call
+// This extends the PropertiesServiceCallOptions struct and adds a TolerateStaleData field.
+// This field is used to determine if the service call can return stale data or not.
+// This is useful for read calls that can tolerate stale data.
+type ReadOptions struct {
+	*CallOptions
```
jhrozek, line 53: The callers read much nicer now, the only question I have is if you considered embedding CallOptions directly and not through a pointer. It seems like it would simplify the methods like WithStoreOrTransaction that now have to initialize CallOptions and then just forward, with an embedded struct you could forward directly.

jhrozek (approved):
one comment, but overall an improvement

internal/entities/properties/service/options.go
JAORMX: I can give it a try!

PR #4461, opened 2024-09-12
Add request id to logs.

### LINKED ISSUE #4611: Trace request ids across RPC calls
We want both end users and our smoke tests to be able to report which requests failed, and then track down logs using a shared request id.

Some work was already done in this regard
* polish/wrap up https://github.com/stacklok/minder/pull/4461
* ensure that the header gets through all our proxies without being dropped (a simple smoke tests would be valuable)

blkt:
# Summary

This change adds a GRPC interceptor that fetches the header `Grpc-Metadata-Request-Id` and adds it to the logger. It also returns the request id to the client in the header, which is helpful for debugging.

Fixes #4611 

## Change Type

# Testing

Manual tests.

# Review Checklist:

### DIFF
--- a/internal/logger/logging_interceptor.go
+++ b/internal/logger/logging_interceptor.go
@@ -21,6 +21,7 @@ import (
 	"path"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
@@ -100,4 +101,46 @@ func Interceptor(cfg config.LoggingConfig) grpc.UnaryServerInterceptor {
 
 		return resp, err
 	}
+}
+
+// RequestIDInterceptor traces request ids.
+//
+// It tries to use the request id from the request context, creating a
+// new one if that is missing. It also sends back in the trailer the
+// request id, ensuring that the client receives it.
+func RequestIDInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		rID := maybeGetRequestID(ctx)
+		ctx = zerolog.Ctx(ctx).With().Str("request_id", rID).Logger().WithContext(ctx)
+
+		resp, err := handler(ctx, req)
+
+		if err := grpc.SetTrailer(ctx, metadata.Pairs("request-id", rID)); err != nil {
+			zerolog.Ctx(ctx).Trace().Err(err).Msg("unable to attach request id to trailer")
+		}
+
+		return resp, err
+	}
+}
+
+func maybeGetRequestID(ctx context.Context) string {
+	var rID string
+	if md, ok := metadata.FromIncomingContext(ctx); ok {
+		if rIDs, ok := md["request-id"]; ok {
+			if len(rIDs) != 0 {
+				rID = rIDs[0]
+			}
+		}
+	}
+
+	if rID == "" {
+		return uuid.New().String()
+	}
+
+	if _, err := uuid.Parse(rID); err != nil {
+		zerolog.Ctx(ctx).Trace().Err(err).Msg("request id is not valid uuid")
+		return uuid.New().String()
+	}
+
+	return rID
 }
--- a/internal/logger/logging_interceptor.go
+++ b/internal/logger/logging_interceptor.go
@@ -105,42 +105,20 @@ func Interceptor(cfg config.LoggingConfig) grpc.UnaryServerInterceptor {
 
 // RequestIDInterceptor traces request ids.
 //
-// It tries to use the request id from the request context, creating a
-// new one if that is missing. It also sends back in the trailer the
-// request id, ensuring that the client receives it.
-func RequestIDInterceptor() grpc.UnaryServerInterceptor {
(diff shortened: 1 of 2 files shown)


cmd/cli/app/history/history_list.go
```diff
@@ -116,11 +118,20 @@ func listCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc
...
+	var header, trailer metadata.MD
+	resp, err := client.ListEvaluationHistory(ctx, req, grpc.Header(&header), grpc.Trailer(&trailer))
 	if err != nil {
 		return cli.MessageAndError("Error getting profile status", err)
 	}
 
+	fmt.Printf("%+v\n", rID)
+	fmt.Printf("%+v\n", header)
+	fmt.Printf("%+v\n", trailer)
+
```
blkt, line 134: This obviously needs to be reverted, but I'm wondering what would be the right way to output the request id in the CLI.

internal/logger/logging_interceptor.go
```diff
@@ -101,3 +102,45 @@ func Interceptor(cfg config.LoggingConfig) grpc.UnaryServerInterceptor {
...
+		}
+
+		return resp, err
+	}
+}
+
+func maybeGetRequestID(ctx context.Context) string {
+	var rID string
+	if md, ok := metadata.FromIncomingContext(ctx); ok {
+		if rIDs, ok := md["request-id"]; ok {
```
JAORMX, line 129: in HTTP what would this translate to? I wonder how it works with the HTTP gateway we use.

blkt: GRPC moves back and forth headers having prefix `Grpc-Metadata-`, so it would become `Grpc-Metadata-Request-Id`.

evankanderson:
We might want to use https://www.w3.org/TR/trace-context/

blkt:
@evankanderson while estimating work to improve logging I had a look at the documents you linked. My take is that the capabilities described in those specs far exceed our current need, which is to track a request execution from start to finish in logs.

My take is that we're better off polishing this work, possibly changing the header's key (not the value), and merge it.

evankanderson:
> @evankanderson while estimating work to improve logging I had a look at the documents you linked. My take is that the capabilities described in those specs far exceed our current need, which is to track a request execution from start to finish in logs.
> 
> My take is that we're better off polishing this work, possibly changing the header's key (not the value), and merge it.

Sounds good; I'd thought this would be something like:

```golang
import "go.opentelemetry.io/otel/trace"

// ...
sc := trace.SpanContextFromContext(ctx)
ctx = zerolog.Ctx(ctx).With().Str("trace_id", sc.TraceID()).Logger().WithContext(ctx)
```

If there's a bunch more content, then I'd agree on doing our own thing.

blkt:
Here is an example of the output

```
curl -v -H "Authorization: Bearer $TOKEN" 'http://localhost:8080/api/v1/user'
> GET /api/v1/user HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> Accept: */*
> Authorization: Bearer xxx
>
* Request completely sent off
< HTTP/1.1 200 OK
< Content-Length: 933
< Content-Type: application/json
< Grpc-Metadata-Content-Type: application/grpc
< Grpc-Metadata-Request-Id: df1b6733-65af-4472-9a69-5b783558322f
< Date: Tue, 15 Oct 2024 12:25:13 GMT
<
...
```

jhrozek (approved):
I think this is a good start, I see the result both on the server side in the middleware and the client side in the headers (verified with curl).

Regarding the server vs client side headers - what if we had another header like e.g. `Request-Correlation-ID` that the client could set in case a single client operation performs multiple requests? In that case the client might set the correlation to request ID of the first operation. There would be still the unique request ID and the correlation would really be just a hint for easier search of requests in the logs?

evankanderson (approved):

evankanderson:
> I think this is a good start, I see the result both on the server side in the middleware and the client side in the headers (verified with curl).
> 
> Regarding the server vs client side headers - what if we had another header like e.g. `Request-Correlation-ID` that the client could set in case a single client operation performs multiple requests? In that case the client might set the correlation to request ID of the first operation. There would be still the unique request ID and the correlation would really be just a hint for easier search of requests in the logs?

I'd wait here for a "bigger" tracing solution, and then have the top-level command use some sort of trace-parent to correlate the requests.

One thing that may not be obvious here, but it looks like the `request_id` header won't pass through any usages of watermill or other asynchronous processing.  I think that's fine for now, but it's worth being aware of given how much async processing we do.

blkt:
> Regarding the server vs client side headers - what if we had another header like e.g. `Request-Correlation-ID` that the client could set in case a single client operation performs multiple requests?

I agree with @evankanderson in this, tracing across multiple requests requires some amount of shared state and more changes to the codebase. This is just a helper to filter down logs quickly.

> One thing that may not be obvious here, but it looks like the `request_id` header won't pass through any usages of watermill or other asynchronous processing.

That's correct, for that case we already track Github's delivery id that sort-of solves the same problem.
https://github.com/mindersec/minder/blob/main/internal/providers/github/webhook/hook.go#L127
https://github.com/mindersec/minder/blob/main/internal/providers/github/webhook/app.go#L151
We might want to spend some effort ensuring this is logged correctly downstream watermill. I'll create a task.

Issue #4329, opened 2024-08-30
Add GitLab webhook handler


JAORMX:
### Please describe the enhancement

Add a webhook handler that issues events into minder about entities. This handler should live in the gitlab provider and should not have access to the central minder database. It should be event-based.

### Solution Proposal

-

PR #4476, opened 2024-09-13
Add general webhook registration mechanism

JAORMX:
# Summary

This adds per provider class paths so each provider class will bring
it's own webhook handling. GitHub is a slightly special case... so we
handle it separately for now.

Note that this will require folks changing their configuration to remove
`github` as a suffix on deployments. If nothing is changed, this should
be a no-op as it does not re-append `github` in the path.

Related-To: #4329

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/controlplane/server.go
+++ b/internal/controlplane/server.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
 	"runtime/debug"
 	"time"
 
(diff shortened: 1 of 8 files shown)


internal/controlplane/server.go
```diff
@@ -352,8 +354,20 @@ func (s *Server) StartHTTPServer(ctx context.Context) error {
 	// This already has _some_ middleware due to the GRPC handling
 	mux.Handle("/", withMaxSizeMiddleware(s.withCORSMiddleware(gwmux)))
 
+	// Register the webhook handlers
+	// Note: The GitHub webhook handler is not registered here.
+	for classpath, handler := range s.providerManager.IterateWebhookHandlers() {
```
jhrozek, line 359: do we not register the GH handler here because we have a single handler for the app and the oauth providers?

jhrozek: ah, no, it's because this adds the API and not the implementation yet, sorry for the noise

jhrozek (approved):
LGTM!

PR #4481, opened 2024-09-13
Add rule evaluations document

ethomson:
# Summary

Include information about the way minder evaluates rules, the possible statuses, and how users can view them.

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/docs/docs/understand/rule_evaluation.md
+++ b/docs/docs/understand/rule_evaluation.md
@@ -1,0 +1,48 @@
+---
+title: Rule evaluations
+sidebar_position: 50
+---
+
+# Rule evaluations
+
+When Minder evaluates the [rules in your profiles](profiles.md), it records the state of those evaluations. When those rules are not satisfied because the criteria that you defined was not met, it will issue [alerts](alerts.md).
+
+Minder evaluates the rules in your profile:
+* When the repository is registered
+* When the profile is updated
+* When activity occurs within the repository
+
+In a rule evaluation, you'll see:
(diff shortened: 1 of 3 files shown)


docs/docs/understand/rule_evaluation.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Viewing rule evaluations
+
+To view the rule evaluations, run [`minder history list`](../ref/cli/minder_history_list.md). You can query the history to only look at certain entities, profiles, or statuses.
+
+### Evaluation status
+
+The _status_ of a rule evaluation describes the outcome of executing the rule against an entity. Possible statuses are:
+
+* **Success**: the entity was evaluated and is in compliance with the rule. For example, given the [`secret_scanning`](../ref/rules/secret_scanning) rule, this means that secret scanning is enabled on the repository being evaluated.
```
jhrozek, line 31: There is also "Pending" which means "this rule was not evaluated yet". I'm not sure if it's even worth documenting because the for repos the only way to have a pending result is to create a profile before you register a repo.

jhrozek (dismissed):
I'll leave it up to you whether you think we should document the pending state. Otherwise looks good!

docs/docs/understand/rule_evaluation.md
ethomson: I think that it's worth being complete here; I added a bullet point, LMK what you think.

jhrozek (approved):

PR #4492, opened 2024-09-16
Add webhook secret bootstrap and verification for Gitlab

JAORMX:
# Summary

Gitlab allows for setting up a shared secret for webhook authentication [1].
This shared secret is passed on via the `X-Gitlab-Token` header, and
it's the receiver's responsibility to verify it.

In our implementation, we take the secret from Gitlab's provider
configuration (it'll be a secret in k8s), append the unique identifier
we create for the webhook, and then apply sha 512. This ensures that we
have a section that's secret, while not passing in the actual secret to
Gitlab. It also ensures that if a secret token get compromised on the
gitlab side, it'll only be applicable to a single entity.

In your local configuration, this would look similar to this:

```
provider:
  gitlab:
      client_id: "XXXX"
      client_secret: "XXXX"
      redirect_uri: "http://localhost:8080/api/v1/auth/callback/gitlab"
      webhook_secret: <A SECRET GOES HERE>
      scopes:
        - api
        - read_api
        - read_repository
        - write_repository
        - read_registry
        - write_registry
        - openid
        - profile
        - email
```

[1] https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#create-a-webhook

## Change Type

# Testing

# Review Checklist:


internal/providers/gitlab/webhooksecret/whsecret.go
```diff
@@ -0,0 +1,57 @@
...
+// New creates a new secret for usage in the gitlab webhook.
+// The secret is generated by combining the base and uniq strings
+// and then hashing the result.
+func New(base string, uniq string) (string, error) {
+	if base == "" || uniq == "" {
+		return "", ErrEmptyBaseOrUniq
+	}
+
+	hash := sum.New()
+	_, err := hash.Write([]byte(base + uniq))
```
jhrozek, line 40: I think this is OK for simplicity sake to bootstrap the gitlab provider, but for production-ready secret handling I think we should go further (for github also). What I would like to be able to do is to rotate the minder secrets without having to rotate the provider secrets and conversely being able to rotate provider without having to rotate the minder secret. We already had a design for this a while ago with @dmjb and I think we even have an issue open.

internal/providers/gitlab/webhooksecret/whsecret.go
```diff
@@ -0,0 +1,57 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package webhooksecret provides a way to generate and verify secrets for GitLab webhooks.
+package webhooksecret
+
+import (
+	sum "crypto/sha512"
```
jhrozek, line 20: I was wondering a bit if we should use HMAC and treat the secret generation as generating an HMAC for the uniq, but...I can't really convince myself that it makes a difference in this scenario.

jhrozek (approved):
I think this is fine for now but we should make sure we are ready to handle rotation before we productize the new provider.

PR #4512, opened 2024-09-17
Add constant comparison operator to jq evaluation

eleftherias:
# Summary

This adds a `constant` field to the `jq` evaluation engine.
This allows us to write ruletypes like the one below, which compares the evaluation result to a constant value, rather than requiring the profile to supply to value.

The example below compares the field `delete_branch_on_merge` in the evaluation result to the constant `true`.
```
def:
  in_entity: repository
  rule_schema:
    type: object
    properties: {}
  ingest:
    type: rest
    rest:
      endpoint: "/repos/{{.Entity.Owner}}/{{.Entity.Name}}"
      parse: json
  eval:
    type: jq
    jq:
    - ingested:
        def: '.delete_branch_on_merge'
      constant: true
```

This also changes the `jq` equality check to not differentiate between number type (`int`, `float64` etc).
The alternative would be to let the user specify the type of their number constant, but I'm making the assumption that the likeliness of someone needing that is small, and not worth the additional overhead of setting a type every time.

Ref https://github.com/stacklok/minder-rules-and-profiles/issues/156

I will update the docs once this is approved.

## Change Type

# Testing

# Review Checklist:


internal/engine/eval/jq/jq.go
```diff
@@ -42,11 +42,11 @@ func NewJQEvaluator(assertions []*pb.RuleType_Definition_Eval_JQComparison) (*Ev
 
 	for idx := range assertions {
 		a := assertions[idx]
-		if a.Profile == nil {
-			return nil, fmt.Errorf("missing profile accessor")
+		if a.Profile == nil && a.Constant == nil {
+			return nil, fmt.Errorf("missing profile or constant accessor")
 		}
```
evankanderson, line 47: Do you want `Profile` and `Constant` to be XOR (at least one set, but not both)?

internal/engine/eval/jq/jq.go
```diff
@@ -88,9 +99,9 @@ func (jqe *Evaluator) Eval(ctx context.Context, pol map[string]any, res *engif.R
...
 
 		// Deep compare
-		if !reflect.DeepEqual(profileVal, dataVal) {
+		if !reflect.DeepEqual(numberToFloat64(expectedVal), numberToFloat64(dataVal)) {
```
evankanderson, line 102: How does this work for booleans (like your example)?  I believe you that it does, but I was surprised to see that it worked.

Oh, I see!  `numberToFloat64` actually returns `any`, and passes through other values.  I'm trying to figure out if there's a better name -- maybe `standardizeNumbers`?

internal/engine/eval/jq/jq_test.go
```diff
@@ -271,6 +317,27 @@ func TestValidJQEvals(t *testing.T) {
...
 			},
 		},
+		{
+			name: "valid single rule evaluates constant bool",
+			assertions: []*pb.RuleType_Definition_Eval_JQComparison{
+				{
+					Constant: &structpb.Value{
+						Kind: &structpb.Value_BoolValue{
+							BoolValue: false,
+						},
```
evankanderson, line 327: I think you can write this as:

```suggestion
					Constant: &structpb.NewBoolValue(false),
```

internal/engine/eval/jq/jq_test.go
```diff
@@ -292,6 +359,45 @@ func TestValidJQEvals(t *testing.T) {
...
+									},
+									{
+										Kind: &structpb.Value_StringValue{
+											StringValue: "c",
+										},
+									},
+								},
+							},
+						},
+					},
```
evankanderson, line 388: I love that we're testing list input. 🎉 

Again, I think the helpers in `structpb` will make this less painful to read and write:

```suggestion
					Constant: &structpb.NewListValue(structpb.ListValue{Values: {
						structpb.NewStringValue("a"),
						structpb.NewStringValue("b"),
						structpb.NewStringValue("c"),
						},
					}}),
```

Unfortunately, `NewList` returns a `*ListValue, error`, so we can't just feed it into `NewListValue` directly.  (We could also do it _outside_ the function, or write an `errorToPanic[T](T, err) T` helper method.)

evankanderson (approved):
This makes a lot of sense, and it's sort of amazing that it was missing for so long.  Thanks for adding the test coverage, too!

internal/engine/eval/jq/jq.go
eleftherias: Good point, I've updated it

internal/engine/eval/jq/jq.go
eleftherias: Updated the name

JAORMX (approved):

Issue #4514, opened 2024-09-17
Parsing a stored selector fails if the selector is using commas


jhrozek:
I'm unsure of the root case, but parsing this selector:
```
repository.properties['github/primary_language'] in ['TypeScript', 'Go']
```
from the database fails. We do have a hand-rolled `Scan` method that converts a JOIN with an ARRAY into a Go type which breaks.

jhrozek:
CC @danbarr who reported the issue.

Marking as P0 because after such selector has been added minder just throws errors attempting to manipulate that profile.

PR #4515, opened 2024-09-17
Use encoding/csv to when parsing selector record

jhrozek:
# Summary

`encoding/csv` allows us to handle selectors with commas in them which
our naive parser embarassingly didn't such as:
```
repository.properties['github/primary_language'] in ['TypeScript', 'Go']
```

Fixes: #4514

## Change Type

# Testing

unit test plus using that selector

# Review Checklist:

### DIFF
--- a/internal/db/profile_selector_scan.go
+++ b/internal/db/profile_selector_scan.go
@@ -39,6 +39,7 @@ func (s *ProfileSelector) Scan(value interface{}) error {
 
 	// Split the string by commas to get the individual field values
 	cr := csv.NewReader(strings.NewReader(str))
+	cr.LazyQuotes = true // Enable LazyQuotes to allow for uneven number of quotes
 	parts, err := cr.Read()
 	if err != nil {
 		return fmt.Errorf("failed to scan SelectorInfo: %v", err)
--- a/internal/db/profile_selector_scan.go
+++ b/internal/db/profile_selector_scan.go
@@ -15,6 +15,7 @@
 package db
 
 import (
+	"encoding/csv"
 	"fmt"
 	"strings"
 )
@@ -37,7 +38,11 @@ func (s *ProfileSelector) Scan(value interface{}) error {
 	str = strings.TrimSuffix(str, ")")
 
 	// Split the string by commas to get the individual field values
-	parts := strings.Split(str, ",")
+	cr := csv.NewReader(strings.NewReader(str))
+	parts, err := cr.Read()
+	if err != nil {
+		return fmt.Errorf("failed to scan SelectorInfo: %v", err)
+	}
 
 	// Assign the values to the struct fields
 	if len(parts) != 5 {
--- a/internal/db/profile_selector_scan_test.go
+++ b/internal/db/profile_selector_scan_test.go
@@ -22,6 +22,9 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// TestScan tests the Scan method of the ProfileSelector struct
+// for more tests that exercise the retrieval of a profile with selectors that also happens to use Scan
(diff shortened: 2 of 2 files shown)


JAORMX (approved):

internal/db/profile_selector_scan_test.go
```diff
@@ -47,6 +47,20 @@ func TestScan(t *testing.T) {
 				Comment:  "comment1",
 			},
 		},
+		{
+			name:  "Valid input with commas in the selector",
+			input: []byte(fmt.Sprintf("(%s,%s,repository,\"repository.properties['github/primary_language'] in ['TypeScript', 'Go']\",\"comment1\")", selectorId, profileId)),
```
evankanderson, line 52: Do you want to use ` for this string to avoid needing to escape single and double quotes?

internal/db/profile_selector_scan.go
```diff
@@ -37,7 +38,11 @@ func (s *ProfileSelector) Scan(value interface{}) error {
 	str = strings.TrimSuffix(str, ")")
 
 	// Split the string by commas to get the individual field values
-	parts := strings.Split(str, ",")
+	cr := csv.NewReader(strings.NewReader(str))
+	parts, err := cr.Read()
+	if err != nil {
+		return fmt.Errorf("failed to scan SelectorInfo: %v", err)
+	}
```
evankanderson, line 45: Do we need to update the reverse (store) code as well?

Do we have some unit tests to cover the round-trip as well as the serialized format?

evankanderson (approved):

internal/db/profile_selector_scan.go
jhrozek: there is no serialization, really. The selectors are used as-is, but they are retrieved by aggregating the selectors table with the profiles table in https://github.com/stacklok/minder/blob/6f89a1de73ec7e42b1730951b645bbfb6c1387a0/database/query/profiles.sql#L68 which uses a SQL type defined at https://github.com/stacklok/minder/blob/6f89a1de73ec7e42b1730951b645bbfb6c1387a0/database/migrations/000072_profile_selectors_type.up.sql#L17 which we force through sqlc (https://github.com/stacklok/minder/blob/6f89a1de73ec7e42b1730951b645bbfb6c1387a0/sqlc.yaml#L31)

The code I modified is a Scan() to convert the Go type for the selectors defined in models.go to the sql type.

internal/db/profile_selector_scan_test.go
jhrozek: will do!

internal/db/profile_selector_scan.go
evankanderson: Aha!  I'm slightly nervous about whether the Postgres `ARRAY_AGG` produces csv output, but this makes sense for  now.

evankanderson (approved):

internal/db/profile_selector_scan.go
jhrozek: turns out that there is `pq.Array` which sounds like something I want to use. I'll take a look into using it - thanks for bringing up the concern!

jhrozek: OK, I confused myself. `pq.Array` would be useful for parsing the whole array, but here we are parsing the `ROW`, not the array. I would still go with the csv approach, but we might want to consider using `LazyQuotes`. I'll do some more tests before we merge this PR.

internal/db/profile_selector_scan_test.go
jhrozek: done!

evankanderson (approved):
I still wish we had a better parser, but I don't want to block this improvement on it.

Issue #4327, opened 2024-08-30
Remove database access from GitHub webhook handler


JAORMX:
### Please describe the enhancement

-

### Solution Proposal

-

jhrozek:
Since the work breakdown we did initially was not great and it might have been a bit opaque what we did I'm going to break this issue into 2 - removing the database access from the GH provider was not strictly needed to enable github, but the generic entity lookup and evaluation enabled us to both implement the GL provider and make the GH provider use the same code. That part is now done and to make it clearer what was done and why I split that part of work into https://github.com/stacklok/minder/issues/4561

jhrozek:
This was fixed by fixing #4618, #4619, #4621, #4666, #4667, #4682, #4703

PR #4517, opened 2024-09-17
Simplify the attributes sent to the entity_add handler

jhrozek:
# Summary

I have a rather large branch that works towards not using database access from
the webhook handler. These two patches were part of that branch and can be submitted separately.

The aim is to decrease the number of attributes we send to the handler as well
as make them more generic.

Related: #4327

## Change Type

# Testing

as part of a larger branch + unit tests

# Review Checklist:

### DIFF
--- a/internal/reconcilers/entity_add.go
+++ b/internal/reconcilers/entity_add.go
@@ -16,7 +16,6 @@ package reconcilers
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 
 	"github.com/ThreeDotsLabs/watermill/message"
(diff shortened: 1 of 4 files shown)


jhrozek:
@blkt would you mind reviewing this? You probably know the code better than I

internal/reconcilers/entity_add.go
```diff
@@ -52,21 +51,17 @@ func (r *Reconciler) handleEntityAddEvent(msg *message.Message) error {
...
-	var repoName string
+	var entName string
 	var ok bool
-	if repoOwner, ok = event.Entity["repoOwner"].(string); !ok {
-		return errors.New("invalid repo owner")
-	}
-	if repoName, ok = event.Entity["repoName"].(string); !ok {
-		return errors.New("invalid repo name")
+	if entName, ok = event.Entity[properties.PropertyName].(string); !ok {
+		return fmt.Errorf("invalid or unset entity name: %s", entName)
```
JAORMX, line 57: Will this continue using the name? Or will this switch to using the upstream ID?

jhrozek: interesting idea, this code goes to repositories.CreateRepository which uses the name as a property for RetrieveAllProperties. I'm not sure if that and especially the underlying code supports retrieving a repo by name. I need to check that.

JAORMX: It's not a big deal what property is used, as long as we use properties and not tie something to names.

jhrozek:
I will take the feedback into account, but I'm also going to close this PR and resend it in another PR that will actually also take this change into use.

PR #4519, opened 2024-09-17
Add a method to retrieve an entity by upstream ID with a hint

jhrozek:
# Summary

When processing a webhook payload, we typically only have the
information from the webhook payload available which is to say
information about the upstream entity. We even don't know exactly which
provider and which project are we handling.

We used to sort of wing it in the github webhook handler and were just
searching by upstream ID as if it was globally unique, but that no
longer works with the introduction of multiple providers where several
entities might have the same upstream ID, just in different providers.

To handle that, we add a new method to the properties service that
searches by upstream ID with a hint. At the moment, the interface
supports providerID, projectID and provider implements as a hint.

The immediate use is that the github provider will search by the
upstream ID with the hint set to ProviderTypeGithub as that's
implemented both by the app and OAuth github providers.

If any attribute of the hint is set, then the entity's attribute must
match the hint, otherwise the entity is filtered out. Additionally, once
the entity is found, the hint might contain a property with a value that
must match. The use-case there would be deleting entities where we want
to ensure that we are deleting an entity with an appropriate upstream
hook ID.

If no entities match the hints or if multiple entities match a hint, an
error is returned.

Related: #4327

## Change Type

# Testing

as part of a larger branch + unit tests

# Review Checklist:

### DIFF
--- a/internal/entities/properties/service/helpers.go
+++ b/internal/entities/properties/service/helpers.go
@@ -20,6 +20,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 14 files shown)


internal/entities/properties/service/service.go
```diff
@@ -337,6 +333,75 @@ func (ps *propertiesService) EntityWithProperties(
...
+		return nil, fmt.Errorf("error getting entity: %w", err)
+	}
+
+	return ps.getEntityWithProperties(ctx, ent, opts)
+}
+
+func (ps *propertiesService) EntityWithPropertiesByUpstreamID(
+	ctx context.Context, entType minderv1.Entity,
+	upstreamID string, hint ByUpstreamIdHint,
+	opts *CallOptions,
```
JAORMX, line 356: I'm a little confused by this function. Don't we want to use properties always instead of having to pass the upstream ID? The provider webhook handler would parse al relevant properties that are necessary to identify an entity, and that would be used here.

jhrozek: This is what I started with actually but then always ended up using only upstreamID as the attribute. This might just be a function of us using the method in the github provider where we typically look up entities by ID.

I'm fine adding properties back and handling only by-ID for now or by-ID and by-name to look up the entity in the cache.

Looking at the code again we could also just reuse getEntityIdByProperties and in the existing callers (RetrieveAllProperties and RetrieveProperty) just set the hint to the projectID and providerID which would point them at a single entity only.

Thanks for the feedback!

jhrozek: Done, now we just loop over the get-by-something properties in a defined order and try to look them up.

internal/entities/properties/service/helpers.go
```diff
@@ -143,35 +144,191 @@ func getEntityIdByName(
...
+	lookupOrder := []string{properties.PropertyUpstreamID, properties.PropertyName}
+	for _, loopupProp := range lookupOrder {
+		prop := props.GetProperty(loopupProp)
+		if prop == nil {
+			continue
+		}
+
+		l.Debug().Str("lookupProp", loopupProp).Msg("fetching by property")
+		ents, err = getAllByProperty(ctx, loopupProp, prop.RawValue(), entType, hint.projectID, hint.providerID, qtx)
+		if errors.Is(err, ErrEntityNotFound) {
```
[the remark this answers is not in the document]
jhrozek: thanks, I moved the logic around so it hopefully reads better

JAORMX (approved):

PR #4584, opened 2024-09-30
Add feature flag for improved evaluation details.

blkt:
# Summary

This change adds feature flag `vulncheck_error_template`, which is currently used by the `vulncheck` evaluator to determine whether to produce "legacy" message format (i.e. comma separated list of package names) or markdown (i.e. unordered list of package names).

Additionally, this change adds support for options when creating new evaluation engines, making them easier to extend and possibly test.

The change is backwards compatible.

## Change Type

# Testing

Manual tests.

# Review Checklist:

### DIFF
--- a/internal/engine/eval/vulncheck/vulncheck.go
+++ b/internal/engine/eval/vulncheck/vulncheck.go
@@ -22,11 +22,14 @@ import (
 	"strings"
 
 	"github.com/hashicorp/go-version"
+	"github.com/open-feature/go-sdk/openfeature"
 	"github.com/rs/zerolog"
 
 	evalerrors "github.com/stacklok/minder/internal/engine/errors"
 	"github.com/stacklok/minder/internal/engine/eval/templates"
 	engif "github.com/stacklok/minder/internal/engine/interfaces"
+	eoptions "github.com/stacklok/minder/internal/engine/options"
+	"github.com/stacklok/minder/internal/flags"
 	pbinternal "github.com/stacklok/minder/internal/proto"
 	provifv1 "github.com/stacklok/minder/pkg/providers/v1"
 )
@@ -38,18 +41,39 @@ const (
 
 // Evaluator is the vulncheck evaluator
 type Evaluator struct {
-	cli provifv1.GitHub
+	cli          provifv1.GitHub
+	featureFlags openfeature.IClient
+}
+
+var _ eoptions.SupportsFlags = (*Evaluator)(nil)
+
+// SetFlagsClient sets the `openfeature` client in the underlying
+// `Evaluator` struct.
+func (e *Evaluator) SetFlagsClient(client openfeature.IClient) error {
+	e.featureFlags = client
+	return nil
 }
 
 // NewVulncheckEvaluator creates a new vulncheck evaluator
-func NewVulncheckEvaluator(ghcli provifv1.GitHub) (*Evaluator, error) {
+func NewVulncheckEvaluator(
+	ghcli provifv1.GitHub,
+	opts ...eoptions.Option,
+) (*Evaluator, error) {
 	if ghcli == nil {
 		return nil, fmt.Errorf("provider builder is nil")
 	}
 
-	return &Evaluator{
+	evaluator := &Evaluator{
 		cli: ghcli,
-	}, nil
+	}
+
+	for _, opt := range opts {
+		if err := opt(evaluator); err != nil {
+			return nil, err
+		}
+	}
+
+	return evaluator, nil
 }
 
 // Eval implements the Evaluator interface.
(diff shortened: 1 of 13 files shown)


internal/flags/constants.go
```diff
@@ -18,10 +18,11 @@ package flags
...
-	// EvalHistory enables logging of evaluation history in the new tables.
-	EvalHistory Experiment = "eval_history"
 	// DockerHubProvider enables the DockerHub provider.
 	DockerHubProvider Experiment = "dockerhub_provider"
 	// GitLabProvider enables the GitLab provider.
 	GitLabProvider Experiment = "gitlab_provider"
+	// ImprovedEvalDetails enables improved evaluation details
+	// messages.
+	ImprovedEvalDetails Experiment = "improved_eval_details"
 )
```
evankanderson, line 28: I'd be inclined to make this experiment specific to the particular improvement, rather than trying to cover _all_ eval conditions with one flag.  In particular, it will make retiring the experiment later easier, because the scope will be limited.

internal/engine/options/options.go
```diff
@@ -0,0 +1,46 @@
...
+
+// SupportsFlags interface advertises the fact that the implementer
+// can use an `openfeature` client to check for flags being set.
+type SupportsFlags interface {
+	SetFlagsClient(client openfeature.IClient) error
+}
+
+// Option is a function that takes an evaluator and does some
+// unspecified operation to it, returning an error in case of failure.
+type Option func(interfaces.Evaluator) error
```
evankanderson, line 33: Interesting question -- do we want to have an `error` return here, or do we want options to have to be successful?

I don't have a strong feeling, but I do wonder whether we want to have the error handling propagate all over the place, or have to be handled by the Option.

internal/engine/eval/vulncheck/vulncheck.go
```diff
@@ -60,9 +84,16 @@ func (e *Evaluator) Eval(ctx context.Context, pol map[string]any, res *engif.Res
 	}
 
 	if len(vulnerablePackages) > 0 {
-		return evalerrors.NewDetailedErrEvaluationFailed(
-			templates.VulncheckTemplate,
-			map[string]any{"packages": vulnerablePackages},
+		if e.featureFlags != nil && flags.Bool(ctx, e.featureFlags, flags.ImprovedEvalDetails) {
```
evankanderson, line 87: Should we improve `flags.Bool` to handle a `nil` IClient?  It sort of feels better than littering `nil` checks elsewhere in the code.

internal/engine/rtengine/cache.go
```diff
@@ -73,17 +76,20 @@ func NewRuleEngineCache(
...
 
-	return &ruleEngineCache{engines: engines}, nil
+	return &ruleEngineCache{engines: engines, opts: opts}, nil
 }
 
-func (r *ruleEngineCache) GetRuleEngine(ctx context.Context, ruleTypeID uuid.UUID) (*RuleTypeEngine, error) {
+func (r *ruleEngineCache) GetRuleEngine(
+	ctx context.Context,
+	ruleTypeID uuid.UUID,
+) (*RuleTypeEngine, error) {
```
evankanderson, line 92: What's causing this reformat?  It's polluting the diff a bit.

I wouldn't fix things for this PR, but it would be nice to avoid these extra diffs for both blame reasons and ease of PR review.

internal/flags/constants.go
evankanderson: e.g. 
```suggestion
	// VulnCheckErrorTemplate enables improved evaluation details
	// messages in the vulncheck rule.
	VulnCheckErrorTemplate Experiment = "vulncheck_error_template"
)
```

evankanderson (dismissed):
A few specific comments, but this approach makes sense to me.

internal/engine/rtengine/cache.go
blkt: This is due to some earlier experiments that made the signature excede 100 characters, I can revert this.

internal/engine/options/options.go
blkt: Usually, this pattern requires the developer to apply `Option` functions in some kind of constructor routine, so you'd have to handle errors anyway.

Besides, they might perform some kind of validation and the constructor should be aware of such errors.

internal/engine/eval/vulncheck/vulncheck.go
evankanderson: I'm okay with that, but I'd also be okay with `nil` Client == `false` result, always.

Needing to think about default polarity of flags in each case is probably an anti-pattern -- I'd rather (like Go) use a default-false polarity, and flags express themselves as positive assertions of some variation from the base behavior.

internal/engine/options/options.go
evankanderson: This is a philosophy question.  I'm willing to go with the "there might be errors, so expose error handling" philosophy, I just wanted to point out the other path ("options must not generate errors and do something sensible themselves") so we could choose.

I think we both agree that _if an option generates an error_, then the underlying object that the option was applied to is in an unspecified state and _must not be used_.

internal/engine/eval/vulncheck/vulncheck.go
blkt: I don't think reasoning about default-false polarity when writing code is the problem, I think the problems begin when trying to figure out why some user facing element, far away from the flag evaluation, does not have the expected form because of Go-style default value semantics.

In general, I'm not a big fan of default values as a solution to memory initialization either.

internal/engine/options/options.go
blkt: > the other path ("options must not generate errors and do something sensible themselves")

I don't think that's useful generally, it might be very surprising to the user.
For example, we use options to configure the filter used in the `history list` endpoint. That endpoint supports filtering by time range, and requires `from` and `to` to be relatively ordered in the right way ([here](https://github.com/stacklok/minder/blob/main/internal/history/models.go#L626-L633)). Not returning an error in that case would be very, very confusing to the user, who would receive results not matching his expectation.

> I think we both agree that if an option generates an error, then the underlying object that the option was applied to is in an unspecified state and must not be used.

I totally agree on that, it might be in an inconsistent state.
I also add that, if all option succeed it's still up to the caller to verify that the object the options were applied to is in a consistent state.

eleftherias (approved):

Issue #4618, opened 2024-10-02
Replace handleArtifactsReconcilerEvent with a more generic reconcile originating entities handler


jhrozek:
At the moment, when reconciling repositories, we do 2 things that are very github specific:
1) we unconditionally reconcile artifacts
2) we use github-provider specific calls to list packages

Instead, we should reconcile originating entities in general. We should move the code that reconciles artifacts using the GH calls to the GH provider and use the new refresh-and-do handlers in the reconciler.

PR #4620, opened 2024-10-02
Remove RefreshRepositoryByUpstreamID from the repo reconciler

### LINKED ISSUE #4595: Remove RefreshRepositoryByUpstreamID from the repo reconciler
This will enable us to use the repo reconciler to work well for both github and gitlab providers as well as remove the RefreshRepositoryByUpstreamID request which is just a temporary hack and causes unnecessary API calls.

jhrozek:
# Summary

- **Rename WithOwner to WithOriginator** - We renamed Owner to Originator in the new refresh-and-do handlers, but forgot to rename the method that sets that attribute.
- **Add a new Get-By-ID-Refresh-And-Do handler** - Since we have the entity ID available, we don't need to refresh by upstream ID, but by entity ID. To do that, we add a new handler that does just that.
- **Use NewEntityRefreshAndDoMessage instead of calling RefreshRepositoryByUpstreamID in reconcilers** - Gets rid of the RefreshRepositoryByUpstreamID request that we are tryign to phase out in favour of the generic new handlers. This will also allow the reconcile handler to be useful for the github reconciler.
- **Handle messages with no entityID** - since we changed the format of the reconcile message, we check if the field we added is set and if not, just drop the message.

Fixes #4595

## Change Type

# Testing

unit tests + manual

# Review Checklist:

### DIFF
--- a/internal/reconcilers/messages/messages.go
+++ b/internal/reconcilers/messages/messages.go
@@ -28,15 +28,18 @@ import (
 type RepoReconcilerEvent struct {
 	// Project is the project that the event is relevant to
 	Project uuid.UUID `json:"project"`
-	// Repository is the repository to be reconciled
-	Repository int64 `json:"repository" validate:"gte=0"`
+	// Provider is the provider that the event is relevant to
+	Provider uuid.UUID `json:"provider"`
+	// EntityID is the entity id of the repository to be reconciled
+	EntityID uuid.UUID `json:"entity_id"`
 }
 
 // NewRepoReconcilerMessage creates a new repos init event
-func NewRepoReconcilerMessage(providerID uuid.UUID, repoID int64, projectID uuid.UUID) (*message.Message, error) {
+func NewRepoReconcilerMessage(providerID uuid.UUID, entityID uuid.UUID, projectID uuid.UUID) (*message.Message, error) {
 	evt := &RepoReconcilerEvent{
-		Repository: repoID,
-		Project:    projectID,
+		Project:  projectID,
+		Provider: providerID,
+		EntityID: entityID,
 	}
 
 	evtStr, err := json.Marshal(evt)
@@ -45,7 +48,6 @@ func NewRepoReconcilerMessage(providerID uuid.UUID, repoID int64, projectID uuid
 	}
 
 	msg := message.NewMessage(uuid.New().String(), evtStr)
-	msg.Metadata.Set("provider_id", providerID.String())
 	return msg, nil
 }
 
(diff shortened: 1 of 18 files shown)


internal/reconcilers/messages/messages.go
```diff
@@ -22,21 +22,28 @@ import (
...
-	// Repository is the repository to be reconciled
-	Repository int64 `json:"repository" validate:"gte=0"`
+	// Provider is the provider that the event is relevant to
+	Provider uuid.UUID `json:"provider"`
+	// Properties is the properties of the repository to be reconciled
+	Properties map[string]any `json:"properties"`
 }
 
 // NewRepoReconcilerMessage creates a new repos init event
 func NewRepoReconcilerMessage(providerID uuid.UUID, repoID int64, projectID uuid.UUID) (*message.Message, error) {
```
jhrozek, line 40: With these changes and especially when issue #4618 is implemented, we can just pass in properties and use this to reconcile a generic entity.

jhrozek:
> oops the tests are failing looking into this

this was just a silly bug - I use a real postgres in the test but was reusing the same ID and triggering duplicate errors

internal/reconcilers/repository.go
```diff
@@ -56,40 +58,71 @@ func (r *Reconciler) handleRepoReconcilerEvent(msg *message.Message) error {
...
+		return nil
+	}
+
 	// first retrieve data for the repository
-	repository, err := r.store.GetRepositoryByRepoID(ctx, evt.Repository)
+	repository, err := r.store.GetRepositoryByRepoID(ctx, repoID)
 	if errors.Is(err, sql.ErrNoRows) {
 		zerolog.Ctx(ctx).Debug().Err(err).
-			Int64("repositoryUpstreamID", evt.Repository).
+			Int64("repositoryUpstreamID", repoID).
```
JAORMX, line 110: Can we use `any` or something of the sort to print this?

jhrozek: This part of the code is past the repository reconcile and now the code is trying to reconcile artifacts which are supported with github only.

Above is this code:
```
	repoID, err := ewp.Properties.GetProperty(properties.PropertyUpstreamID).AsInt64()
	if err != nil {
		log.Printf("error getting property upstreamID: %v", err)
		// no point in retrying, so we just return nil
		return nil
	}
```
so in the case you called out where the upstream ID is a string we'd hit the error case and shortcut.

I can add a unit test with upstream ID as string and split this long function to repo and artifact parts, would it help?

JAORMX (approved):

Issue #4621, opened 2024-10-02
Remove RefreshRepositoryByUpstreamID from publishProfileInitEvents


jhrozek:
RefreshRepositoryByUpstreamID is legacy code that we want to get rid of. This will allow us to evaluate any entity upon profile creation and get rid of the remains of github code in that module.

PR #4634, opened 2024-10-03
Use generic refresh and message forwarding to evaluator for profile init

jhrozek:
# Summary

- **Use generic message forwarding to initialize the profile instead of entity-specific database searches** - Because we were bound to entity-specific protobuf types and didn't have a good way of iterating over all entities, we used to enumerate all repos and then all artifacts just to marshall them and send off to the evaluator. With generic entities we can simply enumerate all entities and send them off for refresh and forwarding to the evaluator.

Fixes: #4621

## Change Type

# Testing

unit test + manually

# Review Checklist:

### DIFF
--- a/internal/reconcilers/run_profile.go
+++ b/internal/reconcilers/run_profile.go
@@ -16,7 +16,6 @@ package reconcilers
 
 import (
 	"context"
-	"database/sql"
 	"encoding/json"
 	"fmt"
 
@@ -24,12 +23,9 @@ import (
 	"github.com/go-playground/validator/v10"
 	"github.com/google/uuid"
 	"github.com/rs/zerolog"
-	"github.com/rs/zerolog/log"
 
-	"github.com/stacklok/minder/internal/artifacts"
-	"github.com/stacklok/minder/internal/db"
-	"github.com/stacklok/minder/internal/engine/entities"
-	"github.com/stacklok/minder/internal/repositories"
+	entityMessage "github.com/stacklok/minder/internal/entities/handlers/message"
+	"github.com/stacklok/minder/internal/events"
 )
 
 // ProfileInitEvent is an event that is sent to the reconciler topic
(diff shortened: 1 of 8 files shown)


internal/reconcilers/run_profile.go
```diff
@@ -92,88 +88,30 @@ func (r *Reconciler) publishProfileInitEvents(
 	ctx context.Context,
 	projectID uuid.UUID,
 ) error {
-	dbrepos, err := r.store.ListRegisteredRepositoriesByProjectIDAndProvider(ctx,
-		db.ListRegisteredRepositoriesByProjectIDAndProviderParams{
-			Provider:  sql.NullString{Valid: false},
-			ProjectID: projectID,
-		})
+	ents, err := r.store.GetEntitiesByProjectHierarchy(ctx, []uuid.UUID{projectID})
```
jhrozek, line 91: 2 questions:
1) Is it OK to list all entities? We used to only list repos and artifacts, this would also include pulls. I don't see why we shouldn't evaluate pulls that are already tracked in our database, but it is a change in behaviour
2) the old code doesn't seem to take hierarchy into account, in other words it would only trigger evaluation of entities in that project only, but not subprojects. Is that intended?

JAORMX: 1. Makes sense and I think it's ok.
2. That was an issue tracked somewhere... thanks for fixing it!

jhrozek: about 2 - I haven't actually searched the child projects to get the hierarchy, I just prepped the code to accept a hierarchy but pass a single-item slice. I'll push another patch to this PR for easier review.

JAORMX: I think not doing the child hierarchy in this PR is the right thing to do. It is a TODO item and we should address it, but let's keep this PR small.

JAORMX (approved):

PR #4651, opened 2024-10-05
Add additional test coverage for helper functions in Util

### LINKED ISSUE #4380: Improve test coverage in parts of Minder with low coverage.
Certain parts of Minder have low unit test coverage. In at least one recent incident, the incident could have been prevented if there had been some unit test coverage of the codepath in question (there was none).

Based on some recent pain points, and looking at stats from Coveralls, some of the files we should focus attention our attention on are:

1. [internal/providers/github/common.go](https://coveralls.io/builds/69632370/source?filename=internal%2Fproviders%2Fgithub%2Fcommon.go)
2. [internal/controlplane/handlers_profile.go](https://coveralls.io/builds/69632370/source?filename=internal%2Fcontrolplane%2Fhandlers_profile.go)
3. [internal/controlplane/handlers_evalstatus.go](https://coveralls.io/builds/69632370/source?filename=internal%2Fcontrolplane%2Fhandlers_evalstatus.go)
4. [internal/verifier/sigstore/container/container.go](https://coveralls.io/builds/69632370/source?filename=internal%2Fverifier%2Fsigstore%2Fcontainer%2Fcontainer.go)
5. [internal/util/helpers.go](https://coveralls.io/builds/69632370/source?filename=internal%2Futil%2Fhelpers.go)
6. [internal/controlplane/handlers_projects.go](https://coveralls.io/builds/69632370/source?filename=internal%2Fcontrolplane%2Fhandlers_projects.go)
7. [internal/controlplane/handlers_user.go](https://coveralls.io/builds/69632370/source?filename=internal%2Fcontrolplane%2Fhandlers_user.go)
8. [internal/controlplane/handlers_oauth.go](https://coveralls.io/builds/69632370/source?filename=internal%2Fcontrolplane%2Fhandlers_oauth.go)
9. [...tions/alert/security_advisory/security_advisory.go](https://coveralls.io/builds/69632370/source?filename=internal%2Fengine%2Factions%2Falert%2Fsecurity_advisory%2Fsecurity_advisory.go)
10. [internal/controlplane/handlers_artifacts.go](https://coveralls.io/builds/69632370/source?filename=internal%2Fcontrolplane%2Fhandlers_artifacts.go)

gajananan:
# Summary

Currently helper functions in Util lacks test coverage

This commit adds test coverage helper functions in Util

Ref #4380

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/util/helpers_test.go
+++ b/internal/util/helpers_test.go
@@ -21,13 +21,14 @@ import (
 	_ "github.com/lib/pq"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/mindersec/minder/internal/util"
 )
 	"google.golang.org/protobuf/proto"
+
+	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
 )
 
 // TestGetConfigDirPath tests the GetConfigDirPath function
@@ -410,9 +411,9 @@ func TestLoadCredentials(t *testing.T) {
 }
 
 // TestRevokeToken tests the RevokeToken function
-func TestRevokeToken(t *testing.T) {
+// func TestRevokeToken(t *testing.T) {
 
-}
+// }
 
 // TestGetJsonFromProto tests the GetJsonFromProto function
 func TestGetJsonFromProto(t *testing.T) {
--- a/internal/util/helpers_test.go
+++ b/internal/util/helpers_test.go
@@ -39,12 +39,14 @@ func setEnvVar(t *testing.T, env string, value string) {
 	envLock.Lock()
 	t.Cleanup(envLock.Unlock)
 
-	originalEnvToken := os.Getenv(env)
+	originalEnvVal := os.Getenv(env)
 	err := os.Setenv(env, value)
 	if err != nil {
 		t.Errorf("error setting %v: %v", env, err)
 	}
-	defer os.Setenv(env, originalEnvToken)
+
+	t.Cleanup(func() { _ = os.Setenv(env, originalEnvVal) })
+
 }
 
 // TestGetConfigDirPath tests the GetConfigDirPath function
@@ -163,9 +165,7 @@ func TestGetGrpcConnection(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-
 			setEnvVar(t, util.MinderAuthTokenEnvVar, tt.envToken)
-
 			conn, err := util.GetGrpcConnection(tt.grpcHost, tt.grpcPort, tt.allowInsecure, tt.issuerUrl, tt.clientId)
 			if (err != nil) != tt.expectedError {
 				t.Errorf("expected error: %v, got: %v", tt.expectedError, err)
--- a/internal/util/helpers_test.go
+++ b/internal/util/helpers_test.go
@@ -36,9 +36,9 @@ var (
 
 	// envLockXdgConfig is a mutex to ensure that all the tests that run os.SetEnv("MINDER_AUTH_TOKEN") need to be prevented from running at the same time as each other.
 	envLockMinderAuthToken = &sync.Mutex{}
-
+	//nolint:gosec // This is not a hardcoded credential
 	XdgConfigHomeEnvVar = "XDG_CONFIG_HOME"
-
+	//nolint:gosec // This is not a hardcoded credential
 	MinderAuthTokenEnvVar = "MINDER_AUTH_TOKEN"
 )
 
--- a/internal/util/helpers_test.go
+++ b/internal/util/helpers_test.go
@@ -411,9 +411,78 @@ func TestLoadCredentials(t *testing.T) {
 }
 
 // TestRevokeToken tests the RevokeToken function
-// func TestRevokeToken(t *testing.T) {
+func TestRevokeToken(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name         string
+		token        string
+		issuerUrl    string
+		clientId     string
+		tokenHint    string
+		expectedPath string
+		expectError  bool
+	}{
+		{
+			name:         "Valid token revocation",
+			token:        "test-token",
+			issuerUrl:    "http://localhost:8081",
+			clientId:     "minder-cli",
+			tokenHint:    "refresh_token",
+			expectedPath: "/realms/stacklok/protocol/openid-connect/revoke",
+			expectError:  false,
+		},
+		{
+			name:         "Invalid issuer URL",
+			token:        "test-token",
+			issuerUrl:    "://invalid-url",
+			clientId:     "minder-cli",
+			tokenHint:    "refresh_token",
+			expectedPath: "",
+			expectError:  true,
+		},
+	}
 
-// }
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			var server *httptest.Server
+			if tt.name != "Invalid issuer URL" {
+				server = httptest.NewServer(http.HandlerFunc(func(_ http.ResponseWriter, r *http.Request) {
+					if r.URL.Path != tt.expectedPath {
+						t.Errorf("expected path %s, got %s", tt.expectedPath, r.URL.Path)
+					}
+
+					if err := r.ParseForm(); err != nil {
+						t.Fatalf("error parsing form: %v", err)
+					}
+
+					if r.Form.Get("client_id") != tt.clientId {
+						t.Errorf("expected client_id %s, got %s", tt.clientId, r.Form.Get("client_id"))
+					}
+
+					if r.Form.Get("token") != tt.token {
+						t.Errorf("expected token %s, got %s", tt.token, r.Form.Get("token"))
+					}
+
+					if r.Form.Get("token_type_hint") != tt.tokenHint {
+						t.Errorf("expected token_type_hint %s, got %s", tt.tokenHint, r.Form.Get("token_type_hint"))
+					}
+				}))
+				defer server.Close()
+			}
+
+			issuerUrl := tt.issuerUrl
+			if tt.name != "Invalid issuer URL" {
+				issuerUrl = server.URL
+			}
+
+			err := RevokeToken(tt.token, issuerUrl, tt.clientId, tt.tokenHint)
+			if (err != nil) != tt.expectError {
+				t.Errorf("RevokeToken() error = %v, expectError %v", err, tt.expectError)
+			}
+		})
+	}
+}
 
 // TestGetJsonFromProto tests the GetJsonFromProto function
 func TestGetJsonFromProto(t *testing.T) {
--- a/internal/util/helpers_test.go
+++ b/internal/util/helpers_test.go
@@ -12,9 +12,9 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -23,17 +23,16 @@ import (
 	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-
-	"github.com/mindersec/minder/internal/util"
-)
 	"google.golang.org/protobuf/proto"
 
-	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
+	"github.com/mindersec/minder/internal/util"
+	pb "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
 )
 
 // TestGetConfigDirPath tests the GetConfigDirPath function
 func TestGetConfigDirPath(t *testing.T) {
 	t.Parallel()
+	var mu sync.Mutex
 	tests := []struct {
 		name           string
 		envVar         string
@@ -57,11 +56,13 @@ func TestGetConfigDirPath(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
+			mu.Lock()
+			defer mu.Unlock()
 			err := os.Setenv("XDG_CONFIG_HOME", tt.envVar)
 			if err != nil {
 				t.Errorf("error setting XDG_CONFIG_HOME: %v", err)
 			}
-			path, err := GetConfigDirPath()
+			path, err := util.GetConfigDirPath()
 			if (err != nil) != tt.expectingError {
 				t.Errorf("expected error: %v, got: %v", tt.expectingError, err)
 			}
@@ -150,11 +151,12 @@ func TestGetGrpcConnection(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			err := os.Setenv(MinderAuthTokenEnvVar, tt.envToken)
+			originalEnvToken := os.Getenv(util.MinderAuthTokenEnvVar)
+			err := os.Setenv(util.MinderAuthTokenEnvVar, tt.envToken)
 			if err != nil {
 				t.Errorf("error setting MinderAuthTokenEnvVar: %v", err)
 			}
-			conn, err := GetGrpcConnection(tt.grpcHost, tt.grpcPort, tt.allowInsecure, tt.issuerUrl, tt.clientId)
+			conn, err := util.GetGrpcConnection(tt.grpcHost, tt.grpcPort, tt.allowInsecure, tt.issuerUrl, tt.clientId)
 			if (err != nil) != tt.expectedError {
 				t.Errorf("expected error: %v, got: %v", tt.expectedError, err)
 			}
@@ -164,6 +166,7 @@ func TestGetGrpcConnection(t *testing.T) {
 					t.Errorf("error closing connection: %v", err)
 				}
 			}
+			defer os.Setenv(util.MinderAuthTokenEnvVar, originalEnvToken)
 		})
 	}
 }
@@ -171,13 +174,13 @@ func TestGetGrpcConnection(t *testing.T) {
 // TestSaveCredentials tests the SaveCredentials function
 func TestSaveCredentials(t *testing.T) {
 	t.Parallel()
-	tokens := OpenIdCredentials{
+	tokens := util.OpenIdCredentials{
 		AccessToken:          "test_access_token",
 		RefreshToken:         "test_refresh_token",
 		AccessTokenExpiresAt: time.Time{}.Add(7 * 24 * time.Hour),
 	}
 
-	cfgPath, err := GetConfigDirPath()
+	cfgPath, err := util.GetConfigDirPath()
 
 	if err != nil {
 		t.Fatalf("error getting config path: %v", err)
@@ -185,7 +188,7 @@ func TestSaveCredentials(t *testing.T) {
 
 	expectedFilePath := filepath.Join(cfgPath, "credentials.json")
 
-	filePath, err := SaveCredentials(tokens)
+	filePath, err := util.SaveCredentials(tokens)
 	if err != nil {
 		t.Fatalf("expected no error, got %v", err)
 	}
@@ -220,19 +223,21 @@ func TestSaveCredentials(t *testing.T) {
 // TestRemoveCredentials tests the RemoveCredentials function
 func TestRemoveCredentials(t *testing.T) {
 	t.Parallel()
-	xdgConfigHome := os.Getenv("XDG_CONFIG_HOME")
-	if xdgConfigHome == "" {
-		homeDir, err := os.UserHomeDir()
-		if err != nil {
-			t.Fatalf("error getting home directory: %v", err)
-		}
-		xdgConfigHome = filepath.Join(homeDir, ".config")
+	// Create a temporary directory
+	testDir := t.TempDir()
+
+	err := os.Setenv("XDG_CONFIG_HOME", testDir)
+	if err != nil {
+		t.Errorf("error setting XDG_CONFIG_HOME: %v", err)
 	}
+
+	xdgConfigHome := os.Getenv("XDG_CONFIG_HOME")
 
 	filePath := filepath.Join(xdgConfigHome, "minder", "credentials.json")
 
 	// Create a dummy credentials file
-	err := os.MkdirAll(filepath.Dir(filePath), 0750)
+	err = os.MkdirAll(filepath.Dir(filePath), 0750)
+
 	if err != nil {
 		t.Fatalf("error creating directory: %v", err)
 	}
@@ -242,7 +247,7 @@ func TestRemoveCredentials(t *testing.T) {
 		t.Fatalf("error writing credentials to file: %v", err)
 	}
 
-	err = RemoveCredentials()
+	err = util.RemoveCredentials()
 	if err != nil {
 		t.Fatalf("expected no error, got %v", err)
 	}
@@ -263,7 +268,7 @@ func TestRefreshCredentials(t *testing.T) {
 		clientId       string
 		responseBody   string
 		expectedError  string
-		expectedResult OpenIdCredentials
+		expectedResult util.OpenIdCredentials
 	}{
 		{
 			name:         "Successful refresh with local server",
@@ -271,7 +276,7 @@ func TestRefreshCredentials(t *testing.T) {
 			issuerUrl:    "http://localhost:8081",
 			clientId:     "minder-cli",
 			responseBody: `{"access_token":"new_access_token","refresh_token":"new_refresh_token","expires_in":3600}`,
-			expectedResult: OpenIdCredentials{
+			expectedResult: util.OpenIdCredentials{
 				AccessToken:          "new_access_token",
 				RefreshToken:         "new_refresh_token",
 				AccessTokenExpiresAt: time.Now().Add(3600 * time.Second),
@@ -314,7 +319,7 @@ func TestRefreshCredentials(t *testing.T) {
 			parsedURL, _ := url.Parse(server.URL)
 			tt.issuerUrl = parsedURL.String()
 
-			result, err := RefreshCredentials(tt.refreshToken, tt.issuerUrl, tt.clientId)
+			result, err := util.RefreshCredentials(tt.refreshToken, tt.issuerUrl, tt.clientId)
 			if tt.expectedError != "" {
 				if err == nil || err.Error() != tt.expectedError {
 					t.Errorf("expected error %v, got %v", tt.expectedError, err)
@@ -338,12 +343,12 @@ func TestLoadCredentials(t *testing.T) {
 		name           string
 		fileContent    string
 		expectedError  string
-		expectedResult OpenIdCredentials
+		expectedResult util.OpenIdCredentials
 	}{
 		{
 			name:        "Successful load",
 			fileContent: `{"access_token":"access_token","refresh_token":"refresh_token","expiry":"2024-10-05T17:46:27+10:00"}`,
-			expectedResult: OpenIdCredentials{
+			expectedResult: util.OpenIdCredentials{
 				AccessToken:          "access_token",
 				RefreshToken:         "refresh_token",
 				AccessTokenExpiresAt: time.Date(2024, 10, 5, 17, 46, 27, 0, time.FixedZone("AEST", 10*60*60)),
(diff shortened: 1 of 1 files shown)


internal/util/helpers_test.go
```diff
@@ -13,21 +13,765 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			// Create a unique directory for each test
+			testDir := fmt.Sprintf("testdir_%s", sanitizeTestName(tt.name))
```
JAORMX, line 699: You could probably leverage golang's utilities instead: https://pkg.go.dev/testing#B.TempDir

This will create a temporary directory for tests and even clean up afterwards.

wdyt?

internal/util/helpers_test.go
```diff
@@ -13,21 +13,765 @@
...
+
+	for _, tc := range testCases {
+		tc := tc
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			// Create a temporary file for testing
+			if tc.filePath == "testfile.txt" {
+				err := os.WriteFile(tc.filePath, []byte(tc.expectedDesc), 0600)
```
evankanderson, line 626: We should do this in `t.TestDir()`

internal/util/helpers_test.go
gajananan: thank you. I agree. that is a clean approach.

internal/util/helpers_test.go
gajananan: Thank you. I assume, you were referring to `t.TempDir()`

evankanderson: Yes, thanks!

internal/util/helpers_test.go
```diff
@@ -13,21 +13,765 @@
...
+	xdgConfigHome := os.Getenv("XDG_CONFIG_HOME")
+	if xdgConfigHome == "" {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			t.Fatalf("error getting home directory: %v", err)
+		}
+		xdgConfigHome = filepath.Join(homeDir, ".config")
+	}
+
+	filePath := filepath.Join(xdgConfigHome, "minder", "credentials.json")
```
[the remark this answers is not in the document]
gajananan: Thank you. it is good to learn about the impact to other tests.

evankanderson:
@gajananan

It looks like this is close to the finish line -- any chance you can incorporate the feedback (and fix conflicts) in the next week?

evankanderson:
Hi @gajananan,

We'd love to get this merged!  If you don't have time to work on this, we'll probably have someone over here at Stacklok apply the requested changes and merge it.

evankanderson:
Hey @gajananan -- we'd love to get this PR in.  I ran into one of your co-workers at KubeCon, and they said that your work schedule might be preventing you from updating.  A quick note of "I can get to this within a week" or "please carry this forward, I'm busy now" would be handy.

Thanks again for your work so far!

gajananan:
> Hey @gajananan -- we'd love to get this PR in. I ran into one of your co-workers at KubeCon, and they said that your work schedule might be preventing you from updating. A quick note of "I can get to this within a week" or "please carry this forward, I'm busy now" would be handy.
> 
> Thanks again for your work so far!

My sincere apologies for long silence on this.  I would take it forward with addressing the comments/feedback

internal/util/helpers_test.go
```diff
@@ -13,21 +13,765 @@
...
+			envVar:         "",
+			expectedPath:   filepath.Join(os.Getenv("HOME"), ".config", "minder"),
+			expectingError: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			err := os.Setenv("XDG_CONFIG_HOME", tt.envVar)
```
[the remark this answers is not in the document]
gajananan: I will add explicit locks.

internal/util/helpers_test.go
```diff
@@ -13,21 +13,765 @@
...
+			clientId:      "minder-cli",
+			envToken:      "",
+			expectedError: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			err := os.Setenv(MinderAuthTokenEnvVar, tt.envToken)
```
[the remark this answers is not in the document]
gajananan: I will do reset the env at the end of test run.

internal/util/helpers_test.go
```diff
@@ -13,21 +13,765 @@
...
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				fmt.Fprintln(w, tt.responseBody)
+			}))
+			defer server.Close()
+
+			parsedURL, _ := url.Parse(server.URL)
+			tt.issuerUrl = parsedURL.String()
+
+			result, err := RefreshCredentials(tt.refreshToken, tt.issuerUrl, tt.clientId)
```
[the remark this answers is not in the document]
gajananan: @evankanderson , could you confirm we still need to do this after my latest change?

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+		},
+		{
+			name:          "Error reading credentials file",
+			fileContent:   "",
+			expectedError: "error reading credentials file",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt // capture range variable
```
evankanderson, line 370: I think this is no longer needed [since Go 1.22](https://forum.golangbridge.org/t/go-1-22-changed-for-loop-variable-semantics/34464).

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			// Create a unique temporary directory for the test
+			// tempDir, err := os.MkdirTemp("", "test_load_credentials_"+tt.name)
+			// if err != nil {
+			// 	t.Fatalf("failed to create temp directory: %v", err)
+			// }
+			// defer os.RemoveAll(tempDir)
```
evankanderson, line 378: Why is this commented out?

internal/util/helpers_test.go
```diff
@@ -13,21 +13,765 @@
...
+				// Create a temporary file with the specified content
+				err := os.WriteFile(filePath, []byte(tt.fileContent), 0600)
+				if err != nil {
+					t.Fatalf("failed to write test file: %v", err)
+				}
+			}
+
+			// Temporarily override the environment variable for the test
+			originalEnv := os.Getenv("XDG_CONFIG_HOME")
+			err = os.Setenv("XDG_CONFIG_HOME", tempDir)
```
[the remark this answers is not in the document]
evankanderson: This still doesn't have a `defer` or switched to `t.Setenv`.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+			} else {
+				if err != nil {
+					t.Errorf("expected no error, got %v", err)
+				}
+				if result.AccessToken != tt.expectedResult.AccessToken || result.RefreshToken != tt.expectedResult.RefreshToken || !result.AccessTokenExpiresAt.Equal(tt.expectedResult.AccessTokenExpiresAt) {
+					t.Errorf("expected result %v, got %v", tt.expectedResult, result)
+				}
+			}
+			defer os.Setenv("XDG_CONFIG_HOME", originalEnv)
+			defer os.RemoveAll(tempDir)
```
evankanderson, line 425: If you're using `defer` (instead of `t.Cleanup`), you should put it next to the resource it creates. By putting these at the end, you'll only perform the cleanup if the test doesn't exit early (e.g. from `t.Error`)

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+			expectedPath: "",
+			expectError:  true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			var server *httptest.Server
+			if tt.name != "Invalid issuer URL" {
```
evankanderson, line 467: Rather than using the test name, why not have a `func(t *testing.T) *httptest.Server` in the test cases which constructs the server?

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+				server = httptest.NewServer(http.HandlerFunc(func(_ http.ResponseWriter, r *http.Request) {
+					if r.URL.Path != tt.expectedPath {
+						t.Errorf("expected path %s, got %s", tt.expectedPath, r.URL.Path)
+					}
```
evankanderson, line 471: We use [require](https://pkg.go.dev/github.com/stretchr/testify/require) in other parts of the codebase to write this as:

```golang
require.Equal(t, tt.expectedPath, r.URL.Path)
```

You're not required to do so, but it (and `NoError`) can be helpful sometimes.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+
+					if r.Form.Get("token") != tt.token {
+						t.Errorf("expected token %s, got %s", tt.token, r.Form.Get("token"))
+					}
+
+					if r.Form.Get("token_type_hint") != tt.tokenHint {
+						t.Errorf("expected token_type_hint %s, got %s", tt.tokenHint, r.Form.Get("token_type_hint"))
+					}
+				}))
+				defer server.Close()
```
evankanderson, line 489: You can't put this inside a constructor function if you take that approach.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+			}
+
+			issuerUrl := tt.issuerUrl
+			if tt.name != "Invalid issuer URL" {
+				issuerUrl = server.URL
+			}
```
evankanderson, line 495: Since most of the body of this test is conditional on the test name, I wonder if using two tests (rather than a data-driven test) would simplify things here.

internal/util/helpers_test.go
evankanderson: Alternatively, you could create `testfile.txt` once outside the `t.Run()`, since it shouldn't need to vary across the test cases.

evankanderson: (You probably still want to put it in `t.TempDir()`, though.)

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+			desc, closer, err := util.OpenFileArg(tc.filePath, tc.dashOpen)
+			if closer != nil {
+				defer closer()
+			}
+
+			if tc.expectError {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				if tc.filePath == "-" || tc.filePath != "-" {
```
evankanderson, line 649: Doesn't this `if` cover all cases?

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+	if len(got) != len(expected) {
+		return false
+	}
+	for i := range got {
+		if got[i].Path != expected[i].Path || got[i].Expanded != expected[i].Expanded {
+			return false
+		}
+	}
+	return true
+}
```
evankanderson, line 740: Consider using [`slices.Equal`](https://pkg.go.dev/slices#Equal), or `slices.EqualFunc` with an anonymous function if you need to.

evankanderson (commented):
Thanks for getting back to this.  Some of the comments are nit-picky, and you can add a comment saying "just trying to get this in", but since I thought part of your goal was learning the Go language, I figured I'd provide more extensive comments and pointers.

internal/util/helpers_test.go
gajananan: I will use slices.EqualFunc with an anonymous function

internal/util/helpers_test.go
gajananan: The condition if tc.expectError { assert.Error(t, err) } covers only the test cases where an error is expected.

The test cases with no expected error ( there are two)  are covered by the subsequent else block, which checks that no error occurred and verifies the file content.

gajananan: Else condition covers the case:
	    {
			name:         "Dash as file path",
			filePath:     "-",
			dashOpen:     strings.NewReader("dash input"),
			expectedDesc: "dash input",
			expectError:  false,
		}

internal/util/helpers_test.go
gajananan: It should be removed.  I will do that.

internal/util/helpers_test.go
gajananan: Yes, no longer need.  I removed it.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+	}
+}
+
+// TestRemoveCredentials tests the RemoveCredentials function
+func TestRemoveCredentials(t *testing.T) {
+	t.Parallel()
+	// Create a temporary directory
+	testDir := t.TempDir()
+
+	err := os.Setenv("XDG_CONFIG_HOME", testDir)
```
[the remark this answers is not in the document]
gajananan: I will reset once the test is completed.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,767 @@
...
+			if (err != nil) != tt.expectedError {
+				t.Errorf("expected error: %v, got: %v", tt.expectedError, err)
+			}
+			if conn != nil {
+				err = conn.Close()
+				if err != nil {
+					t.Errorf("error closing connection: %v", err)
+				}
+			}
+			defer os.Setenv(util.MinderAuthTokenEnvVar, originalEnvToken)
```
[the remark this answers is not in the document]
gajananan: I moved this deferred call up next to the os.Setenv call.

internal/util/helpers_test.go
gajananan: I moved the deferred call as suggested.

internal/util/helpers_test.go
gajananan: I modified to have a func(t *testing.T) *httptest.Server in the test cases which constructs the server.

internal/util/helpers_test.go
gajananan: That is good to learn. I changed accordingly.

internal/util/helpers_test.go
gajananan: I addressed this as I have changed the approach with test server creation.

internal/util/helpers_test.go
gajananan: "Alternatively, you could create testfile.txt once outside the t.Run()"

 testfile.txt  is required only in one case, so it is created only for that case.

evankanderson: Sorry, the comment here is that this line says:

```suggestion
				path := tc.filePath
				if path == "-" || path != "-" {
```

It should always be true that a string is either equal to a constant, or not equal to the same constant.

evankanderson: > "Alternatively, you could create testfile.txt once outside the t.Run()"
> 
> testfile.txt is required only in one case, so it is created only for that case.

(This was from a different comment.)

While it's true that `testfile.txt` is only needed in one case, it doesn't hurt the other cases, and could end up simplifying the setup logic.  On the other hand, I don't feel strongly about it being one place or the other.

gajananan: > Sorry, the comment here is that this line says:
> 
> It should always be true that a string is either equal to a constant, or not equal to the same constant.
I got it, I will fix it.

internal/util/helpers_test.go
gajananan: I significantly  changed the test function, I guess this comment has been addressed with latest change.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,758 @@
...
+	"google.golang.org/protobuf/proto"
 
 	"github.com/mindersec/minder/internal/util"
+	pb "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
 )
 
+// TestGetConfigDirPath tests the GetConfigDirPath function
+func TestGetConfigDirPath(t *testing.T) {
+	t.Parallel()
+	var envLock sync.Mutex
```
evankanderson, line 36: I think you want `envLock` to be at the file level, e.g. on line 32 before this function, because you have multiple tests all trying to set the environment, and you don't want any of them conflicting with each other.

internal/util/helpers_test.go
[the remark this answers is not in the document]
evankanderson: (This still needs to clean up after the `os.Setenv`)

internal/util/helpers_test.go
```diff
@@ -4,18 +4,758 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			originalEnvToken := os.Getenv(util.MinderAuthTokenEnvVar)
+			err := os.Setenv(util.MinderAuthTokenEnvVar, tt.envToken)
+			// reset this the environment variable when complete.
+			defer os.Setenv(util.MinderAuthTokenEnvVar, originalEnvToken)
```
evankanderson, line 158: You'll want to protect this with a mutex as well.  Given the likely test execution time, I'd probably be lazy and re-use `envLock` from `TestGetConfigDirPath`,  but you could also use separate locks for each environment variable if you're feeling particularly ambitious (i wouldn't recommend it, though).

internal/util/helpers_test.go
```diff
@@ -4,18 +4,758 @@
...
+		})
+	}
+}
+
+// TestSaveCredentials tests the SaveCredentials function
+func TestSaveCredentials(t *testing.T) {
+	t.Parallel()
+	var envLock sync.Mutex
+	envLock.Lock()
+	defer envLock.Unlock()
```
evankanderson, line 182: Declaring a new `sync.Mutex` here and then acquiring it doesn't do much, because the mutex isn't shared across multiple goroutines (e.g. `t.Run()` spawns a new goroutine, but this doesn't call `t.Run`).

internal/util/helpers_test.go
```diff
@@ -4,18 +4,758 @@
...
+			expectedError: "error reading credentials file",
+		},
+	}
+
+	for _, tt := range tests {
+
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			tempDir := t.TempDir()
+			defer os.RemoveAll(tempDir)
```
evankanderson, line 383: You shouldn't need this if you use [`t.TempDir()`](https://pkg.go.dev/testing#T.TempDir):

> The directory is automatically removed when the test and all its subtests complete.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,758 @@
...
+			}
+
+			// Temporarily override the environment variable for the test
+			originalEnv := os.Getenv("XDG_CONFIG_HOME")
+			err = os.Setenv("XDG_CONFIG_HOME", tempDir)
+			if err != nil {
+				t.Errorf("error setting XDG_CONFIG_HOME: %v", err)
+			}
+			// reset this the environment variable when complete.
+			defer os.Setenv("XDG_CONFIG_HOME", originalEnv)
```
evankanderson, line 413: You'll want to grab `envLock` here for the rest of the test, too.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,758 @@
...
+	token        string
+	issuerUrl    string
+	clientId     string
+	tokenHint    string
+	expectedPath string
+	expectError  bool
+	createServer func(t *testing.T, tt TestCase) *httptest.Server
+}
+
+func createTestServer(t *testing.T, tt TestCase) *httptest.Server {
```
evankanderson, line 445: golangci-lint likes non-top-level test functions that take a `t *testing.T` to call `t.Helper()` so that the tracing of errors from `t.Errorf()` etc gets a cleaner stacktrace.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,758 @@
...
+		var tempFilePath string
+
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			// Create a temporary file for testing
+			if tc.filePath == "testfile.txt" {
+				tempFilePath = filepath.Join(testDir, tc.filePath)
+				err := os.WriteFile(tempFilePath, []byte(tc.expectedDesc), 0600)
+				assert.NoError(t, err)
+				defer os.Remove(tempFilePath)
```
evankanderson, line 636: Since you're putting `tempFilePath` in a directory from `t.TempDir`, you shouldn't need to delete it yourself later; `t.TempDir` will delete the directory and its contents at the end of the test.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,758 @@
...
+			} else {
+				// When handling the dash (-) as a file path, it should read from the provided io.Reader (tc.dashOpen) instead of trying to open a file
+				tempFilePath = tc.filePath
+			}
```
evankanderson, line 640: If you create `testfile.txt` outside the test case code, you can avoid needing as much conditional logic in the test.  Compare:

```golang
func TestOpenFileArg(t *testing.T) {
	t.Parallel()
	testDir := t.TempDir()
	// Create one test file for the file input case
	tempFilePath := filepath.Join(testDir, "testfile.txt")
	assert.NoError(os.WriteFile(tempFilePath, []byte("test content"), 0600))

	...
		filePath: tempFilePath
```

(vs the current 11 lines with `if`/`else`)

internal/util/helpers_test.go
gajananan: Fixed now.

internal/util/helpers_test.go
gajananan: Removed

internal/util/helpers_test.go
```diff
@@ -4,18 +4,765 @@
...
+		t.Errorf("expected file path %v, got %v", expectedFilePath, filePath)
+	}
+
+	// Verify the file content
+	credsJSON, err := json.Marshal(tokens)
+	if err != nil {
+		t.Fatalf("error marshaling credentials: %v", err)
+	}
+
+	fpath := filepath.Clean(filePath)
```
evankanderson, line 223: Why do you need to run `filepath.Clean` here?

internal/util/helpers_test.go
```diff
@@ -4,18 +4,765 @@
...
+	SetEnvVar(t, "XDG_CONFIG_HOME", testDir)
+
+	cfgPath := filepath.Join(testDir, "minder")
+
+	expectedFilePath := filepath.Join(cfgPath, "credentials.json")
+
+	filePath, err := util.SaveCredentials(tokens)
+	if err != nil {
+		t.Fatalf("expected no error, got %v", err)
+	}
```
evankanderson, line 211: Slight preference for `require.NoError(t, err)` here and below, to break up the common-case flow of the test a bit less.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,765 @@
...
+
+	if string(content) != string(credsJSON) {
+		t.Errorf("expected file content %v, got %v", string(credsJSON), string(content))
+	}
+
+	// Clean up
+	err = os.Remove(filePath)
+	if err != nil {
+		t.Fatalf("error removing file: %v", err)
+	}
```
evankanderson, line 237: You don't need this since your files are in `t.TempDir()`; the test will do it for you automatically.

internal/util/helpers_test.go
gajananan: t.Helper() is added to createTestServer().

internal/util/helpers_test.go
```diff
@@ -4,18 +4,765 @@
...
+			t.Parallel()
+
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				fmt.Fprintln(w, tt.responseBody)
+			}))
+			defer server.Close()
+
+			parsedURL, _ := url.Parse(server.URL)
+			tt.issuerUrl = parsedURL.String()
```
evankanderson, line 339: Why do this, rather than simply:

```
tt.issuerUrl = server.URL
```

(or simply use `server.URL` below, and don't make the `issuerUrl` a field in the testcase struct, since it is never used...)

internal/util/helpers_test.go
```diff
@@ -4,18 +4,765 @@
...
+			}
+
+			filePath := filepath.Join(minderDir, "credentials.json")
+
+			if tt.fileContent != "" {
+				// Create a temporary file with the specified content
+				err := os.WriteFile(filePath, []byte(tt.fileContent), 0600)
+				if err != nil {
+					t.Fatalf("failed to write test file: %v", err)
+				}
```
evankanderson, line 409: Again, prefer `require.NoError()`:

```suggestion
				require.NoError(t, os.WriteFile(filePath, []byte(tt.fileContent), 0600))
```

internal/util/helpers_test.go
```diff
@@ -4,18 +4,765 @@
...
+	tests := []TestCase{
+		{
+			name:         "Valid token revocation",
+			token:        "test-token",
+			issuerUrl:    "http://localhost:8081",
+			clientId:     "minder-cli",
+			tokenHint:    "refresh_token",
+			expectedPath: "/realms/stacklok/protocol/openid-connect/revoke",
+			expectError:  false,
+			createServer: createTestServer,
```
evankanderson, line 471: Another option here is to make createTestServer an anonymous function, which avoids needing to declare it or `TestCase` outside this function:

```suggestion
			createServer: func(t *testing.T) *httptest.Server {
				return httptest.NewServer(http.HandlerFunc(func(_ http.ResponseWriter, r *http.Request) {
					require.NoError(t, r.ParseForm())
					require.Equal(t, "minder-cli", r.Form.Get("client_id"))
					require.Equal(t, "test-token", r.Form.Get("token"))
					require.Equal(t, "refresh_token", r.Form.Get("token_type_hint"))
				}))
			},
```

internal/util/helpers_test.go
gajananan: thank you, I fixed it.

evankanderson (approved):
A few more small nits, but this is looking very close!  In particular, I think you want `SetEnvVar` to take care of restoring the environment variable; the other fixes are purely your choice / passing along some experience maintaining Go code.

internal/util/helpers_test.go
gajananan: yes. I can simplify as suggested.

internal/util/helpers_test.go
gajananan: Yes, I removed that.

internal/util/helpers_test.go
gajananan: will do that.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,765 @@
...
 	"github.com/mindersec/minder/internal/util"
+	pb "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
 )
 
+var (
+	// envLock is a mutex to ensure that all the tests that run os.SetEnv("XDG_CONFIG...") need to be prevented from running at the same time as each other.
+	envLock = &sync.Mutex{}
+)
+
+func SetEnvVar(t *testing.T, env string, value string) {
```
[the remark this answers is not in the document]
gajananan: I fix it.

internal/util/helpers_test.go
gajananan: I tested, we would not need it.

gajananan:
> A few more small nits, but this is looking very close! In particular, I think you want `SetEnvVar` to take care of restoring the environment variable; the other fixes are purely your choice / passing along some experience maintaining Go code.

Thank you for your detailed feedback which helped a lot to learn the code base. I addressed the remainder of the comments as well.

internal/util/helpers_test.go
```diff
@@ -4,18 +4,765 @@
...
+	t.Helper() // Keep golangci-lint happy
+	envLock.Lock()
+	t.Cleanup(envLock.Unlock)
+
+	err := os.Setenv(env, value)
+	if err != nil {
+		t.Errorf("error setting %v: %v", env, err)
+	}
```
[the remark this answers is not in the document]
gajananan: added a cleanup to reset the environment variable to the original value

internal/util/helpers_test.go
```diff
@@ -4,18 +4,765 @@
...
+			envToken:      "",
+			expectedError: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			originalEnvToken := os.Getenv(util.MinderAuthTokenEnvVar)
+			err := os.Setenv(util.MinderAuthTokenEnvVar, tt.envToken)
```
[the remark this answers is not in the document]
gajananan: yes,

evankanderson (dismissed):

evankanderson (approved):

evankanderson (approved):

Issue #4619, opened 2024-10-02
Use the generic AddOriginatingEntity handler instead of directly writing to the database in the github webhook handler for PRs


jhrozek:
We write pull request entities to the DB directly in the github webhook handler. This is bespoke code in the webhook handler that is not reusable and at the same time we already have the new AddOriginatingEntity. Let's use that instead.

PR #4663, opened 2024-10-07
Use the AddOriginatingEntity/RemoveOriginatingEntity handlers to handle PRs in the webhook provider

jhrozek:
# Summary

- **Remove DeleteEntityByName** - With multiple providers, we might have entities with the same name.  Let's just use DeleteEntity which deletes by ID, forcing the API user to do a search before.
- **Add tests for DeleteOriginatingEntity** - We didn't have any tests for this codepath before. Also switches to making an explicit search through getEntityInner for the PR and then deleting by ID
- **Use the originating webhook handlers for handling PRs in the github webhook handler** - Instead of bespoke github-specific code, let's use the new AddOriginatingEntity and RemoveOriginatingEntity handlers.

Fixes: #4619

## Change Type

# Testing

manually + unit tests

# Review Checklist:


internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -983,17 +982,8 @@ func (s *Server) processPullRequestEvent(
...
-		HTMLURL: ghRepo.HTMLURL,
-		Private: ghRepo.Private,
-	})
-	if err != nil {
-		return nil, err
-	}
-
 	pullProps, err := properties.NewProperties(map[string]any{
-		properties.PropertyUpstreamID: event.GetPullRequest().GetID(),
+		properties.PropertyUpstreamID: strconv.FormatInt(event.GetPullRequest().GetID(), 10),
```
JAORMX, line 986: nit: you probably want to create a function to format the ID.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -1003,39 +993,48 @@ func (s *Server) processPullRequestEvent(
 		return nil, fmt.Errorf("error creating pull request properties: %w", err)
 	}
 
-	provider, err := s.providerManager.InstantiateFromID(ctx, repoEnt.Entity.ProviderID)
+	repoProps, err := properties.NewProperties(map[string]any{
+		properties.PropertyUpstreamID: strconv.FormatInt(ghRepo.GetID(), 10),
+		properties.PropertyName:       ghRepo.GetName(),
```
JAORMX, line 998: is the name strictly needed? can we do just with the upstream ID?

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -1003,39 +993,48 @@ func (s *Server) processPullRequestEvent(
...
 	}
 
-	prEntWithProps, err := s.reconcilePrWithDb(ctx, provider, repoEnt, pullProps)
-	if errors.Is(err, errNotHandled) {
-		return nil, err
-	} else if err != nil {
-		return nil, fmt.Errorf("error reconciling PR with DB: %w", err)
+	// it is bit of a code smell to use the fetcher here just to format the name
+	fetcher := ghprop.NewPullRequestFetcher()
+	name, err := fetcher.GetName(pullProps)
```
JAORMX, line 1006: can you call this in-line so you don't need to allocate memory?

e.g.

```
name, err := ghprop.PullRequestFetcher{}.GetName(pullProps)
```

or something of the sort

JAORMX (approved):

blkt (approved):

internal/controlplane/handlers_githubwebhooks.go
blkt: Memory for `fetcher` is going to be allocated regardless of the pattern, but this way is more readable.
Note that fields within `PullRequestFetcher` are global variables, so their allocation is already payed for.

jhrozek: done (in a separate patch for easier re-review)

internal/controlplane/handlers_githubwebhooks.go
jhrozek: done in a separate patch for easier re-review

internal/controlplane/handlers_githubwebhooks.go
jhrozek: I added a function using generics and Go's `x/constraints` module. This might also be useful for the similar functions in the GL provider.

This also uncovered a bug in the original properties, luckily the handlers refresh the PRs so we'd paper over the bug now, plus we also pass the name if we can't match the ID.

JAORMX (approved):

jhrozek:
I did run the following 2 tests locally before I merge:
 - `pr_trusty_check_go`
 - `pr_vulnerability_check_go`

jhrozek:
(and they passed)

PR #4673, opened 2024-10-08
Really fix PR origination

JAORMX:
# Summary

There were several bugs in PR origination that were addressed:

* The legacy origination was hardcoded to use GitHub values. This uses
  the per-provider function so we can get the needed info in a general
  way.

* We were using the `RetrieveAllProperties` function to try to persist
  the new entity's properties, while we could have just used the
  `SaveAllProperties` call from the properties service. This actually
  persists as opposed to passing by the more complicated logic.

* We were passing in the incoming properties to persist as opposed to
  the recently retrieved properties. This didn't store the whole view of
  the originated PR, and so we were unable to delete it later on.

Finally, in addition to these changes, we also now don't fail on
deleting and storing legacy entities since newer entities might not need
to do this. Thus, if no legacy entity is created, we instantiate a new
UUID.

## Change Type

# Testing

# Review Checklist:


internal/entities/handlers/strategies/entity/add_originating_entity.go
```diff
@@ -108,10 +123,9 @@ func (a *addOriginatingEntityStrategy) GetEntity(
 		}
 
 		// Persist the properties
-		upstreamProps, err = a.propSvc.RetrieveAllProperties(ctx, prov,
-			parentEwp.Entity.ProjectID, parentEwp.Entity.ProviderID,
-			childProps, entMsg.Entity.Type,
-			propertyService.ReadBuilder().WithStoreOrTransaction(t),
+		err = a.propSvc.SaveAllProperties(ctx, entID,
```
jhrozek, line 126: not related to this change, but to earlier calling `FetchAllProperties` - this means that the message needs different properties depending on which handler we call. If we call the `TopicQueueRefreshEntityAndEvaluate` we pretty much need an upstream ID. If we call `TopicQueueOriginatingEntityAdd` the message must contain properties that are passed to the upstream call, in github's case repo owner, repo name and ID. I don't have a great idea on how to solve this except comments and clear error messages.

I just hate when I try to design an API and the result turns out this way..

internal/entities/handlers/strategies/entity/add_originating_entity.go
```diff
@@ -135,16 +149,21 @@ func (_ *addOriginatingEntityStrategy) GetName() string {
...
 	case minderv1.Entity_ENTITY_PULL_REQUESTS:
+		pr, ok := pbEnt.(*minderv1.PullRequest)
+		if !ok {
+			return uuid.Nil, fmt.Errorf("unexpected proto message type: %T", pbEnt)
+		}
+
 		dbPr, err := t.UpsertPullRequest(ctx, db.UpsertPullRequestParams{
 			RepositoryID: parentEwp.Entity.ID,
-			PrNumber:     childProps.GetProperty(ghprop.PullPropertyNumber).GetInt64(),
+			PrNumber:     pr.Number,
```
jhrozek, line 166: this is much better than leaking the ghprops module but we should eventually either get rid of the number in the PR protobuf message or even better get rid of the protobuf message completely.

jhrozek (approved):

internal/entities/handlers/strategies/entity/add_originating_entity.go
JAORMX: We'll only get here once we get rid of the PR table.

Issue #4682, opened 2024-10-08
Use properties and the new handlers to add/remove repositories in the GH webhook handler


jhrozek:
This is one of the last pieces in the github webhook handler that still doesn't use properties. This will also make use use properties through the entity add/entity delete process, making that codepath available for other providers.

PR #4685, opened 2024-10-08
Simplify the entity_add handlers in preparation for using the NewGetEntityAndDeleteHandler"

jhrozek:
# Summary

As a follow-up to this PR, I want to use the NewGetEntityAndDeleteHandler for handling repo deletions
and in general I want to get rid of calls to `fetchRepo()`. This is a prep PR that makes the
`MinderEvent` messages more generic and let them work with properties.

- **Remove unused setters of repoID attribute for entity reconciliation** - This just simplifies the code
- **Use minderv1.Entity instead of a string and check the entity type for repos** - Using just strings is error prone. We might need to revisit this as we make entities completely pluggable, but we're not there yet and I don't like to add strings comparison. At the same time I want to only run the entity_add handler on repositories.
- **Use generic properties instead of repoName/repoOwner when adding entities** - The handler required the use of repoName/repoOwner that were eventually converted to `properties.PropertyName` before being sent off. Let's just use properties all the way down.

Related: #4682

## Change Type

# Testing

adding/removing repos, registering all repos

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_entities.go
+++ b/internal/controlplane/handlers_entities.go
@@ -22,9 +22,11 @@ import (
 	"github.com/google/uuid"
 	"github.com/rs/zerolog"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/stacklok/minder/internal/db"
 	"github.com/stacklok/minder/internal/engine/engcontext"
+	"github.com/stacklok/minder/internal/entities/properties"
 	"github.com/stacklok/minder/internal/events"
 	"github.com/stacklok/minder/internal/logger"
 	"github.com/stacklok/minder/internal/providers"
@@ -83,7 +85,7 @@ func (s *Server) ReconcileEntityRegistration(
 				continue
 			}
 
-			msg, err := createEntityMessage(ctx, &l, projectID, providerID, repo.Repo.GetName(), repo.Repo.GetOwner())
+			msg, err := createEntityMessage(ctx, &l, projectID, providerID, repo.Entity.GetEntity().GetProperties())
 			if err != nil {
 				l.Error().Err(err).
 					Int64("repoID", repo.Repo.RepoId).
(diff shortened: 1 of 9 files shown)


internal/reconcilers/entity_delete.go
```diff
@@ -47,6 +48,11 @@ func (r *Reconciler) handleEntityDeleteEvent(msg *message.Message) error {
 		return nil
 	}
 
+	if event.EntityType != pb.Entity_ENTITY_REPOSITORIES {
+		l.Error().Str("entity_type", event.EntityType.String()).Msg("entity type not supported")
+		return nil
+	}
```
JAORMX, line 54: Does this only really work for repos?

internal/reconcilers/entity_add.go
```diff
@@ -52,31 +52,25 @@ func (r *Reconciler) handleEntityAddEvent(msg *message.Message) error {
 		return nil
 	}
 
-	var repoOwner string
-	var repoName string
-	var ok bool
-	if repoOwner, ok = event.Entity["repoOwner"].(string); !ok {
-		return errors.New("invalid repo owner")
+	if event.EntityType != pb.Entity_ENTITY_REPOSITORIES {
```
JAORMX, line 55: Does this really only work for repos?

JAORMX (approved):

internal/reconcilers/entity_add.go
jhrozek: At the moment, those handlers go directly to the repository service, see https://github.com/stacklok/minder/blob/d78ff1c866655ef6cb22fb3b6267f2d3060b0ccc/internal/reconcilers/entity_delete.go#L63 or https://github.com/stacklok/minder/blob/d78ff1c866655ef6cb22fb3b6267f2d3060b0ccc/internal/reconcilers/entity_add.go#L91

The type parameter was really redundant as it was not checked anywhere. This change at least made the code fail predictably with another type, previously you'd send data and the code would still try to create a repo out of them..

internal/controlplane/handlers_entities.go
```diff
@@ -116,19 +118,23 @@ func createEntityMessage(
...
+		return nil, err
+	}
+
 	event := messages.NewMinderEvent().
 		WithProjectID(projectID).
 		WithProviderID(providerID).
-		WithEntityType("repository").
-		WithAttribute("repoName", repoName).
-		WithAttribute("repoOwner", repoOwner)
+		WithEntityType(pb.Entity_ENTITY_REPOSITORIES).
```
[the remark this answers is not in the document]
jhrozek: sorry I don't understand, can you elaborate?

internal/controlplane/handlers_entities.go
```diff
@@ -83,7 +85,7 @@ func (s *Server) ReconcileEntityRegistration(
 				continue
 			}
 
-			msg, err := createEntityMessage(ctx, &l, projectID, providerID, repo.Repo.GetName(), repo.Repo.GetOwner())
+			msg, err := createEntityMessage(ctx, &l, projectID, providerID, repo.Entity.GetEntity().GetProperties())
```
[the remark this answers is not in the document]
jhrozek: no, this is not reminder, this is for when you call `minder repo register --all` or its equivalent from the UI.
I agree about making it generic, but one step at a time - I suspect we'll want some sort of auto-registration for OCI registries.

jhrozek:
@JAORMX thanks for the review. I'm going to merge, but if you find time to reply to the comments, I'll be happy to improve on the work in a subsequent PR.

PR #4688, opened 2024-10-08
Add osv-scalibr as a dependency extraction method for PR remediation

evankanderson:
# Summary

Switches from a home-rolled dependency parser to the (in-development) OSV-Scalibr library from Google, which already has parsers for our current set of dependencies.†

† Except that OSV-Scalibr doesn't seem to parse `dependencies` from `package.json`, and only parses `package-lock.json`.  We'd prefer to get _both_ dependency locations, so we can patch both of them.

**I have not documented the `new-dep` PR ingestion type, as I'd like to replace the existing `dep` type (via experiment) over time, but wasn't quite ready to wire that up in the same PR.**

This also moves us to development branches of both `osv-scalibr` and `go-billy` to take advantage of the in-flight code.  `osv-scalibr`'s development branch has a new API that the Google team recommended, and `go-billy` has an adapter from the `billy.Filesystem` type to `io/fs.FS` native Golang (read-only) filesystem.

## Change Type

# Testing

I did a bunch of manual testing with `cmd/dev` on an actual in-flight PR, and then added unit tests to cover the languages we currently support.

# Review Checklist:


evankanderson:
@oliverchang -- I'd appreciate your review of the usage of `github.com/google/osv-scalibr` in this PR.

Thanks!

oliverchang:
@erikvarga, @another-rex are the actual OSV-Scalibr experts -- would you be able to help here?

internal/engine/ingester/diff/diff.go
```diff
@@ -203,6 +208,124 @@ func (di *Diff) ingestFileForDepDiff(
...
+				},
+			})
+		}
+	}
+
+	return &engif.Result{Object: &deps, Checkpoint: checkpoints.NewCheckpointV1Now()}, nil
+}
+
+func inventorySorter(a *extractor.Inventory, b *extractor.Inventory) int {
+	return cmp.Or(cmp.Compare(a.Name, b.Name), cmp.Compare(a.Version, b.Version))
```
erikvarga, line 249: Can you  get packages whose name and version are the same? (e.g. two identical lockfiles scanned from different files)
We have an existing comparison function for Inventory in SCALIBR that looks at all the relevant fields of the inventory: https://github.com/google/osv-scalibr/blob/main/scalibr.go#L293
If you think it'd beneficial to use for you we can make it public.

internal/engine/ingester/diff/diff.go
```diff
@@ -203,6 +208,124 @@ func (di *Diff) ingestFileForDepDiff(
...
+
+func scanFs(ctx context.Context, memFS billy.Filesystem, _ map[string]string) ([]*extractor.Inventory, error) {
+	// have to down-cast here, because scalibr needs multiple io/fs types
+	wrapped, ok := iofs.New(memFS).(scalibr_fs.FS)
+	if !ok {
+		return nil, fmt.Errorf("error converting filesystem to ReadDirFS")
+	}
+
+	desiredCaps := scalibr_plugin.Capabilities{
+		OS:            scalibr_plugin.OSAny,
```
erikvarga, line 273: OSAny is supposed to be used for specifying plugin requirements only. When specifying the scan environment's capabilities you should pick a specific OS.
You can use the SCALIBR-prodivded platform.OS() function to determine the OS at runtime: https://github.com/google/osv-scalibr/blob/main/binary/cli/cli.go#L418

internal/engine/ingester/diff/diff.go
```diff
@@ -203,6 +208,124 @@ func (di *Diff) ingestFileForDepDiff(
...
+		RunningSystem: false,
+	}
+
+	scalibrFs := scalibr_fs.ScanRoot{FS: wrapped}
+	scanConfig := scalibr.ScanConfig{
+		ScanRoots: []*scalibr_fs.ScanRoot{&scalibrFs},
+		// All includes Ruby, Dotnet which we're not ready to test yet, so use the more limited Default set.
+		FilesystemExtractors: list.FilterByCapabilities(list.Default, &desiredCaps),
+		Capabilities: &desiredCaps,
+		Stats: scalibr_stats.NoopCollector{},
```
erikvarga, line 285: This can also be left empty

internal/engine/ingester/diff/diff.go
```diff
@@ -203,6 +208,124 @@ func (di *Diff) ingestFileForDepDiff(
...
+
+	return scanResults.Inventories, nil
+}
+
+func inventoryToEcosystem(inventory *extractor.Inventory) pbinternal.DepEcosystem {
+	if inventory == nil {
+		zerolog.Ctx(context.Background()).Warn().Msg("nil ecosystem scanning diffs")
+		return pbinternal.DepEcosystem_DEP_ECOSYSTEM_UNSPECIFIED
+	}
+	// This should be inventory.GetEcosystem()... but there isn't a convenience wrapper yet
```
erikvarga, line 306: What about inventory.Ecosystem()?

https://github.com/google/osv-scalibr/blob/main/extractor/extractor.go#L86

internal/engine/ingester/diff/diff.go
```diff
@@ -203,6 +208,124 @@ func (di *Diff) ingestFileForDepDiff(
...
+	ecosystem, err := inventory.Extractor.Ecosystem(inventory)
+	if err != nil {
+		zerolog.Ctx(context.Background()).Warn().Err(err).Msg("error getting ecosystem from inventory")
+		return pbinternal.DepEcosystem_DEP_ECOSYSTEM_UNSPECIFIED
+	}
+
+	// Sometimes Scalibr uses the string "PyPI" instead of "pypi" when reporting the ecosystem.
+	switch strings.ToLower(ecosystem) {
```
erikvarga, line 314: If you're interested in the PURL-defined package type you should use converter.ToPURL(inventory).Type or inventory.Extractor.ToPURL(inventory).Type (https://github.com/google/osv-scalibr/blob/main/converter/converter.go#L45)

Ecosystem is specific to the OSV schema and sometimes use different strings than PURL package types.

erikvarga: Hm, we should probably provide an Inventory.PURL() member function similar to Inventory.Ecosystem() so people don't have to import converter.go.

erikvarga: Right, at the moment we don't really have any cases where we'd need to return an error. I'll discuss internally if we ever would want to do that or if we can just remove it.

evankanderson:
> osv-scalibr does have a 'package.json' extractor at: extractor/filesystem/language/javascript/packagejson/extractor.go

Yes, but that extractor does not unpack the `dependencies` or `devDependencies` values (at the moment).

evankanderson:
> > osv-scalibr does have a 'package.json' extractor at: extractor/filesystem/language/javascript/packagejson/extractor.go
> 
> Yes, but that extractor does not unpack the `dependencies` or `devDependencies` values (at the moment).

Similarly, [there is an extractor for `go.mod`](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/golang/gomod/extractor.go), but it does not also cover `go.sum`.  If the goal is to find dependencies, one file location is sufficient, but if the goal is to locate the files which need fixing, it would be nice to have _all_ the dependency-implicated files (using multiple `Location` values in one `Inventory` for the related files).

erikvarga:
> > > osv-scalibr does have a 'package.json' extractor at: extractor/filesystem/language/javascript/packagejson/extractor.go
> > 
> > 
> > Yes, but that extractor does not unpack the `dependencies` or `devDependencies` values (at the moment).

Adding or modifying extractors in SCALIBR is relatively simple (https://github.com/google/osv-scalibr/blob/main/docs/new_extractor.md) so we're happy to merge in PRs for this functionality.

erikvarga (approved):

evankanderson:
BTW, not Scalibr's fault, but it turns out that `github.com/containerd/containerd` depends on `github.com/containerd/errgroup` at `v0.1.0`, and the containerd folks did a bad versioning thing, and released a `v0.3.0` that breaks the interfaces used by `github.com/containerd/containerd` at `v1.7.22`.

internal/engine/ingester/diff/diff.go
```diff
@@ -203,6 +209,124 @@ func (di *Diff) ingestFileForDepDiff(
...
+		for _, filename := range inventory.Locations {
+			deps.Deps = append(deps.Deps, &pbinternal.PrDependencies_ContextualDependency{
+				Dep: &pbinternal.Dependency{
+					Ecosystem: inventoryToEcosystem(inventory),
+					Name:      inventory.Name,
+					Version:   inventory.Version,
+				},
+				File: &pbinternal.PrDependencies_ContextualDependency_FilePatch{
+					Name:     filename,
+					PatchUrl: "", // TODO: do we need this?
```
jhrozek, line 238: Not to diff the deps, but we'd need this in case we wanted to provide the inline reviews. (Which in their current form where we just propose the version bump w/o proposing also any possible transitive deps is...demo quality)

jhrozek (approved):

internal/engine/ingester/diff/diff.go
evankanderson: Yeah, I want to think through what the abstract interface is for proposing patch comments upstream, but I need to take a look at the GitHub PR and GitLab MR comment APIs before I propose something concrete.  (In any case, I think this is a place where we want to be able to leverage the providers so this can be re-used across multiple repo providers.)

internal/engine/ingester/diff/diff.go
erikvarga: > I'll discuss internally if we ever would want to do that or if we can just remove it.

Update: We ended up removing it: https://github.com/google/osv-scalibr/commit/9f86538b542446508e8fab14eff6eaab11c6e913

So if you pin SCALIBR to after this commit you should be able to get rid of the error handling in your code too.

Issue #4703, opened 2024-10-09
Remove database use from handleRelevantRepositoryEvent


jhrozek:
This is the last one in handlers_githubwebhooks! Once we remove this, we'll be able to close issue #4327.

PR #4705, opened 2024-10-09
Remove database use from handleRelevantRepositoryEvent

jhrozek:
# Summary

- **Add Property.Equal** - Comparing properties will be useful to check if a matching property is a subset of the retrieved properties.
- **Add WithMatchProps to the HandleEntityAndDoMessage** - We'll match the received entity on these properties
- **Don't forward entity if it doesn't match message.MatchProps** - After receiving an entity, check if the `message.MatchProps` map is set.  If yes, make sure that it is a subset of the properties of the entity just received. If it is not, drop the message.  This will allow us to restrict forwarding the message by properties set by the caller.
- **Use the handler forwarding to evaluate repository/meta messages** - Instead of fetching the repo directly in the handler and forwarding to a handler, create a NewEntityRefreshAndDoMessage message and offload the entity processing to minder-core.  This removes the last piece of the github webhook that was directly touching the database with the exception of looking up the github app installation.

Fixes: #4703

## Change Type

# Testing

mostly unit tests + adding and removing repos

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_githubwebhooks.go
+++ b/internal/controlplane/handlers_githubwebhooks.go
@@ -38,18 +38,14 @@ import (
 	"github.com/stacklok/minder/internal/db"
 	"github.com/stacklok/minder/internal/engine/entities"
 	entityMessage "github.com/stacklok/minder/internal/entities/handlers/message"
-	"github.com/stacklok/minder/internal/entities/models"
 	"github.com/stacklok/minder/internal/entities/properties"
 	"github.com/stacklok/minder/internal/events"
-	"github.com/stacklok/minder/internal/projects/features"
 	"github.com/stacklok/minder/internal/providers/github/clients"
 	"github.com/stacklok/minder/internal/providers/github/installations"
 	ghprop "github.com/stacklok/minder/internal/providers/github/properties"
 	ghsvc "github.com/stacklok/minder/internal/providers/github/service"
 	"github.com/stacklok/minder/internal/reconcilers/messages"
-	reposvc "github.com/stacklok/minder/internal/repositories"
 	pb "github.com/stacklok/minder/pkg/api/protobuf/go/minder/v1"
-	provifv1 "github.com/stacklok/minder/pkg/providers/v1"
 )
 
 // errRepoNotFound is returned when a repository is not found
@@ -718,7 +714,7 @@ func (_ *Server) processPackageEvent(
 	return &processingResult{topic: events.TopicQueueOriginatingEntityAdd, wrapper: pkgMsg}, nil
 }
 
-func (s *Server) processRelevantRepositoryEvent(
+func (_ *Server) processRelevantRepositoryEvent(
 	ctx context.Context,
 	payload []byte,
 ) (*processingResult, error) {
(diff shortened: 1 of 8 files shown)


jhrozek:
@JAORMX I think that rebasing #4704 onto this PR will make your tests work.

internal/controlplane/handlers_githubwebhooks.go
```diff
@@ -744,66 +740,47 @@ func (s *Server) processRelevantRepositoryEvent(
...
-			// one we have stored in the DB.
-			return nil, newErrNotHandled("meta event with action %s not handled, hook ID %d does not match stored webhook ID %d",
-				event.GetAction(),
-				event.GetHookID(),
-				hookId,
-			)
+		// If not, this means we got a deleted event for a
+		// webhook ID that doesn't correspond to the
+		// one we have stored in the DB.
+		matchHookProps, err := properties.NewProperties(map[string]any{
```
jhrozek, line 761: This codepath was not easy to test. In the end I resorted to a trick and:
 - registered a repo
 - changed the hookID value in the DB
 - deleted the webhook
 - made sure the message is ignored

internal/entities/handlers/handler.go
```diff
@@ -130,11 +130,45 @@ func (b *handleEntityAndDoBase) handleRefreshEntityAndDo(msg *watermill.Message)
...
+
+	matchProps, err := properties.NewProperties(entMsg.MatchProps)
+	if err != nil {
+		return err
+	}
+
+	for propName, prop := range matchProps.Iterate() {
+		entProp := ewp.Properties.GetProperty(propName)
+		if !prop.Equal(entProp) {
+			return errors.New("properties do not match")
```
JAORMX, line 162: Let's turn this error into a variable `ErrPropertiesDontMatch`.

internal/entities/handlers/handler.go
```diff
@@ -102,7 +102,7 @@ func (b *handleEntityAndDoBase) handleRefreshEntityAndDo(msg *watermill.Message)
 		l.Debug().Msg("entity not retrieved")
 	}
 
-	err = b.forwardEntityCheck(ctx, ewp)
+	err = b.forwardEntityCheck(ctx, entMsg, ewp)
```
JAORMX, line 105: Below here we could check if the properties don't match (via a defined error, mentioned below), and not log an error, since it isn't one.

internal/entities/handlers/handler.go
```diff
@@ -130,11 +130,45 @@ func (b *handleEntityAndDoBase) handleRefreshEntityAndDo(msg *watermill.Message)
...
 	}
 
+	if err := b.matchPropertiesCheck(entMsg, ewp); err != nil {
+		return err
+	}
+
+	return b.repoPrivateOrArchivedCheck(ctx, ewp)
+}
+
+func (_ *handleEntityAndDoBase) matchPropertiesCheck(
```
JAORMX, line 146: Can you add more documentation on why we need to do this matching? e.g. a case that represents the use of this. I'm afraid this will not be easy to follow or use in the future.

internal/entities/handlers/message/message.go
```diff
@@ -43,9 +43,10 @@ type EntityHint struct {
 
 // HandleEntityAndDoMessage is a message that is sent to the entity handler to refresh an entity and perform an action.
 type HandleEntityAndDoMessage struct {
-	Entity     TypedProps `json:"entity"`
-	Originator TypedProps `json:"owner"`
-	Hint       EntityHint `json:"hint"`
+	Entity     TypedProps     `json:"entity"`
+	Originator TypedProps     `json:"owner"`
+	Hint       EntityHint     `json:"hint"`
+	MatchProps map[string]any `json:"match_props"`
```
JAORMX, line 49: Can you add docs about `MatchProps`?

jhrozek: done!

internal/entities/handlers/handler.go
jhrozek: done!

internal/entities/handlers/handler.go
jhrozek: done!

JAORMX (approved):

JAORMX (approved):

Issue #4690, opened 2024-10-08
ListEvaluationResults is slow in some cases


evankanderson:
### Describe the issue

@danbarr says

>I'd been tinkering with the artifact signature rule, so I've been refreshing the Alerts and History tabs a bunch

This caused ListEvaluationResults latency to hit >1s @50th and >2s at 95th.  I also see ~700 log messages of the form:

```
2024-10-08T22:13:12.699Z     properties are valid, skipping provider fetch
2024-10-08T22:13:12.695Z     entity found, fetching properties
2024-10-08T22:13:12.691Z     fetching entity with properties
2024-10-08T22:13:12.691Z     properties fetched
```

Note that this seems to be around 10ms for fetching the properties of a single entity; with 200 entities, we might expect to see times on the order of... 2 seconds.

### What version are you using?

Live (production)

danbarr:
Some details on the project I was testing in...

Total of 3 repos, 5 profiles in the project.

The two profiles I was actively poking have 29 and 1 rules respectively. I was working with the stacklok/artifact_signature and [artifact_attestation_slsa](https://github.com/stacklok/minder-rules-and-profiles/blob/main/rule-types/github/artifact_attestation_slsa.yaml) rules trying to figure out my failures, so I was bouncing between listing my alerts and history (filtering down to Artifact entity type to find my results).

LMK if any other details would be useful!

PR #4717, opened 2024-10-10
Remove redundant propsvc.RetrieveAllPropertiesForEntity when looking up entities for history lookup

jhrozek:
# Summary

We were doing 2 calls when serching for entities during the
`ListEvaluationHistory` lookup:

1) EntityWithPropertiesByID
2) RetrieveAllPropertiesForEntity with the TolerateStaleData option set.

The second is not needed right after the first if we set tolerate stale
data in this case. The flow of the calls is:

- EntityWithPropertiesByID
 - db.GetEntityByID
 - db.GetAllPropertiesForEntity
 - then convert to our data model and return

- RetrieveAllPropertiesForEntity
 - provMan.InstantiateFromID
 - db.GetAllPropertiesForEntity
 - check property validity always shortcuts because we set TolerateStaleData
 - turn into our model
 - write the properties into the entity model representation we passed
   in

Related: #4690

## Change Type

# Testing

I added a debug printf to verify that all the properties are there.

# Review Checklist:

### DIFF
--- a/internal/history/service.go
+++ b/internal/history/service.go
@@ -229,12 +229,6 @@ func (ehs *evaluationHistoryService) ListEvaluationHistory(
 				return nil, fmt.Errorf("error fetching entity for properties: %w", err)
 			}
 
-			err = propsvc.RetrieveAllPropertiesForEntity(ctx, lookupEwp, ehs.providerManager,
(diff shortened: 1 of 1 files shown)


jhrozek:
After talking to @JAORMX I'm going to close this PR because this would break the case where we really have zero properties , which is true for some entities that we are migrating from the old tables.

PR #4729, opened 2024-10-11
Implement an entity cache for the properties service

JAORMX:
# Summary

This adds a wrapper that is able to cache entities for cases that need
to iterate them. This is handy for one-off calls like getting the
history of executions or showing reports. This way, we don't need to
query the database multiple times.

This cache is persistent, so it's only relevant for one execution.

## Change Type

# Testing

# Review Checklist:


jhrozek (approved):
This is a nice API! I tested this by using the API in `ListEvaluationResults` where on my local system the API responses went down from ~500 to ~200 ms when the provider was already initialized. We still probably have some optimization to do elsewhere, but this is a clear improvement.

I'll submit the ListEvaluationResults patch and tests as a separate PR.

PR #4783, opened 2024-10-21
Introduce `ResourceMeta` and general resource parsing

JAORMX:
# Summary

This introduces the concept of resource metadata, which includes three
things:

* `type`: the type of resource
* `version`: The version of a resource
* `name`: The name of a resource

Parsing these three is mandatory for any resource in Minder's domain
model.

This also introduces a general parser which replaces the
rule-type-specific parser.

Finally, the rule type `create`/`apply` commands were refactored to use
this new parser, and will skip files that are effectively not rule
types. Thus allowing the aforementioned sub commands to run on file
directories with objects that aren't necessarily rule types.

## Change Type

# Testing

# Review Checklist:


pkg/api/protobuf/go/minder/v1/api.go
```diff
@@ -0,0 +1,130 @@
...
+		ArtifactResource:       &Artifact{},
+	}
+)
+
+// ResourceMeta defines the basic metadata for a resource within
+// Minder. This is used as a common interface and to determine
+// the type of resource.
+type ResourceMeta struct {
+	// Version is the version of the resource
+	Version string `json:"version" yaml:"version" validate:"required"`
```
evankanderson, line 81: What kind of resources does this cover?  It looks like you want to cover both RuleTypes/Profiles (which would let us just `apply` a directory) _and_ entities / repositories / artifacts -- what's the use-case for those being on-disk, and do they apply in the same way?  Will we have up- and down-conversion on the server side, or is this simply detecting mismatches?

Given that proto can produce `Get` methods, what would you think of the following instead of `ParseResource`?

```golang

type ResourceMeta interface {
  protoreflect.ProtoMessage

  GetVersion() string
  GetType() string
  GetName() string
}

var _ ResourceMeta = (*RuleType)(nil)
var _ ResourceMeta = (*RuleType)(nil)
var _ ResourceMeta = (*EntityInstance)(nil)
var _ ResourceMeta = (*Repository)(nil)
var _ ResourceMeta = (*Artifact)(nil)

func ParseResource(r io.Reader, out ResourceMeta) error {
  // no need for io.TeeReader
  if err := json.NewDecoder(tee).Decode(out); err != nil {
    return fmt.Errorf("Error validating %T: %w", out, err)
  }
  // validate, including checking GetName() matches the type.
}
```

evankanderson (commented):
I'm not quite sure I follow the goals here, but one comment on using Golang interfaces that might reduce the need for re-parsing.

pkg/api/protobuf/go/minder/v1/api.go
JAORMX: Mostly this is useful for projects, rule types and profiles, but given thay we use the same format, I decided to apply it to entities as well. Trying to unify parsing and validation for ease of testing.

This suggestion is super handy, thanks! I'll try it tomorrow

PR #4784, opened 2024-10-21
Add flag-driven event publisher

evankanderson:
# Summary

Adds an event publisher which selects between two event publishing channels based on flag configuration.

## Change Type

# Testing

Manual testing between go-channel and nats, then wrote a unit test.

Note: I used two go channels for the tests, because the setup for those is easier.  I could switch one to NATS if desired.

# Review Checklist:


internal/events/flagged_driver.go
```diff
@@ -0,0 +1,162 @@
+//
+// Copyright 2023 Stacklok, Inc.
```
jhrozek, line 2: I hate to be the person who is commenting on copyright headers of all things in a PR, but isn't this going to make our CI choke on the header (and if it's not maybe it should..)

jhrozek (commented):
I think the fan-in implementation is good and I learned something about nil channels!

The only comment I have is about the copyright header.

internal/events/flagged_driver.go
evankanderson: We don't have anything in CI to choke on this, and I'm not sure we should, because then it would need to understand or choke on renames.  (But then again, I'm not sure we need the date in the copyright header at all, between the Berne convention and various other sources of that information, like `git`.)

Vyom-Yadav:
Needs a rebase + go mod tidy

internal/events/flagged_driver.go
```diff
@@ -0,0 +1,162 @@
...
+			case msg, ok := <-experiment:
+				if !ok {
+					experiment = nil
+					continue
+				}
+				out <- msg
+				f.readMessages.Add(ctx, 1, metric.WithAttributes(attribute.Bool("experiment", true)))
+			}
+		}
+	}()
```
Vyom-Yadav, line 106: We are missing context cancellation here imo:
```go
case <-ctx.Done():
            return
```

internal/events/flagged_driver.go
```diff
@@ -0,0 +1,162 @@
...
+// the drivers managed by the flagged publisher.
+func (f *flaggedDriver) Close() error {
+	if err := f.basePub.Close(); err != nil {
+		return fmt.Errorf("Failed to close base publisher: %w", err)
+	}
+	if err := f.experimentPub.Close(); err != nil {
+		return fmt.Errorf("Failed to close experiment publisher: %w", err)
+	}
+	return nil
+}
```
Vyom-Yadav, line 120: Do we also want to invoke the `baseCloser()` and `experimentCloser()`? (By storing them?)
I think we close everything in eventer, just wanted to highlight this:
```go
	return &eventer{
		router:            router,
		webhookPublisher:  pubWithMetrics,
		webhookSubscriber: subWithMetrics,
		closer: func() {
			//nolint:gosec // It's fine if there's an error as long as we close the router
			pubWithMetrics.Close()
			//nolint:gosec // It's fine if there's an error as long as we close the router
			subWithMetrics.Close()
			// driver close
			cl()
		},
		msgInstruments: metricInstruments,
	}, nil
```

internal/events/flagged_driver.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+	}
+	return nil
+}
+
+// Subscribe implements message.Subscriber.  In this case, it should subscribe
+// to both the base and experiment drivers, because we might get messages for
+// either.
+func (f *flaggedDriver) Subscribe(ctx context.Context, topic string) (<-chan *message.Message, error) {
+	out := make(chan *message.Message)
```
Vyom-Yadav, line 76: This channel, being unbuffered, will block the pipeline until the already published messages have been consumed (slow subscriber, etc.), which in turn blocks whatever we receive from the other channels (fan-in), basically blocking everything. I'm assuming this is what we have right now (with other implementations, too).

internal/events/flagged_driver.go
```diff
@@ -0,0 +1,162 @@
...
+	readMessages      metric.Int64Counter
+}
+
+// Publish implements message.Publisher.  If the message is flagged by the
+// alternate_message_driver experiment, it is published to the experiment
+// driver, otherwise it is published to the base driver.
+func (f *flaggedDriver) Publish(topic string, messages ...*message.Message) error {
+	// Each message has its own context, so they _could_ be in different flag treatments
+	for _, m := range messages {
+		emptyContext := engcontext.EntityContext{}
```
Vyom-Yadav, line 53: `emptyContext` can be moved outside the loop.

Vyom-Yadav (changes requested):
some additional items:

internal/events/flagged_driver_test.go
```diff
@@ -0,0 +1,231 @@
...
+	t.Cleanup(func() {
+		otel.SetMeterProvider(existingProvider)
+		if err := reader.Shutdown(context.Background()); err != nil {
+			t.Fatalf("failed to shutdown reader: %v", err)
+		}
+	})
+	otel.SetMeterProvider(meterProvider)
+	// Tests want t.Parallel(), but we don't want them to increment counters at
+	// the same time, so we need to actually mutex their execution.
+	metricMutex := &sync.Mutex{}
```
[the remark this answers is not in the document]
evankanderson: We are actually using `t.Parallel()` in the subtest, but not in the top-level test.  Updated the comment.

internal/events/flagged_driver.go
evankanderson: `makeFlaggedDriver` returns a `closer` function that calls the returned `baseCloser` and `experimentCloser`.

I'm not entirely sure why we have both `closer` functions and a `Close` method on publisher and subscriber, but if there's a difference, it feels like we shouldn't cross the streams.  (And if there's no difference, we should clean that up in a _separate_ refactor...)

internal/events/flagged_driver.go
evankanderson: Sounds good, though the cost is probably ~zero, since it is stack-allocated as the zero value.  (This is one of the times I wish Golang had `const`)

internal/events/flagged_driver.go
evankanderson: Yes, that's on purpose.  (This is called back-pressure.)  We don't want one process (Pod) to greedily read 100 messages from the topic and then spend a long time digesting them while the other processes are sitting around with no work.  By blocking the read-from-queue thing after the first element, it gives a chance for other processes to pick up the next work items.

If we had 4 threads in a process reading messages from the subscriber, we'd end up with each of them pulling one item, and then hitting back-pressure until a thread became free.  I'm not sure whether we have these kinds of worker pools currently in place, but that feels like the right place to manage parallelism to me.

internal/events/flagged_driver.go
evankanderson: It looks like this context is coming from the Router.RunHandlers method, so we should only get a cancellation when the router is shutting down.  (That should also cause all the subscribers to be closed, but it's safe to add this here as well.)

evankanderson: (I added the shutdown anyway)

internal/events/flagged_driver.go
Vyom-Yadav: Yeah, valid point.

internal/events/flagged_driver.go
Vyom-Yadav: I think this should be fine as publishers (what we use today) are not blocked if there are no active subscribers (as it should be, of course, to a limit). 

This is more of a question than a concern: how about something where subscribers can pick multiple events from a queue (in a load-balanced way) and have the ability to process multiple events parallelly (although I think for Minder, we want to sequentially process events that we get from a repo to make sure we are not creating any implicit race conditions (reading different repo config)).

---

I'm fine with this.

internal/events/flagged_driver.go
Vyom-Yadav: Yeah, just a nit.

Vyom-Yadav (approved):
lgtm, thank you!

PR #4793, opened 2024-10-22
Expose a helper to run a JQ expression on YAML input through our Rego library

jhrozek:
# Summary

The reason I'm adding this helper is that Rego's (OPAs?) YAML parser seems to be coalescing
boolean-like values to bool types which means that:
```
on:
 - push
 - pull_request
```

Becomes:
```
{ "true": ["push", "pull_request"] }
```

The other reason is that the `on` trigger can have many forms from a simple string:
```
on: push
```
to array:
```
on:
 - push
 - pull_request
```
or an object:
```
on:
 - push:
    - branches: ["main", "devel"]
```

All in all, it seemed better to have a predictable implementation that lets you write a JQ
query and get a result rather than working around the idiosyncraties of Rego and OPA.

## Change Type

# Testing

Through unit tests and an upcoming patch to rules-and-profiles which I will submit separately.

# Review Checklist:


jhrozek:
per Slack discussion with @JAORMX I'm going to split the API into 2 calls - safe loading of the YAML and the JQ query

JAORMX (approved):

PR #4814, opened 2024-10-24
Move db from internal to pkg

rdimitrov:
# Summary

The following PR moves db from internal to pkg

## Change Type

# Testing

# Review Checklist:


evankanderson:
What portion of `/internal/db` are you hoping to access?   I'm split on this -- on one hand, having additional tools not "slip" the query and schema seems useful.  On the other hand, I worry about introducing a bunch of SQL for tools in `mindersec/minder` which isn't actually used in that codebase.

On the gripping hand, I worry about queries like `SELECT * FROM ...` getting broken by a schema changed applied by the push process of another tool.  This is actually something that we could suffer from today during a push -- can you check and see what the effect is if you add a migration and then run some pre-migration `sqlc` code that is defined using `SELECT *` into a struct?  e.g.

```
--- name: GetArtifactByID :one
SELECT * FROM artifacts
WHERE artifacts.id = $1 AND artifacts.project_id = $2;
```

```
ALTER TABLE artifacts
  ADD COLUMN a_test IF NOT EXISTS BOOLEAN NOT NULL DEFAULT TRUE;
```

// What happens if I built a tool before the migration was defined, and then run the query _after_ the migration has been run on the database?

rdimitrov:
Closing this as we are going to take another approach in a follow up PR

PR #4829, opened 2024-10-26
Add NATS publisher support to reminder

Vyom-Yadav:
# Summary

Add NATS as an option for `reminder` to publish events. 

## Change Type

# Testing

Tested locally, working fine (`nats stream view` output):
```
...
[269] Subject: minder.internal.repo.reconciler.event Received: 1970-01-01T05:30:02+05:30

  ce-source: minder
  ce-specversion: 1.0
  ce-subject: TODO
  ce-type: minder.internal.repo.reconciler.event
  ce-datacontenttype: application/json
  ce-id: 36f8fc06-6b36-48e8-ad92-3cde66759a13
  ce-minderpublished0at: 2024-10-26T18:39:07Z

{"entity_id":"ca6f63c7-672b-4906-ae2d-ce050cd0b6bc","project":"f6bf23d3-ace6-4652-b858-8f9addf7bea5","provider":"a3c63545-59d7-4d46-9a4c-5028876ad236"}

...
```

# Review Checklist:


internal/events/events.go
```diff
@@ -153,7 +150,7 @@ func instantiateDriver(
 		return eventersql.BuildPostgreSQLDriver(ctx, cfg)
 	case constants.NATSDriver:
 		zerolog.Ctx(ctx).Info().Msg("Using NATS driver")
-		return nats.BuildNatsChannelDriver(cfg)
+		return nats.BuildNatsChannelDriver(&cfg.Nats)
```
evankanderson, line 153: All the other constructors take the entire `cfg`, so that's the model I followed here.

internal/reminder/publisher.go
```diff
@@ -13,12 +13,33 @@ import (
...
+func (r *reminder) getMessagePublisher(ctx context.Context) (message.Publisher, common.DriverCloser, error) {
+	switch r.cfg.EventConfig.Driver {
+	case constants.NATSDriver:
+		return r.setupNATSPublisher(ctx)
+	case constants.SQLDriver:
+		return r.setupSQLPublisher(ctx)
+	default:
+		return nil, nil, fmt.Errorf("unknown publisher type: %s", r.cfg.EventConfig.Driver)
+	}
+}
```
evankanderson, line 29: I'm nervous about having this diverge from `internal/events/events.go`.  Can we use an [`eventer.Interface` from `eventer.New()`](https://pkg.go.dev/github.com/mindersec/minder@v0.0.71/pkg/eventer) here?

(I need to do some refactoring in that package to move `eventer/interface` into `eventer`, but I'll wait until this code is landed to reduce the amount of conflicts.)

pkg/config/common.go
```diff
@@ -149,3 +149,14 @@ func ReadKey(keypath string) ([]byte, error) {
...
+// NatsConfig is the configuration when using NATS as the event driver
+type NatsConfig struct {
+	// URL is the URL for the NATS server
+	URL string `mapstructure:"url" default:"nats://localhost:4222"`
+	// Prefix is the prefix for the NATS subjects to subscribe to
+	Prefix string `mapstructure:"prefix" default:"minder"`
+	// Queue is the name of the queue group to join when consuming messages
+	// queue groups allow multiple process to round-robin process messages.
+	Queue string `mapstructure:"queue" default:"minder"`
+}
```
evankanderson, line 162: _If_ we want to move this up to `common` (rather than just having a split between `client` and `server` configuration), let's put it in its own file.  It feels kind of peculiar to have this in `common` while having the rest of the event configuration still in `server`, so I'd argue for moving the code back, particularly in the context of having smaller, more focused PRs.

pkg/config/reminder/events.go
evankanderson, line 1: What do you think of moving the `reminder` configuration into the same package as `server`, and calling them all "server-side components`?  It feels like it would allow for more sharing between the two implementations.

internal/events/nats/natschannel.go
```diff
@@ -111,25 +116,21 @@ func (c *cloudEventsNatsAdapter) ensureTopic(ctx context.Context, topic string,
...
-			Subjects: []string{c.cfg.Prefix + ".>"},
-		})
-		return err
-	}
-	return nil
+	_, err = js.CreateOrUpdateStream(ctx, jetstream.StreamConfig{
+		Name:     c.cfg.Prefix,
+		Subjects: []string{c.cfg.Prefix + ".>"},
+	})
+	return err
```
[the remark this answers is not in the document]
evankanderson: 👍  if this works; I think when I started looking, this wasn't cleaned up / operable with CloudEvents yet.

evankanderson (commented):
There's a separate fix going in for the one test failure.

In general, I think you could have separated the changes to `internal/events/nats` from the changes to `reminder`, which makes it easier to roll back or hold one change or the other if there needs to be discussion.  But I'm willing to approve this and then iterate more going forward, modulo the comments, particularly about re-using `eventer.New`, rather than duplicating the setup code at this point.

pkg/config/reminder/events.go
Vyom-Yadav: My initial thoughts with reminder were that it'll be a separate micro service (remember when we had the talk of maybe let's write it in python). I agree with it all being server side components, but I'm not sure about moving that to server config (I wanted to keep the dependency tree separate, as they'd be packaged as separate binaries):

```
$ go list -f '{{range .Imports}}{{.}} # These are just imports and not deps
{{end}}' ./pkg/config/server
context
crypto/rsa
errors
fmt
github.com/go-playground/validator/v10
github.com/golang-jwt/jwt/v4
github.com/mindersec/minder/internal/auth/jwt
github.com/mindersec/minder/internal/util
github.com/mindersec/minder/pkg/config
github.com/rs/zerolog
github.com/rs/zerolog/log
github.com/spf13/pflag
github.com/spf13/viper
golang.org/x/oauth2
golang.org/x/oauth2/clientcredentials
io
net/http
net/url
os
path/filepath
strings
time
```

```
$ go list -f '{{range .Imports}}{{.}}
{{end}}' ./pkg/config/reminder
fmt
github.com/mindersec/minder/internal/util
github.com/mindersec/minder/pkg/config
github.com/rs/zerolog
github.com/spf13/pflag
github.com/spf13/viper
os
strings
time
```

>  It feels like it would allow for more sharing between the two implementations.

I agree with that. Currently, there is some duplication regarding how we manage components, which I try to tackle by moving things to a common place (which doesn't feel ideal in some situations). But, I'm of the opinion to try to keep things separate.

internal/events/events.go
Vyom-Yadav: This has become a tedious problem to refactor. I have to change the signature of `BuildNatsChannelDriver` as `cloudEventsNatsAdapter` is not externally exposed. 

Either I change the signature of every driver function to take in not `serverconfig`, but only the relevant config (in order to follow a pattern), which again isn't pretty as it does not truly achieve separation from `serverconfig` as functions like `BuildPostgreSQLDriver` would still rely on components purely in `serverconfig` (like `cfg.SQLPubSub.AckDeadline`) which isn't there in `reminderconfig`. Either it would result in breaking the pattern or some form of duplication of the instantiation code. 

I see how this can slowly become just duplication of all instantiation code in order to keep minder and reminder separate, but I'm torn between keeping them separate (as minder is the main engine and reminder is a completely different service) v/s keeping them in the same code for ease of maintainability (and for de duplication - which again strikes the thought that this duplication would _not_ have been there if reminder was in another language). 

What do you think about it?

internal/reminder/publisher.go
Vyom-Yadav: > I'm nervous about having this diverge from internal/events/events.go

`internal/events/events.go` is all about `serverconfig`, it would diverge from it as this is `reminderconfig`. I don't think we should use the [interfaces.Interface](https://pkg.go.dev/github.com/mindersec/minder@v0.0.71/pkg/eventer/interfaces#Interface):

```go
// Interface is a combination of the Publisher, Registrar, and Service interfaces.
// This is handy when spawning the eventer in a single function for easy setup.
type Interface interface {
	Publisher
	Registrar
	Service
}
```

as reminder only needs publishing capability, I'd be happy to implement [interfaces.Publisher](https://pkg.go.dev/github.com/mindersec/minder@v0.0.71/pkg/eventer/interfaces#Publisher) for `reminder`.

pkg/config/common.go
Vyom-Yadav: > If we want to move this up to common (rather than just having a split between client and server configuration), let's put it in its own file

That still breaks the pattern as all the structs in `common` don't have their own file. 

> so I'd argue for moving the code back,

To duplicate vs 'have common code'. Do you want me to duplicate the `NatsConfig` code in that case for `reminderconfig`

Vyom-Yadav:
Sorry for the delay on this @evankanderson. Please ignore the failing CI, that's not the main point. I'm confused between how to refactor this properly, because there isn't any easy and good way imo, in all approaches, we either have some duplication or break some pattern (unless we refactor a lot of code in serverconfig and internal/events/events.go).

Three problems right now:

1. Where to move `NatsConfig`?
2. Changing signature of `BuildNatsChannelDriver`?
3. Should we move `reminderconfig` to the `serverconfig` section?

pkg/config/common.go
evankanderson: I think there are a couple options here:

1. The way that structs are all piled into `common.go` today is probably a historical mistake.  I'm not asking you fix that mistake in this PR (because we should be making small, targeted changes), but we shouldn't make that mistake worse.  By putting this in `nats` or `eventing` files, it will make the future expansion more clear.
2. I think it's fine to have `pkg/config/reminder` depend on `pkg/config/server`, as long as we keep the graph acyclic.  In particular, if we have common server-side configuration for e.g. flags and messaging, it feels like it would make sense to have those be shared between the two.  Given that you can reasonably run a `minder` server without `reminder`, but the opposite is not very useful, I'd argue for having `pkg/config/reminder` import _parts_ of the config from `pkg/config/server`.  In the future, if we split the Minder server into e.g. `rpc_server` and `event_processor`, I'd suggest that both would want to use parts of the current `ServerConfig`, but would probably each have their own top-level struct for configuration (possibly in a separate module, or possibly in the `pkg/config/server` module).

I'm more concerned about leaking the dependencies on e.g. NATS into the `minder` client, which is statically linked and downloaded by many more users than the server.  Right now, this take that dependency, but I could see add a helper to create a NATS connection to this library in the future.

internal/reminder/publisher.go
evankanderson: I think it's fine to use the `interfaces.Publisher` in the reminder code, but it feels like it might be easiest if the reminder code expected an `interfaces.Publisher` (which is satisfied by the current eventing implementations), but the construction code used `NewEventer` to construct the `interfaces.Publisher`.

internal/events/events.go
evankanderson: I think it's fine to re-use the _implementation_ of the construction code in reminder, even as it runs as a separate service.  (For example, most Kubernetes controllers use the same informers library and construction client to create their connection to the K8s apiserver.)

So, my suggestion would be to embed a `serverconfig.EventConfig` into the `ReminderConfig`, rather than attempting to mirror only selected options into the reminder config.  This will mean that it's possible to have something meaningless like the Go channel in reminder; you can avoid that by adding a check in the top-level `ReminderConfig.Validate` that the event configuration isn't using the go module.

evankanderson: (This pattern will also give you access to the flag-driven eventer, and future improvements in flagging that I'm planning -- using a central flag server, rather than repeating the flag file for each binary.)

evankanderson (commented):
I hope using parts of the server-side configuration in reminder makes sense -- if necessary, we could have a `server/common` that both components use, but that feels like a lot of extra refactoring without a lot of extra value.

Vyom-Yadav:
Importing `serverconfig.EventConfig` directly made things simpler (although the init code does things not required by reminder, but it gets the job done). CI failure is the golang net CVE. I also added simple checking for not selecting the unsupported driver.

internal/reminder/publisher.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/mindersec/minder/pkg/eventer"
+)
+
+func (r *reminder) getMessagePublisher(ctx context.Context) (message.Publisher, error) {
+	pub, err := eventer.New(ctx, nil, &r.cfg.EventConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create publisher: %w", err)
+	}
+
+	return pub, nil
```
evankanderson, line 21: I have a slight preference for just leaving this code in-line, given that it could generally be reduced to:

```suggestion
	return eventer.New(ctx, nil, &r.cfg.EventConfig)
```

I don't feel very strongly, however.  (Leaving it in-line reduces the need to "peek" into the implementation to see if there's anything fancy going on when tracing the code.  This is admittedly less of an issue for one-time initialization code.)

pkg/config/reminder/config.go
```diff
@@ -52,3 +60,14 @@ func RegisterReminderFlags(v *viper.Viper, flags *pflag.FlagSet) error {
...
+func validateEventConfig(cfg serverconfig.EventConfig) error {
+	switch cfg.Driver {
+	case constants.NATSDriver:
+	case constants.SQLDriver:
+	default:
+		return fmt.Errorf("events.driver %s is not supported", cfg.Driver)
+	}
+
+	return nil
+}
```
evankanderson, line 73: This prevents using e.g. the flag driver.  I think that's okay, but I'd tend to write this as a deny-list of configuration that we know won't work rather than an allow-list (because we'll need to go back and expand the allow-list as we add new messaging configurations).

evankanderson (approved):
A few minor nits, but I'm going to merge as-is, because I think this will function well enough, and the other comments are just stylistic touches that could reasonably be argued either way.

PR #4831, opened 2024-10-27
Record Metrics for Reminder

Vyom-Yadav:
# Summary

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/reminder/metrics/metrics.go
+++ b/internal/reminder/metrics/metrics.go
@@ -1,0 +1,84 @@
+package metrics
+
+import (
+	"context"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// Default bucket boundaries in seconds for the send delay histogram
+var sendDelayBuckets = []float64{
+	0,    // immediate
+	10,   // 10 seconds
+	20,   // 20 seconds
+	40,   // 40 seconds
+	80,   // 1m 20s
+	160,  // 2m 40s
+	320,  // 5m 20s
+	640,  // 10m 40s
+	1280, // 21m 20s
+}
+
+type Metrics struct {
+	// Time between when a reminder became eligible and when it was sent
+	SendDelay metric.Float64Histogram
+
+	// Current number of reminders in the batch
+	BatchSize metric.Int64Gauge
+
+	// Average batch size (updated on each batch)
+	AvgBatchSize metric.Float64Gauge
+
+	// For tracking average calculation
+	// TODO: consider persisting this to avoid reset on restart (maybe)
+	totalBatches   int64
+	totalReminders int64
+}
+
+func NewMetrics(meter metric.Meter) (*Metrics, error) {
+	sendDelay, err := meter.Float64Histogram(
+		"reminder_send_delay",
+		metric.WithDescription("Time between reminder becoming eligible and actual send (seconds)"),
+		metric.WithUnit("s"),
+		metric.WithExplicitBucketBoundaries(sendDelayBuckets...),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	batchSize, err := meter.Int64Gauge(
+		"reminder_batch_size",
+		metric.WithDescription("Current number of reminders in the batch"),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	avgBatchSize, err := meter.Float64Gauge(
+		"reminder_avg_batch_size",
+		metric.WithDescription("Average number of reminders per batch"),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Metrics{
+		SendDelay:    sendDelay,
+		BatchSize:    batchSize,
+		AvgBatchSize: avgBatchSize,
+	}, nil
+}
+
+func (m *Metrics) RecordBatch(ctx context.Context, size int64) {
+	// Update current batch size
+	m.BatchSize.Record(ctx, size)
+
+	// Update running average
+	m.totalBatches++
+	m.totalReminders += size
+	avgSize := float64(m.totalReminders) / float64(m.totalBatches)
+	m.AvgBatchSize.Record(ctx, avgSize)
+}
+
+func (m *Metrics) RecordSendDelay(ctx context.Context, delaySeconds float64) {
+	m.SendDelay.Record(ctx, delaySeconds)
+}
--- a/internal/reminder/metrics/provider.go
+++ b/internal/reminder/metrics/provider.go
@@ -1,0 +1,132 @@
+package metrics
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/mindersec/minder/pkg/config/reminder"
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/rs/zerolog/log"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/prometheus"
+	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/resource"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+)
+
+const (
+	metricsPath       = "/metrics"
+	readHeaderTimeout = 2 * time.Second
+)
+
+// Provider manages the metrics server and OpenTelemetry setup
+type Provider struct {
+	server  *http.Server
+	mp      *sdkmetric.MeterProvider
+	metrics *Metrics
+}
+
+// NewProvider creates a new metrics provider
+func NewProvider(cfg *reminder.MetricsConfig) (*Provider, error) {
+	if cfg == nil {
+		return nil, errors.New("metrics config is nil")
+	}
+
+	if !cfg.Enabled {
+		return &Provider{}, nil
+	}
+
+	// Create Prometheus exporter
+	prometheusExporter, err := prometheus.New(
+		prometheus.WithNamespace("reminder_service"),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Prometheus exporter: %w", err)
+	}
+
+	// Create resource with service information
+	res := resource.NewWithAttributes(
+		semconv.SchemaURL,
+		semconv.ServiceName("reminder-service"),
+		semconv.ServiceVersion("v0.1.0"),
+	)
+
+	// Create meter provider
+	mp := sdkmetric.NewMeterProvider(
+		sdkmetric.WithReader(prometheusExporter),
+		sdkmetric.WithResource(res),
+	)
+
+	// Set global meter provider
+	otel.SetMeterProvider(mp)
+
+	// Create metrics
+	meter := mp.Meter("reminder-service")
+	metrics, err := NewMetrics(meter)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create metrics: %w", err)
+	}
+
+	// Create HTTP server
+	mux := http.NewServeMux()
+	mux.Handle(metricsPath, promhttp.Handler())
+
+	server := &http.Server{
+		Addr:              fmt.Sprintf("%s:%d", cfg.Host, cfg.Port),
+		Handler:           mux,
+		ReadHeaderTimeout: readHeaderTimeout,
+	}
+
+	return &Provider{
+		server:  server,
+		mp:      mp,
+		metrics: metrics,
+	}, nil
+}
+
+// Start starts the metrics server if enabled
+func (p *Provider) Start(ctx context.Context) error {
+	if p.server == nil {
+		return nil // Metrics disabled
+	}
+
+	errCh := make(chan error)
+	go func() {
+		log.Info().Str("address", p.server.Addr).Msg("Starting metrics server")
+		if err := p.server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			errCh <- fmt.Errorf("metrics server error: %w", err)
+		}
+	}()
+
+	select {
+	case err := <-errCh:
+		return err
+	case <-ctx.Done():
+		return p.Shutdown(ctx)
+	}
+}
+
+// Shutdown gracefully shuts down the metrics server
+func (p *Provider) Shutdown(ctx context.Context) error {
+	if p.server == nil {
+		return nil
+	}
+
+	shutdownCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	log.Info().Msg("Shutting down metrics server")
+	if err := p.mp.Shutdown(shutdownCtx); err != nil {
+		log.Error().Err(err).Msg("Error shutting down meter provider")
+	}
+
+	return p.server.Shutdown(shutdownCtx)
+}
+
+// Metrics returns the metrics instance
+func (p *Provider) Metrics() *Metrics {
+	return p.metrics
+}
--- a/internal/reminder/reminder.go
+++ b/internal/reminder/reminder.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/mindersec/minder/internal/reminder/metrics"
 	"sync"
 	"time"
 
@@ -44,6 +45,8 @@ type reminder struct {
 
 	eventPublisher message.Publisher
 	eventDBCloser  common.DriverCloser
+
+	metricsProvider *metrics.Provider
 }
 
 // NewReminder creates a new reminder instance
@@ -66,6 +69,13 @@ func NewReminder(ctx context.Context, store db.Store, config *reminderconfig.Con
 
 	r.eventPublisher = pub
 	r.eventDBCloser = cl
+
+	metricsProvider, err := metrics.NewProvider(&config.MetricsConfig)
+	if err != nil {
+		return nil, fmt.Errorf("error creating metrics provider: %w", err)
+	}
+
+	r.metricsProvider = metricsProvider
 	return r, nil
 }
 
@@ -76,6 +86,11 @@ func (r *reminder) Start(ctx context.Context) error {
 	case <-r.stop:
 		return errors.New("reminder stopped, cannot start again")
 	default:
+	}
+
+	err := r.metricsProvider.Start(ctx)
+	if err != nil {
+		return fmt.Errorf("error starting metrics provider: %w", err)
 	}
 
 	interval := r.cfg.RecurrenceConfig.Interval
@@ -119,6 +134,11 @@ func (r *reminder) Stop() {
 	r.stopOnce.Do(func() {
 		close(r.stop)
 		r.eventDBCloser()
+
+		err := r.metricsProvider.Shutdown(context.Background())
+		if err != nil {
+			zerolog.Ctx(context.Background()).Error().Err(err).Msg("error shutting down metrics provider")
+		}
 	})
 }
 
@@ -143,6 +163,11 @@ func (r *reminder) sendReminders(ctx context.Context) error {
 		return fmt.Errorf("error creating reminder messages: %w", err)
 	}
 
+	remMetrics := r.metricsProvider.Metrics()
+	if remMetrics != nil {
+		remMetrics.RecordBatch(ctx, int64(len(repos)))
+	}
+
 	err = r.eventPublisher.Publish(constants.TopicQueueRepoReminder, messages...)
 	if err != nil {
 		return fmt.Errorf("error publishing messages: %w", err)
@@ -151,13 +176,18 @@ func (r *reminder) sendReminders(ctx context.Context) error {
 	repoIds := make([]uuid.UUID, len(repos))
 	for _, repo := range repos {
 		repoIds = append(repoIds, repo.ID)
+		if remMetrics != nil {
+			// sendDelay = Now() - ReminderLastSent - MinElapsed
+			reminderLastSent := repo.ReminderLastSent
+			if reminderLastSent.Valid {
+				remMetrics.SendDelay.Record(ctx, (time.Now().Sub(reminderLastSent.Time) - r.cfg.RecurrenceConfig.MinElapsed).Seconds())
+			} else {
+				// TODO: Should the send delay be zero if the reminder has never been sent?
+				remMetrics.SendDelay.Record(ctx, 0)
+				//remMetrics.SendDelay.Record(ctx, r.cfg.RecurrenceConfig.MinElapsed.Seconds())
+			}
+		}
 	}
-
-	// TODO: Collect Metrics
-	// Potential metrics:
-	// - Gauge: Number of reminders in the current batch
-	// - UpDownCounter: Average reminders sent per batch
-	// - Histogram: reminder_last_sent time distribution
 
 	err = r.store.UpdateReminderLastSentForRepositories(ctx, repoIds)
 	if err != nil {
--- a/pkg/config/reminder/metrics.go
+++ b/pkg/config/reminder/metrics.go
@@ -1,0 +1,7 @@
+package reminder
+
+type MetricsConfig struct {
+	Enabled bool   `mapstructure:"enabled" default:"true"`
+	Host    string `mapstructure:"host" default:"127.0.0.1"`
+	Port    int    `mapstructure:"port" default:"8080"`
+}
--- a/pkg/config/reminder/config.go
+++ b/pkg/config/reminder/config.go
@@ -19,6 +19,7 @@ type Config struct {
 	RecurrenceConfig RecurrenceConfig      `mapstructure:"recurrence"`
 	EventConfig      EventConfig           `mapstructure:"events"`
 	LoggingConfig    LoggingConfig         `mapstructure:"logging"`
+	MetricsConfig    MetricsConfig         `mapstructure:"metrics"`
 }
 
 // Validate validates the configuration


internal/reminder/metrics/provider.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, errors.New("metrics config is nil")
+	}
+
+	if !cfg.Enabled {
+		return &Provider{}, nil
+	}
+
+	// Create Prometheus exporter
+	prometheusExporter, err := prometheus.New(
+		prometheus.WithNamespace("reminder_service"),
```
evankanderson, line 44:
```suggestion
		prometheus.WithNamespace("reminder"),
```

Skip "service" here -- it's implicit in the context, and the namespace may get stuck onto a bunch of metric names, and make them even longer and more difficult to read.  😁

pkg/config/reminder/metrics.go
```diff
@@ -0,0 +1,7 @@
+package reminder
+
+type MetricsConfig struct {
+	Enabled bool   `mapstructure:"enabled" default:"true"`
+	Host    string `mapstructure:"host" default:"127.0.0.1"`
```
evankanderson, line 5: You may want to leave the address as `""`, so that Prometheus (for example) running in another network namespace can reach the metrics endpoint.

internal/reminder/metrics/provider.go
```diff
@@ -0,0 +1,132 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Prometheus exporter: %w", err)
+	}
+
+	// Create resource with service information
+	res := resource.NewWithAttributes(
+		semconv.SchemaURL,
+		semconv.ServiceName("reminder-service"),
```
evankanderson, line 53: Ditto on removing "-service" noise.
```suggestion
		semconv.ServiceName("reminder"),
```

internal/reminder/metrics/provider.go
```diff
@@ -0,0 +1,132 @@
...
+		semconv.ServiceVersion("v0.1.0"),
```
evankanderson, line 54: Can you copy over the:

```
// TODO: Make this auto-generated
```

if we're going to include it.  😁   (I think this was my fault.)

internal/reminder/metrics/provider.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// Set global meter provider
+	otel.SetMeterProvider(mp)
+
+	// Create metrics
+	meter := mp.Meter("reminder-service")
+	metrics, err := NewMetrics(meter)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create metrics: %w", err)
+	}
```
evankanderson, line 71: If you break the setup out into a static function, I'd probably have the calling reminder code look like:

```
if err := setupMetrics(ctx, cfg); err != nil {
  return err
}
reminder.metrics, err = NewMetrics(otel.Meter("reminder"))
if err != nil {
  return err
}
```

That separates the "system setup" of the metrics exporter from the provisioning of metrics for the main reminder loop, and makes it easier in the future to test the metrics without accidentally setting up a whole prometheus exporter.

internal/reminder/metrics/provider.go
```diff
@@ -0,0 +1,132 @@
...
+	mp := sdkmetric.NewMeterProvider(
+		sdkmetric.WithReader(prometheusExporter),
+		sdkmetric.WithResource(res),
+	)
+
+	// Set global meter provider
+	otel.SetMeterProvider(mp)
+
+	// Create metrics
+	meter := mp.Meter("reminder-service")
```
evankanderson, line 67:
```suggestion
	meter := mp.Meter("reminder")
```

internal/reminder/metrics/provider.go
```diff
@@ -0,0 +1,132 @@
...
+	select {
+	case err := <-errCh:
+		return err
+	case <-ctx.Done():
+		return p.Shutdown(ctx)
+	}
+}
+
+// Shutdown gracefully shuts down the metrics server
+func (p *Provider) Shutdown(ctx context.Context) error {
```
evankanderson, line 113: If you make this method private, then cancelling the context to `Start` or `NewProvider` becomes the only way to reach this code, which simplifies your lifecycle over having multiple callers.

evankanderson: (among other things, you can guarantee that `p` is not `nil`, because you don't call the `shutdown` method unless `p` is non-nil.)

internal/reminder/metrics/provider.go
```diff
@@ -0,0 +1,132 @@
...
+		log.Error().Err(err).Msg("Error shutting down meter provider")
+	}
+
+	return p.server.Shutdown(shutdownCtx)
+}
+
+// Metrics returns the metrics instance
+func (p *Provider) Metrics() *Metrics {
+	return p.metrics
+}
```
evankanderson, line 132: My main suggestion is to remove the `Provider` object in favor of a static setup function with cancellation, but if you do keep it, I'd suggesting making `Metrics` exported, rather than needing to add an accessor function.

internal/reminder/reminder.go
```diff
@@ -143,6 +163,11 @@ func (r *reminder) sendReminders(ctx context.Context) error {
 		return fmt.Errorf("error creating reminder messages: %w", err)
 	}
 
+	remMetrics := r.metricsProvider.Metrics()
+	if remMetrics != nil {
+		remMetrics.RecordBatch(ctx, int64(len(repos)))
```
evankanderson, line 168: If you make batches a histogram, this can just be a `Record()` on that histogram.

internal/reminder/metrics/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	// Update running average
+	m.totalBatches++
+	m.totalReminders += size
+	avgSize := float64(m.totalReminders) / float64(m.totalBatches)
+	m.AvgBatchSize.Record(ctx, avgSize)
+}
+
+func (m *Metrics) RecordSendDelay(ctx context.Context, delaySeconds float64) {
+	m.SendDelay.Record(ctx, delaySeconds)
+}
```
evankanderson, line 84: It looks like this function isn't doing anything -- you're already exposing `SendDelay`, and there's not much semantic benefit to `foo.RecordSendDelay(...)` over `foo.SendDelay.Record(...)`.  That also saves you from needing to write a comment.

If you transform `BatchSize` into a histogram as well, then you might be able to get rid of all the methods on `Metrics`, which would be nice.

evankanderson (commented):
A bunch of comments, mostly aiming to help simplify your life.  😁

Vyom-Yadav:
Opening this, I'm working on this.

internal/reminder/reminder.go
```diff
@@ -151,14 +173,19 @@ func (r *reminder) sendReminders(ctx context.Context) error {
 	repoIds := make([]uuid.UUID, len(repos))
 	for _, repo := range repos {
 		repoIds = append(repoIds, repo.ID)
+		if r.metrics != nil {
+			// sendDelay = Now() - ReminderLastSent - MinElapsed
+			reminderLastSent := repo.ReminderLastSent
+			if reminderLastSent.Valid {
+				r.metrics.SendDelay.Record(ctx, (time.Since(reminderLastSent.Time) - r.cfg.RecurrenceConfig.MinElapsed).Seconds())
```
Vyom-Yadav, line 180: I'm still not sure about this metric. Let's say a repo has been frequently updated every hour (due to GH webhook events) and `r.cfg.RecurrenceConfig.MinElapsed` is 2 hours. In this case, let's say after sending a reminder 18 hours back, every hour the repo was updated by GH webhook and wasn't eligible for a reminder. Now when it becomes eligible for the reminder `SendDelay`  would be recorded as `18h` which seems to be very wrong in this case. (I think we discussed this previously too)

internal/reminder/reminder.go
```diff
@@ -151,14 +173,19 @@ func (r *reminder) sendReminders(ctx context.Context) error {
...
+			} else {
+				// Recording metric for first time reminders (after how long was the initial reminder sent)
+				newSendDelay := time.Since(repoToLastUpdated[repo.ID]) - r.cfg.RecurrenceConfig.MinElapsed
+				r.metrics.NewSendDelay.Record(ctx, newSendDelay.Seconds())
```
Vyom-Yadav, line 184: Added a new metric for first time reminders. (I'm thinking this is _the_ definition of `sendDelay` and what we have above is wrong)

Vyom-Yadav (commented):
@evankanderson Because we reused eventer code in https://github.com/mindersec/minder/commit/a6d9b2295746fed45ad7fef0c24b021d3a7b1a6a, which created `pubWithMetrics` using `minder` namespace, there are some non-reminder metrics being outputted by reminder (e.g. `minder_eventer_publish_time_seconds_bucket`). 

I think this can cause problems, but I assume them to use the same infra, so not sure about how big of a problem that is. (Is this something we want to tackle in this PR?)

internal/reminder/metrics/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	// Current number of reminders in the batch
+	BatchSize metric.Int64Gauge
+
+	// Average batch size (updated on each batch)
+	AvgBatchSize metric.Float64Gauge
+
+	// For tracking average calculation
+	// TODO: consider persisting this to avoid reset on restart (maybe)
+	totalBatches   int64
+	totalReminders int64
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

internal/reminder/metrics/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	640,  // 10m 40s
+	1280, // 21m 20s
+}
+
+type Metrics struct {
+	// Time between when a reminder became eligible and when it was sent
+	SendDelay metric.Float64Histogram
+
+	// Current number of reminders in the batch
+	BatchSize metric.Int64Gauge
```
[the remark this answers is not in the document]
Vyom-Yadav: Done.

internal/reminder/metrics/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+var sendDelayBuckets = []float64{
+	0,    // immediate
+	10,   // 10 seconds
+	20,   // 20 seconds
+	40,   // 40 seconds
+	80,   // 1m 20s
+	160,  // 2m 40s
+	320,  // 5m 20s
+	640,  // 10m 40s
+	1280, // 21m 20s
```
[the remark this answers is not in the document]
Vyom-Yadav: Updated with:

```go
var delayBuckets = []float64{
	60,    // 1 minute
	300,   // 5 minutes
	600,   // 10 minutes
	1800,  // 30 minutes
	3600,  // 1 hour
	7200,  // 2 hours
	10800, // 3 hours
	18000, // 5 hours
	25200, // 7 hours
	36000, // 10 hours
}
```

internal/reminder/metrics/metrics.go
```diff
@@ -0,0 +1,84 @@
...
+	}, nil
+}
+
+func (m *Metrics) RecordBatch(ctx context.Context, size int64) {
+	// Update current batch size
+	m.BatchSize.Record(ctx, size)
+
+	// Update running average
+	m.totalBatches++
+	m.totalReminders += size
```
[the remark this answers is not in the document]
Vyom-Yadav: Good point, I think this should be doable now for multiple reminder processes as we can aggregate metrics which are being exposed.

internal/reminder/reminder.go
```diff
@@ -151,14 +176,19 @@ func (r *reminder) sendReminders(ctx context.Context) error {
...
 	for _, repo := range repos {
 		repoIds = append(repoIds, repo.ID)
+		if remMetrics != nil {
+			// sendDelay = Now() - ReminderLastSent - MinElapsed
+			reminderLastSent := repo.ReminderLastSent
+			if reminderLastSent.Valid {
+				remMetrics.SendDelay.Record(ctx, (time.Now().Sub(reminderLastSent.Time) - r.cfg.RecurrenceConfig.MinElapsed).Seconds())
+			} else {
+				// TODO: Should the send delay be zero if the reminder has never been sent?
+				remMetrics.SendDelay.Record(ctx, 0)
```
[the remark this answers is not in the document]
Vyom-Yadav: Exposing a new metric now.

pkg/config/reminder/metrics.go
Vyom-Yadav: I have configured this to localhost by default (changed the port). Can be overridden with configuration.

evankanderson (commented):
A few notes -- I think you're right about determining the metrics for "sendDelay".

If you want to be particularly cute, you could write:

```golang
sendDelay := time.Since(...) - ...MinElapsed
recorder := r.metrics.SendDelay
if !reminderLastSent.Valid {
  // Use a different metric for first-time reminders
  recorder = r.metrics.NewSendDelay
}
recorder.Record(ctx, sendDelay.Seconds())
```

internal/reminder/reminder.go
```diff
@@ -80,6 +84,20 @@ func (r *reminder) Start(ctx context.Context) error {
...
 
+	if r.cfg.MetricsConfig.Enabled {
+		go func() {
+			if err := r.startMetricServer(ctx); err != nil {
+				logger.Error().Err(err).Msg("error starting metrics server")
+			}
+		}()
+
+		var err error
+		r.metrics, err = metrics.NewMetrics(otel.Meter("reminder"))
```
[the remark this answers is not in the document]
Vyom-Yadav: No, you should feel strongly. There is a race condition here: (I wouldn't have checked until you pointed 🥇 )

```go
	if r.cfg.MetricsConfig.Enabled {
		go func() {
			if err := r.startMetricServer(ctx); err != nil {
				logger.Error().Err(err).Msg("error starting metrics server")
			}
		}()

		// Race condition here

		var err error
		r.metrics, err = metrics.NewMetrics(otel.Meter("reminder"))
		if err != nil {
			return err
		}
	}
```

`otel.SetMeterProvider(mp)` might not have been executed by the time `otel.Meter("reminder")` is run.

internal/reminder/metrics_server.go
```diff
@@ -0,0 +1,84 @@
...
+	shutdownCtx, shutdownRelease := context.WithTimeout(context.Background(), 5*time.Second)
+	defer shutdownRelease()
+
+	logger.Info().Msg("shutting down metrics server")
+
+	if err := mp.Shutdown(shutdownCtx); err != nil {
+		logger.Err(err).Msg("error shutting down metrics provider")
+	}
+
+	return server.Shutdown(shutdownCtx)
```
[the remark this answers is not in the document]
Vyom-Yadav: Yeah, the main goroutine goes home without waiting for his younger sibling, let me change it to send both of them back home together.

internal/reminder/reminder.go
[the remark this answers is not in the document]
Vyom-Yadav:
```go
// If this is called before a global MeterProvider is registered the returned
// Meter will be a No-op implementation of a Meter. When a global MeterProvider
// is registered for the first time, the returned Meter, and all the
// instruments it has created or will create, are recreated automatically from
// the new MeterProvider.
```

Hmm, although technically, it is a race condition, but I think it should be fine (still I'll change it). Evan, can it be possible that few metrics get recorded on the no-op implementation, and then when the MeterProvider is registered for the first time, those metrics are lost?

internal/reminder/metrics_server.go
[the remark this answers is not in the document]
evankanderson: As mentioned, I'm not too worried about hanging up in the middle of a metrics request, so don't contort the code too much to make it happen.  `exit()` is a _very_ effective garbage collector.

internal/reminder/reminder.go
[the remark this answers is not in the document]
evankanderson: That's a reasonable implementation for OTel, in my opinion.

My main reaction was that you went through the process of creating a `MeterProvider` in `startMetricsServer` and then 2 lines later in the outer function you go through OTel globals to fish it out again.  If the code was more separated, I'd say the globals made sense, but at this proximity, I'd go for explicit over spooky action at a distance.

internal/reminder/metrics/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, err
+	}
+
+	totalBatches, err := meter.Int64Counter(
+		"total_batches",
+		metric.WithDescription("Total number of batches processed"),
+	)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
Vyom-Yadav: `reminder_total_batches_total` and `reminder_batch_size_count` give the same result. `reminder_total_reminders_total` and `reminder_batch_size_sum` give the same result. So, I don't think those two extra metrics are useful. I've removed them.

internal/reminder/reminder.go
```diff
@@ -213,10 +243,11 @@ func (r *reminder) getEligibleRepositories(ctx context.Context, repos []db.Repos
 	for _, repo := range repos {
 		if t, ok := idToLastUpdate[repo.ID]; ok && t.Before(cutoff) {
 			eligibleRepos = append(eligibleRepos, repo)
+			eligibleReposLastUpdated[repo.ID] = t
 		}
 	}
 
-	return eligibleRepos, nil
+	return eligibleRepos, eligibleReposLastUpdated, nil
```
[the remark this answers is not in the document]
Vyom-Yadav: I could just return `idToLastUpdate` as well, but that contained non eligible repos as well (so my initial thought was to create a new map of only eligible repos - helps in debugging with breakpoints - but it's fine either way). I've changed now to return `idToLastUpdate`.

internal/reminder/reminder.go
```diff
@@ -74,21 +78,52 @@ func (r *reminder) Start(ctx context.Context) error {
...
 
+	metricsServerDone := make(chan struct{})
+
+	if r.cfg.MetricsConfig.Enabled {
+		metricsProviderReady := make(chan struct{})
+
+		go func() {
+			if err := r.startMetricServer(ctx, metricsProviderReady); err != nil {
+				logger.Fatal().Err(err).Msg("failed to start metrics server")
+			}
```
Vyom-Yadav, line 96: This will call `exit()` even when `server.Shutdown(shutdownCtx)` return an error, is this correct? (This will prevent the deferred stop call)

internal/reminder/reminder.go
```diff
@@ -74,21 +78,52 @@ func (r *reminder) Start(ctx context.Context) error {
...
+		case <-metricsProviderReady:
+			var err error
+			r.metrics, err = metrics.NewMetrics(otel.Meter("reminder"))
+			if err != nil {
+				return err
+			}
+		case <-ctx.Done():
+			logger.Info().Msg("reminder stopped")
+			return nil
+		}
```
Vyom-Yadav, line 110: I hope this isn't me too complicating stuff.

internal/reminder/reminder.go
```diff
@@ -74,21 +78,52 @@ func (r *reminder) Start(ctx context.Context) error {
...
+
 	r.ticker = time.NewTicker(interval)
-	defer r.Stop()
 
 	for {
 		select {
 		case <-ctx.Done():
+			if r.cfg.MetricsConfig.Enabled {
+				<-metricsServerDone
+			}
```
Vyom-Yadav, line 120: Same for this one (making sure I'm not complicating things)

internal/reminder/reminder.go
```diff
@@ -143,6 +178,10 @@ func (r *reminder) sendReminders(ctx context.Context) error {
 		return fmt.Errorf("error creating reminder messages: %w", err)
 	}
 
+	if r.metrics != nil {
+		r.metrics.RecordBatch(ctx, int64(len(repos)))
+	}
```
Vyom-Yadav, line 183: I could change this to a direct call, but have kept it as it is for now.

internal/reminder/reminder.go
```diff
@@ -151,13 +190,16 @@ func (r *reminder) sendReminders(ctx context.Context) error {
...
-	// - Gauge: Number of reminders in the current batch
-	// - UpDownCounter: Average reminders sent per batch
-	// - Histogram: reminder_last_sent time distribution
+			recorder := r.metrics.SendDelay
+			if !repo.ReminderLastSent.Valid {
+				recorder = r.metrics.NewSendDelay
+			}
+			recorder.Record(ctx, sendDelay.Seconds())
+		}
+	}
```
Vyom-Yadav, line 202: Now this all brings back the question that how useful is `ReminderLastSent` metric that we store in the main DB? Is there any other application for it? Right now it's just being used as a boolean for checking whether a repo has been reconciled or not (through reminder)

Vyom-Yadav:
@evankanderson Did you see this:

> Because we reused eventer code in https://github.com/mindersec/minder/commit/a6d9b2295746fed45ad7fef0c24b021d3a7b1a6a, which created pubWithMetrics using minder namespace, there are some non-reminder metrics being outputted by reminder (e.g. minder_eventer_publish_time_seconds_bucket).
> 
> I think this can cause problems, but I assume them to use the same infra, so not sure about how big of a problem that is. (Is this something we want to tackle in this PR?)

internal/reminder/reminder.go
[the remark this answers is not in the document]
evankanderson: My thought was primarily to avoid a bunch of copies and allocations.  You could also clear out the entries from `idToLastUpdate` as you filter them if you don't want to return the extra data?

evankanderson:
> @evankanderson Did you see this:
> 
> > Because we reused eventer code in [a6d9b22](https://github.com/mindersec/minder/commit/a6d9b2295746fed45ad7fef0c24b021d3a7b1a6a), which created pubWithMetrics using minder namespace, there are some non-reminder metrics being outputted by reminder (e.g. minder_eventer_publish_time_seconds_bucket).
> > I think this can cause problems, but I assume them to use the same infra, so not sure about how big of a problem that is. (Is this something we want to tackle in this PR?)

This isn't actually a big problem, because the pod / job labels (and the OTel MetricProvider Resource, if you're using the Otel binary protocol) will be different.  In fact, using the same metric name across different components makes it easier to build a dashboard that lets you "zoom out" and "zoom in" on how different components are handling common concerns (like queue lengths, latencies, etc).

internal/reminder/metrics/metrics.go
```diff
@@ -0,0 +1,79 @@
...
+		SendDelay:    sendDelay,
+		NewSendDelay: newSendDelay,
+		BatchSize:    batchSize,
+	}, nil
+}
+
+// RecordBatch records the metrics for a batch of reminders
+func (m *Metrics) RecordBatch(ctx context.Context, size int64) {
+	m.BatchSize.Record(ctx, size)
+}
```
evankanderson, line 79: This is a super-minor nit, but it feels like you could remove this method, and just have callers invoke `m.BatchSize.Record(ctx, size)`.

internal/reminder/metrics_server.go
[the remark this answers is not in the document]
evankanderson: Oh, now I see the other concern -- you may want to shut down the HTTP server _before_ the metrics provider.  Generally, I follow a LIFO order on teardown, with the idea that items constructed earlier probably don't depend on later items, but the other way may happen.

internal/reminder/reminder.go
evankanderson: Since `startMetricServer` blocks, this seems reasonable.  The other option would be to have `startMetricServer` return a `chan error` and execute asynchronously within the function (i.e. move 68-86 into the goroutine started by `startMetricServer`).

internal/reminder/reminder.go
evankanderson: I do think you're being overly conscientious for shutdown from `main`, but I'll allow it.

internal/reminder/reminder.go
evankanderson: Here, you might more easily do the following:

```golang
if r.cfg.MetricsConfig.Enabled {
  ...
} else {
  close(metricsServerDone)
}
```

And then you won't need the conditional here.

internal/reminder/reminder.go
evankanderson: I think the only advantage of using an indirect call is that you could hide the `if r.metrics != nil` inside the `RecordBatch` call if you use a pointer receiver.

Another option is to decide that metrics will simply always be on in Reminder -- I don't know if that will simplify the code, but I've always found the "maybe you don't want metrics" to be a strange notion in Minder that was there before I got here.

internal/reminder/reminder.go
evankanderson: Ah, you mean the column in the database?  Given where we're at now, maybe we don't need it?

Vyom-Yadav: Let's keep it for some time around, till we setup reminder with metrics to see if this can be potentially valuable. We can cleanup later if not. (Adding code requires more testing)

internal/reminder/reminder.go
Vyom-Yadav: I removed this function with a direct call. `if r.metrics != nil` is down in the code as well, so replacing it at one place breaks the consistency.

internal/reminder/reminder.go
Vyom-Yadav: I didn't understand that. The current shutdown sequence waits for the metrics server to be done. (waiting so that deferred call cannot be executed i.e. the `Stop()` call for reminder)

internal/reminder/reminder.go
Vyom-Yadav: Thank you, kindly!

internal/reminder/reminder.go
Vyom-Yadav: I feel the current option is fine.

I replaced `logger.Fatal().Err(err)` to be not `Fatal()`. OS will eventually clean up the socket and release the port.

internal/reminder/metrics/metrics.go
Vyom-Yadav: Yeah, I kept it initially, but I've removed it now.

Vyom-Yadav:
CI, flaking TestNatsChannel

```
❌ TestNatsChannel (5.09s)
      natschannel_test.go:95: Got {"msg":"hello"} from out1
      natschannel_test.go:95: Got {"msg":"hola"} from out1
      natschannel_test.go:100: timeout waiting for message 2
      natschannel_test.go:114: expected 4 messages, got 2: [0xc0003c1020 0xc000160240]
```

Vyom-Yadav:
Hey @evankanderson, sorry for the delay on this, I got caught with one thing after the other. I addressed the review comments, ptal and let me know if anything else needs to be changed. Thanks!

internal/reminder/reminder.go
```diff
@@ -74,21 +78,52 @@ func (r *reminder) Start(ctx context.Context) error {
 		return errors.New("reminder stopped, cannot start again")
 	default:
 	}
+	defer r.Stop()
```
evankanderson, line 81: This is kind of odd, since `Start` doesn't return, and would normally be exited by calling `Stop()`.  Add a comment here on why this is needed?

(I'm not saying it's wrong, I'm saying to drop a hint to prevent future authors from deleting the line.

internal/reminder/reminder.go
```diff
@@ -74,21 +78,52 @@ func (r *reminder) Start(ctx context.Context) error {
...
+
+	if r.cfg.MetricsConfig.Enabled {
+		metricsProviderReady := make(chan struct{})
+
+		go func() {
+			if err := r.startMetricServer(ctx, metricsProviderReady); err != nil {
+				logger.Err(err).Msg("failed to start metrics server")
+			}
+			close(metricsServerDone)
+		}()
```
evankanderson, line 98: It feels weird to me to end up needing to call `startMetricServer` in another thread and then deal with that, given that `startMetricServer` _also_ sets up a background thread (and then waits for it).

We end up with:

```
main
 \--> startMetricServer
          \--> http.ListenAndServe
```

When it feels like we should have:

```
main
\--> http.ListenAndServe
```

Possibly with a "closer" callback.

I only mention this because `Join`-ing goroutines is a bunch of code that distracts from the rest of what's going on, so I tend to prefer to minimize how often I need to block on a goroutine exiting.

internal/reminder/reminder.go
evankanderson: If you `close(metricsServerDone)` in the `else` block on 111, 118-120 and 124-126 can simply do `<-metricsServerDone` whether or not `r.cfg.MetricsConfig.Enabled` is true or false, because either case will close the channel.

I'd also be okay with leaking the metrics HTTP server if it doesn't cause harm elsewhere.

internal/reminder/metrics_server.go
```diff
@@ -0,0 +1,87 @@
...
+		sdkmetric.WithResource(res),
+	)
+
+	otel.SetMeterProvider(mp)
+
+	// Indicates that a global MeterProvider is available
+	close(mpReady)
+
+	mux := http.NewServeMux()
+	mux.Handle(metricsPath, promhttp.Handler())
```
evankanderson, line 54: Why do you need an `http.ServeMux` here, rather than simply passing `promhttp.Handler` to the `http.Server` below?

evankanderson (dismissed):
I'm willing to merge this as-is if you want, or you can try to simplify the startup code.  I can understand just wanting to be done with it, though.  :grin:

internal/reminder/metrics_server.go
Vyom-Yadav: Do we want to serve the metrics on the root path? I don't see any problem with that, but it's just that I've seen `/metrics` being commonly used.

internal/reminder/reminder.go
Vyom-Yadav: This defer here is mostly to handle context cancellation part. Depending on the invoker to clean up reminder after context cancellation isn't something I wanted to do. I added a comment on top.

internal/reminder/reminder.go
Vyom-Yadav: Yeah, that's a good point. It was confusing especially with the part of getting the otel meter provider ready and set. I modified the implementation to be like:

```
main
\--> http.ListenAndServer
\--> watch for stopping the metrics server
```

It can't be done in a single go routine, but I think this solution is better than what was there previously.

internal/reminder/reminder.go
Vyom-Yadav: Yeah, good point. I implemented this change.

Vyom-Yadav:
>  I can understand just wanting to be done with it, though. 😁

Not really, your reviews always teach me something, so deeply appreciated 🙇🏼‍♂️

internal/reminder/metrics_server.go
evankanderson: (On the other hand, I don't have a strong complaint about putting this at `/metrics`)

evankanderson (approved):

evankanderson:
I'm going to merge this, since I think it's as close to perfect as we're going to manage.

PR #4835, opened 2024-10-28
Validate the user invitation HTML templates

rdimitrov:
# Summary

The following PR:
* Adds validation for the user invitation HTML templates against HTML and JS injection
* Adds email validation for the user invitation feature based on RFC 5322
* Adds unit tests

## Change Type

# Testing

# Review Checklist:


internal/email/email.go
```diff
@@ -61,56 +99,88 @@ func NewMessage(
...
-		project, role, sponsor, inviteURL)
-}
-
 // getEmailSubject returns the subject for the email based on the message payload
 func getEmailSubject(project string) string {
 	return fmt.Sprintf("You have been invited to join the %s organization in Minder", project)
 }
+
+// isValidField checks if the string is empty or contains HTML or JavaScript injection
+func isValidField(str string) error {
```
evankanderson, line 145: I think this is also covered by html/template -- do we need both?

internal/email/email.go
```diff
@@ -61,56 +99,88 @@ func NewMessage(
...
-		RecipientEmail:   inviteeEmail,
-		MinderURL:        minderURL,
-		TermsURL:         DefaultMinderTermsURL,
-		PrivacyURL:       DefaultMinderPrivacyURL,
-		SignInURL:        minderURL,
-		RoleName:         role,
-		RoleVerb:         authz.AllRolesVerbs[authz.Role(role)],
+// getHTMLBodyString returns the HTML body for the email based on the message payload.
+// If there is an error creating the HTML body, it will try to return the text body instead
+func getEmailBodyHTML(ctx context.Context, data bodyData) string {
```
evankanderson, line 104: Why does this function not return an error?  In particular, it seems like we want to distinguish between "template rendered successfully" and "template failed rendering" before we send the email.

internal/email/email.go
rdimitrov: I decided to add it because html/template was not erroring if it finds html in the template input, instead it escapes it and so we end up with things like `&lt;script&gt;`

internal/email/email.go
rdimitrov: Fixed it 👍

internal/email/email.go
```diff
@@ -46,12 +63,42 @@ func NewMessage(
...
+		MinderURL:        minderURLBase,
+		TermsURL:         DefaultMinderTermsURL,
+		PrivacyURL:       DefaultMinderPrivacyURL,
+		SignInURL:        minderURLBase,
+		RoleName:         role,
+		RoleVerb:         authz.AllRolesVerbs[authz.Role(role)],
+	}
+
+	// Validate the data source template for HTML injection attacks or empty fields
+	err = validateDataSourceTemplate(&data)
```
evankanderson, line 81: You could add this method to `bodyData` now that the struct is declared at the top-level.  (Probably as `Validate`)

internal/email/email.go
```diff
@@ -46,12 +63,42 @@ func NewMessage(
...
+	if err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error validating data source")
+		return nil, fmt.Errorf("error validating data source: %w", err)
```
evankanderson, line 84: General rule of thumb for error handling: either pass the error up the chain, _or_ log the error and return `nil`.  If you do both, you'll get a lot of duplicate error messages in the logs, as each components logs the error _and_ passes it up to the next level for it to _also_ log the error.

internal/email/email.go
```diff
@@ -61,56 +108,84 @@ func NewMessage(
...
 
-	// Parse the template
-	tmpl, err := templates.Parse(bodyHTML)
+	bodyHTMLTmpl, err := util.NewSafeHTMLTemplate(&bHTML, "body-invite-html")
 	if err != nil {
-		zerolog.Ctx(ctx).Error().Err(err).Msg("error parsing the HTML template for email invitations")
-		// Default to the text body
-		return getEmailBodyText(inviteURL, sponsor, project, role)
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error creating the HTML template for email invitations")
+		return getEmailBodyText(ctx, data)
```
evankanderson, line 120: Is it safe to return a text body template if the HTML template fails to load?  I suspect not, as there are characters that are safe in text that are unsafe in HTML.

internal/email/email.go
```diff
@@ -61,56 +108,84 @@ func NewMessage(
...
 	}
-	// Execute the template
-	var b strings.Builder
-	if err := tmpl.Execute(&b, data); err != nil {
-		return ""
+	err = bodyHTMLTmpl.Execute(ctx, &b, data, EmailBodyMaxLength)
+	if err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error executing the HTML template for email invitations")
+		return getEmailBodyText(ctx, data)
```
evankanderson, line 125: Ditto.

Note that if you switch this to a `return "", err`, then you'll be returning an error, and won't need the log line above.  (Currently, you're _handling_ the error by attempting to call the text template.)

internal/email/email.go
```diff
@@ -61,56 +108,84 @@ func NewMessage(
...
-	// TODO: Load the email template from elsewhere
+// getHTMLBodyString returns the HTML body for the email based on the message payload.
+// If there is an error creating the HTML body, it will try to return the text body instead
+func getEmailBodyHTML(ctx context.Context, data bodyData) (string, error) {
+	var b strings.Builder
+	bHTML := bodyHTML
 
-	// Parse the template
-	tmpl, err := templates.Parse(bodyHTML)
+	bodyHTMLTmpl, err := util.NewSafeHTMLTemplate(&bHTML, "body-invite-html")
```
evankanderson, line 117: Don't fix this now, but we should be able to load this template _once_ and then invoke it a number of times, rather than loading it every single time.  (And then, if we get a nice data push set up, we'll need to be able to load it dynamically, but right now this is either being put there by an init container, or being put there by the container install.)

Also, rather than using `bHTML`, you should be able to put the file in `/cmd/server/kodata/templates/email.tmpl` and then read it from `$KO_DATA_PATH/templates/email.tmpl`.  I'd hold off on that for a different PR, though.

internal/email/email.go
```diff
@@ -61,56 +108,84 @@ func NewMessage(
...
+	if err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error executing the text template for email invitations")
+		return "", err
+	}
+	return b.String(), nil
 }
 
 // getEmailSubject returns the subject for the email based on the message payload
 func getEmailSubject(project string) string {
 	return fmt.Sprintf("You have been invited to join the %s organization in Minder", project)
```
evankanderson, line 150: Do we need to escape / bound the length and contents of `project` here?

internal/email/email.go
evankanderson: Can you add a comment that we want to detect and error, rather than escape?

internal/email/email.go
```diff
@@ -61,56 +108,84 @@ func NewMessage(
...
+// isValidField checks if the string is empty or contains HTML or JavaScript injection
+func isValidField(str string) error {
+	// Check for HTML content
+	escapedHTML := template.HTMLEscapeString(str)
+	if escapedHTML != str {
+		return fmt.Errorf("string %s contains HTML injection", str)
+	}
+
+	// Check for JavaScript content separately
+	escapedJS := template.JSEscaper(str)
```
evankanderson, line 162: It looks like `JSEscapeString` should be simpler / faster here.

internal/email/email.go
```diff
@@ -61,56 +108,84 @@ func NewMessage(
...
+			strVal := field.String()
+
+			// Execute your function on the field value
+			err := isValidField(strVal)
+			if err != nil {
+				return fmt.Errorf("field %s failed validation - %s", v.Type().Field(i).Name, strVal)
+			}
+		}
+	}
+	return nil
```
evankanderson, line 190: I'm not thrilled with the reflection here -- in particular, we should know that each field is settable and of type String, so we should be able to error if it is not.  Similarly, I'd prefer to have a more concrete argument than `any`, because this doesn't sanitize other fields, like if we added something which contained further string fields, or implements `fmt.Stringer()`, but is not a string itself.

internal/email/email_test.go
```diff
@@ -0,0 +1,160 @@
...
+		{"Just plain text", false, ""},
+
+		// Test case 2: String with HTML tags
+		{"<b>Bold Text</b>", true, "string <b>Bold Text</b> contains HTML injection"},
+
+		// Test case 3: String with HTML entity
+		{"This is a test &amp; example.", true, "string This is a test &amp; example. contains HTML injection"},
+
+		// Test case 4: String with multiple HTML entities
+		{"This &amp; that &lt; should &gt; work.", true, "string This &amp; that &lt; should &gt; work. contains HTML injection"},
```
evankanderson, line 26: Looking at this, what prevents us from getting a 1MB string that we try to put into the error message?

(Also, it feels like we should use `%q` here for the text that we're objecting to for clarity.)

internal/invites/service.go
```diff
@@ -342,3 +355,9 @@ func getInviteUrl(emailCfg serverconfig.EmailConfig, userInvite db.UserInvite) (
 	}
 	return inviteURL, nil
 }
+
+// isEmail is used to validate if the email is a valid email address according to RFC 5322
+func isEmail(emailString string) bool {
+	_, err := mail.ParseAddress(emailString)
```
evankanderson, line 361: Should we be using only the `address` part for the rest of the function?  I'm not sure we've thought about `"Joe Bob <target@foo.com>"` and `"Oh gosh, I really dislike you <target@foo.com>"` in the rest of our code.

evankanderson: (If so, we could just return `string` or `""`)

evankanderson (commented):
Sorry, just got a chance to review this on computer rather than on phone, and I've got a few more comments.

internal/invites/service.go
rdimitrov: We can, but do we want to support such kind of inputs? For now at least I assume it's much cleaner to fail if the input is not strictly an email.

internal/email/email_test.go
rdimitrov: What do you think is a sane length limitation for these values? Is 200 chars too much?

internal/email/email.go
rdimitrov: Thanks 👍 I updated the reflect part to not take any as input but also to enforce that all fields are of type string. If this happens to be wrong, we'll fail the validation and error out.

internal/email/email.go
```diff
@@ -43,15 +61,47 @@ func NewMessage(
...
+		TermsURL:         DefaultMinderTermsURL,
+		PrivacyURL:       DefaultMinderPrivacyURL,
+		SignInURL:        minderURLBase,
+		RoleName:         role,
+		RoleVerb:         authz.AllRolesVerbs[authz.Role(role)],
 	}
 
+	// Validate the data source template for HTML injection attacks or empty fields
+	err = data.Validate()
+	if err != nil {
```
evankanderson, line 83:
```suggestion
	if err := data.Validate(); err != nil {
```

internal/email/email.go
```diff
@@ -43,15 +61,47 @@ func NewMessage(
...
+		return nil, err
+	}
+	// Create the HTML and text bodies
+	strBodyHTML, err := data.GetEmailBodyHTML(ctx)
```
evankanderson, line 87: Oh, these are really nice as methods on the data struct!

internal/email/email.go
```diff
@@ -61,56 +111,88 @@ func NewMessage(
...
+			strVal := field.String()
+			// Execute your function on the field value
+			err := isValidField(strVal)
+			if err != nil {
+				return fmt.Errorf("field %s failed validation - %s", v.Type().Field(i).Name, strVal)
+			}
+		} else {
+			// Return an error if the field is not settable or not a string
+			return fmt.Errorf("field %s is not settable or a string", v.Type().Field(i).Name)
+		}
```
evankanderson, line 195: You can avoid some indentation if you reverse this and take advantage of the early-exit of `return`:

```suggestion
		if !field.CanSet() || field.Kind() != reflect.String || field.Type() != reflect.TypeOf("") {
			return fmt.Errorf("field %s is not settable or a string", v.Type().Field(i).Name)
		}
		strVal := field.String()
		err := isValidField(strVal)
		if err != nil {
			return fmt.Errorf("field %s failed validation - %s", v.Type().Field(i).Name, strVal)
		}
```

internal/email/email_test.go
evankanderson: I think 200 characters is probably sufficient.

internal/email/email_test.go
```diff
@@ -0,0 +1,167 @@
...
+		})
+	}
+}
+
+func TestValidateDataSourceTemplate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		input          bodyData
+		expectedErr    bool
```
evankanderson, line 65: Do you need `expectedErr` separately from `expectedErrMsg != ""`?

internal/invites/service.go
evankanderson: My point is that `mail.ParseAddress` _will_ accept `"An unpleasant joke from recent politics <target@foo.com>"`, and the `Name` field will be set to `"An unpleasant joke from recent politics"`.

(And I was suggesting we should return just the `parsed.Address` part of the `ParseAddress` result.

evankanderson (commented):
A few nits (which you can ignore) aligning with what I think of as idiomatic go, and one bit on `isEmail` that we probably _do_ want to fix.

internal/invites/service.go
evankanderson: See https://github.com/bufbuild/protovalidate-go/blob/1446f1e1d576e67382edcb4863f76414e2eb2256/celext/lib.go#L426 for fairly-complete email address validation.

evankanderson: (And, I'm okay removing this bit of code in favor of #4840 )

internal/email/email_test.go
rdimitrov: Thanks 👍 Removed the obsolete bool for expected error 👍

internal/email/email.go
rdimitrov: Applied the change 👍

internal/email/email.go
rdimitrov: Applied 👍

internal/invites/service.go
rdimitrov: I decided to remove it in favour of the validation checks we'll introduce 👍

evankanderson (approved):

PR #4849, opened 2024-10-31
Add pkg/querier package

rdimitrov:
# Summary

The following PR adds a querier package to pkg. Its purpose is to expose a set of tools to interact with the Minder database.
The wrappers for managing the ruletypes and profiles are wrapped around their service alternatives as making changes to these entries without validating them can cause conflicts in the profile references and evaluations.
 
## Change Type

# Testing

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -2756,3 +2758,11 @@ message UpstreamEntityRef {
...
     google.protobuf.Struct properties = 3;
 }
+
+// BundleSubscription represents a subscription to a bundle.
+message BundleSubscription {
+    string id = 1;
+    string project_id = 2;
+    string bundle_id = 3;
+    string current_version = 4;
+}
```
evankanderson, line 2781: It feels like this should be someplace different from the external API, since it's not referenced by any clients.

We have an `internal/proto`, but maybe it makes sense to have a pkg/proto (and exclude that from coverage)?

pkg/querier/querier.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+// Store interface provides functions to execute db queries and transactions
+type Store interface {
+	ProjectHandlers
+	RuleTypeHandlers
+	ProfileHandlers
+	BundleHandlers
+	Close()
+	Commit() error
```
evankanderson, line 31: Do you need a way to start a transaction, given that you have a `Commit` method?

evankanderson: Oh, it looks like you need to throw away the object (and the connection pool) after each transaction?

evankanderson (commented):
How are you feeling about this approach having started it?

I'm a bit nervous about the interaction with transactions; I think I'd rather expose methods to manage a `sql.Tx` outside the interface than try to bundle it into the interface if you need transaction support.  (Or you could include the transaction in the context, which might be a nice API)

database/query/projects.sql
```diff
@@ -101,4 +101,8 @@ WHERE id = $1 RETURNING *;
 -- name: UpdateProjectMeta :one
 UPDATE projects
 SET metadata = $2
-WHERE id = $1 RETURNING *;
\ No newline at end of file
+WHERE id = $1 RETURNING *;
+
+-- name: ListAllParentProjects :many
+SELECT * FROM projects
+WHERE parent_id IS NULL AND is_organization = FALSE;
```
JAORMX, line 108: The `is_organization` boolean is now unused. We should probably just get rid of this column.

rdimitrov: I'll file an issue so we fix this separately 👍

rdimitrov: https://github.com/mindersec/minder/issues/4853

JAORMX: I'd rename this to `ListRootProjects` to avoid confusion. Since a parent project may itself have a parent.

eleftherias (approved):

pkg/querier/querier.go
```diff
@@ -0,0 +1,136 @@
...
+	BeginTx() (Querier, error)
+}
+
+// Querier interface provides functions to interact with the Minder database using transactions
+type Querier interface {
+	ProjectHandlers
+	RuleTypeHandlers
+	ProfileHandlers
+	BundleHandlers
+	CommitTx() error
```
evankanderson, line 39: I'd probably call this `Commit` and `Cancel`, and document that the object can't be used after calling these.

pkg/querier/querier.go
```diff
@@ -0,0 +1,136 @@
...
+		err := t.store.Rollback(t.tx)
+		// Clear the transaction and reset the querier to the store
+		t.tx = nil
+		t.querier = t.store
+		return err
+	}
+	return fmt.Errorf("no transaction to cancel")
+}
+
+// initDb function initializes the database connection and transaction details, if needed
```
evankanderson, line 115: This doesn't do anything with transactions, does it?

pkg/querier/querier.go
```diff
@@ -0,0 +1,136 @@
...
+	querier    db.ExtendQuerier
+	ruleSvc    ruletypes.RuleTypeService
+	profileSvc profiles.ProfileService
+}
+
+// Ensure Type implements the Querier interface
+var _ Querier = (*Type)(nil)
+
+// New creates a new instance of the querier
+func New(ctx context.Context, config *server.Config) (Store, func(), error) {
```
evankanderson, line 56: You may want to define a type for the second return argument (it's not clear what you do with a `func()` on its own).

pkg/querier/querier.go
```diff
@@ -0,0 +1,136 @@
...
+	ProjectHandlers
+	RuleTypeHandlers
+	ProfileHandlers
+	BundleHandlers
+	CommitTx() error
+	CancelTx() error
+}
+
+// Type represents the querier type
+type Type struct {
```
evankanderson, line 44: Why is this type exported?

proto/minder/v1/minder.proto
```diff
@@ -1023,6 +1023,8 @@ message Project {
     // display_names are short *non-unique* strings to provide
     // a user-friendly name for presentation in lists, etc.
     string display_name = 5;
+    bool is_organization = 8;
```
evankanderson, line 1026: Discussed offline, but I don't think we want to change our public API in this PR.

pkg/querier/querier.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to setup database: %w", err)
+	}
+	// Get a watermill event handler
+	evt, err := eventer.New(ctx, nil, &config.Events)
+	if err != nil {
+		return nil, nil, fmt.Errorf("unable to setup eventer: %w", err)
+	}
+	// Return the new Type
+	return &querierType{
```
evankanderson, line 83: Oh, this is cute -- you're using querierType as both `Store` and `Querier`.  I _think_ that won't hurt you; anyone who goes to cast one to the other gets what's coming to them.

pkg/querier/querier.go
```diff
@@ -0,0 +1,151 @@
...
+		store:      store,
```
evankanderson, line 84: You probably don't need both `querier` and `store` with the new implementation, but I'm not going to begrudge you 8 bytes.

pkg/querier/querier.go
```diff
@@ -0,0 +1,151 @@
...
+// Commit commits the transaction
+func (q *querierType) Commit() error {
+	if q.tx != nil {
+		err := q.store.Commit(q.tx)
+		// Clear the transaction and the querier
+		q.tx = nil
+		q.querier = nil
+		return err
+	}
+	return fmt.Errorf("no transaction to commit")
```
evankanderson, line 115: No need to fix here, but I'd tend to use the "early exit" model for these sorts of checks:

```suggestion
	if q.tx == nil {
		return fmt.Errorf("no transaction to commit")
	}
	err := q.store.Commit(q.tx)
	// Clear the transaction and the querier
	q.tx = nil
	q.querier = nil
	return err
```

When reading code, I find that I need to keep a mental stack open whenever I'm in a conditional to recall what the code path is with and without the conditional.  Having conditionals that close quickly with early returns lets me retire that call stack in my head, and leaves the "most common" code path running to the end of the function.

pkg/querier/querier.go
```diff
@@ -0,0 +1,151 @@
...
+
+// initDatabase function initializes the database connection
+func initDatabase(ctx context.Context, cfg *server.Config) (db.Store, func(), error) {
+	zerolog.Ctx(ctx).Debug().
+		Str("name", cfg.Database.Name).
+		Str("host", cfg.Database.Host).
+		Str("user", cfg.Database.User).
+		Str("ssl_mode", cfg.Database.SSLMode).
+		Int("port", cfg.Database.Port).
+		Msg("connecting to minder database")
```
evankanderson, line 138: We log these inside `GetDBConnection` too, don't we?

evankanderson (approved):

evankanderson:
To be clear, the above are nits, and I'm fine with the code going in as-is.

pkg/querier/querier.go
rdimitrov: You know this unlocks my embedded programming side if you do that 😁

evankanderson (approved):

PR #4852, opened 2024-10-31
Allow proto email to be empty

rdimitrov:
# Summary

The following PR updates the email proto validation to allow for email to be empty.

## Change Type

# Testing

# Review Checklist:


jhrozek (approved):
the code is fine but can you help me understand the use-case of when we want that field to be empty?

rdimitrov:
> the code is fine but can you help me understand the use-case of when we want that field to be empty?

It's because we reuse the same message between sending an invitation (by email) and making role assignment (by subject). I think we should fix this in the future and make them separate

proto/minder/v1/minder.proto
```diff
@@ -2313,27 +2313,42 @@ message RoleAssignment {
     // subject is the subject to which the role is assigned.
     string subject = 2;
     // display_name is the display name of the subject.
-    string display_name = 5 [(buf.validate.field).string = {
-        pattern: "[A-Za-z][- [:word:]\\[\\]\\(\\)]*",
-        max_len: 200,
-    }, (buf.validate.field).ignore_empty = true];
+    string display_name = 5 [
+        (buf.validate.field).string = {
+            pattern: "[A-Za-z][- [:word:]\\[\\]\\(\\)]*",
```
evankanderson, line 2318: I'm not sure you want names `[ to include brackets ]`

proto/minder/v1/minder.proto
```diff
@@ -2313,27 +2313,42 @@ message RoleAssignment {
...
+        (buf.validate.field).ignore_empty = true
+    ];
     // project is the project in which the role is assigned.
-    optional string project = 4 [(buf.validate.field).string = {
-        pattern: "[A-Za-z][- [:word:]\\[\\]\\(\\)]*",
-        max_len: 200,
-    }, (buf.validate.field).ignore_empty = true];
+    optional string project = 4 [
+        (buf.validate.field).string = {
+            pattern: "[A-Za-z][- [:word:]\\[\\]\\(\\)]*",
```
evankanderson, line 2326: Is the project a UUID?

proto/minder/v1/minder.proto
rdimitrov: I should test regex before using them 😃 This is the regex you suggested in the previous PR but I admit I did not verified it. Let me check 👍

evankanderson: That regex allows for capital or lowercase, followed by `-`, space, A-Z, a-z, 0-9, `_`, square brackets, and parenthesis.  But I'm not sure that's what we want for the `display_name` field, which seems to be taken from the GitHub username / slug, which may be closer to `[-[:word:]]+`, or `[A-Za-z][-[:word:]]*`.

proto/minder/v1/minder.proto
rdimitrov: oh, correct 👍 I'll fix that 👍

proto/minder/v1/minder.proto
rdimitrov: I think we also allow the user to set this, so I guess [ ] can happen? Of course if we want to allow for it.

PR #4891, opened 2024-11-06
Add proto validation for cursor indexes

rdimitrov:
# Summary

The following PR adds proto validation for cursor indexes ensuring only base64 encoded strings are accepted.

Ref https://github.com/stacklok/minder-stories/issues/94

## Change Type

# Testing

# Review Checklist:


eleftherias (approved):

proto/minder/v1/minder.proto
```diff
@@ -90,8 +90,14 @@ extend google.protobuf.MethodOptions {
 message Cursor {
     // cursor is the index to start from within the collection being
     // retrieved. It's an opaque payload specified and interpreted on
-    // an per-rpc basis.
-    string cursor = 1;
+    // an per-rpc basis. Note it is base64 encoded.
+    string cursor = 1 [
+        (buf.validate.field).string = {
+            pattern: "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$",
```
evankanderson, line 96: Can we simplify this to `"^[[:word:]=]*$"`?

rdimitrov: The one you suggest seems more loose?

proto/minder/v1/minder.proto
```diff
@@ -90,8 +90,14 @@ extend google.protobuf.MethodOptions {
 message Cursor {
     // cursor is the index to start from within the collection being
     // retrieved. It's an opaque payload specified and interpreted on
-    // an per-rpc basis.
-    string cursor = 1;
+    // an per-rpc basis. Note it is base64 encoded.
```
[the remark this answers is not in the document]
rdimitrov: I'll update the comment because we do enforce the base64 part in the handler 👍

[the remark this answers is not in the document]
evankanderson: My point was that saying "it's opaque" means it can be base64-encoded now, and some other pattern later.

proto/minder/v1/minder.proto
evankanderson: It is.  We're concerned with very unexpected input, but presumably we will validate the cursor more completely after base64-decoding it, so we just want to rule out truly unexpected characters here.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
evankanderson: I'd leave the comment as-is, and enforce length + general character limits, rather than overspecifying on base64 now, and having clients break later if we change the design.

proto/minder/v1/minder.proto
rdimitrov: I see. Alright, make sense 👍

proto/minder/v1/minder.proto
```diff
@@ -91,7 +91,13 @@ message Cursor {
     // cursor is the index to start from within the collection being
     // retrieved. It's an opaque payload specified and interpreted on
     // an per-rpc basis.
-    string cursor = 1;
+    string cursor = 1 [
+        (buf.validate.field).string = {
+            pattern: "^[[:word:]=]*$",
+            max_len: 200,
+        },
+        (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
```
evankanderson, line 99: If you use `*` as the pattern, you don't actually need this, since it will match a zero-length string.  It's fine to leave it, though.

evankanderson (approved):

proto/minder/v1/minder.proto
rdimitrov: Fixed it in https://github.com/mindersec/minder/pull/4895 👍

PR #4910, opened 2024-11-07
Exclude profiles with labels from history.

blkt:
# Summary

History list showed evaluation results even for profiles having labels. This is different from the behaviour of Profile list and status, which only showed results for profiles without labels.

This change makes the behaviour of History list uniform with Profile list and status.

Fixes stacklok/minder-stories#102

## Change Type

# Testing

Unit tested.

# Review Checklist:


database/query/eval_history.sql
```diff
@@ -158,6 +158,8 @@ SELECT s.id::uuid AS evaluation_id,
    AND (sqlc.narg(tots)::timestamp without time zone IS NULL OR  s.evaluation_time < sqlc.narg(tots))
    -- implicit filter by project id
    AND j.id = sqlc.arg(projectId)
+   -- implicit filter, exclude all labelled profiles
+   AND p.labels = ARRAY[]::TEXT[]
```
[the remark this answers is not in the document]
blkt: It's a matter of choice, let me bring this up internally.

proto/minder/v1/minder.proto
```diff
@@ -3271,6 +3271,23 @@ message ListEvaluationHistoryRequest {
     // Timestamp representing the end time of the selection window.
     google.protobuf.Timestamp to = 9;
 
+    // Filter evaluation history to only those matching the specified labels.
+    //
+    // The default is to return all user-created profiles; the string "*" can
+    // be used to select all profiles, including system profiles.  This syntax
+    // may be expanded in the future.
+    repeated string label_filter = 11 [
```
eleftherias, line 3279: Should this be a single string, similar to what the evaluation history and list profiles requests have?

internal/history/models.go
```diff
@@ -357,6 +374,29 @@ func (filter *listEvaluationFilter) ExcludedProfileNames() []string {
 	return filter.excludedProfileNames
 }
 
+func (filter *listEvaluationFilter) AddLabel(label string) error {
+	if label == "!*" {
```
eleftherias, line 378: This looks like the same logic that exists in `LabelsFromFilter`, should it be centralized somewhere so it's not duplicated? https://github.com/mindersec/minder/blob/main/internal/db/domain.go#L49

proto/minder/v1/minder.proto
blkt: Filters usable in the history-list endpoint are all repeated string so that they can be sent (e.g. via a REST GET) as separate entries (e.g. `?label_filter=foo&label_filter=bar`), as comma separated strings (e.g. `?label_filter=foo,bar`), or as a combination of the two.
The handler is responsible for splitting the string and populating the struct input to the service/db layer as an array of strings. Layers outside of the control plane only reason in terms of array of strings.

My understanding of `label_filter` in Profiles is that it can be only passed as [comma separated string](https://github.com/mindersec/minder/blob/main/internal/db/domain.go#L58), which is supported in this case as well. Additionally, the database layer is responsible for splitting the string, while here the responsibility is separate, which hopefully makes lower interfaces clearer.

TL;DR: it already support comma separated strings of labels, I would keep it this way.

internal/history/models.go
blkt: I agree that this logic should be shared, but history list RPC splits the responsibility between the controlplane and the service layer, while `LabelsFromFilter` wraps it in the database layer.

I would update how the `LabelsFromFilter` works by doing the same as done here, but in a separate PR. I can open a task for proper prioritization.

eleftherias (approved):

eleftherias (approved):

PR #4921, opened 2024-11-08
Support the release entity for the GitHub provider

JAORMX:
# Summary

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/providers/github/properties/release.go
+++ b/internal/providers/github/properties/release.go
@@ -21,10 +21,6 @@ const (
 	ReleasePropertyOwner = "github/owner"
 	// ReleasePropertyRepo represents the github repo
 	ReleasePropertyRepo = "github/repo"
-	// ReleasePropertyTag represents the github release tag name.
-	ReleasePropertyTag = "github/tag"
-	// ReleasePropertyBranch represents the github release branch
-	ReleasePropertyBranch = "github/branch"
 )
 
 // ReleaseFetcher is a property fetcher for releases
@@ -43,8 +39,10 @@ func NewReleaseFetcher() *ReleaseFetcher {
 						properties.PropertyName,
 						properties.PropertyUpstreamID,
 						// general release
-						ReleasePropertyTag,
-						ReleasePropertyBranch,
+						properties.ReleasePropertyTag,
+						properties.ReleasePropertyBranch,
+						ReleasePropertyOwner,
+						ReleasePropertyRepo,
 					},
 					wrapper: getReleaseWrapper,
 				},
@@ -62,7 +60,7 @@ func (_ *ReleaseFetcher) GetName(props *properties.Properties) (string, error) {
 		return "", fmt.Errorf("failed to get repo name: %w", err)
 	}
 
-	tag, err := props.GetProperty(ReleasePropertyTag).AsString()
+	tag, err := props.GetProperty(properties.ReleasePropertyTag).AsString()
 	if err != nil {
 		return "", fmt.Errorf("failed to get tag name: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


internal/providers/github/properties/release.go
```diff
@@ -0,0 +1,154 @@
...
+	tag, err := props.GetProperty(ReleasePropertyTag).AsString()
+	if err != nil {
+		return "", fmt.Errorf("failed to get tag name: %w", err)
+	}
+
+	return getReleaseNameFromParams(owner, repo, tag), nil
+}
+
+func getReleaseNameFromParams(owner, repo, tag string) string {
+	if owner == "" {
```
evankanderson, line 74: This seems like a weird case -- does it actually happen, or could we put in a complaint?

evankanderson (commented):
Does this create release entities, or just set us up for doing so in the future?

It looks (right now), like there's still more work before you can write a rule against these release objects -- are they created automatically (like PRs) after registering a repo?

I'm assuming (like PRs), that we will only start looking at releases that are modified after Minder is installed?

JAORMX:
> Does this create release entities, or just set us up for doing so in the future?
> 

It creates the release entities.

> It looks (right now), like there's still more work before you can write a rule against these release objects -- are they created automatically (like PRs) after registering a repo?
> 
> I'm assuming (like PRs), that we will only start looking at releases that are modified after Minder is installed?


Like PRs, these are created after registering a repo, as they get created. You could already start writing policies for these; the main issue we have right now is that we don't have an API to get general entities like these.

internal/providers/github/properties/release.go
```diff
@@ -0,0 +1,154 @@
...
+	"net/http"
+
+	go_github "github.com/google/go-github/v63/github"
+
+	"github.com/mindersec/minder/internal/entities/properties"
+	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+	v1 "github.com/mindersec/minder/pkg/providers/v1"
+)
+
+// Release Properties
```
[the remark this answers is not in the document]
JAORMX: that's right. This applies to all of our entity properties.

internal/providers/github/properties/release.go
evankanderson: Let's make this an error and/or log at warning/error -- I think it's valid input to the GitHub API, but we shouldn't be getting _output_ in this non-normalized form.

evankanderson (approved):

internal/providers/github/properties/release.go
```diff
@@ -0,0 +1,154 @@
...
+	if fetchErr != nil {
+		if result != nil && result.StatusCode == http.StatusNotFound {
+			return nil, v1.ErrEntityNotFound
+		}
+		return nil, fmt.Errorf("failed to fetch release: %w", fetchErr)
+	}
+
+	return map[string]any{
+		properties.PropertyUpstreamID: properties.NumericalValueToUpstreamID(release.GetID()),
+		properties.PropertyName:       getReleaseNameFromParams(owner, repo, release.GetTagName()),
```
evankanderson, line 115: It looks like GitHub's `name` here is non-unique, but may still be worth exposing for e.g. filtering with CEL.  (Apply / omit evalution only to releases with certain names)

internal/providers/github/properties/release.go
```diff
@@ -0,0 +1,154 @@
...
+// Release Properties
+const (
+	// ReleasePropertyOwner represents the github owner
+	ReleasePropertyOwner = "github/owner"
+	// ReleasePropertyRepo represents the github repo
+	ReleasePropertyRepo = "github/repo"
+	// ReleasePropertyTag represents the github release tag name.
+	ReleasePropertyTag = "github/tag"
+	// ReleasePropertyBranch represents the github release branch
+	ReleasePropertyBranch = "github/branch"
```
evankanderson, line 27: It was suggested we should have a common-top-level property for these, and for the commit SHA.

evankanderson (commented):
A few more comments.

It looks like *GitHub* uses the tag name as the unique identifier for releases, and you can have duplicate "name" (titles, really) for releases, but that my not be universal -- it feels like `getReleaseNameFromParams` will probably need to vary based on provider.

evankanderson (approved):
One nit on https://github.com/mindersec/minder/pull/4921/files#r1871541959 that I'd like to see us track (because having two different name formats makes _everything else_ harder).

evankanderson:
If we merge this, we should make sure to follow up with a PR to restore the -0.4% code coverage drop from adding this new code.

rdimitrov (approved):

PR #4953, opened 2024-11-12
Rule Type schema validation: change library and apply defaults

JAORMX:
# Summary

This changes the underlying library to something that's actually
maintained: github.com/santhosh-tekuri/jsonschema/v6

It also adds the feature of applying defaults if they were defined in
the JSON schema.

Closes https://github.com/mindersec/minder/issues/1657

## Change Type

# Testing

Added a relevant unit test.

# Review Checklist:


JAORMX:
> Would it make sense to call `applyDefaults` separately instead of making it part of both validation functions? Since `ValidateRuleDef` and `ValidateParams` are already separate, I expected `ApplyDefaults` to be separate as well.

That might make sense, I just wanted to simplify the API by doing both validation and applying defaults. Else the implementor gets a more verbose setup.

eleftherias (dismissed):

pkg/profiles/rule_validator.go
```diff
@@ -67,36 +67,55 @@ func (r *RuleValidator) ValidateParamsAgainstSchema(params map[string]any) error
...
 	for _, desc := range errs {
-		problems = append(problems, desc.String())
+		problems = append(problems, desc.Error())
 	}
-
 	return fmt.Errorf("invalid json schema: %s", strings.TrimSpace(strings.Join(problems, "\n")))
 }
+
+// applyDefaults recursively applies default values from the schema to the object.
+func applyDefaults(schema *jsonschema.Schema, obj map[string]any) {
```
evankanderson, line 104: I wish the library had this function, but it looks like it doesn't.  😞

evankanderson (dismissed):

evankanderson:
> > Would it make sense to call `applyDefaults` separately instead of making it part of both validation functions? Since `ValidateRuleDef` and `ValidateParams` are already separate, I expected `ApplyDefaults` to be separate as well.
> 
> That might make sense, I just wanted to simplify the API by doing both validation and applying defaults. Else the implementor gets a more verbose setup.

It feels like we should do `applyDefaults` first and then validate, but it should be safe to call `applyDefaults` as part of validation.  The one weird thing is that you would expect validation to be non-mutating, but `applyDefaults` is clearly mutating

pkg/profiles/rule_validator.go
```diff
@@ -67,36 +67,55 @@ func (r *RuleValidator) ValidateParamsAgainstSchema(params map[string]any) error
...
 
-func buildValidationError(errs []gojsonschema.ResultError) error {
+func buildValidationError(errs []*jsonschema.ValidationError) error {
 	problems := make([]string, 0, len(errs))
 	for _, desc := range errs {
-		problems = append(problems, desc.String())
+		problems = append(problems, desc.Error())
 	}
-
 	return fmt.Errorf("invalid json schema: %s", strings.TrimSpace(strings.Join(problems, "\n")))
```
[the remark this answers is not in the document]
JAORMX: In all honesty, I don't remember. It may not be needed anymore.

pkg/profiles/rule_validator.go
JAORMX: No library I found has one :cry:

eleftherias (approved):

PR #4962, opened 2024-11-13
Add data sources protobuf

JAORMX:
# Summary

This adds the protobuf definition of data sources. We'll re-use this
protobuf around Minder to create the CRUD interface and reason about the
rest of the implementation.

## Change Type

# Testing

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+
+    oneof body {
+        // body is the body of the request.
+        google.protobuf.Struct bodyobj = 4;
+        // bodystr is the body of the request as a string.
+        string bodystr = 5 [
+            (buf.validate.field).string = {
+                max_len: 1000,
+            },
+            (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
```
evankanderson, line 3496: You shouldn't need this if length 0 is a valid length.

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+        (buf.validate.field).string = {
+            in: ["GET", "POST", "PUT", "PATCH", "DELETE"]
+        },
+        (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
+    ];
+
+    // headers is a map of headers to send with the request.
+    google.protobuf.Struct headers = 3;
+
+    oneof body {
```
evankanderson, line 3488: Do you want to do a message-level CEL expression to allow `body` only for POST/PUT/PATCH requests?

evankanderson: (Later is a fine answer)

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+    // method is the HTTP method to use for the request.
+    string method = 2 [
+        (buf.validate.field).string = {
+            in: ["GET", "POST", "PUT", "PATCH", "DELETE"]
+        },
+        (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
+    ];
+
+    // headers is a map of headers to send with the request.
+    google.protobuf.Struct headers = 3;
```
evankanderson, line 3486: Should this be a `map[string]string`, rather than any type of JSON object?

Are any of the following valid?

```yaml
headers: help
```

```yaml
headers:
- one
- two
```

```yaml
headers:
  accept:
    json: ["text", "xml"]
```

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+
+    // parse is the parse configuration for the response.
+    // This allows us to serialize the response into a structured format,
+    // or not.
+    // If left unset, the response will be treated as a string.
+    string parse = 6 [
+        (buf.validate.field).string = {
+        pattern: "^[a-z_]+$",
+            max_len: 10,
+            },
```
evankanderson, line 3508: Do you want to start with `in` for the validation, and expand later?

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+                int32: {
+                    gte: 100,
+                    lte: 599,
+                }
+            }
+        },
+        (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
+    ];
+
+    DataSourceAuth auth = 9;
```
evankanderson, line 3545: Do we need this proto and field now, or can we add it later and the default empty value means "no auth"?

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+
+    // context is the context in which the data source is evaluated.
+    // Note that in this case we only need the project in the
+    // context, since data sources are not provider-specific.
+    ContextV2 context = 3;
+
+    // name is the name of the data source.
+    string name = 4 [
+        (buf.validate.field).string = {
+            pattern: "^[A-Za-z][-[:word:]]*$",
```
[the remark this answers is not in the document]
JAORMX: They are case insensitive within a project. I'm thinking they'd have the same mechanics as profiles. That is, they apply and trickle down to subprojects, but are not modifiable by child projects. That allows for child projects to leverage data sources in profiles targetted at those child projects.

I suspect data sources will be more nuanced that this, so we may want to make this configurable in the future.

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+
+    oneof driver {
+        // rest is the REST data source driver.
+        RestDataSource rest = 7;
+    }
+}
+
+// RestDataSource is the REST data source driver.
+message RestDataSource {
+    // endpoint is the URL of the REST API.
```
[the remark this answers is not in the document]
JAORMX: I should have added this as a comment. I intended to use RFC6570 as you suggested in the design doc.

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+    string endpoint = 1 [
+        (buf.validate.field).string = {
+            pattern: "^https?://.*$",
+            max_len: 800,
+        }
```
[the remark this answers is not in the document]
JAORMX: It should be! thanks

proto/minder/v1/minder.proto
JAORMX: ack, I'm still learning to use these validation annotations

proto/minder/v1/minder.proto
JAORMX: Good catch!

proto/minder/v1/minder.proto
JAORMX: We can add it later, no worries.

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+    ];
+
+    // method is the HTTP method to use for the request.
+    string method = 2 [
+        (buf.validate.field).string = {
+            in: ["GET", "POST", "PUT", "PATCH", "DELETE"]
+        },
+        (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
```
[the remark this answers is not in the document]
JAORMX: Yes

proto/minder/v1/minder.proto
JAORMX: How does one do that?

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,137 @@ message UpstreamEntityRef {
...
+            (buf.validate.field).string = {
+                max_len: 1000,
+            },
+            (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
+        ];
+    }
+
+    // fallback is the fallback configuration for the response in case
+    // of an unexpected status code.
+    repeated Fallback fallback = 7;
```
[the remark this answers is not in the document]
JAORMX: Yes, here it is https://github.com/mindersec/minder-rules-and-profiles/blob/main/rule-types/github/no_open_security_advisories.yaml#L57

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,144 @@ message UpstreamEntityRef {
...
+    // If left unset, the response will be treated as a string.
+    string parse = 6 [
+        (buf.validate.field).string = {
+        pattern: "^[a-z_]+$",
+            max_len: 10,
+            },
+        (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
+    ];
+
+    message Fallback {
```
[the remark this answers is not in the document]
blkt: It's ok for me, just raising awareness.

evankanderson (approved):
LGTM, a couple comments, but nothing that needs to block.

I would prefer that the server be able to ignore the `type` field rather than requiring all clients to set it to a static value, but I don't feel super strongly.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
JAORMX: You are right, I'll add a comment here.

proto/minder/v1/minder.proto
JAORMX: That sounds handy, though I'm still not quite sure how to use it. Mind adding this in a subsequent PR?

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,139 @@ message UpstreamEntityRef {
...
+
+    // id is the unique identifier of the data source.
+    string id = 5 [
+        (buf.validate.field).string = {uuid: true},
+        (google.api.field_behavior) = OUTPUT_ONLY
+    ];
+
+    oneof driver {
+        // rest is the REST data source driver.
+        RestDataSource rests = 6;
```
JAORMX, line 3461: @blkt note that I'm not able to use a `map` in a `oneof` in protobuf [1]. So I needed to nest the map within the `RestDataSource` message.

[1] https://github.com/protocolbuffers/protobuf/issues/7542

proto/minder/v1/minder.proto
```diff
@@ -3411,3 +3412,139 @@ message UpstreamEntityRef {
...
+                        lte: 599,
+                    }
+                }
+            },
+            (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
+        ];
+    }
+
+    // defs is the list of definitions for the REST API.
+    map<string, Def> defs = 1;
```
JAORMX, line 3549: @blkt this is the nesting. It'll end up like:

```
rests:
  def:
    foo:
       <endpoint config>...
    bar:
       ...
```

blkt: I think that's OK, it also solves the problem of where to put extra fields specific to the `rest(s)` and/or other types.

proto/minder/v1/minder.proto
blkt: **nit:** sorry for bringing this up so late, but given that we don't have a `rest` field anymore, I would rename `rests` into `rest`.

blkt (approved):

blkt (approved):

PR #4963, opened 2024-11-13
Add entitlements assignment at the time of project creation

teodor-yanev:
# Summary

This PR adds the functionality of automatically creating entitlements records for new projects. This will be done for both project creation paths - parent and sub-projects. 

The process is as follows:

1) In the config we have (e.g.:)
```
features:
  role_feature_mapping:
    stacklok-employees: "stacklok"
    default-roles-stacklok: "default"
    offline_access: "offline"
```

2) From the user's JWT, we get Keycloak-related data in the form of:
```
  "realm_access": {
    "roles": [
      "stacklok-employees",
      "default-roles-stacklok",
      "offline_access",
      "uma_authorization"
    ]
  },
```
3) For each of the three features (stacklok, default, offline), a new entitlement record will be created, mapping the ID of the newly created project.

This not only introduces flexibility in setting up Minder by simply adjusting a config mapping but also helps have separate configurations for each environment and adequate telemetry data at the time of creation instead of aggregating and analysing batched data periodically.

## Change Type

# Testing

Tested locally and with updated Unit tests.

Screenshot 2024-11-14 at 15 54 34
Screenshot 2024-11-14 at 15 54 37
Screenshot 2024-11-14 at 15 54 41

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_projects.go
+++ b/internal/controlplane/handlers_projects.go
@@ -161,12 +161,6 @@ func (s *Server) CreateProject(
 			"project does not allow project hierarchy operations")
 	}
 
-	// Retrieve the role-to-feature mapping from the configuration
-	projectFeatures, err := s.cfg.Features.GetFeaturesForRoles(ctx)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "error getting features for roles: %v", err)
-	}
-
 	tx, err := s.store.BeginTransaction()
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "error starting transaction: %v", err)
@@ -207,6 +201,8 @@ func (s *Server) CreateProject(
 		return nil, status.Errorf(codes.Internal, "error creating subproject: %v", err)
 	}
 
+	// Retrieve the role-to-feature mapping from the configuration
+	projectFeatures := s.cfg.Features.GetFeaturesForRoles(ctx)
 	if err := features.CreateEntitlements(ctx, qtx, subProject.ID, projectFeatures); err != nil {
 		return nil, status.Errorf(codes.Internal, "error creating entitlements: %v", err)
 	}
--- a/internal/controlplane/handlers_projects.go
+++ b/internal/controlplane/handlers_projects.go
@@ -161,6 +161,12 @@ func (s *Server) CreateProject(
 			"project does not allow project hierarchy operations")
 	}
 
+	// Retrieve the role-to-feature mapping from the configuration
+	projectFeatures, err := s.cfg.Features.GetFeaturesForRoles(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error getting features for roles: %v", err)
+	}
+
 	tx, err := s.store.BeginTransaction()
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "error starting transaction: %v", err)
@@ -199,6 +205,10 @@ func (s *Server) CreateProject(
 			return nil, util.UserVisibleError(codes.AlreadyExists, "project named %s already exists", req.Name)
 		}
 		return nil, status.Errorf(codes.Internal, "error creating subproject: %v", err)
+	}
+
+	if err := features.CreateEntitlements(ctx, qtx, subProject.ID, projectFeatures); err != nil {
+		return nil, status.Errorf(codes.Internal, "error creating entitlements: %v", err)
 	}
 
 	if err := s.authzClient.Adopt(ctx, parent.ID, subProject.ID); err != nil {
--- a/internal/controlplane/handlers_projects.go
+++ b/internal/controlplane/handlers_projects.go
@@ -203,7 +203,10 @@ func (s *Server) CreateProject(
 
 	// Retrieve the membership-to-feature mapping from the configuration
 	projectFeatures := s.cfg.Features.GetFeaturesForMemberships(ctx)
-	if err := features.CreateEntitlements(ctx, qtx, subProject.ID, projectFeatures); err != nil {
+	if err := qtx.CreateEntitlements(ctx, db.CreateEntitlementsParams{
+		Column1: projectFeatures,
+		Column2: subProject.ID,
+	}); err != nil {
 		return nil, status.Errorf(codes.Internal, "error creating entitlements: %v", err)
 	}
 
--- a/internal/controlplane/handlers_projects.go
+++ b/internal/controlplane/handlers_projects.go
@@ -201,8 +201,8 @@ func (s *Server) CreateProject(
 		return nil, status.Errorf(codes.Internal, "error creating subproject: %v", err)
 	}
 
-	// Retrieve the role-to-feature mapping from the configuration
-	projectFeatures := s.cfg.Features.GetFeaturesForRoles(ctx)
+	// Retrieve the membership-to-feature mapping from the configuration
+	projectFeatures := s.cfg.Features.GetFeaturesForMemberships(ctx)
 	if err := features.CreateEntitlements(ctx, qtx, subProject.ID, projectFeatures); err != nil {
 		return nil, status.Errorf(codes.Internal, "error creating entitlements: %v", err)
 	}
--- a/internal/controlplane/handlers_projects.go
+++ b/internal/controlplane/handlers_projects.go
@@ -204,8 +204,8 @@ func (s *Server) CreateProject(
 	// Retrieve the membership-to-feature mapping from the configuration
 	projectFeatures := s.cfg.Features.GetFeaturesForMemberships(ctx)
 	if err := qtx.CreateEntitlements(ctx, db.CreateEntitlementsParams{
-		Column1: projectFeatures,
-		Column2: subProject.ID,
+		Features:  projectFeatures,
+		ProjectID: subProject.ID,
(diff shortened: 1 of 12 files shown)


evankanderson (commented):
A few top-level comments:

* I'm not sure I'd use the terms "Role" and "Feature" given that we also have OpenFGA roles and OpenFeature features.  Maybe "Claim" or "membership" and "Entitlement"?
* This needs a test which covers the new code you've added to actually add some entitlements to a project.
* I'd tend to prefer "return empty" over "return error" for cases where fields are not found unless we can't proceed without that information.

JAORMX:
> * I'm not sure I'd use the terms "Role" and "Feature" given that we also have OpenFGA roles and OpenFeature features.  Maybe "Claim" or "membership" and "Entitlement"?

We do have the term feature within Minder and it's a database table which is linked to projects via entitlements. So the term feature might not be too bad after all in this context. I certainly would advice against using "role" here, though.

pkg/config/server/features.go
```diff
@@ -0,0 +1,55 @@
...
+	return features, nil
+}
+
+// extractRolesFromContext extracts roles from the JWT in the context
+func extractRolesFromContext(ctx context.Context) ([]string, error) {
+	var realmAccess map[string]interface{}
+	if claim, ok := jwt.GetUserClaimFromContext[map[string]interface{}](ctx, "realm_access"); ok {
+		realmAccess = claim
+	} else {
+		return nil, fmt.Errorf("realm_access claim not found")
```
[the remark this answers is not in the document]
teodor-yanev: Removed all errors returned from this func, I had the same thought.

pkg/config/server/features.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	var roles []string
+	if rolesInterface, ok := realmAccess["roles"].([]interface{}); ok {
+		for _, role := range rolesInterface {
+			if roleStr, ok := role.(string); ok {
+				roles = append(roles, roleStr)
+			}
+		}
```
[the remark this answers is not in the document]
teodor-yanev: This type assertion was not possible --tested in runtime locally.
Still, refactored this part.

internal/projects/features/features.go
```diff
@@ -30,6 +30,22 @@ func ProjectAllowsProjectHierarchyOperations(ctx context.Context, store db.Store
...
 }
 
+// CreateEntitlements creates entitlements for a project
+// It takes a 'qtx' because it is usually called within a transaction
+func CreateEntitlements(ctx context.Context, qtx db.Querier, projectID uuid.UUID, features []string) error {
+	for _, feature := range features {
+		err := qtx.CreateEntitlement(ctx, db.CreateEntitlementParams{
+			ProjectID: projectID,
+			Feature:   feature,
+		})
```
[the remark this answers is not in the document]
teodor-yanev: Yes, that is another thing I also had in mind, the new query cleverly handles this.

teodor-yanev:
Thanks for the review @evankanderson, I've addressed all the concerns we had.

pkg/config/server/features.go
```diff
@@ -0,0 +1,53 @@
...
+	MembershipFeatureMapping map[string]string `mapstructure:"membership_feature_mapping"`
+}
+
+// GetFeaturesForMemberships returns the features associated with the memberships in the context
+func (fc *FeaturesConfig) GetFeaturesForMemberships(ctx context.Context) []string {
+	memberships := extractMembershipsFromContext(ctx)
+
+	var features []string
+	for _, m := range memberships {
+		if feature, ok := fc.MembershipFeatureMapping[m]; ok {
```
evankanderson, line 24: What a about:

```suggestion
		if feature := fc.MembershipFeatureMapping[m]; feature != "" {
```

Which will cover both "not found" and "was an empty string", which we probably don't want.

pkg/config/server/features.go
```diff
@@ -0,0 +1,53 @@
...
+	if !ok {
+		return nil
+	}
+
+	memberships := make([]string, len(rawMemberships))
+	for i, membership := range rawMemberships {
+		if membershipStr, ok := membership.(string); ok {
+			memberships[i] = membershipStr
+		}
+	}
```
evankanderson, line 50: This will leave empty-string memberships if fields can't be converted to string.  (Which shouldn't happen, but if we're checking, we should assume it could fail.)

```suggestion
	memberships := make([]string, 0, len(rawMemberships))
	for i, membership := range rawMemberships {
		if membershipStr, ok := membership.(string); ok {
			memberships = append(memberships, membershipStr)
		}
	}
```

pkg/config/server/features.go
```diff
@@ -0,0 +1,53 @@
...
+type FeaturesConfig struct {
+	// MembershipFeatureMapping maps a membership to a feature
+	MembershipFeatureMapping map[string]string `mapstructure:"membership_feature_mapping"`
+}
+
+// GetFeaturesForMemberships returns the features associated with the memberships in the context
+func (fc *FeaturesConfig) GetFeaturesForMemberships(ctx context.Context) []string {
+	memberships := extractMembershipsFromContext(ctx)
+
+	var features []string
```
evankanderson, line 22: On line 45, you pre-allocate the slice.  Do the same thing here?

```suggestion
	features := make([]string, 0, len(memberships))
```

internal/projects/creator_test.go
```diff
@@ -107,3 +128,16 @@ func TestProvisionSelfEnrolledProjectInvalidName(t *testing.T) {
 	}
 
 }
+
+// prepareTestToken creates a JWT token with the specified roles and returns the context with the token.
+func prepareTestToken(t *testing.T, roles []any) context.Context {
```
evankanderson, line 133: This should probably take an existing context, so you can use it as a wrapper.  (Just a standard pattern, like `context.WithValue()`)

internal/projects/creator.go
```diff
@@ -105,6 +108,15 @@ func (p *projectCreator) ProvisionSelfEnrolledProject(
 		return nil, fmt.Errorf("failed to create default project: %v", err)
 	}
 
+	// Retrieve the membership-to-feature mapping from the configuration
+	projectFeatures := p.featuresCfg.GetFeaturesForMemberships(ctx)
+	if err := qtx.CreateEntitlements(ctx, db.CreateEntitlementsParams{
+		Column1: projectFeatures,
+		Column2: project.ID,
```
evankanderson, line 115: Can we rename these parameters from `Column1` and `Column2`?  😁

evankanderson (commented):
The only item I feel strongly enough to not approve is the names of the `db.CreateEntitlementsParams` -- I'd like names that help us avoid swapping parameters.

internal/projects/creator.go
teodor-yanev: Absolutely 😄

internal/controlplane/handlers_projects.go
```diff
@@ -201,6 +201,15 @@ func (s *Server) CreateProject(
...
 	}
 
+	// Retrieve the membership-to-feature mapping from the configuration
+	projectFeatures := s.cfg.Features.GetFeaturesForMemberships(ctx)
+	if err := qtx.CreateEntitlements(ctx, db.CreateEntitlementsParams{
+		Features:  projectFeatures,
+		ProjectID: subProject.ID,
+	}); err != nil {
+		return nil, status.Errorf(codes.Internal, "error creating entitlements: %v", err)
+	}
```
evankanderson, line 211: This is sort-of-weird: this is creating a child project free-form here, separate from the `internal/projects` code.  It's not clear to me whether these child projects should:

1. Copy entitlements from the parent project
2. Use entitlements based on the specific parent project member who creates them
3. Have an empty set of entitlements, and the entitlement queries are extended to check parents.

Right now, it looks like were doing option 2, but it's not clear that is correct.

Let's file an issue and assign it to @ethomson to figure out the desired way to handle entitlements in sub-projects, because those are currently behind an entitlement, and we _probably_ know most of the users of that feature.

evankanderson: Also, file a bug to move this code to `internal/projects`, so all project creation happens in that one module, and `internal/controlplane` gets a bit smaller.

internal/projects/creator_test.go
```diff
@@ -33,10 +37,28 @@ func TestProvisionSelfEnrolledProject(t *testing.T) {
...
+			if !reflect.DeepEqual(params.Features, expectedFeatures) {
+				t.Errorf("expected features %v, got %v", expectedFeatures, params.Features)
+			}
+			return nil
+		})
+
+	ctx := prepareTestToken(context.Background(), t, []any{
+		"teamA",
+		"teamB",
+		"teamC",
```
evankanderson, line 52: Nice addition to have a membership that doesn't convert to a feature.

evankanderson (approved):

internal/controlplane/handlers_projects.go
teodor-yanev: I did find it weird that we have two separate paths for project creation, especially one of them being exclusive to the handler and was going to follow-up on that, thanks for noting it as well.

Here are the new issues:

https://github.com/mindersec/minder/issues/5001
https://github.com/mindersec/minder/issues/5002

PR #4991, opened 2024-11-18
Validate rule type alert during creation/update

eleftherias:
# Summary

Ensure early validation of the alert in ruletypes.

## Change Type

# Testing

# Review Checklist:


rdimitrov (dismissed):

pkg/api/protobuf/go/minder/v1/validators.go
```diff
@@ -286,6 +293,32 @@ func (ing *RuleType_Definition_Ingest) Validate() error {
 	return nil
 }
 
+// Validate validates a rule type definition alert
+func (alert *RuleType_Definition_Alert) Validate() error {
+	if alert == nil {
```
JAORMX, line 298: Wait, alert should be nullable... shouldn't it?

JAORMX: Nevermind, noticed the check above

JAORMX (approved):

pkg/api/protobuf/go/minder/v1/validators.go
eleftherias: Yes, but I figured if it gets to the point of being validated then it shouldn't be null, as in the caller should decide on the nullability. I can also change this to `return nil` if the alert is `nil`, I don't feel strongly.

JAORMX: I think changing it to return `nil` is fine, as alert configuration is optional after all.

eleftherias: Updated

JAORMX (approved):

PR #4997, opened 2024-11-19
Implement data source interfaces & registration in rego engine

JAORMX:
# Summary

This implements the basic interfaces that data sources must fulfil in
order to be called and used within Minder. This is:

* Listing the functions they provide
* Validating updates

Each function itself must provide a unique key for the engine to use, as
well as argument validation and the data source call itself.

Finally, this implements an engine option that allows data sources to be
dynamically registered into the rego engine.

## Change Type

# Testing

# Review Checklist:


internal/engine/eval/rego/datasources.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/open-policy-agent/opa/types"
+
+	v1datasources "github.com/mindersec/minder/pkg/datasources/v1"
+)
+
+// RegisterDataSource implements the Eval interface.
+func (e *Evaluator) RegisterDataSource(ds v1datasources.DataSource) {
+	for key, dsf := range ds.GetFuncs() {
+		fmt.Printf("Registering data source %s\n", key)
+		e.regoOpts = append(e.regoOpts, buildFromDataSource(key, dsf))
```
blkt, line 21: **question:** I believe there might be a potential issue here that we have to address: we need to make sure functions are in the "namespace" of the Data Source they're declared in, and that they do not shadow other identifiers in REGO.

As an example, assume I have two Data Sources, `foo` and `bar`, and I want to use them both in a single Rule; additionally, let's assume both Data Sources expose a function named  `garble` with the same signature; I would expect to be able to use both `foo.garble(...)` and `bar.garble(...)`.

Where is the final name defined? My understanding is that it would be the value of `key`, am I right? If that's the case, we need to either document that the name MUST contain the data source name, or we must turn registration inside-out.

Proposal: assuming `key` is the final name, Minder must be in control of it, and not the final User, and we could just change the interface to return a collection of `func(_ rego.BuiltinContext, obj *ast.Term) (*ast.Term, error)` and let the upper layer figure out the final name.

blkt: Solved in `func (reg *DataSourceRegistry) RegisterDataSource(name string, ds DataSource) (err error)`

blkt (dismissed):
LGTM

internal/engine/eval/rego/datasources.go
```diff
@@ -0,0 +1,78 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			return ast.NewTerm(val), nil
+		},
+	)
+}
+
+// This converts the data source key into a format that can be used in the rego query.
```
eleftherias, line 62: nit
```suggestion
// This converts the data source function key into a format that can be used in the rego query.
```

eleftherias (approved):

eleftherias (approved):

internal/engine/eval/rego/datasources.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/open-policy-agent/opa/rego"
+	"github.com/open-policy-agent/opa/types"
+
+	v1datasources "github.com/mindersec/minder/pkg/datasources/v1"
+)
+
+// RegisterDataSources implements the Eval interface.
+func (e *Evaluator) RegisterDataSources(dsr *v1datasources.DataSourceRegistry) {
+	for key, dsf := range dsr.GetFuncs() {
+		fmt.Printf("Registering data source %s\n", key)
```
evankanderson, line 20: Did you mean to leave in a `Printf`?

internal/engine/eval/rego/datasources.go
```diff
@@ -0,0 +1,78 @@
...
+		e.regoOpts = append(e.regoOpts, buildFromDataSource(key, dsf))
```
evankanderson, line 21: Question: does it make sense for `dsf` from `dsr.GetFuncs()` to actually have a `func(*rego.Rego)` method (i.e. to move the `buildFromDataSource` into a function on `DataSourceFuncDef`)?

It pushes the Rego types into a slighly larger area of the code, but it helps avoid having a big ball of code in `eval/rego` (and if we name the method `Rego` or `RegoOpt`, we would have a pattern for any other future evaluation engines).

internal/engine/eval/rego/datasources.go
```diff
@@ -0,0 +1,78 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			if err := dsf.ValidateArgs(obj); err != nil {
+				return nil, err
+			}
+
+			// Call the data source function
+			ret, err := dsf.Call(jsonObj)
```
evankanderson, line 47: If you move `buildFromDataSource` to DataSourceFuncDef, you might not need to expose `ValidateArgs` or `Call` in the interface.

internal/engine/eval/rego/rego_test.go
```diff
@@ -454,3 +458,59 @@ violations[{"msg": msg}] {
...
+	require.NoError(t, err, "could not evaluate")
+
+	// Doesn't match
+	fdsf.EXPECT().Call(gomock.Any()).Return("bar", nil)
+	err = e.Eval(context.Background(), emptyPol, nil, &interfaces.Result{
+		Object: map[string]any{
+			"data": "bar",
+		},
+	})
+	require.ErrorIs(t, err, engerrors.ErrEvaluationFailed, "should have failed the evaluation")
```
evankanderson, line 515: I'm wondering if it would make sense to use a real registry and DataSource here, to provide an example for future users (and to verify the lifecycle more deeply than Mocks).

(You'll also get a bit more test coverage of `registry.go` for "free")

pkg/datasources/v1/registry.go
```diff
@@ -0,0 +1,66 @@
...
+			if err == nil {
+				reg.r.Store(funckey, f)
+			}
+		}()
+	}
+
+	return nil
+}
+
+// GetFuncs returns all functions that data sources provide globally.
```
evankanderson, line 53: This could eventually pull from the database to get project-specific data sources, right?

evankanderson (commented):
Looks very promising!  A couple comments you can take or leave as you please.

internal/engine/eval/rego/datasources.go
JAORMX: I did not

internal/engine/eval/rego/datasources.go
JAORMX: I wanted to have data sources to be quite independent from the eval engine. The eval engine would then be in charge of interpreting the data source and making sense of it. It brings a lot more responsibility into the eval engine, but it allows for us to include more engines with ease in the future. e.g. if we'd introduce a WASM engine, we wouldn't need to mess around with rego structures.

internal/engine/eval/rego/rego_test.go
JAORMX: You have a good point, I'll add a testing implementation

pkg/datasources/v1/registry.go
JAORMX: That's right, the idea is that we'd do an initial parsing from the db, turn this into in-memory objects, and then pass these around.

PR #5022, opened 2024-11-21
Kick off data sources service

JAORMX:
# Summary

This service will contain all the business logic to deal with data
sources.

For the moment this only kicks off the `get` calls. But populating this should
allow others to work on other function definitions.

## Change Type

# Testing

# Review Checklist:


database/query/datasources.sql
```diff
@@ -0,0 +1,73 @@
...
+
+-- name: CreateDataSource :one
+INSERT INTO data_sources (project_id, name, display_name)
+VALUES ($1, $2, $3) RETURNING *;
+
+-- AddDataSourceFunction adds a function to a datasource.
+
+-- name: AddDataSourceFunction :one
+INSERT INTO data_sources_functions (data_source_id, name, type, definition)
+VALUES ($1, $2, $3, $4) RETURNING *;
```
blkt, line 11: **issue:** `INSERTS` should always specify `project_id`.

database/query/datasources.sql
```diff
@@ -0,0 +1,73 @@
...
+RETURNING *;
+
+-- UpdateDataSourceFunction updates a function in a datasource. We're
+-- only able to update the type and definition of the function.
+
+-- name: UpdateDataSourceFunction :one
+UPDATE data_sources_functions
+SET type = $3, definition = $4, updated_at = NOW()
+WHERE data_source_id = $1 AND name = $2
+RETURNING *;
```
blkt, line 28: **issue:** these statements should filter by `project_id` as well.

database/query/datasources.sql
```diff
@@ -0,0 +1,73 @@
...
+
+-- name: DeleteDataSource :one
+DELETE FROM data_sources
+WHERE id = $1 AND project_id = $2
+RETURNING *;
+
+-- name: DeleteDataSourceFunction :one
+DELETE FROM data_sources_functions
+WHERE data_source_id = $1 AND name = $2
+RETURNING *;
```
blkt, line 38: **issue:** these statements should filter by `project_id` as well.

database/query/datasources.sql
```diff
@@ -0,0 +1,73 @@
...
+
+-- name: ListDataSources :many
+SELECT * FROM data_sources
+WHERE project_id = ANY(sqlc.arg(projects)::uuid[]);
+
+-- ListDataSourceFunctions retrieves all functions for a datasource.
+
+-- name: ListDataSourceFunctions :many
+SELECT * FROM data_sources_functions
+WHERE data_source_id = $1;
```
blkt, line 72: **issue:** this statement should join data sources to filter by `project_id`.

database/query/datasources.sql
JAORMX: @blkt there is no project_id in this table. It points to a data_sources row which already has a project ID check

database/query/datasources.sql
JAORMX: same comment as above.

database/query/datasources.sql
blkt: Should we add it?

JAORMX:
@blkt @teodor-yanev and I talked about this and agreed to include the project ID in the functions and relationships tables. @blkt will bring this in and I'll rebase on top of it.

database/query/datasources.sql
```diff
@@ -0,0 +1,73 @@
+-- CreateDataSource creates a new datasource in a given project.
```
rdimitrov, line 1: Should we have delete on cascade based on the data source id for functions?

internal/datasources/service/convert.go
```diff
@@ -0,0 +1,62 @@
...
+			ProjectId: ds.ProjectID.String(),
+		},
+	}
+
+	if len(dsfuncs) == 0 {
+		return nil, errors.New("data source is invalid and has no defintions")
+	}
+
+	// All data source types should be equal... so we'll just take the first one.
+	dsfType := dsfuncs[0].Type
```
rdimitrov, line 33: Should we ensure/validate that all sources have the same type?

internal/datasources/service/service.go
```diff
@@ -0,0 +1,207 @@
...
+	return d.getDataSourceSomehow(
+		ctx, project, opts, func(ctx context.Context, tx db.ExtendQuerier, projs []uuid.UUID,
+		) (db.DataSource, error) {
+			ds, err := tx.GetDataSourceByName(ctx, db.GetDataSourceByNameParams{
+				Name:     name,
+				Projects: projs,
+			})
+			if err != nil {
+				if errors.Is(err, sql.ErrNoRows) {
+					return db.DataSource{}, util.UserVisibleError(codes.NotFound,
```
rdimitrov, line 95: Isn't it better to return nil instead of empty object in case of errors? 

_Comment is valid for the other use cases below_

internal/datasources/service/service.go
```diff
@@ -0,0 +1,207 @@
...
+				return db.DataSource{}, fmt.Errorf("failed to get data source by name: %w", err)
+			}
+
+			return ds, nil
+		})
+}
+
+func (d *dataSourceService) List(
+	ctx context.Context, project uuid.UUID, opts *ReadOptions) ([]*minderv1.DataSource, error) {
+	stx, err := d.txBuilder(d, opts)
```
rdimitrov, line 128: Question: Why do we want to list within a transaction?

database/query/datasources.sql
JAORMX: We have that https://github.com/mindersec/minder/blob/main/database/migrations/000108_data_sources.up.sql#L47

internal/datasources/service/convert.go
JAORMX: We should. I was thinking of doing this on the `create` and `update` parts though. There's not much we can do on the `get` side.

teodor-yanev (approved):

internal/datasources/service/service.go
JAORMX: this sub-function returns an object and not a pointer, since that's the type of object that the database wrappers usually returns. we do return pointers and thus `nil` underneath though.

internal/datasources/service/service.go
JAORMX: So, data sources (and their functions) are part of separate tables, we want the states of these to be consistent when listing/getting them, and not get into funky intermediate states if parallel update happens.

Note that I do set the transaction to be "read-only" so this doesn't block other reads.

rdimitrov (approved):

PR #5029, opened 2024-11-22
Include provider ID in minder CLI response

### LINKED ISSUE #5000: Cannot delete a provider by its ID using the CLI
Deleting a provider by its ID is not possible, because the minder CLI does not output the provider ID (it's missing in the response).

navnitms:
Fixed the issue where deleting a provider by its ID was not possible due to the missing provider ID in the CLI response

# Summary

Fixes #5000 

## Change Type

# Testing

Tested the flow using CLI command ``` minder provider get -n <provider-name> ```

Response :

image

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -3348,6 +3348,12 @@ message Provider {
     // credentials_state is the state of the credentials for the provider.
     // This is an output-only field. It may be: "set", "unset", "not_applicable".
     string credentials_state = 9;
+
+    // id is the unique identifier of the provider.
+    string id = 10 [
+        (buf.validate.field).string = {uuid: true},
+        (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
+    ];
```
evankanderson, line 3356: Not strictly required, but you might want to also add the `(google.api.field_behavior) = OUTPUT_ONLY` annotation, since I expect that we would reject changing the ID on create/update.

Here's an example:

https://github.com/mindersec/minder/blob/main/proto/minder/v1/minder.proto#L3646

cmd/cli/app/provider/provider_get.go
```diff
@@ -81,6 +81,7 @@ func GetProviderCommand(ctx context.Context, cmd *cobra.Command, _ []string, con
 		impls := getImplementsAsStrings(p)
 		afs := getAuthFlowsAsStrings(p)
 
+		t.AddRow("ID", p.GetId())
 		t.AddRow("Name", p.GetName())
 		t.AddRow("Project", p.GetProject())
 		t.AddRow("Version", p.GetVersion())
```
evankanderson, line 87: This is fine for now, but these wide tables end up being a bit ugly, so we'll probably (at some future point) need to figure out how to collapse them more elgantly.  (And a 36-character wide UUID doesn't help, but it's not the only contributor)

evankanderson (approved):
Two minor comments, but I'm happy to merge as-is, and you can do a follow-up PR if you want, since our usage of `google.api.field_behavior` is not very consistent, so we'll need to do a pass over all the fields anyway.

evankanderson:
(The security scan failure is an upstream Trivy issue where they are triggering on their own examples...)

proto/minder/v1/minder.proto
navnitms: Sure will do a follow up PR to address this 
Thanks

PR #5030, opened 2024-11-22
Add repo dependency ingester

evankanderson:
# Summary

Add a dedicate ingestion mode for repository dependencies which uses https://github.com/google/osv-scalibr to extract the dependencies from various lockfile and other dependency files in the repository.

See [this design doc](https://docs.google.com/document/d/1pqOEnlUazta5MZgTtQAkMH8JS4826uc4YvLcqNPzq2Q/edit?tab=t.0) for the long term direction on dependency ingest; this is part of "stage 1" (but needs tests)

## Change Type

# Testing

Still work in progress...  Feel free to pick this up and patch it in while I am out.

# Review Checklist:


internal/engine/ingester/deps/deps.go
```diff
@@ -0,0 +1,213 @@
...
+	for _, inv := range scanResults.Inventories {
+		node := &sbom.Node{
+			Type:    sbom.Node_PACKAGE,
+			Id:      uuid.New().String(),
+			Name:    inv.Name,
+			Version: inv.Version,
+			Identifiers: map[int32]string{
+				int32(sbom.SoftwareIdentifierType_PURL): inv.Extractor.ToPURL(inv).String(),
+				// TODO: scalibr returns a _list_ of CPEs, but protobom will store one.
+				// use the first?
```
puerco, line 199: Ideally the most specific one. Now that package URL has official support for ranges, the one-purl-per-package rule is broken for purls. This means that perhaps we should create a new identifier list to capture more than one.

evankanderson: I'm slightly confused about what you're suggesting here.  I can see why PURLs supporting ranges would be useful _in general_, but I'm not sure if you're suggesting:

1. That upstream `protobom` needs a change in the `identifiers` type.
2. That Scalibr needs a change in the `ToPURL` function.
3. That this code should be parsing the purl / Inventory and creating multiple `Node` objects when the Inventory contains certain types of data.

Looking at the scalibr code, it looks like CPEs are only reported from the SPDX and CDX extractors, so multiple (or > 0) CPEs seems moot for our current use cases.  :puzzled:

puerco: I meant number 1, but there are use cases when you may also want to return purls with ranges, more than one purl, and/or more than one CPE. It all depends on what you are trying to match.

puerco:
I finished this one in #5058, thanks!

PR #5058, opened 2024-11-27
Dependency ingester

puerco:
# Summary

This PR finishes Evan's ingester started in #5030. It enables the dependency ingester using ov-scalibr. This allows us to write rules on the ingested protobom structs.

Here's the example ruletype I have been using to test:

```yaml
version: v1
release_phase: alpha
type: rule-type
name: avoid_malicious_dependency
display_name: Check a project's dependencies against a denylist
short_failure_message: Malicious dependencies in use
severity:
  value: high
context:
  provider: github
description: Alert if a repository uses a malicious dependency from a blocklist.
guidance: |
  Using the configured list of prohibited dependencies, Minder will check the 
  dependencies at use in a repository. If any package name is included in the
  disallowed list, the profile will fail.
def:
  in_entity: repository
  rule_schema:
    type: object
    properties:
      prohibited:
        type: array
        items:
          type: string
        description: |
          Names of packages not allowed in the repository
      branch:
        type: string
        description: "The name of the branch to check out. Defaults to 'main'"
    required:
      - prohibited
  # Defines the configuration for ingesting data relevant for the rule
  ingest:
    type: deps
  eval:
    type: rego
    rego:
      type: deny-by-default
      def: |
        package minder

        import rego.v1

        # Default deny rule
        default allow := false

        # Create a set of prohibited items from the profile definition
        prohibited := {pkg | pkg := input.profile.prohibited[_]}

        # Extract dependencies from the input
        deps := {node.name | node := input.ingested.node_list.nodes[_]}

        # Allow rule
        allow if {
            deps == deps - prohibited
        }

```

Supersedes #5030

## Change Type

# Testing

Here's a simple profile to test:

```yaml
---
version: v1
type: profile
name: deps-ingest
display_name: Test dependency ingestion
context:
  provider: github
alert: "off"
remediate: "off"
repository:
   - type: avoid_malicious_dependency
     def:
      prohibited:
        - "invokehttp"
```

# Review Checklist:


internal/engine/ingester/deps/deps.go
```diff
@@ -0,0 +1,225 @@
...
+// and scanning it for dependencies with scalibr.
+func (gi *Deps) Ingest(ctx context.Context, ent protoreflect.ProtoMessage, params map[string]any) (*interfaces.Result, error) {
+	switch entity := ent.(type) {
+	case *pb.Repository:
+		return gi.ingestRepository(ctx, entity, params)
+	default:
+		return nil, fmt.Errorf("deps is only supported for repositories")
+	}
+}
+func (gi *Deps) ingestRepository(ctx context.Context, repo *pb.Repository, params map[string]any) (*interfaces.Result, error) {
```
JAORMX, line 84: `nit`: I wonder if we could call the `git` ingester directly from here and THEN process the result of that.

internal/engine/ingester/deps/deps.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	// have to down-cast here, because scalibr needs multiple io/fs types
+	wrapped, ok := iofs.New(memFS).(scalibr_fs.FS)
+	if !ok {
+		return nil, fmt.Errorf("error converting filesystem to ReadDirFS")
+	}
+
+	desiredCaps := scalibr_plugin.Capabilities{
+		OS:            scalibr_plugin.OSLinux,
+		Network:       true,
```
JAORMX, line 178: what do we need the network option for? We might wanna add a small comment for this.

JAORMX (approved):

internal/engine/ingester/deps/deps.go
puerco: Sgmt, I'll refactor it in follow up.

JAORMX: yeah, doesn't need to be in this PR

internal/engine/ingester/deps/deps.go
puerco: It's to select the plugins that require network access. I'll add a note.

PR #5060, opened 2024-11-27
Instantiate data sources from the policy engine

JAORMX:
# Summary

This hooks data source registration into Minder's policy engine. This is
done by populating individual data source registries into each rule type
engine, which pertains to a single rule type. This way, each rule type
would only have access to the data sources it instantiates.

This also caught a segfault we were unaware of. For some reason, the
builder of the ruletype engine cache was not populating all the needed
members of the struct which would segfault in case the cache did not
have a hit. This fixes that.

Closes: https://github.com/mindersec/minder/issues/5043

## Change Type

# Testing

# Review Checklist:

### DIFF
--- a/internal/engine/rtengine/cache.go
+++ b/internal/engine/rtengine/cache.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/google/uuid"
 
+	datasourceservice "github.com/mindersec/minder/internal/datasources/service"
 	"github.com/mindersec/minder/internal/db"
 	"github.com/mindersec/minder/internal/engine/ingestcache"
 	eoptions "github.com/mindersec/minder/internal/engine/options"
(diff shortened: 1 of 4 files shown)


internal/engine/rtengine/cache.go
```diff
@@ -121,6 +134,21 @@ func cacheRuleEngine(
...
 
+	// Build a registry instance per rule type. This allows us to have an
+	// isolated data source list per instance of the rule type engine which is
+	// what we want. We don't want rule types using data sources they haven't
+	// instantiated. It is in this spot that we would add something like a cache
+	// so data sources could optimize in a per-execution context.
+	//
+	// TODO: Do we need to pass in a transaction here?
+	// TODO: We _might_ want to pass in a slice of the hierarchy here.
+	dsreg, err := dssvc.BuildDataSourceRegistry(ctx, pbRuleType, nil)
```
rdimitrov, line 145: It's not possible for this to conflict right? I.e. one ruletype having access to the data sources for another one?

JAORMX: That's right. This is why I make the instantiation of the registry here and per rule type. The registry is not shared.

rdimitrov: Perfect! 💯 😃

PR #5068, opened 2024-11-27
Validate Data Sources when creating rules.

### LINKED ISSUE #5049: Hook data source validation into rule types API
The idea is that when a rule type is created or updated, we should always validate that the appropriate data soruces are available and valid.

We should also populate the `rule_type_data_sources` table with all of the rule type / data source relationships that are found, if all validations pass.

blkt:
# Summary

Rules can now specify a list of data sources names that they need for their evaluation. This change ensures that data sources required by the rule are available when creating the rule.

A similar change is being added in the code path implementing data source deletion ensuring that a data source is not deleted if there exist rules depending on it.

Fixes #5049

## Change Type

# Testing

Unit tested.

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_ruletype.go
+++ b/internal/controlplane/handlers_ruletype.go
@@ -21,6 +21,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	datasourcesvc "github.com/mindersec/minder/internal/datasources/service"
 	"github.com/mindersec/minder/internal/db"
 	"github.com/mindersec/minder/internal/engine/engcontext"
 	"github.com/mindersec/minder/internal/flags"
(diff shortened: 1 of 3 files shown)


internal/controlplane/handlers_ruletype.go
```diff
@@ -175,9 +177,10 @@ func (s *Server) CreateRuleType(
 		return nil, util.UserVisibleError(codes.InvalidArgument, "%s", err)
 	}
 
-	ds := crt.GetRuleType().GetDef().GetEval().GetDataSources()
-	if len(ds) > 0 && !flags.Bool(ctx, s.featureFlags, flags.DataSources) {
```
JAORMX, line 179: Where's the feature flag check now? We still need that

JAORMX (changes requested):

internal/controlplane/handlers_ruletype.go
```diff
@@ -175,9 +177,10 @@ func (s *Server) CreateRuleType(
...
-		return nil, status.Errorf(codes.Unavailable, "DataSources feature is disabled")
+	ruleDS := crt.GetRuleType().GetDef().GetEval().GetDataSources()
+	if err := s.validateDataSources(ctx, projectID, ruleDS); err != nil {
```
JAORMX, line 181: Wouldn't it make sense to validate within the transaction?

internal/controlplane/handlers_ruletype.go
```diff
@@ -373,3 +377,48 @@ func validateMarkdown(md string) error {
...
+		return status.Errorf(codes.Unavailable, "DataSources feature is disabled")
+	}
+
+	requested := make([]string, 0)
+	for _, ds := range ruleDS {
+		requested = append(requested, ds.Name)
+	}
+
+	// List Data Sources defined in the whole project hierarchy.
+	availableDS, err := s.dataSourcesService.List(
```
JAORMX, line 401: There could be many data sources and only one or two references by a rule type. Wouldn't it make more sense to do it the other way around and query only the ones being instantiated by the rule?

JAORMX (changes requested):

internal/controlplane/handlers_ruletype.go
blkt: It's hard to tell whether this is going to be a performance hit without some data, but the trade off is
* current implementation: potentially many rule types, but only one round trip with the database
* your suggestion: fewer data loaded at rule creation time, but potentially more round trips to the database

Given it's reasonable to think that rules will use few data sources rather than many, I'll change this to load only the necessary ones, but I won't add a new statement for bulk-loading and will use [`GetDataSourceByName`](https://github.com/mindersec/minder/blob/main/database/query/datasources.sql#L55-L57) instead.

blkt:
@JAORMX I should have addressed all your concerns, PTAL.

JAORMX (approved):

PR #5094, opened 2024-11-29
Dependency Extraction Data Source Driver

### LINKED ISSUE #5090: Dependency Data Source Driver
Following up on the dependency ingestion feature, minder should also support dependency extraction as a data source. This would allow rule writers more control over which part of the source code is analyzed to extract dependencies.

puerco:
# Summary

This PR introduces the dependency extraction data source. 

Not ready for review, checking it in early and needs other PRs and a rebase to work.

This needs ~#5091~ ~#5092~ ~#5093~

Fixes  #5090 

## Change Type

# Testing

# Review Checklist:


puerco:
OK, this should be ready for review. The options in the data source definitions don't do anything now, we need #5127 and #5128 for them to work

proto/minder/v1/minder.proto
```diff
@@ -4017,9 +4017,25 @@ message DataSource {
...
+        // Path where the dependency extractor will look for dependency data
+        string path = 1;
+        
+        // List of ecosystems to check, defaults all supported by the extractor
+        repeated string ecosystems = 2;
+    }
+    
+    map<string, Def> def = 1;
+}
+
```
blkt, line 4038: **nit:** is there some form of validation that we can add here?

internal/engine/eval/rego/datasources.go
```diff
@@ -56,6 +57,7 @@ func buildFromDataSource(
 			}
 
 			if err := dsf.ValidateArgs(jsonObj); err != nil {
+				log.Error().Interface("dsfunction", k).Err(err).Msg("error validating datasource function arguments")
```
blkt, line 60: **nit:** I would avoid logging so deep in the stack, it would be much better to raise a proper error and log in upper layers instead.

Given that Data Sources are meant to be user-defined, I don't think there's much value to log an error, as the user would not have access to it. Besides, we have no idea how often this function is going to be called by REGO code in the rule type, this might be spammy.

internal/engine/eval/rego/datasources.go
```diff
@@ -69,6 +71,7 @@ func buildFromDataSource(
 			)
 			ret, err := dsf.Call(ctx, jsonObj)
 			if err != nil {
+				log.Error().Interface("dsfunction", k).Err(err).Msgf("function call returned error")
```
blkt, line 74: **nit:** see comment above.

internal/datasources/service/service.go
```diff
@@ -448,6 +448,23 @@ func addDataSourceFunctions(
...
+			if _, err := tx.AddDataSourceFunction(ctx, db.AddDataSourceFunctionParams{
+				DataSourceID: dsID,
+				ProjectID:    projectID,
+				Name:         name,
+				Type:         v1datasources.DataSourceDriverDeps,
+				Definition:   defBytes,
+			}); err != nil {
+				return fmt.Errorf("failed to create data source function: %w", err)
+			}
+		}
```
blkt, line 467: **question:** is there a way to share this code path regardless of the type?

internal/datasources/deps/handler.go
```diff
@@ -0,0 +1,118 @@
...
+	var errs = []error{}
+	for _, es := range ecosystems {
+		if _, ok := purl.KnownTypes[es]; !ok {
+			errs = append(errs, fmt.Errorf("unkown ecosystem: %q", es))
+		}
+	}
+	return errs
+}
+
+func (_ *depsDataSourceHandler) ValidateUpdate(_ any) error { return nil }
```
blkt, line 95: **question:** I guess we could verify something here, like the cardinality of the functions, but might be impractical. Any ideas?

blkt (commented):
Left a few comments.
Could you share an example of a definition for this data source? It helps me a lot understanding its use from the perspective of the end user.

internal/engine/eval/rego/datasources.go
puerco: This is mostly for us to track the rego eval engine, it's hard to observe what it's doing without them. Would it be better if I set the level to Debug?

puerco:
I'll prioritize https://github.com/mindersec/minder/pull/5165 because we need that data source sooner, and then come back to this one.

evankanderson:
I borrowed much of the patterns of this code for #5326, which was an alternate approach.  Thanks for the work on this, @puerco !

Issue #4999, opened 2024-11-19
Cannot get profile status by profile ID (CLI flag is missing)


rdimitrov:
The issue is that we cannot get the status of a profile by its ID because the CLI flag is missing.

navnitms:
I would like to work on this issue

rdimitrov:
hey, @navnitms, of course, I've assigned it to you 👍 Thanks! 💯 🚀

navnitms:
@rdimitrov Just to clarify, am I correct in understanding that the requirement is to add a profileId flag to the "Get Profile Status" command?

rdimitrov:
Yes, one part of the task is to have something similar to the `-i/--id` flag we have for the `profile get` command. 

The other one is to add a new `GetProfileStatusById` handler similar to `GetProfileStatusByName` so you can tie all of this together 👍

navnitms:
I've started working on this issue

JAORMX:
Thank you!

rdimitrov:
Closing as completed 👍 Thank you @navnitms 💯

PR #5100, opened 2024-11-29
Add support for Get Profile Status By ID in Cli and Api

navnitms:
# Summary
Added support to get profile status by Profile Id in CLI
Reformatted GetProfileStatusByNameAndProject code to accommodate GetProfileStatusByIdAndProject
Reformatted GetProfileStatusByNameAndProject to reduce complexity to a permissible level (<15)

Fixes (#4999 )

## Change Type

# Testing

# Review Checklist:


navnitms:
Cant find any changes after running make gen

proto/minder/v1/minder.proto
```diff
@@ -606,6 +606,17 @@ service ProfileService {
         };
     }
 
+    rpc GetProfileStatusById (GetProfileStatusByIdRequest) returns (GetProfileStatusByIdResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/profile/id/{id=**}/status"
```
eleftherias, line 611: Following the convention set by the other API endpoints, this should be
```suggestion
            get: "/api/v1/profile/{id}/status"
```

internal/controlplane/handlers_profile.go
```diff
@@ -644,3 +554,273 @@ func (s *Server) UpdateProfile(ctx context.Context,
...
+	if err := entityCtx.ValidateProject(ctx, s.store); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "error in entity context: %v", err)
+	}
+
+	// Retrieve profile status
+	var dbProfileStatus interface{}
+	var getProfileErr error
+
+	switch id := profileIdentifier.(type) {
+	case string:
```
eleftherias, line 628: I worry that the switch statement here is fragile and makes it harder to follow the codepath.
Do you see a way to call `GetProfileStatusByNameAndProjectParams` or `GetProfileStatusByIdAndProjectParams` respectively in their corresponding top level functions and then pass the results to the helper?

cmd/cli/app/profile/status/status_get.go
```diff
@@ -76,6 +70,54 @@ func getCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc.
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Convert to the common response type
+		resp = &minderv1.GetProfileStatusResponse{
+			ProfileStatus:        nameResp.ProfileStatus,
+			RuleEvaluationStatus: nameResp.RuleEvaluationStatus,
+		}
+	}
```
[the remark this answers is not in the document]
navnitms: Updated the code and added the else case

internal/controlplane/handlers_profile.go
navnitms: Have refactored the code , to make the flow more simple to follow

navnitms:
@eleftherias Thanks for the review

proto/minder/v1/minder.proto
```diff
@@ -1919,6 +1930,55 @@ message GetProfileStatusByNameRequest {
...
+    // id is the id of the profile to get
+    string id = 2 [
+        (buf.validate.field).string = {uuid: true}
+    ];
+
+    EntityTypedId entity = 3;
+    bool all = 4;
+
+    // rule is the type of the rule. Deprecated in favor of rule_type
+    string rule = 5 [deprecated=true];
```
eleftherias, line 1954: Since this is a new API, we shouldn't add deprecated fields

internal/controlplane/handlers_profile.go
```diff
@@ -644,3 +554,256 @@ func (s *Server) UpdateProfile(ctx context.Context,
...
+
+	if extractFiltersErr != nil {
+		return nil, extractFiltersErr
+	}
+
+	// Retrieve rule evaluation statuses if needed
+	if selector != nil || isAllRequested(ctx) {
+		dbRuleEvaluationStatuses, err := s.store.ListRuleEvaluationsByProfileId(ctx, db.ListRuleEvaluationsByProfileIdParams{
+			ProfileID:    profileID,
+			EntityID:     *selector,
```
eleftherias, line 662: `selector` could be nil here which would result in a panic

internal/controlplane/handlers_profile.go
```diff
@@ -644,3 +554,256 @@ func (s *Server) UpdateProfile(ctx context.Context,
...
+
+func extractProfileDetails(dbProfileStatus interface{}) (
+	profileName string,
+	profileID uuid.UUID,
+	lastUpdated *timestamppb.Timestamp,
+	profileStatus string,
+	err error) {
+
+	switch ps := dbProfileStatus.(type) {
+	case db.GetProfileStatusByNameAndProjectRow:
```
eleftherias, line 722: The isn't much benefit in having a common helper function if it has switch statements which separate the code paths. 
In this case, we can change the signature of `processProfileStatus` to `processProfileStatus(ctx context.Context, profileName string, profileID uuid.UUID, lastUpdated *timestamppb.Timestamp, profileStatus string)` and then just pass the fields that we need:
```
	resp, err := s.processProfileStatus(
		ctx,
		dbProfileStatus.Name,
		dbProfileStatus.ID,
		timestamppb.New(dbProfileStatus.LastUpdated),
		string(dbProfileStatus.ProfileStatus),
	)
```

eleftherias (changes requested):
Thanks for the quick updates @navnitms!
I've left a few more comments inline.

eleftherias:
Hey @navnitms, happy new year! 
Do you think you'll have time to continue working on this enhancement?

navnitms:
Hey @eleftherias , Sorry for the delay
I was caught up in some work for a while
Sure i would love to complete this 
Also looking forward to contribute more

eleftherias:
No problem @navnitms and thank you for your work so far. Feel free to reach out if you have any questions.

navnitms:
Hey @eleftherias, Thanks for the support,
I have made the necessary updates

eleftherias (changes requested):
Thanks for the updates! I've left one comment inline.
There is also a change that needs to be made to the validators. Currently, the validator still requires a profile name. We should update it to require either a name or an ID:
https://github.com/mindersec/minder/blob/a02f7548b08561e4a9f4a1ddec629a90092c618b/pkg/api/protobuf/go/minder/v1/validators.go#L460-L471

navnitms:
Hi @eleftherias 

> There is also a change that needs to be made to the validators. Currently, the validator still requires a profile name. We should update it to require either a name or an ID:

I have some concerns about this proposed change. 
Are we planning for the validation to pass if only one of the ID or name is provided? Isn't the profile name supposed to be mandatory? Additionally, I couldn't find any instance where the validate function is called in relation to GetProfileStatus. Could you please clarify?

Additionally, I noticed that the check for whether profileId or profileName is passed is happening in the minder/cmd/cli/app/profile/get.go 


Apologies for the multiple iterations of review required for this PR.

eleftherias:
Thanks for you continued effort on this @navnitms! I really appreciate the conversation.

> Are we planning for the validation to pass if only one of the ID or name is provided? Isn't the profile name supposed to be mandatory? Additionally, I couldn't find any instance where the validate function is called in relation to GetProfileStatus. Could you please clarify?
Additionally, I noticed that the check for whether profileId or profileName is passed is happening in the minder/cmd/cli/app/profile/get.go

I see what you mean.
Since the `--id` flag was added to `status.go`, it will apply to both sub-commands, `profile status get` and `profile status list`. With the current changes if we run `minder profile status list --help` then it shows the option to use the `--id` flag, however it doesn't actually support it.
I see 2 options. Option 1 is to move the `id` flag to `status_get.go`. Option 2 is to implement getting the profile status by ID in `status_list.go` as well. With option 2 the validation will also need to be updated.
Do you have a preference between those options?

navnitms:
I find the second option to be more appropriate in this case.

Since we are already providing a method to get the profile status by ID, it makes sense to have the same functionality available when listing profiles as well.

However, I would prefer raising that as a separate PR, as this PR has become a bit cluttered.
That said, if you prefer merging those changes into this PR, that could also work.

eleftherias:
> I would prefer raising that as a separate PR, as this PR has become a bit cluttered.

That makes sense to me. Before merging this though, could you temporarily move the `id` flag to only the `profile get` command, or mark it as hidden?
The reasoning is that we release new versions of Minder frequently and we don't want to accidentally release a version with partially implemented code.

navnitms:
Sure, @eleftherias.
I have refactored the flags so that nothing breaks.

eleftherias (approved):

eleftherias:
Thank you for your contribution @navnitms!

navnitms:
Thanks for the active review @eleftherias , Will raise a separate PR for list profile status by id

Issue #4952, opened 2024-11-12
Set "REQUIRED" field behaviour flag for fields in minder protos 


eleftherias:
For clients using the openAPI spec, it would be useful to know which input fields are required and which output fields are non-nullable.

We can use the `REQUIRED` [field_behaviour flag](https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/#using-googleapifield_behavior) to annotate these types of fields.

For example by setting the behaviour to `REQUIRED`, the typescipt generated field is non-optional:
```
message Provider {
    // name is the name of the provider.
    string name = 1 [
        (google.api.field_behavior) = REQUIRED
    ];
```

produces the following typescipt type (note that it does not have a `?` marking it as optional):
```
type v1Provider = {
    /**
     * name is the name of the provider.
     */
    name: string;
}
```

This task involves finding all the fields that are both required as input **AND** non-nullable as output.

⚠️ Note that we reuse several messages as inputs and outputs. 
We can only set the `REQUIRED` flag on fields that are both required on input **AND** never null on output.
If a field is required on input, but may be null on output, we cannot use this flag.

eleftherias:
PR created for:
- `ListArtifactsRequest`
- `ListArtifactsResponse`
- `Artifact`
- `ArtifactVersion`
- `GetArtifactByIdRequest`
- `GetArtifactByIdResponse`
- `GetArtifactByNameRequest`
- `GetArtifactByNameResponse`

eleftherias:
PR created for:
- `GetInviteDetailsRequest`
- `GetInviteDetailsResponse`
- `CheckHealthResponse`
- `GetAuthorizationURLRequest`
- `GetAuthorizationURLResponse`
- `StoreProviderTokenRequest`

eleftherias:
PR created for:
- Project
- ListRemoteRepositoriesFromProviderResponse
- RegistrableUpstreamEntityRef
- UpstreamRepositoryRef
- Repository
- RegisterRepositoryRequest
- RegisterRepositoryResponse
- GetRepositoryByIdRequest
- GetRepositoryByIdResponse
- DeleteRepositoryByIdRequest
- DeleteRepositoryByIdResponse
- GetRepositoryByNameRequest
- GetRepositoryByNameResponse
- DeleteRepositoryByNameRequest
- DeleteRepositoryByNameResponse
- ListRepositoriesRequest
- ListRepositoriesResponse

PR #5110, opened 2024-12-02
Set required proto fields for rules and invitations

eleftherias:
# Summary

Ref #4952

## Change Type

# Testing

# Review Checklist:


proto/minder/v1/minder.proto
```diff
@@ -3170,7 +3170,9 @@ message ListRolesRequest {
 }
 
 message ListRolesResponse {
-    repeated Role roles = 1;
+    repeated Role roles = 1 [
+        (google.api.field_behavior) = REQUIRED
```
evankanderson, line 3174: Sorry for approving the other PR, it turns out that [`REQUIRED`](https://github.com/googleapis/googleapis/blob/master/google/api/field_behavior.proto#L58) applies to requests, but it's not clear what it means for replies.

https://google.aip.dev/203 is probably a good reference here, despite being Google-specific.  In that context, it means:

> A field should only be described as required if either:
> 
> * It is a field on a resource that a user provides somewhere as input. In this case, the resource is only valid if a "truthy" value is stored.
> * When creating the resource, a value must be provided for the field on the create request.
> * When updating the resource, the user may omit the field provided that the field is also absent from the field mask, indicating no change to the field (otherwise it must be provided).
> * It is a field on a request message (a message that is an argument to an RPC, with a name usually ending in Request). In this case, a value must be provided as part of the request, and failure to do so must cause an error (usually INVALID_ARGUMENT).
> 
> We define the term "truthy" above as follows:
> 
> * For primitives, values other than 0, 0.0, empty string/bytes, and false
> * For repeated fields maps, values with at least one entry
> * For messages, any message with at least one "truthy" field.

So ListRoles is _probably_ required here (as the user themselves should have a role), but invitations would not be because the list of outstanding invitations might be zero-length.

proto/minder/v1/minder.proto
```diff
@@ -3181,27 +3183,34 @@ message ListRoleAssignmentsRequest {
...
-    repeated RoleAssignment role_assignments = 1;
+    repeated RoleAssignment role_assignments = 1 [
+        (google.api.field_behavior) = REQUIRED
+    ];
 
     // invitations contains outstanding role invitations which have not yet
     // been accepted by a user.
-    repeated Invitation invitations = 2;
+    repeated Invitation invitations = 2 [
+        (google.api.field_behavior) = REQUIRED
```
evankanderson, line 3193: See above, probably not required.

evankanderson (commented):
This looks like it improves the swagger output so a general 👍 , but one concern about invitations.

proto/minder/v1/minder.proto
eleftherias: Empty lists are tricky because some languages consider them truthy and some don't. 
From a typescript perspective, an empty list is truthy and different than a nil list.
In the generated typescript type, setting a list field in a response as `REQUIRED` means it's not optional, and the client can safely do things like get the length, or iterate over it.
If we didn't set the field as required the client would have to perform a null check before using the field.

evankanderson: Oh, interesting!  it sounds like this is a difference between swagger semantics and proto semantics.  TIL!

evankanderson (approved):

PR #5182, opened 2024-12-11
Add a shared actions context structure

### LINKED ISSUE #5119: Combine all Minder PR comments into a single comment
When a new PR comment is added, we should combine it with existing comments from other rules, rather than having Minder create separate comments for each rule.

Out of scope:

- Gitlab

JAORMX:
# Summary

This shared space will allow actions to gather structures or results
that will allow it to aggregate and flush. In more solid terms, we'd be
able to aggregate PR comments, iterate them, and issue just one big
comment.

The intention is for drivers (e.g. the pull request comment alert type) to fetch this from the context and push shared statements (e.g. comments), plus an aggregation handling function that meets the relevant signature. The aggregation function will be called in the end, thus allowing it to gather all comments and publish just one result.

Fixes #5119

# Testing

# Review Checklist:

### DIFF
--- a/internal/engine/executor.go
+++ b/internal/engine/executor.go
@@ -139,21 +139,23 @@ func (e *executor) EvalEntityEvent(ctx context.Context, inf *entities.EntityInfo
 		return fmt.Errorf("error while retrieving profiles and rule instances: %w", err)
 	}
 
+	sacctx, sac := actions.WithSharedActionsContext(ctx)
+
 	// For each profile, get the profileEvalStatus first. Then, if the profileEvalStatus is nil
 	// evaluate each rule and store the outcome in the database. If profileEvalStatus is non-nil,
 	// just store it for all rules without evaluation.
 	for _, profile := range profileAggregates {
 
-		profileEvalStatus := e.profileEvalStatus(ctx, inf, profile)
+		profileEvalStatus := e.profileEvalStatus(sacctx, inf, profile)
 
 		for _, rule := range profile.Rules {
-			if err := e.evaluateRule(ctx, inf, provider, &profile, &rule, ruleEngineCache, profileEvalStatus); err != nil {
+			if err := e.evaluateRule(sacctx, inf, provider, &profile, &rule, ruleEngineCache, profileEvalStatus); err != nil {
 				return fmt.Errorf("error evaluating entity event: %w", err)
 			}
 		}
 	}
 
-	return nil
+	return sac.Flush()
 }
 
 func (e *executor) evaluateRule(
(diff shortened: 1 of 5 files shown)


internal/engine/executor.go
```diff
@@ -138,21 +138,23 @@ func (e *executor) EvalEntityEvent(ctx context.Context, inf *entities.EntityInfo
 		return fmt.Errorf("error while retrieving profiles and rule instances: %w", err)
 	}
 
+	sacctx, sac := actions.WithSharedActionsContext(ctx)
```
evankanderson, line 141: Why not still use `ctx` for `sacctx`?

evankanderson (commented):
How do we see this aggregation working?  To quote from https://github.com/mindersec/minder/issues/5119#issuecomment-2531787062

> If license scanning takes 100ms and static analysis takes 5 minutes, do we add the license scan report right away and then update the comment later when static analysis comes in?

I'm having a little trouble visualizing how this works -- I think this assumes that all the rule eval and actions for a given event happen on the same Minder server?  And we'll end up with something on the event-handling context that allows "enqueue this to be done in the future"?

Thinking about audit logs and splitting action execution from action computation, I'd love to have a _data structure_ for collecting these actions, rather than a list of functions.  Ideallly, this would look like:

```
{notification / reminder edge} --> [visit $foo resource]
[visit $foo resource] -> (per-rule) {rule-eval} -> (per-rule) [eval results] & [action plan?]
[action plan] -> {action-processor} -> [action result]
[eval results] -> {audit-log}
[action result] -> {audit-log}
```

Where `[message]`s are in brackets, and `{computation}` is in braces.

JAORMX:
> How do we see this aggregation working? To quote from [#5119 (comment)](https://github.com/mindersec/minder/issues/5119#issuecomment-2531787062)
> 
> > If license scanning takes 100ms and static analysis takes 5 minutes, do we add the license scan report right away and then update the comment later when static analysis comes in?
> 
> I'm having a little trouble visualizing how this works -- I think this assumes that all the rule eval and actions for a given event happen on the same Minder server?

Right, it assumes an set of profiles will be evaluated for one event in a single Minder server. As is the case today.

> And we'll end up with something on the event-handling context that allows "enqueue this to be done in the future"?
> 

I was thinking of kicking it off immediately for now. We probably need to add such mechanism later when the use-case comes.


> Thinking about audit logs and splitting action execution from action computation, I'd love to have a _data structure_ for collecting these actions, rather than a list of functions. Ideallly, this would look like:
> 
> ```
> {notification / reminder edge} --> [visit $foo resource]
> [visit $foo resource] -> (per-rule) {rule-eval} -> (per-rule) [eval results] & [action plan?]
> [action plan] -> {action-processor} -> [action result]
> [eval results] -> {audit-log}
> [action result] -> {audit-log}
> ```
> 
> Where `[message]`s are in brackets, and `{computation}` is in braces.

The eval params gets us partly there, the main missing piece is the action results. Let me give that part some thought.

JAORMX:
superseded by https://github.com/mindersec/minder/pull/5223

PR #5228, opened 2024-12-19
Add `mindev ruletype init` to kick off a rule type

JAORMX:
# Summary

This helps folks set up the basic skeleton for ruletype writing.

Fixes #(related issue)

## Change Type

# Testing

# Review Checklist:


cmd/dev/app/rule_type/init.go
```diff
@@ -0,0 +1,188 @@
...
+
+			return fmt.Errorf("file %s already exists", file)
+		}
+	}
+
+	return nil
+}
+
+func createRuleTypeFile(fileName, name string) error {
+	return createFileWithContent(fileName, fmt.Sprintf(`---
```
evankanderson, line 112: It feels like we should use `text/template` here rather than `fmt.Sprintf`, particularly if we start to input more data before outputting the rule.

cmd/dev/app/rule_type/init.go
```diff
@@ -0,0 +1,188 @@
...
+    expect: "pass"
+    entity: &test-repo
+      type: repository
+      entity:
+        owner: "coolhead"
+        name: "haze-wave"
+    # http:
+    #   body_file: HTTP_BODY_FILE
+    # git:
+    #   repo_base: REPO_BASE_PATH
```
evankanderson, line 164: It feels like these could benefit from a bit more documentation in a comment, e.g.

```suggestion
    # When testing a rule, additional content can be supplied
    # from files in the `{{ .RuleName }}.testdata` directory.
    # File paths below are relative to this directory.
    # http:
    #  # Input from the `http` ingest type.
    #  body_file: HTTP_BODY_FILE
    # git:
    #   # Input from the `git` ingest type.  Base paths contain
    #   # directory contents, but do not actually need to be a
    #   # git repository.
    #   repo_base: REPO_BASE_PATH
```

cmd/dev/app/rule_type/init.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating file %s: %w", fileName, err)
+	}
+	defer file.Close()
+
+	if _, err := file.WriteString(content); err != nil {
+		return fmt.Errorf("error writing to file %s: %w", fileName, err)
+	}
+
+	return nil
```
evankanderson, line 187: Can we simply use [`os.WriteFile(fileName, byte[](content), 0644)`](https://pkg.go.dev/os#WriteFile) for this?

cmd/dev/app/rule_type/init.go
```diff
@@ -0,0 +1,188 @@
...
+		return errors.New("name must only contain alphanumeric characters and underscores")
+	}
+
+	return nil
+}
+
+func assertFilesDontExist(files ...string) error {
+	for _, file := range files {
+		if _, err := os.Stat(file); err == nil {
+
```
evankanderson, line 103: Odd empty newline here.

cmd/dev/app/rule_type/init.go
```diff
@@ -0,0 +1,188 @@
...
+			return fmt.Errorf("file %s already exists", file)
```
evankanderson, line 104: Not sure if you want to check `os.IsExist(err)` here before returning this error -- there could be a different reason the stat fails, like a missing parent directory.

cmd/dev/app/rule_type/init.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return nil
+}
+
+func validateRuleTypeName(name string) error {
+	if name == "" {
+		return errors.New("name cannot be empty")
+	}
+
+	validName := regexp.MustCompile(`^[a-zA-Z0-9_]+$`)
```
evankanderson, line 90: This is a different regex from https://github.com/mindersec/minder/blob/main/proto/minder/v1/minder.proto#L2496.  Not sure if we want to try to re-use that validation?

cmd/dev/app/rule_type/init.go
```diff
@@ -0,0 +1,188 @@
...
+
+	if err := initCmd.MarkFlagRequired("name"); err != nil {
+		fmt.Fprintf(os.Stderr, "Error marking flag as required: %s\n", err)
+		os.Exit(1)
+	}
+
+	return initCmd
+}
+
+func initCmdRun(cmd *cobra.Command, args []string) error {
```
evankanderson, line 41: What do you think about having this operate in a more "wizard" style in terms of prompting for missing data rather than people needing to figure out the flag arguments up front and then getting a file with a bunch of comments to fill in?

JAORMX: not sure I see much value in that. There is one argument to figure out.

cmd/dev/app/rule_type/init.go
JAORMX: I'm not sure how to check this other than trying to create the files or directories.

evankanderson: Oh, I mis-read the code (thought it was `err != nil`), and thought you were trying to call `os.Create` and then returning an error if the create failed.

cmd/dev/app/rule_type/init.go
evankanderson: A wizard-type mode (or extra command-line flags) could fill out the commented sections of the template.  For example, if we had a `display_name`, we could automatically create the `RuleName` based on the display name.  We could also prompt for the critical fields like `description`, `short_failure_message`, and `in_entity`.

evankanderson (approved):
Approving, but also explaining the wizard idea a bit more clearly.

I'm also still wondering whether Rego rule types should be able to have a top-level `.rego` file, so that it can work with language servers like [Regal](https://github.com/styrainc/regal).

rdimitrov (approved):
LGTM 👍 If we want we can add Evan's suggestions regarding the wizard-type UX as follow up PRs

PR #5292, opened 2025-01-13
Add alias to data source reference

eleftherias:
# Summary

Add an alias to a data source reference in the rule type, so the data source can be referenced in the Rego code with an alias different than its name.

If we include data sources in a bundle, then the name will be namespaced, e.g "custom/my-data-source". This syntax is not easy to refer to in rego code, so instead we can use an alias.

Example:
```
  eval:
    type: rego
    data_sources:
      - name: osv
        alias: vulnerability-db
```

## Change Type

# Testing

# Review Checklist:


rdimitrov (commented):
What would happen in case there's 2 data sources with the same alias? I suppose one way is to make ensure they're unique and validate this upon data source creation.

eleftherias:
> What would happen in case there's 2 data sources with the same alias? I suppose one way is to make ensure they're unique and validate this upon data source creation.

Thanks for catching that. Right now it will error on evaluation, but you're right it would be much better to error on creation. I'll update the PR.

PR #5320, opened 2025-01-17
Add documentation style guide and tool configs

danbarr:
# Summary

This PR proposes a documentation style guide and configurations for Prettier, markdownlint, and ESLint tooling. The style guide is a subset of Google's documentation and Markdown style guides.

If accepted, I will follow up with another PR to format/lint all of the current docs files.

A final step will be to propose automated checks by way of GitHub workflows to validate formatting on PRs that modify docs.

## Change Type

# Testing

Tested that the prettier, markdownlint, and eslint configurations are reporting issues as expected. Auto-generated docs (API/CLI/protocol) have been excluded/ignored.

# Review Checklist:

- [ ] (N/A) Included tests that validate the fix or feature.
- [ ] (N/A) Checked that related changes are merged.


.gitattributes
```diff
@@ -5,7 +5,7 @@
 # linguist-generated causes generated files to not be expanded during code review.
 
 /database/mock/*.go linguist-generated=true
-/docs/docs/cli/*.md linguist-generated=true
+/docs/docs/ref/cli/*.md linguist-generated=true
```
evankanderson, line 8: Hah, thank you!

docs/.markdownlint.json
```diff
@@ -0,0 +1,37 @@
...
+  "hr-style": {
+    "style": "---"
+  },
+  "line-length": {
+    "code_blocks": false,
+    "line_length": 80,
+    "tables": false
+  },
+  "no-bare-urls": false,
+  "no-duplicate-heading": false,
```
evankanderson, line 23: Why is this false?

docs/.markdownlint.json
```diff
@@ -0,0 +1,37 @@
...
+    "names": ["GitHub", "GitLab", "Bitbucket"]
+  },
+  "strong-style": {
+    "style": "asterisk"
+  },
+  "table-pipe-style": {
+    "style": "leading_and_trailing"
+  },
+  "ul-style": {
+    "style": "dash"
```
evankanderson, line 35: I always use `*`, but this dog _can_ learn new tricks!

docs/README.md
```diff
@@ -1,25 +1,66 @@
...
 ```
 
-Build the docs
-```
-cd docs
-```
+Run a preview server:
 
+```bash
+npm run start
```
evankanderson, line 30: This needs to be in the `docs` repo, right?  Maybe that's obvious, but the Makefile is at the top level.

```suggestion
cd docs
npm run start
```

docs/STYLE-GUIDE.md
```diff
@@ -0,0 +1,273 @@
...
+**don’t** capitalize features or generic terms. For non-Minder terms, follow the
+norms of the third-party project/company (ex: npm is stylized in lowercase, even
+when it begins a sentence).
+
+:white_check_mark: Yes: Minder profiles are a collection of rule types that are
+applied to entities.\
+:x: No: Minder Profiles are a collection of Rule Types that are applied to
+Entities.
+
+Use **sentence case** in titles and headings.
```
evankanderson, line 74: Oh man, I am gonna get this one wrong.  Thanks!

evankanderson (approved):

docs/.markdownlint.json
danbarr: Not sure which one you mean, but

- tables for line-length: wrapping tables makes them really, really hard to edit
- no-bare-urls: the auto-fix for this (wrapping a bare URL in <>'s) breaks Docusaurus rendering since it runs everything thru an MDX parser, even .md files. I'd like to  be able to check for these but not auto-fix them, but currently markdownlint doesn't support making that distinction.
- no-duplicate-heading: I think we can actually turn this one back on, the rule reference pages have a lot of these but I put this into a more-specific config file just in that folder so I don't think any of this rule gets flagged elsewhere. I'll do that on the next PR.

docs/.markdownlint.json
danbarr: I used to also; the rationale I drew this from was that it's easier to read when you do things like this:

```markdown
* **A bold bullet**

vs.

- **A bold bullet**
```

But frankly, this is a nit - if `*` are more comfortable/natural I'm not opposed to changing this - consistency is more important than _which_ character.

Markdownlint also support a `sublist` style which uses different symbols as you nest:

```markdown
* Item 1
  + Item 2
    - Item 3
  + Item 4
* Item 4
  + Item 5
```

docs/README.md
danbarr: Whoops good catch, I added this in and missed that the `cd docs` is below it now. I'll fix in the next PR.

docs/.markdownlint.json
danbarr: Fixed in #5336

docs/README.md
danbarr: Fixed in #5336

PR #5326, opened 2025-01-18
Add dependency extraction functions

### LINKED ISSUE #5090: Dependency Data Source Driver

evankanderson:
# Summary

#5094 attempted to add a data source for extracting dependencies.  On reflection, this seems simpler to implement as a standard rego function (available in all projects without needing to create a specific data source).

Fixes #5090

## Change Type

# Testing

Added unit tests.

# Review Checklist:

### DIFF
--- a/internal/engine/eval/rego/lib.go
+++ b/internal/engine/eval/rego/lib.go
@@ -806,7 +806,7 @@ func BaseDependencyExtract(res *interfaces.Result) func(*rego.Rego) {
 	)
 }
 
-func fsExtractDeps(vfs billy.Filesystem) func (rego.BuiltinContext, *ast.Term) (*ast.Term, error) {
+func fsExtractDeps(vfs billy.Filesystem) func(rego.BuiltinContext, *ast.Term) (*ast.Term, error) {
 	return func(bctx rego.BuiltinContext, op1 *ast.Term) (*ast.Term, error) {
 		var path string
 		if err := ast.As(op1.Value, &path); err != nil {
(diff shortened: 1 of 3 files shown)


internal/engine/eval/rego/lib.go
```diff
@@ -763,3 +771,105 @@ func ParseToml(_ *interfaces.Result) func(*rego.Rego) {
...
 	)
 }
+
+// DependencyExtract is a rego function that extracts dependencies from a file
+// or subtree of the filesystem being evaluated (which comes from the ingester).
+// It takes one arguments: the path to the file or subtree to be scanned.
+// It returns the extracted dependencies as an AST term in the form of a
+// protobom SBOM with the "nodes" fields but not "edges".
+// It's exposed as `file.deps`.
+func DependencyExtract(res *interfaces.Result) func(*rego.Rego) {
```
JAORMX, line 781: What's the usecase of this function? Can't we just do the same with the dependency ingest?

Also, we have the ingest result as part of the rego context we pass already, you could probably leverage that. Wdyt?

evankanderson: The idea is that you might want to be able to do either a more-targeted dependency ingest, or do a conditional dependency ingest.  See #5090 for the general idea, but a specific application would be:

* Check to see whether a pages publishing action exists
* If so, check for NPM dependencies specifically in that tree, as pages documentation is generally on a different release cycle than the main binary

JAORMX (approved):

PR #5283, opened 2025-01-10
mindev: Add utility to generate a data source definition from a Swagger doc

JAORMX:
# Summary

This adds the following command: `mindev datasource generate ...`

The aforementioned command will read a swagger doc (OpenAPI v2) and will
generate the appropriate data source definition out of it.

I went for v2 since there was an easy-to-use implementation ready.
Adding v3 support is left as a TODO item.

Here's a demo of how usage would look like:

```
$ cat sample.yaml
swagger: "2.0"

info:
  title: Sample API
  description: Optional multiline or single-line description in [CommonMark](http://commonmark.org/help/) or HTML.
  version: 0.1.9

basePath: http://api.example.com/v1

paths:
  /users:
    get:
      summary: Returns a list of users.
      description: Optional extended description in CommonMark or HTML.
      responses:
        "200": # status code
          description: A JSON array of user names
          content:
            application/json:
              schema:
                type: array
                items:
                  type: string

$ go run cmd/dev/main.go datasource generate sample.yaml
version: v1
type: data-source
context: {}
name: Sample-API
rest:
  def:
    get_users:
      endpoint: http://api.example.com/v1/users
      method: GET
      parse: json
      inputSchema: {}

```

## Change Type

***Mark the type of change your PR introduces:***

- [ ] Bug fix (resolves an issue without affecting existing features)
- [x] Feature (adds new functionality without breaking changes)
- [ ] Breaking change (may impact existing functionalities or require documentation updates)
- [ ] Documentation (updates or additions to documentation)
- [ ] Refactoring or test improvements (no bug fixes or new functionality)

# Testing

***Outline how the changes were tested, including steps to reproduce and any relevant configurations. 
Attach screenshots if helpful.***

# Review Checklist:

- [x] Reviewed my own code for quality and clarity.
- [ ] Added comments to complex or tricky code sections.
- [ ] Updated any affected documentation.
- [ ] Included tests that validate the fix or feature.
- [ ] Checked that related changes are merged.

PR #5439, opened 2025-02-14
Add unit test for generating data source

gajananan:
# Summary

Add unit tests for Swagger datasource definition generation (#5283)

- Implement unit tests to validate datasource definition generation from Swagger files.
- Ensure hardcoded OpenAPI v2 docs produce expected datasource definitions.
- Maintain robustness of generation command to prevent accidental breaks.

Ref  #(5300)

## Change Type

# Testing

# Review Checklist:


cmd/dev/main_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+			assert.NoError(t, err, "Command execution should not produce an error")
+
+			// Handle the case for missing OpenAPI file
+			if tt.name == "missing OpenAPI file" {
+				mu.Lock()
+				assert.FileExists(t, tt.openAPIFile, "OpenAPI file should not exist")
+				mu.Unlock()
+				return
+			}
```
evankanderson, line 127: Do we ever get here, given that `tt.expectedError == true` for the "missing OpenAPI file" case, and that exits on line 109?

(This will allow you to simplify the test code, which is another win)

cmd/dev/main_test.go
```diff
@@ -0,0 +1,141 @@
...
+			assert.NoError(t, err, "Command execution should not produce an error")
+
+			// Normalize and compare the YAML strings
+			expectedYAML := strings.Join(strings.Fields(string(tt.expectedData)), "")
+			generatedYAML := strings.Join(strings.Fields(string(output)), "")
+			// Compare the YAML strings directly
+			assert.Equal(t, expectedYAML, generatedYAML, "Generated datasource definition should match expected")
+
+			// Add a slight delay to ensure the output is captured correctly
+			time.Sleep(100 * time.Millisecond)
```
evankanderson, line 138: Why do we need a "Sleep" after we've read and compared all the content?

evankanderson (commented):
Thanks for this!  This looks fairly good, but a couple questions about the implementation.

I also noticed a data race warning; you might want to try running with `-race -count 20` to see if you can shake it out.

cmd/dev/main_test.go
```diff
@@ -0,0 +1,141 @@
...
+			os.Stdout = originalStdout
+
+			// Read the captured output
+			output := <-outC
+
+			assert.NoError(t, err, "Command execution should not produce an error")
+
+			// Handle the case for missing OpenAPI file
+			if tt.name == "missing OpenAPI file" {
+				mu.Lock()
```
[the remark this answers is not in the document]
gajananan: This section should not be there, the mutex. should not be.

cmd/dev/main_test.go
gajananan: Due to the following code to capture output from Command execution, at times the output was not captured properly, I tried few option to see, it is one of them. I agree it does not make sense when I look at it.
                       `// Create a pipe to capture the output,
			r, w, _ := os.Pipe()
			os.Stdout = w

			// Redirect the output to the buffer in a separate goroutine
			outC := make(chan string)
			go func() {
				var buf bytes.Buffer
				_, err = io.Copy(&buf, r)
				assert.NoError(t, err, "Buffer copy should not produce an error")

				outC <- buf.String()
			}()`

PR #5478, opened 2025-03-04
Remove hard-coded 'stacklok' realm from minder

evankanderson:
# Summary

While we were getting started, we hard-coded the "stacklok" realm in a bunch of places in the Minder source code.  While the code is still entwined with Keycloak, this change should allow other users to use a different Keycloak realm than "stacklok".

## Change Type

# Testing

All unit tests pass, did a quick integration pass.

# Review Checklist:


cmd/cli/app/auth/auth_logout.go
```diff
@@ -39,7 +39,8 @@ var logoutCmd = &cobra.Command{
 		// See https://github.com/spf13/cobra/issues/340#issuecomment-374617413
 		cmd.SilenceUsage = true
 
-		logoutUrl := parsedURL.JoinPath("realms/stacklok/protocol/openid-connect/logout")
+		// TODO: we should probably get this from the server in a header in the future.
+		logoutUrl := parsedURL.JoinPath("realms", clientConfig.Identity.CLI.Realm, "/protocol/openid-connect/logout")
```
JAORMX, line 43: Should we just do a breaking change and remove the realm entirely? That's a keycloak concept that we shouldn't have added to our configurations.

cmd/cli/app/auth/offline_token/offline_use.go
```diff
@@ -57,8 +57,9 @@ func offlineUseCommand(_ context.Context, cmd *cobra.Command, _ []string, _ *grp
 
 	issuerUrlStr := clientConfig.Identity.CLI.IssuerUrl
 	clientID := clientConfig.Identity.CLI.ClientId
+	realm := clientConfig.Identity.CLI.Realm
 
-	creds, err := cli.RefreshCredentials(tok, issuerUrlStr, clientID)
+	creds, err := cli.RefreshCredentials(tok, issuerUrlStr, realm, clientID)
```
JAORMX, line 62: Is it possible for us to not require a realm here? would be ideal to move away from that.

JAORMX (dismissed):

cmd/cli/app/auth/auth_logout.go
evankanderson: I put in a TODO, because it would be great if we didn't expose the Keycloak config information to our clients.  That requires a bit more thought, and I was checking the process of standing up an independent Minder and noticed that the realm needed to be called "stacklok".

I definitely would like it if this was less Keycloak-specific, but I think that's separate from making the realm configurable.

Ideally, I think we would use a header from the API service like `WWW-Authenticate: OAuth realm=https://auth.stacklok.com/realms/stacklok/protocol/openid-connect/auth` as [per RFC 5849](https://www.rfc-editor.org/rfc/rfc5849.html#section-3.5.1) to communicate the login URL, rather than hard-coding that knowledge into the client.  That's what the TODO is alluding to, but would be a much larger piece of work.

cmd/cli/app/auth/offline_token/offline_use.go
evankanderson: Agree that it would be ideal.  I may try to hack that together this weekend.

JAORMX (approved):

PR #5480, opened 2025-03-04
Add API for dealing directly with entity instances

JAORMX:
# Summary

This will eventually replace the API for repositories and is intended as
a general-purpose API for entity instances. As Minder moves to having a
more general model for entities themselves, this is required to be able
to work with new entities.

## Change Type

# Testing

# Review Checklist:


internal/controlplane/handlers_entity_instances.go
```diff
@@ -0,0 +1,311 @@
...
+	ctx context.Context,
+	in *pb.GetEntityByIdRequest,
+) (*pb.GetEntityByIdResponse, error) {
+	// Parse entity ID
+	entityID, err := uuid.Parse(in.GetId())
+	if err != nil {
+		return nil, util.UserVisibleError(codes.InvalidArgument, "invalid entity ID")
+	}
+
+	projectID := GetProjectID(ctx)
```
evankanderson, line 99: Interesting that List and Get-by-Name require a provider and a type, but Get-by-ID does not.  Might be worth documenting that.

internal/controlplane/handlers_entity_instances.go
```diff
@@ -0,0 +1,311 @@
...
+	qtx := s.store.GetQuerierWithTransaction(tx)
+
+	// Convert pb.Entity to db.Entities
+	dbEntityType, err := entities.EntityTypeToDBType(entityType)
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert entity type: %w", err)
+	}
+
+	// Create entity
+	entity, err := qtx.CreateEntity(ctx, db.CreateEntityParams{
```
evankanderson, line 255: Does this allow creating a repository which Minder doesn't have access to (or doesn't exist)?

I'd suggest keeping `CreateEntity` out of this PR, as I think the create path is substantially more complicated than the "Get" path.

internal/entities/service/service.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+
+		// Convert to protobuf
+		pbEntity := entityInstanceToProto(ewp)
+
+		results = append(results, pbEntity)
+	}
+
+	// Commit transaction
+	if err := tx.Commit(); err != nil {
```
evankanderson, line 179: Why do we need to commit the transaction?  There shouldn't be any writes here, correct?

evankanderson (commented):
A few questions, mostly about `CreateEntity`, which I think we might want to put off for now.

proto/minder/v1/minder.proto
```diff
@@ -4024,6 +4028,194 @@ message EntityInstance {
...
+        option (google.api.http) = {
+            post: "/api/v1/entity"
+            body: "*"
+        };
+
+        option (rpc_options) = {
+            target_resource: TARGET_RESOURCE_PROJECT
+            relation: RELATION_ENTITY_CREATE
+        };
+    }
```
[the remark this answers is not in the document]
JAORMX: Perhaps I should rename it `RegisterEntity`? We'll be registering organizations too.

internal/authz/model/minder.fga
```diff
@@ -56,6 +56,11 @@ type project
 
     define entity_reconcile: editor
 
+    define entity_get: viewer
+    define entity_create: editor
+    define entity_update: editor
+    define entity_delete: editor
```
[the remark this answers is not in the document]
JAORMX: They are, I thought I'll add an explicit `entity_*` and later on we'll deprecate the `<entity>_*` relationships in favor of the `entity_*` ones.

internal/controlplane/handlers_entity_instances.go
JAORMX: I actually still don't know how to feel about that. In a multi-provider world, getting by ID might be an easy way to fetch an entity for a particular provider. Perhaps I should allow for further filtering with provider ID.

internal/controlplane/handlers_entity_instances.go
JAORMX: You're right, let me remove it to make life easier. Mind if I leave this (or rename to `RegisterEntity` and just throw an error always?

proto/minder/v1/minder.proto
```diff
@@ -4024,6 +4028,194 @@ message EntityInstance {
...
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // entity_type is the type of entity to create
+    Entity entity_type = 3 [
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // properties is a map of properties of the entity
+    google.protobuf.Struct properties = 4;
```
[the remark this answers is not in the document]
JAORMX: So... When registering an entity, we need a set of "identifying properties" which I've talked about in presentations but never actually documented (oops!). These properties are meant to aid the provider to look up the entity upstream and to register them. I could probably rename this field to `identifying_properties` to make that clearer and actually add documentation.

We'll need this in the general entity flow since the base attributes of entities aren't enough to reliably identify an entity upstream. e.g. for GitLab one often needs the `upstream_id` to identify the project, which is not part of the base set of attributes (context, name, entity_type), but can be looked for with the properties.

Great catch! I should make an effort to portray and document this a lot better.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
evankanderson: I feel like `RegisterEntity` may make it more clear that this needs to line up with things that already exist in the outside world.

proto/minder/v1/minder.proto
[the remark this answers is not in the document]
evankanderson: Hmm... I think `identifying_properties` may be better, but I'm wondering whether making this single entry point which is deeply polymorphic in its contents will be confusing to use.

It's easy for me to approve the other parts; this API is one that I need to think a bit more about because the implementation seems tricky.

internal/controlplane/handlers_entity_instances.go
evankanderson: I think throwing UNIMPLEMENTED is fine, but I'm not sure how I feel about committing to the parameters of the call when we don't understand the implementation.

(For example, we can use name + type + provider to pick out an entity once it's been registered -- why can't we do that for the registration as well?)

proto/minder/v1/minder.proto
```diff
@@ -4024,6 +4028,197 @@ message EntityInstance {
...
+            pattern: "^[A-Za-z][-/[:word:]]*$",
+            max_len: 200,
+        },
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // entity_type is the type of entity to get
+    Entity entity_type = 3 [
+        (google.api.field_behavior) = REQUIRED
+    ];
```
evankanderson, line 4157: Should the URL for this be `/api/v1/entity/{type}/{name=**}`?

I think it's fine to leave this like this for now, and possibly change the URL later

internal/entities/service/service.go
evankanderson: (this is still open)

internal/entities/service/service.go
```diff
@@ -0,0 +1,332 @@
...
+	pb "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+)
+
+//go:generate go run go.uber.org/mock/mockgen -package mock_$GOPACKAGE -destination=./mock/$GOFILE -source=./$GOFILE
+
+// ErrEntityNotFound is returned when an entity is not found
+var ErrEntityNotFound = errors.New("entity not found")
+
+// EntityService encapsulates logic related to entity instances
+type EntityService interface {
```
evankanderson, line 33: We're relying on GPRC implementing the "unimplemented" result for RegisterEntity, correct?

evankanderson (commented):
I saw some changes, so I added a couple comments, but it looks like you're still working on this.

internal/entities/service/service.go
JAORMX: Giving this more thought, I think it's actually a good idea to leave this and explicitly COMMIT. Even if nothing is being written, this service _might_ be used in such a way that this is called first and then a write operation, which would have an unintended effect of having an unclosed transaction. But maybe I'm overthinking this, what do you think? Is this a valid concern?

proto/minder/v1/minder.proto
JAORMX: I like this!

internal/entities/service/service.go
JAORMX: duh! yes, we could do that. I left the implementation and returned an error explicitly. Let me remove that entirely.

internal/controlplane/handlers_entity_instances.go
JAORMX: I removed the name since it's something that the provider will derive. However, I left entity type since I figured that using the properties to derive that might be fragile. wdyt?

internal/entities/service/service.go
evankanderson: I was wondering why not simply rollback

proto/minder/v1/minder.proto
```diff
@@ -4024,6 +4028,188 @@ message EntityInstance {
...
+    // entity_type is the type of entity to create
+    Entity entity_type = 2 [
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // identifying_properties is a map of the minimal set of properties
+    // the provider needs in order to uniquely identify an entity.
+    // In most cases, upstream_id will suffice, but this is really
+    // provider-dependent.
+    google.protobuf.Struct identifying_properties = 3;
```
evankanderson, line 4202: This feels "weird" to me.  Specifying a type and a "name in the provider context" feels more natural and less surprising to me.  Consider the following request:

```
context:
  project: "ABC123"
  provider: "gitlab"
entity_type: REPOSITORY
identifying_properties:
  map:
    "github/repo_owner": "JAORMX"  # From github repo
    "tag": "fancybranch"  # From github release
    "gitlab/license": "Apache-2.0"  # From gitlab repo
    "gitlab/project_id": "12460"  # From gitlab MR
    "upstream_id": "12460"  # The thing you mention here
```

Given that we don't expect users to be able to set the other properties, taking a full-blown object (not even a string-string map), and then throwing out all the other keys feels like a clumsy interface on both sides.

evankanderson (commented):
I'm still concerned that we don't understand how users will want to use `RegisterEntity` well enough to handle it.  For the GitHub repository case (which is the only one we've thought about so far), we have a special call to get the "view" of the available repositories through the lens of the provider.

proto/minder/v1/minder.proto
JAORMX: Yeah, it's not ideal that you could pass anything; but the multi-provider nature of this makes it kinda complicated. Ideally it should look as follows:

```
identifying_properties:
  map:
    "upstream_id": "12460"  # The thing you mention here
```

Since you only need to pass the bare minimum and the rest of the properties will be ignored. This works in _some_ providers, but GitHub... again... is kinda special since it does need the owner and name to properly identify a resource.

evankanderson: For GitHub repos, you could use the `owner/repo` slug, which is common enough that it works on the gh CLI and elsewhere.

I feel like there could be a "list unregistered" that gives strings in a provider-specific format, and then a "register" that takes strings in the same format.

JAORMX: So, you think we should replace the github upstream_id to leverage that slug?

evankanderson: For input from users, the slug feels correct and friendly.  For storage, the numeric upstream id is correct to support repo renames and moves.  For output, we should include both.

I'm not particularly worried about a race between a user supplying a slug in registration and our looking up the id through the provider.  I *do* think we need to hook through the provider during the create, which will give the opportunity to resolve the upstream id and also to register webhooks, etc where needed.

evankanderson (approved):

PR #5502, opened 2025-03-13
Determine authentication host based on WWW-Authenticate header

evankanderson:
# Summary

With these changes, I no longer need to set the client-side `identity` parameters, and I can call Minder as:

```
./bin/minder --grpc-host localhost --grpc-port 8090 --grpc-insecure auth login
```

And it will log in with the local keycloak instance.  For other production instances, it should be possible to simply use `--grpc-host` to select them, and not need to explicitly know about their Keycloak configuration.  (And, at some point, it could be non-keycloak...)

On the server side, this requires something like the following `identity` configuration:

```yaml
identity:
  server:
    issuer_url: http://keycloak:8080 # Use http://localhost:8081 instead for running minder outside of docker-compose
    issuer_claim: http://localhost:8081/realms/stacklok # This needs to match keycloak's configured issuer, not the network path from Minder
    realm: stacklok
    client_id: minder-server
    client_secret: secret
    audience: minder
    scope: minder-audience
```

## Change Type

# Testing

Manual testing with a number of different configurations.  Along the way, I discovered that #5478 actually broke `minder auth login` for Stacklok / the default configuration, because we weren't setting the `mapstructure` defaults on the client, only on the server.  (There's a fix in here for that).

# Review Checklist:

### DIFF
--- a/cmd/cli/app/auth/auth_token.go
+++ b/cmd/cli/app/auth/auth_token.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"google.golang.org/grpc"
 
 	"github.com/mindersec/minder/internal/util/cli"
 	"github.com/mindersec/minder/pkg/config"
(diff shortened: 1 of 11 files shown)


evankanderson:
Let me know if you'd like me to break out the config changes from the other changes... I sort of ended up refactoring a bunch of stuff that had felt awkward, but I can see that it makes a large total set of things to keep in your head.

JAORMX (dismissed):

JAORMX (approved):

PR #5586, opened 2025-04-15
Enable CreateProject and DeleteProject on top-level projects

### LINKED ISSUE #3173: Allow creation of top-level projects
Currently, we allow the creation of sub-projects, projects that are children of another project.
This issue tracks creating top-level projects, projects without a parent project.

We can repurpose the existing `minder project create` command. 
We should deprecate the existing `--project` flag and introduce a new `--parent` flag to make it explicit that the new project will be a sub-project.
If no `--parent` flag is provided, then we should assume that it is a top-level project.

### LINKED ISSUE #3174: Allow deletion of top-level projects
Depends on https://github.com/stacklok/minder/issues/3173, because without it, users may end up with no projects and no way to create a new project.

We currently block the deletion of top-level projects, projects that don't have a parent project.
Once https://github.com/stacklok/minder/issues/3173 is complete, we should allow all projects to be deleted.

evankanderson:
# Summary

Allow CreateProject (with auth but no permission check) on top-level projects.

Allow DeleteProject (with required project permissions) on top-level projects.

Both changes are behind a new feature flag.

Fixes #3173
Fixes #3174

## Change Type

# Testing

Added a few unit tests, also doing some manual testing.

# Review Checklist:

### DIFF
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -80,13 +80,8 @@ func ProjectAuthorizationInterceptor(ctx context.Context, req interface{}, info 
 
 	relation := opts.GetRelation()
 
-	relationValue := relation.Descriptor().Values().ByNumber(relation.Number())
-	if relationValue == nil {
-		return nil, status.Errorf(codes.Internal, "error reading relation value %v", relation)
-	}
-	extension := proto.GetExtension(relationValue.Options(), minder.E_Name)
-	relationName, ok := extension.(string)
-	if !ok {
+	relationName := relationAsName(relation)
+	if relationName == "" {
 		return nil, status.Errorf(codes.Internal, "error getting name for requested relation %v", relation)
 	}
 
@@ -103,6 +98,21 @@ func ProjectAuthorizationInterceptor(ctx context.Context, req interface{}, info 
 	return handler(ctx, req)
 }
 
+// relationAsName returns the OpenFGA relation name for the given relation enum.
+// It returns an empty string in the case of error.
+func relationAsName(relation minder.Relation) string {
+	relationValue := relation.Descriptor().Values().ByNumber(relation.Number())
+	if relationValue == nil {
+		return ""
+	}
+	extension := proto.GetExtension(relationValue.Options(), minder.E_Name)
+	relationName, ok := extension.(string)
+	if !ok {
+		return ""
+	}
+	return relationName
+}
+
 // populateEntityContext populates the project in the entity context, by looking at the proto context or
 // fetching the default project
 func populateEntityContext(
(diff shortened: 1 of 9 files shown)


internal/controlplane/handlers_authz.go
```diff
@@ -80,13 +80,8 @@ func ProjectAuthorizationInterceptor(ctx context.Context, req interface{}, info
...
 
-	relationValue := relation.Descriptor().Values().ByNumber(relation.Number())
-	if relationValue == nil {
-		return nil, status.Errorf(codes.Internal, "error reading relation value %v", relation)
-	}
-	extension := proto.GetExtension(relationValue.Options(), minder.E_Name)
-	relationName, ok := extension.(string)
-	if !ok {
+	relationName := relationAsName(relation)
+	if relationName == "" {
```
JAORMX, line 84: This seems unrelated. Should it be in a different PR?

internal/controlplane/handlers_projects.go
```diff
@@ -154,83 +154,87 @@ func (s *Server) getImmediateChildrenProjects(ctx context.Context, projectID uui
...
+			return nil, status.Errorf(codes.Internal, "error committing transaction: %v", err)
+		}
 
-	if err := projects.ValidateName(req.Name); err != nil {
-		return nil, util.UserVisibleError(codes.InvalidArgument, "invalid project name: %v", err)
-	}
+	} else {
+		// This is a top-level project creation request.
+		// We need to check if the user has the right to create projects in the system.
+		if !flags.Bool(ctx, s.featureFlags, flags.ProjectCreateDelete) {
```
JAORMX, line 202: Do we need a feature flag for this?

JAORMX (approved):
minor comments. Looks good otherwise.

internal/controlplane/handlers_authz.go
evankanderson: I wanted to re-use this logic in `handlers_projects.go` to get the OpenFGA relation for a specific relation (`create`), so I extracted this bit of code.

internal/controlplane/handlers_authz.go
```diff
@@ -111,7 +121,7 @@ func populateEntityContext(
 	authzClient authz.Client,
 	req any,
 ) (context.Context, error) {
-	projectID, err := getProjectIDFromContext(req)
+	projectID, err := getProjectIDFromRequest(req)
```
evankanderson, line 124: This rename was probably too much -- I can roll it back if you want (`Context` being an overloaded name between the request field and the Golang concept).

internal/controlplane/handlers_projects.go
evankanderson: It feels like good practice, even if I'm going to turn it on shortly.

internal/controlplane/handlers_authz.go
JAORMX: gotcha

internal/controlplane/handlers_authz.go
JAORMX: nah, it's fine

JAORMX (approved):

PR #5702, opened 2025-06-23
Add templates for REST and YQ eval

evankanderson:
# Summary

Adds the ability to use Go templates in REST methods and YQ expressions.

The former can be used as a more flexible version of the branch protection remediation, to switch between POST and PUT or PATCH depending on whether a rule already exists.

The latter (in combination with rule eval output) will make it easier to fix cases like `dependabot.yml` with the list of missing ecosystems.

Also did a little cleanup of exported -> not-exported methods and types.

## Change Type

# Testing

Added unit tests, wrote some rules to use the new features.

# Review Checklist:

### DIFF
--- a/internal/engine/actions/remediate/pull_request/types_yq.go
+++ b/internal/engine/actions/remediate/pull_request/types_yq.go
@@ -13,10 +13,14 @@ import (
 	"github.com/go-git/go-git/v5/plumbing/filemode"
 	"github.com/mikefarah/yq/v4/pkg/yqlib"
 	"github.com/rs/zerolog"
+	"google.golang.org/protobuf/proto"
 	gologging "gopkg.in/op/go-logging.v1"
 
 	"github.com/mindersec/minder/internal/engine/interfaces"
+	textutil "github.com/mindersec/minder/internal/util"
 )
+
+// TODO: document the YQ remediation model.
 
 // unfortunately yqlib does seem to be using global variables...
 func init() {
@@ -82,7 +86,19 @@ func newYqExecute(
 	}, nil
 }
 
-func (yq *yqExecute) createFsModEntries(ctx context.Context, _ interfaces.ActionsParams) error {
+// templateParams are the parameters for template expansion for the expression template
+type templateParams struct {
+	// Entity is the metadata for the entity which was evaluated
+	Entity any
+	// Profile is the parameters to be used in the template
+	Profile map[string]any
+	// Params are the rule instance parameters
+	Params map[string]any
+	// EvalResultOutput is the output from the rule evaluation
+	EvalResultOutput any
+}
+
+func (yq *yqExecute) createFsModEntries(ctx context.Context, ent proto.Message, params interfaces.ActionsParams) error {
 	matchingFiles := make([]string, 0)
 	for _, pattern := range yq.config.Patterns {
 		if pattern.Type != string(patternTypeGlob) {
@@ -100,8 +116,30 @@ func (yq *yqExecute) createFsModEntries(ctx context.Context, _ interfaces.Action
 		matchingFiles = append(matchingFiles, patternMatches...)
 	}
 
+	templateData := templateParams{
+		Entity:  ent,
+		Profile: params.GetRule().Def,
+		Params:  params.GetRule().Params,
+	}
+	if params.GetEvalResult() != nil {
+		templateData.EvalResultOutput = params.GetEvalResult().Output
+	}
+
+	expression := ""
+	if yq.config.Expression != "" {
+		expressionBytes := new(bytes.Buffer)
+		expressionTmpl, err := textutil.NewSafeTextTemplate(&yq.config.Expression, "expression")
+		if err != nil {
+			return fmt.Errorf("unable to parse templates in expression: %w", err)
+		}
+		if err := expressionTmpl.Execute(ctx, expressionBytes, templateData, 3000); err != nil {
+			return fmt.Errorf("unable to expand templates in expression: %w", err)
+		}
+		expression = expressionBytes.String()
+	}
+
 	for _, file := range matchingFiles {
-		newContent, err := yq.executeYq(file, yq.config.Expression)
+		newContent, err := yq.executeYq(file, expression)
 		if err != nil {
 			return fmt.Errorf("cannot execute yq: %w", err)
 		}
--- a/internal/engine/actions/remediate/pull_request/types_yq_test.go
+++ b/internal/engine/actions/remediate/pull_request/types_yq_test.go
@@ -15,7 +15,10 @@ import (
 	"golang.org/x/exp/maps"
 	"google.golang.org/protobuf/types/known/structpb"
 
+	"github.com/mindersec/minder/internal/engine/interfaces"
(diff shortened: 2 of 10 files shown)


JAORMX (approved):

internal/engine/actions/remediate/rest/rest_test.go
```diff
@@ -124,6 +126,18 @@ func TestNewRestRemediate(t *testing.T) {
...
+			args: args{
+				restCfg: &pb.RestType{
+					// No {{end}}
+					Method:   "{{if .EvalResultOutput.doPatch}}patch{{else}}put",
+					Endpoint: "/graphql",
+					Body:     &simpleBodyTemplate,
+				},
+			},
+			wantErr: true,
+		},
```
Vyom-Yadav: The error from this test is: `action type cannot be empty`. However, I think it should be:

Screenshot from 2025-06-30 10-42-00

I'd suggest to change the wantErr to Err string here and do something like err Contains rather than just checking whether the method errored or  not to prevent such things in the future.

internal/engine/actions/remediate/rest/rest.go
```diff
@@ -29,17 +31,20 @@ const (
 	// RemediateType is the type of the REST remediation engine
 	RemediateType = "rest"
 
-	// EndpointBytesLimit is the maximum number of bytes for the endpoint
-	EndpointBytesLimit = 1024
+	// methodBytesLimit is the maximum number of bytes for the HTTP method
+	methodBytesLimit = 10
```
Vyom-Yadav, line 35: Is the max size 10? Isn't it 7 bytes?

```
string method = 2 [
            pattern: '^(?i)(GET|POST|PUT|PATCH|DELETE)$'
        },
        (buf.validate.field).ignore = IGNORE_IF_DEFAULT_VALUE
    ];
```

internal/engine/actions/remediate/pull_request/types_yq.go
```diff
@@ -100,8 +116,30 @@ func (yq *yqExecute) createFsModEntries(ctx context.Context, _ interfaces.Action
...
+
+	expression := ""
+	if yq.config.Expression != "" {
+		expressionBytes := new(bytes.Buffer)
+		expressionTmpl, err := textutil.NewSafeTextTemplate(&yq.config.Expression, "expression")
+		if err != nil {
+			return fmt.Errorf("unable to parse templates in expression: %w", err)
+		}
+		if err := expressionTmpl.Execute(ctx, expressionBytes, templateData, 3000); err != nil {
+			return fmt.Errorf("unable to expand templates in expression: %w", err)
```
Vyom-Yadav: nit: 3000 - magic number, how about extracting it to a variable.

internal/engine/actions/remediate/pull_request/types_yq_test.go
```diff
@@ -227,6 +232,101 @@ func TestYQExecute(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "TestTemplatedExpression",
+			yqParams: func(t *testing.T) *modificationConstructorParams {
+				t.Helper()
+
```
Vyom-Yadav, line 239: Might be stretching too much here, but how about another test that has a yq expression inside the go template so both conditions are hit? (If this is too much, feel free to ignore)

Vyom-Yadav (commented):
Overall lgtm, just some items:

internal/engine/actions/remediate/rest/rest_test.go
evankanderson: Thanks, I did so, and cleaned up some of the "stutter":

```
invalid method: cannot parse template: template: method:1: unexpected EOF
```

internal/engine/actions/remediate/rest/rest.go
evankanderson: Actually, it's currently 17 bytes if we use the types registered with IANA:

https://www.iana.org/assignments/http-methods/http-methods.xhtml

(From [section 8.1 of RFC7231](https://datatracker.ietf.org/doc/html/rfc7231#section-8.1))

But yes, I needed to update the the validation for `RestType` to allow for templates.  Example usage:

```
  remediate:
    type: rest
    rest:
      method: '{{if .EvalResultOutput.id}}PUT{{else}}POST{{end}}'
      endpoint: '/repos/{{.Entity.Owner}}/{{.Entity.Name}}/rulesets{{if .EvalResultOutput.id}}/{{.EvalResultOutput.id}}{{end}}'
```

Which allows you to update an existing rule if it exists, or create a new rule if not.  This is needed with Repository Rulesets because the name must be unique, but the rulesets are addressed in the API by their database ID.

internal/engine/actions/remediate/pull_request/types_yq.go
evankanderson: Done, though it's still pulled from a wild guess.

internal/engine/actions/remediate/pull_request/types_yq_test.go
evankanderson: I'm confused.  The Go template is generating the following YQ expression:

```
.updates += [
{"package-ecosystem": "gomod",
    "directory": "/go",
    "schedule": {"interval": "weekly"}},
{"package-ecosystem": "github-actions",
    "schedule": {"interval": "weekly"}},
]
```

The rule is then evaluating the YQ patch against the original `dependabot.yml`, which ends up adding the new entries (with the correct indentation and preserving comments!).

evankanderson:
(I was trying to get Coveralls to re-generate the coverage info, not delete it):

> 63 of 78 new or added lines in 7 files covered. (80.77%)
> 23 existing lines in 4 files now uncovered.
> 
> 18596 of 32396 relevant lines covered (57.4%)
> 
> 37.18 hits per line

internal/engine/actions/remediate/pull_request/types_yq_test.go
Vyom-Yadav: Missed that, thanks for pointing.

internal/engine/actions/remediate/rest/rest.go
Vyom-Yadav: I wasn't aware of those other types (good to know). 10 is ok, if those other methods are needed in the future, this can be increased.

proto/minder/v1/minder.proto
```diff
@@ -2358,9 +2358,11 @@ message RestType {
     ];
 
     // method is the method to use to fetch data.
+    // Go templates may be used to vary the method using the same parameters
+    // as the endpoint.
     string method = 2 [
         (buf.validate.field).string = {
-            pattern: '^(?i)(GET|POST|PUT|PATCH|DELETE)$'
+            max_len: 50
```
Vyom-Yadav, line 2365: Method templates can have `max_len = 50` is a weird enforcement, but it's fine for now given that most of the expressions would fit in this length.

evankanderson: That was the thought process, yes.  (And, writing a regex for a go template seemed like a bit of overkill -- instead, we attempt to parse the template to make sure it's valid.)

PR #5913, opened 2025-10-15
Phase 1 + 2: Remove legacy table writes and migrate reads

JAORMX:
Implements Phase 1 and Phase 2 of the [legacy table removal plan](https://gist.github.com/JAORMX/4c8e35cd24659edd3e83472958d692f7): stops writing to deprecated `repositories`, `artifacts`, and `pull_requests` tables AND migrates all read operations to the unified `entity_instances` and `properties` tables.

**Note**: Originally this PR only included Phase 1 (stop writes). After team review, Phase 2 (migrate reads) was added to ensure safe deployment without service disruption.

## Phase 1: Stop Writes to Legacy Tables
- Remove dual writes to legacy tables in repository/artifact creation and deletion
- All writes now go only to `entity_instances` and `properties` tables
- Update test mocks accordingly

### Repository Service
- Updated `GetRepositoryById` to query entity_instances + properties
- Updated `GetRepositoryByName` to search by name property using `GetTypedEntitiesByProperty`
- Updated `DeleteByName` to use property-based entity search
- Changed interface return types from `db.Repository` to `*pb.Repository`

### Artifact Handlers
- Updated `GetArtifactById` and `GetArtifactByName` to use entity_instances
- Updated `ListArtifacts` to query entity_instances directly
- Removed legacy helper functions that used repositories table

### Repository Handlers
- Simplified `GetRepositoryById` and `GetRepositoryByName` to use updated service
- Updated `ListRepositories` to convert EntityWithProperties to protobuf
- Updated reconciliation task handler to use `GetEntityByID`

### Reminder Service (Critical Background Process)
- Changed to work with `db.EntityInstance` instead of `db.Repository`
- Uses new cursor-based queries: `ListEntitiesAfterID`, `EntityExistsAfterID`
- Uses `ListOldestRuleEvaluationsByEntityID` for evaluation lookups
- Removed `UpdateReminderLastSentForRepositories` call (reminder throttling removed)

### Metrics
- Updated repository count metric to use `CountEntitiesByType`
- Updated project deletion telemetry to use `CountEntitiesByTypeAndProject`

### New SQL Queries
Added to `database/query/entities.sql` and `database/query/profile_status.sql`:
- `ListEntitiesAfterID` - cursor-based pagination for entities
- `EntityExistsAfterID` - check if more entities exist after cursor
- `CountEntitiesByType` - count all entities of a given type
- `CountEntitiesByTypeAndProject` - count entities for specific project
- `ListOldestRuleEvaluationsByEntityID` - entity-based evaluation lookup

## Testing
- All repository service tests passing
- All controlplane handler tests passing
- All reminder service tests passing
- All projects tests passing
- Full build successful
- Updated test mocks to reflect new query patterns

## Deployment Safety
✅ **No database migrations required** - All `entity_instances` and `properties` tables already exist
✅ **Pure code change** - Safe for normal pod rollout
✅ **Backward compatible** - Existing data already in new tables (from previous dual writes)
✅ **Zero downtime** - Reads and writes both work with new tables

## Why Both Phases Together?

Originally planned to deploy Phase 1 separately, but team review identified that stopping writes before migrating reads would cause newly created repositories to be invisible to read operations. By combining both phases:

1. **Before deployment**: Dual writes active, reads from legacy tables
2. **After deployment**: Single writes to new tables, reads from new tables
3. **Result**: Continuous service, no data loss, no downtime

Next steps: Deploy to staging, monitor, then proceed with Phase 3 (clean up evaluation tables) and Phase 4 (drop legacy tables).


internal/controlplane/handlers_artifacts.go
```diff
@@ -60,63 +60,77 @@ func (s *Server) GetArtifactByName(ctx context.Context, in *pb.GetArtifactByName
...
+	}
+
 	// the artifact name is the rest of the parts
 	artifactName := strings.Join(nameParts[2:], "/")
-	artifact, err := s.store.GetArtifactByName(ctx, db.GetArtifactByNameParams{
-		ArtifactName: artifactName,
-		ProjectID:    projectID,
+
+	// Search for artifact by name property
+	nameJSON, err := json.Marshal(artifactName)
```
evankanderson: Given that `artifactName` is a string, I'm not sure this query is going to work.  I think you need something like:

```golang
artifactQuery := struct {
  Value: string
}{ Value: artifactName }
```

And then you'll need to json.Marshal that object.

evankanderson: (or specifically encode looking up by name, or change the GetTypedEntitiesByProperty query)

evankanderson: (or use `GetTypedEntitiesByPropertyV1` from the db helpers)

internal/repositories/service.go
```diff
@@ -280,18 +309,68 @@ func (r *repositoryService) GetRepositoryByName(
...
+	}
+
+	// Search for repository by name property using GetTypedEntitiesByProperty
+	// Marshal the name value as JSON for the query
+	nameJSON, err := json.Marshal(fullName)
+	if err != nil {
+		return nil, fmt.Errorf("error marshaling name: %w", err)
+	}
+
+	entities, err := r.store.GetTypedEntitiesByProperty(ctx, db.GetTypedEntitiesByPropertyParams{
```
evankanderson: This also needs to use the "V1" version, or do more encoding of nameJSON.

internal/repositories/service.go
```diff
@@ -322,28 +401,52 @@ func (r *repositoryService) DeleteByName(
...
+		providerID = prov.ID
+	}
+
+	// Search for repository by name property
+	nameJSON, err := json.Marshal(fullName)
+	if err != nil {
+		return fmt.Errorf("error marshaling name: %w", err)
+	}
+
+	entities, err := r.store.GetTypedEntitiesByProperty(ctx, db.GetTypedEntitiesByPropertyParams{
```
evankanderson: I think this also needs to be the "V1" version of the method

evankanderson:
I may follow this up at some point by moving some of the calls to `GetTypedEntitiesByPropertyV1` to use `GetEntityByName` instead, but I don't want to hold up this PR.  Merging!

PR #5933, opened 2025-10-25
Use rule short_failure_message as fallback for evaluation failure message

### LINKED ISSUE #4718: Use rule short failure message as fallback for evaluation failure message
In some cases, for example when using the `rego/deny-by-default` evaluation engine, we expect the writer of the rule to set the evaluation failure message.
If they don't, we currently default to "denied".

Instead, we should default to the `short_failure_message` if that is provided in the rule definition.

JAORMX:
## Summary

In deny-by-default rego evaluations, when no custom message is provided by the rule writer, the evaluator now falls back to the rule's `short_failure_message` if available, instead of always defaulting to "denied".

This provides better error messages to users when rule evaluations fail, as they will see the descriptive `short_failure_message` from the rule definition rather than a generic "denied" message.

## Changes

- Added `shortFailureMessage` field to `denyByDefaultEvaluator` struct
- Modified `parseResult` to use `cmp.Or()` with `short_failure_message` as fallback before "denied"
- Created `WithShortFailureMessage()` Option function to pass the message to the evaluator
- Updated `NewRuleEvaluator` to pass `short_failure_message` option when creating rego evaluators

## Example

**Before**: Rule failures would show `"denied"`

**After**: Rule failures show the descriptive message from the rule type, e.g.:
- `"License file does not match the expected license type"`
- `"Malicious package found"`
- Falls back to `"denied"` only if no `short_failure_message` is defined

## Test Plan

- ✅ All existing tests pass
- ✅ Linter passes with auto-fixes applied
- ✅ Manual verification of fallback chain: custom message → short_failure_message → "denied"

## Related Issues

Fixes #4718

### DIFF
--- a/internal/entities/handlers/handler_test.go
+++ b/internal/entities/handlers/handler_test.go
@@ -575,6 +575,7 @@ func TestRefreshEntityAndDoHandler_HandleRefreshEntityAndEval(t *testing.T) {
 			mockStoreFunc: df.NewMockStore(
 				df.WithTransaction(),
 				df.WithSuccessfulUpsertPullRequestWithParams(
+					db.PullRequest{ID: pullRequestID},
 					db.EntityInstance{
 						ID:         pullRequestID,
 						EntityType: db.EntitiesPullRequest,


internal/engine/eval/eval.go
```diff
@@ -40,6 +40,10 @@ func NewRuleEvaluator(
 		}
 		return jq.NewJQEvaluator(e.GetJq(), opts...)
 	case rego.RegoEvalType:
+		// Add short_failure_message as an option if available
```
evankanderson, line 43: You can actually store this always, since you use `cmp.Or` to combine this and the "denied".

PR #5971, opened 2025-11-15
Remove: machine_accounts feature flag.

### LINKED ISSUE #5435: Remove machine_accounts feature flag
Remove the `machine_accounts` feature flag, either by releasing the feature or removing it.

Mujib-Ahasan:
# Summary
The default value of `machine_accounts` flag is false now. But related functionalities sets it to true. This PR is raised for the release of the flag.

Fixes #5435

### DIFF
--- a/internal/controlplane/handlers_authz_test.go
+++ b/internal/controlplane/handlers_authz_test.go
@@ -791,21 +791,12 @@ func TestAssignRole(t *testing.T) {
 				Provider: &keycloak.KeyCloak{},
 			},
 		}, {
-			name:    "machine accounts disabled",
-			subject: "githubactions/repo:mindersec/community:ref:refs/heads/main",
-			userIdentity: &auth.Identity{
-				UserID:   "repo:mindersec/community:ref:refs/heads/main",
-				Provider: &githubactions.GitHubActions{},
-			},
-		},
-		{
 			name:    "grant permission to GitHub Action",
 			subject: "githubactions/repo:mindersec/community:ref:refs/heads/main",
 			userIdentity: &auth.Identity{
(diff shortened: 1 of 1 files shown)


internal/controlplane/handlers_authz.go
```diff
@@ -398,7 +398,7 @@ func (s *Server) AssignRole(ctx context.Context, req *minder.AssignRoleRequest)
 		if !isMachine && flags.Bool(ctx, s.featureFlags, flags.UserManagement) {
 			return nil, util.UserVisibleError(codes.Unimplemented, "human users may only be added by invitation")
 		}
-		if isMachine && !flags.Bool(ctx, s.featureFlags, flags.MachineAccounts) {
+		if isMachine {
```
evankanderson: Ooh, this is tricky, since this is a negative use of the flag -- in this case, I think `flags.Bool(ctx, s.featureFlags, flags.MachineAccounts)` should effectively always return `true`, so `if isMachine && !true` will always evaluate to `false`, and the entire `if false {...}` block can be removed.

internal/controlplane/handlers_authz_test.go
```diff
@@ -797,7 +797,6 @@ func TestAssignRole(t *testing.T) {
 				UserID:   "repo:mindersec/community:ref:refs/heads/main",
 				Provider: &githubactions.GitHubActions{},
 			},
-			expectedError: "Description: Unimplemented\nDetails: machine accounts are not enabled",
 		},
```
evankanderson: You can probably remove this entire test case, and remove the `flagData` from the next test case on line 808.

evankanderson (approved):

PR #5970, opened 2025-11-13
Removed: user_management feature flag

Mujib-Ahasan:
# Summary
The default value of `user_management` flag is false now. But related functionalities sets it to true. This PR is raised for the release of the flag.
Fixes #5436

PR #6052, opened 2026-01-12
Replace mock service with simpler-to-configure fake

evankanderson:
# Summary

To simplify e.g. testing in #5970, let's use an in-memory fake (implemented by Claude) to replace the mock (which needs to be programmed in the test with behaviors matching the state-tracking in the real version).

This also includes replacing direct SQL queries in the controlplane with calls through the (extended) InviteService, and some refactoring of the InviteService interface to avoid duplicating List functionality which felt like it belonged in the controlplane.

# Testing

Added unit tests and did several manual tests:
* Create invite
* Attempt to resolve invite from same account (fails, correct error message)
* Resolve invite from another account which did not previously exist
* Attempt to resolve invite which does not exist
* List invites

### DIFF
--- a/internal/controlplane/handlers_user.go
+++ b/internal/controlplane/handlers_user.go
@@ -361,7 +361,7 @@ func (s *Server) ResolveInvitation(ctx context.Context, req *pb.ResolveInvitatio
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "failed to get or create user: %s", err)
 		}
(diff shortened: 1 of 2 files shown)


evankanderson:
This is still WIP, do not review yet.

evankanderson:
> Is this ready for review?

I've finished manually testing this, so yes, I think it's ready for review

PR #6090, opened 2026-02-05
feat: adds scorecard workflow to publish openssf scores

### LINKED ISSUE #6089: Add OpenSSF Scorecard GitHub Actions workflow to publish security scores

The repository currently does not publish OpenSSF Scorecard results to the public Scorecard API (https://api.securityscorecards.dev/). As a result, users and downstream projects cannot easily discover or track the project’s security best-practice posture in a standardized, automated way. The absence of published results reduces visibility and makes it harder for consumers to assess the project using common tooling.

### Solution Proposal

Add the official OpenSSF Scorecard GitHub Actions workflow to the repository so that scorecard runs automatically on a schedule and on the default branch and the results are published to the OpenSSF Scorecard API.

_No response_

### Additional context

_No response_

### Acceptance Criteria

_No response_

gaganhr94:
# Summary

The repository currently does not publish OpenSSF Scorecard results, which means users and downstream consumers cannot easily discover the project’s security best-practice signals via Scorecard API.

Publishing Scorecard results:
- Improves transparency for users and integrators
- Provides early visibility into missing or improvable security practices

Fixes #6089


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,42 @@
+name: Scorecard supply-chain security
+on:
+  push:
+    branches: [ main ]
+  schedule:
+    - cron: '0 0 * * 1'
+
+permissions:
+    contents: read
```
evankanderson: github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml uses `read-all` instead of `contents: read`.  Will `contents: read` prevent any of scorecard's operation?

gaganhr94: Thanks for pointing this out! I looked into it a bit more, and it seems that using contents: read can cause some checks to be skipped or return incomplete results, since scorecard is a complete analysis of the project. I'll keep it aligned with the scorecard workflow by changing it to `permissions: read-all`

PR #6106, opened 2026-02-12
Remove user_management feature flag

### LINKED ISSUE #5436: Remove user_management feature flag
Remove the `user_management` feature flag, either by releasing the feature or deleting it.

navnitms:
Remove the user_management feature flag and make its behavior always enabled.

# Summary

Fix this issue https://github.com/mindersec/minder/issues/5436

Fixes #5436 

# Testing

Tested using unit tests

Also referred the conversation in this [PR](https://github.com/mindersec/minder/pull/5970)

### DIFF
--- a/internal/controlplane/handlers_authz.go
+++ b/internal/controlplane/handlers_authz.go
@@ -15,14 +15,12 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
-	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/mindersec/minder/internal/auth"
 	"github.com/mindersec/minder/internal/auth/jwt"
 	"github.com/mindersec/minder/internal/authz"
 	"github.com/mindersec/minder/internal/db"
 	"github.com/mindersec/minder/internal/engine/engcontext"
-	"github.com/mindersec/minder/internal/invites"
 	"github.com/mindersec/minder/internal/util"
 	minder "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
 )
@@ -310,23 +308,14 @@ func (s *Server) ListRoleAssignments(
 	}
 
 	// Add invitations, which are only stored in the Minder DB
-	projectInvites, err := s.store.ListInvitationsForProject(ctx, targetProject)
+	projectInvites, err := s.invites.ListInvitationsForProject(ctx, s.store, targetProject)
 	if err != nil {
 		// return the information we can and log the error
 		zerolog.Ctx(ctx).Error().Err(err).Msg("error getting invitations")
 	}
-	for _, i := range projectInvites {
-		invitations = append(invitations, &minder.Invitation{
-			Role:           i.Role,
-			Email:          i.Email,
-			Project:        targetProject.String(),
-			CreatedAt:      timestamppb.New(i.CreatedAt),
-			ExpiresAt:      invites.GetExpireIn7Days(i.UpdatedAt),
-			Expired:        invites.IsExpired(i.UpdatedAt),
-			Sponsor:        i.IdentitySubject,
-			SponsorDisplay: mapIdToDisplay[i.IdentitySubject],
-			// Code is explicitly not returned here
-		})
+	for _, inv := range projectInvites {
+		inv.SponsorDisplay = mapIdToDisplay[inv.Sponsor]
+		invitations = append(invitations, inv)
 	}
 
 	return &minder.ListRoleAssignmentsResponse{
--- a/internal/controlplane/handlers_authz_test.go
+++ b/internal/controlplane/handlers_authz_test.go
@@ -398,14 +398,23 @@ func TestRoleManagement(t *testing.T) {
 	user1 := uuid.New()
 	user2 := uuid.New()
 
+	// Non-zero fixed time for deterministic invitation timestamps.
+	// Using a time in the past so invitations show as expired in tests.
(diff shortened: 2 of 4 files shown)


internal/controlplane/handlers_authz.go
```diff
@@ -310,26 +309,24 @@ func (s *Server) ListRoleAssignments(
...
-				CreatedAt:      timestamppb.New(i.CreatedAt),
-				ExpiresAt:      invites.GetExpireIn7Days(i.UpdatedAt),
-				Expired:        invites.IsExpired(i.UpdatedAt),
-				Sponsor:        i.IdentitySubject,
-				SponsorDisplay: mapIdToDisplay[i.IdentitySubject],
-				// Code is explicitly not returned here
-			})
-		}
+	// Add invitations, which are only stored in the Minder DB
+	projectInvites, err := s.store.ListInvitationsForProject(ctx, targetProject)
```
evankanderson: Sorry to have missed this before when I was refactoring!

This should go through a (new) method on [InviteService](https://github.com/mindersec/minder/blob/main/internal/invites/service.go#L33), which will also mean updating the fake.  The direct access to database tables from the service logic is not a great pattern which we still have a in a few places. The database accesses are generally complex enough that programming a mock for them doesn't actually emulate real behavior for testing -- as I'll point out in the tests.  😁

internal/controlplane/handlers_authz_test.go
```diff
@@ -401,102 +399,44 @@ func TestRoleManagement(t *testing.T) {
...
 		},
-		stored: []*minder.RoleAssignment{{
-			Role:    authz.RoleAdmin.String(),
-			Subject: user1.String(),
-			Project: proto.String(project.String()),
-		}, {
-			Role:    authz.RoleAdmin.String(),
-			Subject: user2.String(),
-			Project: proto.String(project.String()),
-		}},
```
evankanderson, line 442: Using a mock for the `ListInvitationsForProject` produces this odd result -- we add two users, send the invitations, and then when we list all the outstanding roles and invitations on the project, there's nothing.  What _should_ be happening is that we should see the two outstanding invites (and test the ListInvitation code for that transformation).  The **old** way of doing this was to inject database rows that _hopefully_ match what would actually be stored; the **new** way is to route these calls through the (Fake)InviteService, and have it actually track state.  That would also allow removing the `invites` fields on the test case.

internal/controlplane/handlers_authz_test.go
```diff
@@ -401,102 +399,44 @@ func TestRoleManagement(t *testing.T) {
...
-		result: &minder.ListRoleAssignmentsResponse{
-			RoleAssignments: []*minder.RoleAssignment{{
-				Role:        authz.RoleAdmin.String(),
-				DisplayName: "user1",
-				Subject:     user1.String(),
-				Project:     proto.String(project.String()),
-			}},
-		},
-	}, {
-		name:    "IDP resolution",
```
evankanderson, line 465: Unfortunately, it looks like we've lost coverage of the identity provider name resolution: https://coveralls.io/builds/77786036/source?filename=internal%2Fcontrolplane%2Fhandlers_authz.go#L294

Two ways to fix this would be:
1. Add a machine account, which should be a direct add
2. Extend the test to be able to pre-load a user into the `authzClient` before the rest of the test execution.

evankanderson (commented):
Thanks for doing this!  Your changes make sense, and I'm willing to merge them as-is if the suggestions seem to intensive; on the other hand, I think using a fake which would echo back the correct invitation state would make the tests more robust.

evankanderson:
I'm mostly-out this week, but will try to take a look if you push a commit or have questions about the interface for the invite service.

internal/controlplane/handlers_authz.go
navnitms: Sorry for the late reply and resolutions, Was caught up in some work,

Added ListInvitationsForProject to the InviteService, Remove the direct store call

internal/controlplane/handlers_authz_test.go
navnitms: Removed the invites field from the test struct,

internal/controlplane/handlers_authz_test.go
navnitms: Added a "machine account with invitations" test case that loads a  identity into the resolver and a role assignment into the authz client

internal/controlplane/handlers_authz.go
```diff
@@ -310,26 +307,15 @@ func (s *Server) ListRoleAssignments(
...
-			})
-		}
+	// Add invitations, which are only stored in the Minder DB
+	projectInvites, err := s.invites.ListInvitationsForProject(ctx, s.store, targetProject)
+	if err != nil {
+		// return the information we can and log the error
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error getting invitations")
+	}
+	for _, inv := range projectInvites {
+		inv.SponsorDisplay = mapIdToDisplay[inv.Sponsor]
```
evankanderson, line 317: There's some subtle but probably correct behavior here.  Just commenting for your edification.

```suggestion
		// Invitations are created (sponsored) by users who have permissions on the project.
		// We already resolved them in the "assignments" section, so we can re-use that work here.
		// If a user has _left_ the project after sending an invitation, we won't have that mapping,
		// we'll have an empty string (default value) for display name.  This is *correct*, because
		// we don't want to leak their identity after they have left the project.
		inv.SponsorDisplay = mapIdToDisplay[inv.Sponsor]
```

navnitms:
Thanks @evankanderson

PR #6107, opened 2026-02-13
Add device authorization flow and request debugging logs

evankanderson:
# Summary

While debugging an issue (which seemed to be firewall related), the logs provided misleading information about whether _any_ outbound requests were working (they were not, and the CLI was falling back to the built-in defaults).  Adding the log message reporting the swallowed error in `GetRealmUrl` the necessitated plumbing the `cobra.Command` through several more layers of login to enable `cmd.Printf()` rather than `fmt.Printf()`.

Additionally, while attempting to work around the firewall issues with `--skip-browser` from a cloud shell, I discovered that `--skip-browser` still uses the my-browser-can-reach-minder-cli's-ports assumption with the authentication code flow.  The solution to this "headless login" problem is either port-forwarding or to use the OAuth "Device Authorization Flow", used for devices which can't necessarily run a full browser on their own.  The addition of `loginWithDeviceFlow` enables the case where the browser and the minder CLI can both reach the auth server (and minder API), but can't reach each other, by giving the user a 8-12 character time-limited code they can use to authorize the CLI from some other device.

Testing the fix required enabling the device flow on the `minder-cli` client, and uncovered that the bitnami `keycloak-config-cli` images had been removed -- if you hadn't already pulled them locally, `make run-docker` would fail.  This covers the `stacklok.yaml` and `docker-compose.yaml` changes.

During testing with the device flow, it became clear that `minder auth login`'s use of `cli.GRPCClientWrapRunE` as well as `cli.LoginAndSaveCreds` was causing the login flow to be run twice (unnecessarily).  Since `minder auth login` manages the login flow explicitly, I removed that call.

# Testing

Tested manually against both the `custcodian.dev` and a local dev instance.

### DIFF
--- a/cmd/cli/app/quickstart/quickstart.go
+++ b/cmd/cli/app/quickstart/quickstart.go
@@ -142,7 +142,7 @@ var cmd = &cobra.Command{
 //
 //nolint:gocyclo
 func quickstartCommand(
-	ctx context.Context,
+	_ context.Context,
 	cmd *cobra.Command,
 	_ []string,
 	conn *grpc.ClientConn,
@@ -395,7 +395,7 @@ func getQuickstartContext(ctx context.Context, v *viper.Viper) (context.Context,
 
 func loginPromptErrWrapper(
 	cmnd *cobra.Command,
-	conn *grpc.ClientConn,
+	_ *grpc.ClientConn,
 	inErr error,
 ) error {
 	// Check if the error is unauthenticated, if so, prompt the user to log in
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -146,7 +146,6 @@ services:
   keycloak-config:
     container_name: keycloak_config
     image: quay.io/adorsys/keycloak-config-cli:6.4.1-26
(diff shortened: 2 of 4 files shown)


internal/util/cli/cli.go
```diff
@@ -83,14 +83,14 @@ func GRPCClientWrapRunE(
...
 
-		ctx, cancel := GetAppContext(cmd.Context(), viper.GetViper())
-		defer cancel()
-
 		c, err := GrpcForCommand(cmd, viper.GetViper())
 		if err != nil {
 			return err
 		}
 
+		ctx, cancel := GetAppContext(cmd.Context(), viper.GetViper())
```
evankanderson, line 91: Moving this call here moves the potential login time (possibly several minutes) _outside_ the request timeout set by GetAppContext (default 20s), which seemed generally useful.

evankanderson: (We probably also should be loading this context back into `cmd.Context`, but I don't want to disturb more items right now.)

evankanderson:
> This PR seems to do quite a lot of changes even if it's small. Do you wanna update the description or split it?

I updated the description a bit more.  I could split it, but given that this is all manually tested, I'd prefer not to introduce intermediate states that I'd want to test again.

PR #6203, opened 2026-03-26
Add output-only annotations to provider proto fields

XiaoPengMei:
## Summary
- add `OUTPUT_ONLY` field behavior annotations to output-only provider fields in `minder.proto`
- regenerate protobuf and OpenAPI outputs so generated clients mark these fields as read-only

## Testing
- `make lint-buf`
- `make buf`
- verified generated `pkg/api/openapi/minder/v1/minder.swagger.json` now contains `readOnly: true` for `organization`, `organizationId`, and `credentialsState`


evankanderson (changes requested):
Marking as "request changes" (needs `make gen` re-run), so I can more easily track which PRs need maintainer attention.

evankanderson:
I think #6221 covered these fields, closing.

PR #6207, opened 2026-03-26
Add with_output_detail to EvalResultsService

### LINKED ISSUE #6193: Inventory: API changes
This is part of #6105.

Extend the existing `EvalResultsService` endpoints to support fetching evaluation output, in addition to current pass/fail and message content.  This should cover the `ListEvaluationResults`, `ListEvaluationHistory`, and `GetEvaluationHistory` endpoints.  Because the evaluation output may be large JSON-structured data, this should only be returned when requested; we should add a boolean such as `with_output_detail` (name may need wordsmithing) to request this output.  We do *not* need to extend the existing legacy `GetProfileStatus*` endpoints; we should prefer the new API.

krrish175-byte:
# Summary

This PR extends the [EvalResultsService](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/proto/minder/v1/minder.proto:787:0-823:1) API to optionally provide evaluation output details, addressing issue #6193.

It adds a `with_output_detail` boolean flag to the [ListEvaluationResultsRequest](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/proto/minder/v1/minder.proto:2275:0-2324:1), [ListEvaluationHistoryRequest](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/proto/minder/v1/minder.proto:3806:0-3903:1), and [GetEvaluationHistoryRequest](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/proto/minder/v1/minder.proto:3787:0-3799:1) protobuf messages. It also introduces an `output` string to the [EvaluationHistoryStatus](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/proto/minder/v1/minder.proto:4005:0-4023:1) and [RuleEvaluationStatus](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/proto/minder/v1/minder.proto:1832:0-1888:1) response messages. 

As per discussion in the tracking issue, the backend table storage (the JSONB output checkpoint column) is not yet fully available for use. Therefore, this PR focuses solely on establishing the API contract. The control plane handler logic is stubbed out for now and ignores the flag to ensure no breaking changes or errors occur prior to the backend implementation. Standard code generation via `buf generate` has been run to update Go bindings, OpenAPI specs, and docs.

Fixes #6193

# Testing

Changes were validated primarily through unit testing and compilation checks, as this is an API contract update with stubbed handlers.

Steps taken to verify:
1. Ran `buf generate` to confirm the protobuf definitions compile successfully and generate valid artifacts.
2. Formatted and linted code locally.
3. Ran `go test ./internal/controlplane/... -v` to ensure the generated handler changes and updated function signatures did not break any existing test coverage for the `EvalResultsService`.

cc - @evankanderson

### DIFF
--- a/proto/minder/v1/minder.proto
+++ b/proto/minder/v1/minder.proto
@@ -1881,11 +1881,11 @@ message RuleEvaluationStatus {
     RuleTypeReleasePhase release_phase = 20 [
         (google.api.field_behavior) = REQUIRED
     ];
-    // output contains the evaluation output, if requested.
-    // This is a JSON-structured string containing the evaluation
-    // output data and is only populated when with_output_detail
-    // is set to true in the request.
-    string output = 21;
+    // output optionally contains the structured rule evaluation output.
+    // Because output may be multiple KB, it is only returned
+    // if include_outputs is set. Historical evaluations may
+    // discard structured output sooner than status results.
+    google.protobuf.Struct output = 21;
 }
 
 // EntityTypedId is a message that carries an ID together with a type to uniquely identify an entity
@@ -2318,10 +2318,11 @@ message ListEvaluationResultsRequest {
         }
     ];
 
-    // If true, include the full evaluation output in the response.
-    // Because the evaluation output may be large JSON-structured
-    // data, it is only returned when this flag is set.
-    bool with_output_detail = 6;
+    // If true, include structured rule output for the matched evaluations.
+    // Not all ruletypes may generate structured outputs.
+    // Because the evaluation output may be large, it is only returned
+    // when explicitly requested.
+    bool include_outputs = 6;
 }
 
 message ListEvaluationResultsResponse {
@@ -3793,10 +3794,11 @@ message GetEvaluationHistoryRequest {
 
     Context context = 2;
 
-    // If true, include the full evaluation output in the response.
-    // Because the evaluation output may be large JSON-structured
-    // data, it is only returned when this flag is set.
-    bool with_output_detail = 3;
+    // If true, include structured rule output for the matched evaluations.
+    // Not all ruletypes may generate structured outputs.
+    // Because the evaluation output may be large, it is only returned
+    // when explicitly requested.
+    bool include_outputs = 3;
 }
 
 // ListEvaluationHistoryRequest represents a request message for the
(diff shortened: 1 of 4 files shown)


proto/minder/v1/minder.proto
```diff
@@ -1881,6 +1881,11 @@ message RuleEvaluationStatus {
     RuleTypeReleasePhase release_phase = 20 [
         (google.api.field_behavior) = REQUIRED
     ];
+    // output contains the evaluation output, if requested.
+    // This is a JSON-structured string containing the evaluation
+    // output data and is only populated when with_output_detail
+    // is set to true in the request.
+    string output = 21;
```
evankanderson: Let's use `google.protobuf.Struct` for these -- it represents JSON data in a structured format, rather than adding the requirement for JSON deserialization for proto clients.

proto/minder/v1/minder.proto
```diff
@@ -2312,6 +2317,11 @@ message ListEvaluationResultsRequest {
             }
         }
     ];
+
+    // If true, include the full evaluation output in the response.
+    // Because the evaluation output may be large JSON-structured
+    // data, it is only returned when this flag is set.
+    bool with_output_detail = 6;
```
evankanderson: While I may have suggested this name in the bug, let's prefer shorter names.  What about `with_output` or `include_outputs`?

I'd also be slightly more precise in the comment:

```suggestion
    // If true, include structured rule output for the matched evaluations.
    // Not all ruletypes may generated structured outputs.
    // Because the evaluation output may be large, it is only returned
    // when explicitly requested.
    bool include_outputs = 6;
```

proto/minder/v1/minder.proto
```diff
@@ -3995,6 +4015,12 @@ message EvaluationHistoryStatus {
     string details = 2 [
         (google.api.field_behavior) = REQUIRED
     ];
+
+    // output contains the evaluation output, if requested.
+    // This is a JSON-structured string containing the evaluation
+    // output data and is only populated when with_output_detail
+    // is set to true in the request.
+    string output = 3;
```
evankanderson: This should also use `google.protobuf.Struct`.  We may also choose to discard historical output results sooner (e.g. store fewer historical records) than for general evaluation status, so it's worth mentioning that this may be "sparse" for certain queries.


```suggestion

    // output optionally contains the structured rule evaluation output.
    // Because output may be multiple KB, it is only returned
    // if include_outputs is set.  Historical evaluations may
    // discard structured output sooner than status results.
    google.protobuf.Struct output = 3;
```

PR #6208, opened 2026-03-29
Test(actions): add unit tests for rule actions

dakshhhhh16:
## Summary
This PR improves the test coverage of the rule engine actions logic by introducing comprehensive unit tests for [internal/engine/actions/actions.go](cci:7://file:///Users/dakshpathak/Desktop/Minder/minder/internal/engine/actions/actions.go:0:0-0:0). The goal is to ensure the reliability of [shouldRemediate](cci:1://file:///Users/dakshpathak/Desktop/Minder/minder/internal/engine/actions/actions.go:137:0-179:1), [shouldAlert](cci:1://file:///Users/dakshpathak/Desktop/Minder/minder/internal/engine/actions/actions.go:181:0-237:1), and metadata extraction functions. We test across all evaluation states (skipped, error, success, failure) to prevent accidental regressions in auto-remediation and alert firing behaviors. There are no new dependencies required.
## Testing
- Ran `go test ./internal/engine/actions -v`. 
- Verified all fallback paths, edge cases involving previously skipped actions, and PR remediation specific cases trigger the [On](cci:2://file:///Users/dakshpathak/Desktop/Minder/minder/internal/engine/eval/vulncheck/review.go:603:0-603:34), `Off`, and `DoNothing` commands securely.

### DIFF
--- a/internal/engine/actions/actions_test.go
+++ b/internal/engine/actions/actions_test.go
@@ -1,0 +1,220 @@
+// SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package actions
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/mindersec/minder/internal/db"
+	"github.com/mindersec/minder/internal/engine/actions/remediate/pull_request"
+	enginerr "github.com/mindersec/minder/internal/engine/errors"
+	engif "github.com/mindersec/minder/internal/engine/interfaces"
+)
+
+func TestShouldRemediate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		prevEval db.RemediationStatusTypes
+		nilRow   bool
+		evalErr  error
+		expected engif.ActionCmd
+	}{
+		// Happy path: eval success
+		{
+			name:     "eval success, prev success -> off",
+			prevEval: db.RemediationStatusTypesSuccess,
+			evalErr:  nil,
+			expected: engif.ActionCmdOff,
+		},
+		{
+			name:     "eval success, prev skipped -> do nothing",
+			prevEval: db.RemediationStatusTypesSkipped,
+			evalErr:  nil,
+			expected: engif.ActionCmdDoNothing,
+		},
+		// Happy path: eval failure triggers remediation
+		{
+			name:     "eval failure, prev skipped -> on",
+			prevEval: db.RemediationStatusTypesSkipped,
+			evalErr:  enginerr.NewErrEvaluationFailed("failed"),
+			expected: engif.ActionCmdOn,
+		},
+		{
+			name:     "eval failure, prev success -> do nothing",
+			prevEval: db.RemediationStatusTypesSuccess,
+			evalErr:  enginerr.NewErrEvaluationFailed("failed"),
+			expected: engif.ActionCmdDoNothing,
+		},
+		// Expected errors: eval error cases
+		{
+			name:     "eval error, prev skipped -> do nothing",
+			prevEval: db.RemediationStatusTypesSkipped,
+			evalErr:  errors.New("random error"),
+			expected: engif.ActionCmdDoNothing,
+		},
+		{
+			// NOTE: EvalStatusTypesError has an empty case body in shouldRemediate,
+			// so eval errors fall through to the default DoNothing. This may be a bug
+			// (see the comment on cases Error/Success in shouldRemediate).
+			name:     "eval error, prev success -> do nothing",
+			prevEval: db.RemediationStatusTypesSuccess,
+			evalErr:  errors.New("random error"),
+			expected: engif.ActionCmdDoNothing,
+		},
+		{
+			name:     "eval failure, prev error -> do nothing",
+			prevEval: db.RemediationStatusTypesError,
+			evalErr:  enginerr.NewErrEvaluationFailed("failed"),
+			expected: engif.ActionCmdDoNothing,
+		},
+		{
+			name:     "eval error, prev error -> do nothing",
+			prevEval: db.RemediationStatusTypesError,
+			evalErr:  errors.New("random error"),
+			expected: engif.ActionCmdDoNothing,
+		},
+		// Edge cases
+		{
+			name:     "eval skipped -> do nothing",
+			prevEval: db.RemediationStatusTypesSkipped,
+			evalErr:  enginerr.ErrEvaluationSkipSilently,
+			expected: engif.ActionCmdDoNothing,
+		},
+		{
+			name:     "nil DB row, eval failure -> on",
+			nilRow:   true,
+			evalErr:  enginerr.NewErrEvaluationFailed("failed"),
+			expected: engif.ActionCmdOn,
+		},
+		{
+			name:     "nil DB row, eval success -> do nothing",
+			nilRow:   true,
+			evalErr:  nil,
+			expected: engif.ActionCmdDoNothing,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			var prevRow *db.ListRuleEvaluationsByProfileIdRow
+			if !tt.nilRow {
+				prevRow = &db.ListRuleEvaluationsByProfileIdRow{RemStatus: tt.prevEval}
+			}
+			got := shouldRemediate(prevRow, tt.evalErr)
+			assert.Equal(t, tt.expected, got)
+		})
+	}
+}
+
+func TestShouldAlert(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name      string
+		prevAlert db.AlertStatusTypes
+		nilRow    bool
+		evalErr   error
+		remErr    error
+		remType   string
+		expected  engif.ActionCmd
+	}{
+		// Happy path: eval success
+		{
+			name:      "eval success, alert on -> off",
+			prevAlert: db.AlertStatusTypesOn,
+			evalErr:   nil,
+			remErr:    nil,
+			// Using pull_request.RemediateType so we reach the switch instead of
+			// hitting the early-return in Case 1 (non-PR successful remediation).
+			remType:  pull_request.RemediateType,
+			expected: engif.ActionCmdOff,
+		},
+		{
+			name:      "eval success, alert already off -> do nothing",
+			prevAlert: db.AlertStatusTypesOff,
+			evalErr:   nil,
+			remErr:    nil,
+			remType:   pull_request.RemediateType,
+			expected:  engif.ActionCmdDoNothing,
+		},
+		// Happy path: eval failure triggers alert
+		{
+			name:      "pr remediation eval failure, alert skipped -> on",
+			prevAlert: db.AlertStatusTypesSkipped,
+			evalErr:   enginerr.NewErrEvaluationFailed("failed"),
+			remErr:    nil,
+			remType:   pull_request.RemediateType,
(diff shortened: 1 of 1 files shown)


internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,195 @@
...
+	assert.Empty(t, remMeta)
+
+	var alertMeta map[string]any
+	require.NoError(t, json.Unmarshal(res.AlertMeta, &alertMeta))
+	assert.Empty(t, alertMeta)
+}
+
+func TestGetRemediationMeta(t *testing.T) {
+	t.Parallel()
+	t.Run("nil profile", func(t *testing.T) {
```
evankanderson: If you aren't using a table test (data-driven test), I don't think there's a need to use `t.Run()` here (which will avoid the warnings on `t.Parallel` usage).

evankanderson: (here and below)

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	tests := []struct {
+		name           string
+		prevEvalFromDb *db.ListRuleEvaluationsByProfileIdRow
+		evalErr        error
+		expected       engif.ActionCmd
+	}{
+		{
+			name:           "eval success, prev skipped -> off",
+			prevEvalFromDb: &db.ListRuleEvaluationsByProfileIdRow{RemStatus: db.RemediationStatusTypesSuccess},
```
evankanderson: Given that you're only ever setting `RemStatus`, this would read more simply if you made the table data `RemediationStatusTypes`, and then created the row in the test.

```golang
			t.Parallel()
			prevRow := db.ListRuleEvaluationsByProfileIdRow{RemStatus: tt.prevEval}
			got := shouldRemediate(&prevRow, tt.evalErr)
			assert.Equal(t, tt.expected, got)
```

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,195 @@
...
+	assert.ErrorIs(t, res.AlertErr, enginerr.ErrActionSkipped)
+
+	var remMeta map[string]any
+	require.NoError(t, json.Unmarshal(res.RemediateMeta, &remMeta))
+	assert.Empty(t, remMeta)
+
+	var alertMeta map[string]any
+	require.NoError(t, json.Unmarshal(res.AlertMeta, &alertMeta))
+	assert.Empty(t, alertMeta)
+}
```
evankanderson: FWIW, `getDefaultResult` is already tested transitively, which seems sufficient for an internal implementation detail function with no real way to return error.  (In fact, we should probably just use a static string rather than `json.Marshal` in that case.)

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+func TestGetRemediationMeta(t *testing.T) {
```
evankanderson: IMO, `getRemediationMeta` isn't really helpful to test on its own, as it's basically a nil-check helper.  Getting coverage for the `!skipRemediate` and `!skipAlert` lines 96-110 in `actions.go` would be much more meaningful (and would incidentally cover these lines).

internal/engine/actions/actions_test.go
evankanderson: Alternatively, you might want to test the case of a nil DB row (e.g. no previous evaluation result).

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,195 @@
...
+			name:           "eval success, prev skipped -> do nothing",
+			prevEvalFromDb: &db.ListRuleEvaluationsByProfileIdRow{RemStatus: db.RemediationStatusTypesSkipped},
+			evalErr:        nil,
+			expected:       engif.ActionCmdDoNothing,
+		},
+		{
+			name:           "eval failure, prev skipped -> on",
+			prevEvalFromDb: &db.ListRuleEvaluationsByProfileIdRow{RemStatus: db.RemediationStatusTypesSkipped},
+			evalErr:        enginerr.NewErrEvaluationFailed("failed"),
+			expected:       engif.ActionCmdOn,
```
evankanderson: Can you add a test for `eval failure, prev error` and `eval err, prev err`?  I feel like there might be something wrong in those combinations.

evankanderson: (As in, we have an actual logic bug that testing would have revealed...)

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,195 @@
...
+		name           string
+		prevEvalFromDb *db.ListRuleEvaluationsByProfileIdRow
+		evalErr        error
+		remErr         error
+		remType        string
+		expected       engif.ActionCmd
+	}{
+		{
+			name:           "successful non-pr remediation with alert not off",
+			prevEvalFromDb: &db.ListRuleEvaluationsByProfileIdRow{AlertStatus: db.AlertStatusTypesOn},
```
evankanderson: Ditto on this type here.

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,195 @@
...
+			remErr:         nil,
+			remType:        pull_request.RemediateType,
+			expected:       engif.ActionCmdDoNothing,
+		},
+		{
+			name:           "eval success and alert on -> off",
+			prevEvalFromDb: &db.ListRuleEvaluationsByProfileIdRow{AlertStatus: db.AlertStatusTypesOn},
+			evalErr:        nil,
+			remErr:         nil,
+			remType:        pull_request.RemediateType,
```
evankanderson: Does the remediation type here matter?

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,195 @@
...
+			expected:       engif.ActionCmdOff,
+		},
+		{
+			name:           "eval success and alert already off -> do nothing",
+			prevEvalFromDb: &db.ListRuleEvaluationsByProfileIdRow{AlertStatus: db.AlertStatusTypesOff},
+			evalErr:        nil,
+			remErr:         nil,
+			remType:        pull_request.RemediateType,
```
evankanderson: Ditto

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	tests := []struct {
+		name           string
+		prevEvalFromDb *db.ListRuleEvaluationsByProfileIdRow
+		evalErr        error
+		remErr         error
+		remType        string
+		expected       engif.ActionCmd
+	}{
+		{
```
evankanderson, line 148: Slight preference for ordering things as follows:

1. "Success" / "happy path" cases - start with "rule passed", then "rule eval detected a failure"
2. "Expected errors" - e.g. "rule eval errored" (like a bad rule definition)
3. "Unexpected errors" / "edge cases" - e.g. `remErr` is non-nil.

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,220 @@
...
+			name:      "eval error -> do nothing",
+			prevAlert: db.AlertStatusTypesOff,
+			evalErr:   errors.New("generic error"),
+			remErr:    nil,
+			remType:   pull_request.RemediateType,
+			expected:  engif.ActionCmdDoNothing,
+		},
+		// Edge cases
+		{
+			name:      "nil DB row, eval failure -> on",
```
evankanderson: Lint says spacing here is off by one, sorry!

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,220 @@
...
+			name:     "eval failure, prev skipped -> on",
+			prevEval: db.RemediationStatusTypesSkipped,
+			evalErr:  enginerr.NewErrEvaluationFailed("failed"),
+			expected: engif.ActionCmdOn,
+		},
+		{
+			name:     "eval failure, prev success -> do nothing",
+			prevEval: db.RemediationStatusTypesSuccess,
+			evalErr:  enginerr.NewErrEvaluationFailed("failed"),
+			expected: engif.ActionCmdDoNothing,
```
evankanderson, line 52: This seems weird... shouldn't we be saying to turn on the remediation at this point?

As an aside, we're basically doing some state tracking here to figure out how to handle the remediation given previous status.  For example, if we have a PR type remediation:

* start failing -> create a PR
* still failing -> already have a PR, don't need to act
* start passing, was failing -> close the PR
* keep passing -> close the PR if one was open

I'm not sure whether we should consider changing this to be level set ("there should be a PR open") or keep this as-is, but that's a larger change.

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,220 @@
...
+		{
+			name:     "eval error, prev skipped -> do nothing",
+			prevEval: db.RemediationStatusTypesSkipped,
+			evalErr:  errors.New("random error"),
+			expected: engif.ActionCmdDoNothing,
+		},
+		{
+			// NOTE: EvalStatusTypesError has an empty case body in shouldRemediate,
+			// so eval errors fall through to the default DoNothing. This may be a bug
+			// (see the comment on cases Error/Success in shouldRemediate).
```
evankanderson, line 64: I think this is *do nothing*, because the evaluation just shrugged.  I think "eval error, prev skipped" is wrong -- can we change that to `ActionCmdOn`?

internal/engine/actions/actions_test.go
dakshhhhh16: Yeah, it definitely looks a bit weird! The reason it's `do nothing` here is to prevent infinite remediation loops (where it remediates, passes, fails again, and then tries to remediate again). I've added a note to the test name to make that a bit clearer.

I totally agree that a proper state machine or level-setting approach would be much cleaner long-term, but probably best to leave it as-is for now so we don't balloon the scope of this PR!

internal/engine/actions/actions_test.go
evankanderson: This line is still off.  Running `make lint-fix` should auto-fix it for you, assuming you've run `make bootstrap` to get the tools installed.

dakshhhhh16: I was working on the tests and, I have by mistakenly pushed some prs which are having non-test files.

dakshhhhh16: > This line is still off. Running `make lint-fix` should auto-fix it for you, assuming you've run `make bootstrap` to get the tools installed.

Fixing this

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,220 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			var prevRow *db.ListRuleEvaluationsByProfileIdRow
+			if !tt.nilRow {
+				prevRow = &db.ListRuleEvaluationsByProfileIdRow{RemStatus: tt.prevEval}
+			}
+			got := shouldRemediate(prevRow, tt.evalErr)
+			assert.Equal(t, tt.expected, got)
+		})
+	}
```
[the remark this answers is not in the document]
evankanderson: We're using go 1.25, so this footgun is very much locked away.  Sorry copilot.

internal/engine/actions/actions_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+	tests := []struct {
+		name      string
+		prevAlert db.AlertStatusTypes
+		nilRow    bool
+		evalErr   error
+		remErr    error
+		remType   string
+		expected  engif.ActionCmd
+	}{
```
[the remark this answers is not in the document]
evankanderson: (alternatively, you can remove `remErr`)

I'd noticed this but wasn't bothered by it.

dakshhhhh16:
@evankanderson 
Any amendments into this?

PR #6210, opened 2026-03-29
Test(util): Add unit tests for rest curl generator

dakshhhhh16:
## Summary
This PR introduces unit tests for [internal/util/rest.go](cci:7://file:///Users/dakshpathak/Desktop/Minder/minder/internal/util/rest.go:0:0-0:0). The new tests check [GenerateCurlCommand](cci:1://file:///Users/dakshpathak/Desktop/Minder/minder/internal/util/rest.go:19:0-62:1) to ensure that it safely constructs curl strings for manual REST API debugging. This guards against potential templating issues, missing methods, or malformed API base URLs silently creeping into the cli logic. There are no new dependencies required.
Fixes #<related issue>
## Testing
- Ran `go test ./internal/util -run TestGenerateCurlCommand -v`.
- Provided test cases explicitly cover missing endpoints, invalid URL formats (`http://[::1]:namedport`), empty methods, and completely valid schema rendering. All executions pass as expected.

### DIFF
--- a/internal/util/rest_test.go
+++ b/internal/util/rest_test.go
@@ -15,13 +15,13 @@ func TestGenerateCurlCommand(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name        string
-		method      string
-		apiBaseURL  string
-		endpoint    string
-		body        string
-		wantError   bool
-		wantString  string
+		name       string
+		method     string
+		apiBaseURL string
+		endpoint   string
+		body       string
+		wantError  bool
+		wantString string
 	}{
 		{
 			name:       "valid curl command standard",
@@ -80,7 +80,7 @@ func TestGenerateCurlCommand(t *testing.T) {
 				require.Error(t, err)
 				return
 			}
-			
+
 			require.NoError(t, err)
 			assert.Equal(t, tt.wantString, got)
 		})


internal/util/rest_test.go
```diff
@@ -0,0 +1,68 @@
...
+		wantString  string
+	}{
+		{
+			name:       "valid curl command standard",
+			method:     "POST",
+			apiBaseURL: "https://api.github.com",
+			endpoint:   "/repos/mindersec/minder/issues",
+			body:       `{"title":"test"}`,
+			wantError:  false,
+			wantString: "curl -L -X POST \\\n -H \"Accept: application/vnd.github+json\" \\\n -H \"Authorization: Bearer $TOKEN\" \\\n -H \"X-GitHub-Api-Version: 2022-11-28\" \\\n https://api.github.com/repos/mindersec/minder/issues \\\n -d '{\"title\":\"test\"}'",
```
evankanderson: Strings with a lot of quotes generally benefit from using Go's backtick strings: 


```suggestion
			wantString: `curl -L -X POST \
 -H "Accept: application/vnd.github+json" \
 -H "Authorization: Bearer $TOKEN" \
 -H "X-GitHub-Api-Version: 2022-11-28" \
 https://api.github.com/repos/mindersec/minder/issues \
 -d '{"title":"test"}'`,
```

internal/util/rest_test.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+		{
+			name:       "empty method",
+			method:     "",
+			apiBaseURL: "https://api.github.com",
+			wantError:  true,
```
evankanderson, line 58: Slight preference for using a non-empty string to indicate both that an error is expected and to check the error substring.  I've found that this tends to produce more useful errors (e.g. "invalid port" vs "validation error").


```suggestion
			errorString: "method is empty",
```

(Or whatever the correct error match would be -- `assert.ErrorContains` works well for this)

internal/util/rest_test.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			got, err := GenerateCurlCommand(context.Background(), tt.method, tt.apiBaseURL, tt.endpoint, tt.body)
+			if tt.wantError {
+				assert.Error(t, err)
```
evankanderson: It's not particularly important here, but using early-return on error cases (rather than else) can help avoid deeply-nested "if" blocks.

internal/util/rest_test.go
dakshhhhh16: okayy!

internal/util/rest_test.go
dakshhhhh16: Yeahh, not an issue!

internal/util/rest_test.go
dakshhhhh16: Added

internal/util/rest_test.go
```diff
@@ -0,0 +1,73 @@
...
+			method:     "GET",
+			apiBaseURL: "",
+			wantError:  true,
+		},
+		{
+			name:       "invalid url format",
+			method:     "GET",
+			apiBaseURL: "http://[::1]:namedport", // invalid port
+			wantError:  true,
+		},
```
evankanderson, line 71: I'm noticing that we didn't test a body like:

```golang
			body:       `{"text": "can't do this; rm -rf *"}`,
```

Which would reveal that the string output wasn't actually safe to `exec`.  (I don't think we do, I think this is diagnostic only.)

Sorry for not catching this over the weekend, I the coverage report doesn't work well from my phone.

evankanderson: Possible "fixes" (from least to most effort):

1. Document this in a comment
2. Use some sort of substitution fix to make the body single-quote safe
3. Possibly remove the functionality

dakshhhhh16: Good catch! No worries at all about the weekend. I've gone ahead and added a test case exactly like what you suggested `{"text": "can't do this; rm -rf *"} `so that we now explicitly verify that edge case.

dakshhhhh16: Also I ended up going with option 2! I've added a substitution fix using 
`strings.ReplaceAll(body, "'", "'\\''")` to properly escape the single quotes, so the resulting string output is now safe if it were to be executed.

internal/util/rest.go
```diff
@@ -48,11 +49,14 @@ func GenerateCurlCommand(ctx context.Context, method, apiBaseURL, endpoint, body
 	}
 	u = u.JoinPath(endpoint)
 
+	// Escape single quotes in the body so that it is safe to be enclosed in single quotes in a shell command
+	safeBody := strings.ReplaceAll(body, "'", `'\''`)
```
evankanderson: You're gonna cry, but at least for Bash, this isn't correct:

https://stackoverflow.com/questions/1250079/how-to-escape-single-quotes-within-single-quoted-strings

In bash, `'\'` is simply `\`, and `'\''` is invalid (unterminated `'`).  I'm not sure about other shells.  Shell quoting safety is tricky; you can do something like `'can'"'"'t do this'` (Which is `'can'` then `"'"` then `'t do this'`), but it's clearly very ugly.)

dakshhhhh16: Felt the same while I was writing the code!
making the changes

internal/util/rest_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenerateCurlCommand(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name        string
```
evankanderson: You have an extra space here beyond what `gofmt` would suggest, which makes the linter sad.

dakshhhhh16: Damn! Thanks for catching that.
This is my first time working with such an accurate mainatiner like you

evankanderson: In this case, I'm just reading what the CI says.  In general, I wouldn't be as bothered by it, except that it will trigger in the future.

dakshhhhh16: Yeah, so is it ready for merge?

Issue #6105, opened 2026-02-10
Inventory gathering: discover and catalog occurrences across entities


JAORMX:
## Please describe the enhancement

Minder excels at policy enforcement (pass/fail evaluation) across repos, artifacts, and PRs. However, it currently **discards** the rich data it ingests after each evaluation cycle -- only the pass/fail status is persisted. Users need a complementary capability: **discovering and cataloging what exists** across their repos without necessarily making a judgment about it.

**Use cases this would unlock:**

| Use Case | Question Answered |
|----------|-------------------|
| **Vulnerability response** | "Which repos are affected by CVE-2024-XXXX?" (cross-ref dep inventory with OSV) |
| **Configuration drift** | "Which repos diverge from our blessed Dockerfile base images?" |
| **Compliance reporting** | "What licenses exist across all our dependencies?" |
| **Supply chain mapping** | "What Actions, tools, and deps does each repo use?" |
| **Onboarding/discovery** | "What do we have?" -- a new engineer's first question |
| **Shadow dependency detection** | "Are any repos pulling from unauthorized registries?" |

Today Minder can answer these per-repo with individual rule evaluations, but can't aggregate across repos or persist the collected data for querying.

**The core gap:** The evaluation pipeline produces structured data (via `EvaluationResult.Output`) but throws it away. Only status + error text reaches the database. The building blocks for collection exist (ingesters, Rego extraction, event pipeline) -- what's missing is **persistence and querying** of structured results.

### Extend Profiles with Inventory Mode

Profiles gain an optional `mode` field. When set to `inventory`, the profile produces **occurrences** (structured findings) instead of pass/fail evaluations. Everything else -- rule types, ingesters, Rego extraction, scheduling, authorization, selectors -- works the same.

**An occurrence** is the atomic unit of inventory -- a specific finding in a specific entity:
- **Where**: which entity (repo, artifact) it was found in
- **What**: a normalized key (`actions/checkout@v4`) + structured metadata (file path, version, etc.)
- **When**: first_seen / last_seen timestamps
- **How**: which rule type + ingester collected it

### What gets reused

| Component | Reused As-Is | Notes |
|-----------|-------------|-------|
| **Ingesters** (REST, Git, Deps, Artifact) | Yes | Already pull Actions workflows, SBOMs, file contents |
| **Rego evaluator** | Extended | New `type: inventory` mode that emits items instead of violations |
| **Profile system** | Extended | New `mode: inventory` field |
[...]

PR #6211, opened 2026-03-29
feat(inventory): add evaluation_outputs table for persisting structur…

### LINKED ISSUE #6192: Inventory: Save Evaluation Output
This is part of #6105.

We should define a new table (`evaluation_outputs`) which contains the `output` (and possibly `debug` text) from rule evaluations.  This would be keyed alongside (as a child of) `evaluation_statuses` to contain larger data types from rule evaluation, which may be collected more frequently than evaluation status information, which is generally compact.  Size for the new data:

* Output: at least 8KB, 16KB would be preferred
* Debug: collect `printf` statements to enable tracing (optional), max 2-5KB

This should be able to re-use the `evaluation_time` and other content from `evaluation_statuses`, but should have a more limited duration -- for example, the last 5 executions, with no more than 30 days retained.  Given 100K rule-entity combos evaluated (e.g. 5K entities and 20 rules each), this would be maximum 10GB of on-disk storage, which is **reasonable** given the scale, particularly given the functionality enabled.

dakshhhhh16:
# Summary

This pr adds persistent storage for structured evaluation output data, which is the first step toward the Inventory Gathering feature (i.e issues #6105, #6192).

Right now, when Minder evaluates a rule, it produces rich structured data but only saves the pass/fail status. All the detailed output gets thrown away. This PR fixes that by creating a new `evaluation_outputs` table that stores the full output alongside the evaluation status, so it can be queried later.

**What changed:**

- New database migration (`000116`) that creates the `evaluation_outputs` table, linked to `evaluation_statuses` via a foreign key with cascade delete
- New sqlc queries for inserting (with upsert), reading, and bulk-deleting evaluation outputs
- Extended [StoreEvaluationStatus](cci:1://file:///Users/dakshpathak/Desktop/Minder/minder/internal/history/service.go:27:1-40:21) in the history service to accept an optional structured output payload and persist it automatically
- Updated all callers and regenerated mocks to match the new interface

This lays the groundwork for future phases where we can build inventory queries and aggregation on top of this stored output data.

Fixes #6192
Related to #6105

# Testing

- Ran `go build ./internal/history/... ./internal/engine/... ./internal/db/...` with zero errors
- Ran `go test ./internal/history/... -v -count=1` and all existing tests pass
- Verified that the new `output` parameter is wired through correctly with `nil` at the current call site, ready for future Rego output extraction

### DIFF
--- a/database/query/eval_outputs.sql
+++ b/database/query/eval_outputs.sql
@@ -8,11 +8,11 @@ INSERT INTO evaluation_outputs(
     debug
 ) VALUES (
     $1,
-    sqlc.narg(output)::jsonb,
+    sqlc.arg(output)::jsonb,
     sqlc.narg(debug)
 )
 ON CONFLICT (id) DO UPDATE
-SET output = COALESCE(sqlc.narg(output)::jsonb, evaluation_outputs.output),
+SET output = COALESCE(sqlc.arg(output)::jsonb, evaluation_outputs.output),
     debug  = COALESCE(sqlc.narg(debug), evaluation_outputs.debug);
 
 -- name: GetEvaluationOutput :one
--- a/internal/history/service.go
+++ b/internal/history/service.go
@@ -13,7 +13,7 @@ import (
 	"slices"
 
 	"github.com/google/uuid"
-	"github.com/sqlc-dev/pqtype"
+	"github.com/rs/zerolog"
 
 	"github.com/mindersec/minder/internal/db"
 	evalerrors "github.com/mindersec/minder/internal/engine/errors"
@@ -131,17 +131,13 @@ func (e *evaluationHistoryService) StoreEvaluationStatus(
 	// Persist structured output if provided
 	if output != nil {
 		outputJSON, err := json.Marshal(output)
-		if err == nil {
-			if err := qtx.UpsertEvaluationOutput(ctx, db.UpsertEvaluationOutputParams{
-				ID: evaluationID,
-				Output: pqtype.NullRawMessage{
-					RawMessage: outputJSON,
-					Valid:      true,
-				},
-			}); err != nil {
-				// Non-fatal: log but don't fail the evaluation
-				_ = err
-			}
+		if err != nil {
+			zerolog.Ctx(ctx).Error().Err(err).Msg("failed to convert rule output to JSON")
+		} else if err := qtx.UpsertEvaluationOutput(ctx, db.UpsertEvaluationOutputParams{
+			ID:     evaluationID,
+			Output: outputJSON,
+		}); err != nil {
+			return evaluationID, fmt.Errorf("error storing extended output for rule/entity %s: %w", ruleEntityID, err)
 		}
 	}
 
--- a/database/mock/store.go
+++ b/database/mock/store.go
@@ -1107,18 +1107,18 @@ func (mr *MockStoreMockRecorder) GetEvaluationHistory(ctx, arg any) *gomock.Call
 }
 
 // GetEvaluationOutput mocks base method.
-func (m *MockStore) GetEvaluationOutput(ctx context.Context, evaluationID uuid.UUID) (db.EvaluationOutput, error) {
+func (m *MockStore) GetEvaluationOutput(ctx context.Context, id uuid.UUID) (db.EvaluationOutput, error) {
 	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "GetEvaluationOutput", ctx, evaluationID)
+	ret := m.ctrl.Call(m, "GetEvaluationOutput", ctx, id)
 	ret0, _ := ret[0].(db.EvaluationOutput)
 	ret1, _ := ret[1].(error)
 	return ret0, ret1
 }
 
 // GetEvaluationOutput indicates an expected call of GetEvaluationOutput.
-func (mr *MockStoreMockRecorder) GetEvaluationOutput(ctx, evaluationID any) *gomock.Call {
+func (mr *MockStoreMockRecorder) GetEvaluationOutput(ctx, id any) *gomock.Call {
 	mr.mock.ctrl.T.Helper()
-	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetEvaluationOutput", reflect.TypeOf((*MockStore)(nil).GetEvaluationOutput), ctx, evaluationID)
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetEvaluationOutput", reflect.TypeOf((*MockStore)(nil).GetEvaluationOutput), ctx, id)
 }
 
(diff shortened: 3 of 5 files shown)


database/migrations/000116_evaluation_outputs.up.sql
```diff
@@ -0,0 +1,21 @@
...
+CREATE TABLE evaluation_outputs (
+    id                   UUID NOT NULL DEFAULT gen_random_uuid() PRIMARY KEY,
+    evaluation_id        UUID NOT NULL REFERENCES evaluation_statuses(id) ON DELETE CASCADE,
+    output               JSONB,
+    debug                TEXT,
+    UNIQUE(evaluation_id)
+);
+
+CREATE INDEX idx_evaluation_outputs_evaluation_id
+    ON evaluation_outputs(evaluation_id);
```
evankanderson: Since we only have a single `evaluation_outputs` row per `evaluation_statuses`, let's re-use the `id` from `evaluation_statuses`, rather than adding an additional UUID column.  This _should_ help align the storage of the two columns for index scans when we're e.g. retrieving all the evaluations and outputs for a given entity or ruletype.

database/query/eval_outputs.sql
```diff
@@ -0,0 +1,24 @@
+-- SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+-- SPDX-License-Identifier: Apache-2.0
+
+-- name: InsertEvaluationOutput :exec
```
evankanderson: Slight preference for `UPSERT` if that's our operation:


```suggestion
-- name: UpsertEvaluationOutput :exec
```

internal/engine/eval_status.go
```diff
@@ -125,6 +125,7 @@ func (e *executor) createOrUpdateEvalStatus(
 			params.EntityID,
 			params.GetEvalErr(),
 			chkpjs,
+			nil, // TODO: pass structured output from evaluation result
```
evankanderson: You should be able to pass `params.GetEvalResult().Output` here (though you may need a `nil` check, because we return a pointer to the interface...).

internal/history/service.go
```diff
@@ -83,6 +87,7 @@ func (e *evaluationHistoryService) StoreEvaluationStatus(
 	entityID uuid.UUID,
 	evalError error,
 	marshaledCheckpoint []byte,
+	output json.RawMessage,
```
evankanderson: I would pass an `any` here documented as being a go struct to be JSON encoded, and then perform the JSON encoding in this function.

database/migrations/000116_evaluation_outputs.up.sql
evankanderson: Adding a second UUID index that's not correlated with the first UUID index is going to add a bunch of extra database maintenance:

* On each write, we will have O(logN) reads for `evaluation_statuses(id)` + O(logN) reads for `idx_evaluation_outputs_evaluation_id` + 1 write to the index for the index, AND O(logN) reads for `id` + 1 write for `id`.  Removing the index removes O(logN) reads for the index and a write.  (We'll still need the `REFERENCES` validation and the index on the primary key.)
* On deletes, we'll end up with O(logN) reads from `idx_evaluation_outputs_evaluation_id` to find the `id`s to delete, and then O(logN) reads to find the `id`s, and then N writes to delete them.  This is basically a 2x cost on the delete side.

If we make `id` be `= evaluation_statuses(id)` or use `evaluation_id` as the primary key, we can avoid the extra index costs and simplify the table schema.  (I realize we haven't always done this, but I'd prefer to start modelling the database more intentionally for new tables and columns).

If we share the primary key with `evaluation_statuses`, then we also get the `UNIQUE` constraint "for free".

dakshhhhh16: Just updated the table to share the `id` primary key with `evaluation_statuses` directly. Thanks for the review!"

internal/history/service.go
```diff
@@ -122,6 +128,23 @@ func (e *evaluationHistoryService) StoreEvaluationStatus(
...
+				ID: evaluationID,
+				Output: pqtype.NullRawMessage{
+					RawMessage: outputJSON,
+					Valid:      true,
+				},
+			}); err != nil {
+				// Non-fatal: log but don't fail the evaluation
+				_ = err
+			}
+		}
```
evankanderson, line 141: If you're willing to be slightly wasteful, you can write this as:


```suggestion
	// Persist structured output if provided
	if outputJSON, err := json.Marshal(output); output != nil {
		if err != nil {
			// this really shouldn't happen
			zerolog.Ctx(ctx).Error(err).Msg("failed to convert rule output to JSON")
			// making this a function would let you early-return here
		} else {
			err := qtx.UpsertEvaluationOutput(ctx, db.UpsertEvaluationOutputParams{
				ID: evaluationID,
				Output: pqtype.NullRawMessage{
					RawMessage: outputJSON,
					Valid:      true,
				},
			})
			if err != nil {
				return evaluationID, fmt.Errorf("error storing extended output for rule/entity %s: %w", ruleEntityID, err)
			}
		}
```

In particular, we generally want to do exactly one of two things with errors:

1. Escalate them up the stack (usually with early-return)
2. Handle them - with something like a retry or by logging them (and then NOT returning up the stack)

If we follow this pattern, errors should either be returned to the user ("handle it" by setting a useful error code or storing the error someplace user-findable) OR they should show up in the server error logs.

The previous version of the code would silently disappear EvalResultOutput that failed to JSON marshal, and silently disappear failed database writes to the output table.  (You had no way of knowing these rules until I told you.)

database/query/eval_outputs.sql
```diff
@@ -0,0 +1,24 @@
...
+-- SPDX-License-Identifier: Apache-2.0
+
+-- name: UpsertEvaluationOutput :exec
+INSERT INTO evaluation_outputs(
+    id,
+    output,
+    debug
+) VALUES (
+    $1,
+    sqlc.narg(output)::jsonb,
```
evankanderson: Does this need to be an `sqlc.narg()`, or can this be required on input (always not null)?  That might simplify some of the caller code.

database/query/eval_outputs.sql
```diff
@@ -0,0 +1,24 @@
...
+    sqlc.narg(debug)
+)
+ON CONFLICT (id) DO UPDATE
+SET output = COALESCE(sqlc.narg(output)::jsonb, evaluation_outputs.output),
+    debug  = COALESCE(sqlc.narg(debug), evaluation_outputs.debug);
+
+-- name: GetEvaluationOutput :one
+SELECT * FROM evaluation_outputs
+WHERE id = $1;
```
evankanderson, line 20: I expect we'll probably prefer to do a join in some of the queries on evaluation history, but it's fine to leave this query in for now.

dakshhhhh16: Okayy

PR #6218, opened 2026-03-31
Support reporting status checks and comments from PR rules

### LINKED ISSUE #5977: Support reporting status checks from PR rules

Currently, all PR rules use custom evaluators (`vulncheck`, `trusty`, `homoglyphs`), and the rule evaluation has side effects of adding comments to the PR, even if alert and remediate are false.

PR comments (and status checks) should be available as standard remediation or alerting flows within PRs, so that PR evaluations can also be written in Rego.

### Solution Proposal

PR comments should be available as a `remediate` type, and Check success / failure should be available as an `alert` type.  Since we allow both remediation and alerts (but only one of each), this best fits in our current model, and would support cases like mindersec/minder-rules-and-profiles#354.

We could support multiple remediations (as a list) such that comments and status checks could both be remediations.  We would need to define semantics for "apply all" or "apply until succeeded" for chained remediations, which we don't consider today.

We could also have a single "PR feedback" type which permits both a list of comments and a list of status check outcomes in a single execution.  This would have the benefit of not conflating enabling security alerts with enabling PR feedback.  Avoiding this conflation could also be addressed by putting the rules with security alert and the rules covering PR feedback in separate profiles.

### Additional context

_No response_

### Acceptance Criteria

_No response_

krrish175-byte:
# Summary

This pull request implements the enhancement requested in Issue #5977, allowing generic PR rules (e.g., Rego rules) to report status checks and leave PR comments natively without relying on custom Go evaluation engines like `vulncheck`.

**Changes included:**
* **Protobuf ([minder.proto](cci:7://file:///Users/krrishbiswas/Desktop/LFX/minder/proto/minder/v1/minder.proto:0:0-0:0))**: 
  * Added `commit_status` as an available [Alert](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/internal/engine/actions/alert/commit_status/commit_status.go:33:0-38:1) type.
  * Added `pull_request_comment` as an available [Remediate](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/proto/minder/v1/minder.proto:2801:8-2908:9) type.
* **Alert Engine**: Created a new `commit_status` action under `internal/engine/actions/alert/commit_status` which interfaces with the GitHub `SetCommitStatus` API to surface a success/failure check run directly on the PR.
* **Remediate Engine**: Linked the existing `pull_request_comment` alert module into the `Remediate` registry so it can be utilized in the remediation phase without duplicating logic.

Fixes #5977

# Testing

***Outline how the changes were tested, including steps to reproduce and any relevant configurations.***

* Implemented full unit test coverage for the new `commit_status` action (`commit_status_test.go`) utilizing `mockghclient`.
* Verified that `pull_request_comment` correctly registers inside the Remediator engine.
* Ran `make buf-generate` to ensure all generated protobuf Go bindings are clean and up to date.
* Ran `go test ./...` across the entire `internal/engine/actions/...` package, with all tests passing successfully.


proto/minder/v1/minder.proto
```diff
@@ -2802,7 +2791,7 @@ message RuleType {
         message Remediate {
             string type = 1 [
                 (buf.validate.field).string = {
-                    in: ["rest", "gh_branch_protection", "pull_request"],
+                    in: ["rest", "gh_branch_protection", "pull_request", "pull_request_comment"],
```
evankanderson, line 2809: We should probably document which remediation types match up with which entity types in a comment. on the `type` field.

proto/minder/v1/minder.proto
```diff
@@ -2905,13 +2894,14 @@ message RuleType {
             optional RestType rest = 2;
             optional GhBranchProtectionType gh_branch_protection = 3;
             optional PullRequestRemediation pull_request = 4;
+            optional Alert.AlertTypePRComment pull_request_comment = 5;
```
evankanderson, line 2912: Try pulling the `AlertTypePRComment` message out of the `Alert` message, and see whether `buf` flags that as a breaking change.  We had some early use of nested messages that went a bit overboard for these use cases.

Also, if we are going to add PR comments as a remediation type, it would be good to understand (document) in an issue or the PR description when people should use the "remediate" vs "alert" type.  A few options:

1. We'll automatically convert "alert: pr_comment" to "remediate: pr_comment"
   * What happens to people who have policies remediation enabled but not alerting, or vice-versa?
2. We'll support all options on both types
   * how do we explain the differences to people?
4. We'll have different options on the two types -- e.g. only remediate can block PR merges
   * Should this be a separate type then?

I'm not trying to be a pain, but it's a lot easier to jump _into_ a mess than to get ourselves back out.  😁

proto/minder/v1/minder.proto
```diff
@@ -2905,13 +2894,14 @@ message RuleType {
...
         }
         Remediate remediate = 6;
 
         message Alert {
             string type = 1 [
```
evankanderson, line 2920: We probably should have included a comment here about which types correspond to which entities as well.

proto/minder/v1/minder.proto
```diff
@@ -2935,6 +2925,9 @@ message RuleType {
                 ];
             }
             optional AlertTypePRComment pull_request_comment = 3;
+
+            message AlertTypeCommitStatus {}
+            optional AlertTypeCommitStatus commit_status = 4;
```
evankanderson: What is this / what is it used for?

internal/engine/actions/alert/alert.go
```diff
@@ -61,6 +62,18 @@ func NewRuleAlert(
...
 			ActionType, alertCfg.GetPullRequestComment(), client, setting)
+	case commit_status.AlertType:
+		if alertCfg.GetCommitStatus() == nil {
+			return nil, fmt.Errorf("alert engine missing commit_status configuration")
+		}
+		client, err := provinfv1.As[provinfv1.GitHub](provider)
+		if err != nil {
+			zerolog.Ctx(ctx).Debug().Str("rule-type", ruletype.GetName()).
+				Msg("provider is not a GitHub provider. Silently skipping alerts.")
+			return noop.NewNoopAlert(ActionType)
```
evankanderson, line 73: Ooof, this is a cleanup opportunity.  The Security Alerts feature is GitHub specific, but it _should_ be implemented as a Provider API such that we can do the equivalent GitLab, Forgejo, etc operation.  Commit Statuses should be handled similarly, though probably as a different "facet", so a provider could implement one but not the other.

Given the other set of changes here, I think **filing an issue** to track this in a future PR is the right step.

internal/engine/actions/alert/commit_status/commit_status.go
```diff
@@ -0,0 +1,267 @@
...
+	actionParams interfaces.ActionsParams,
+) (json.RawMessage, error) {
+	logger := zerolog.Ctx(ctx)
+
+	// Context string for the commit status check (e.g. "minder/rule_name")
+	// If rule_name isn't available, we fallback to "minder"
+	contextStr := "minder"
+	if rule := actionParams.GetRule(); rule != nil {
+		contextStr = fmt.Sprintf("minder/%s", rule.Name)
+	}
```
evankanderson: Oh, this is interesting!  I would have expected the name of the status check to be in the rule definition.  You may also want to distinguish between rule type and rule name (our nomenclature here isn't always consistent, you basically have several "name-like" things:

1. The `name` field from the profile, if set.  If not set, I think there's some defaulting here from either 3 or 2, and I don't recall the exact rules
2. The `type` field from the profile, which matches the `name` from the ruletype
3. The `display_name` from the ruletype

evankanderson (commented):
Very nice work, though we may want to consider adding either `target_url` or `description` as fields from the proto message, possibly with Go templating for the latter, and RFC6570 templates for the URL.

With respect to the `target_url`, I'm also trying to figure out whether we should offer some sort of "Rule Status URL" endpoint that could be auto-filled based on server-known parameters.  We'd also need to figure out permissions on that endpoint, which is a separate work track.  For now, I'd simply allow users to template their own URL, and we can add a separate `automagic_url` option (bool) in the future that's exclusive with the `detail_url` on the commit status message.

evankanderson:
While setting up testing for another PR, I realized that we may also need to change the [GitHub App recommended permissions](https://docs.mindersec.dev/run_minder_server/config_provider#permissions) to allow updating status checks.

You also have some lint errors that might be fixed by `make lint-fix`, or might need to be handled by hand.

proto/minder/v1/minder.proto
```diff
@@ -2800,9 +2800,13 @@ message RuleType {
         ];
 
         message Remediate {
+            // type is the type of the remediation.
+            // * 'rest' can be used with any entity type.
+            // * 'gh_branch_protection' can only be used with the 'repository' entity type.
+            // * 'pull_request' and 'pull_request_comment' can only be used with the 'pull_request' entity type.
```
evankanderson: I think `pull_request` opens a new PR on the repo, so it applies to the "repository" type, not the "pull_request" type.

tools/go.mod
```diff
@@ -397,13 +395,15 @@ require (
...
-	modernc.org/sqlite v1.46.1 // indirect
+	modernc.org/sqlite v1.44.0 // indirect
 	mvdan.cc/unparam v0.0.0-20251027182757-5beb8c8f8f15 // indirect
 	mvdan.cc/xurls/v2 v2.6.0 // indirect
 	pluginrpc.com/pluginrpc v0.5.0 // indirect
 )
 
 replace github.com/mindersec/minder => ../.
+
+replace github.com/docker/docker => github.com/docker/docker v28.0.0+incompatible
```
evankanderson: This is probably not right -- I'm assuming this was in hopes of fixing some of the security warnings?

internal/engine/actions/alert/alert.go
evankanderson: Let's file this issue before we submit the PR, so we don't forget about it.

internal/engine/actions/alert/commit_status/commit_status.go
```diff
@@ -0,0 +1,332 @@
...
+
+	if pr.Number > math.MaxInt {
+		return nil, fmt.Errorf("pr number is too large")
+	}
+	result.Number = int(pr.Number)
+
+	// Create template params
+	tmplParams := &CommitStatusTemplateParams{
+		EvalErrorDetails: enginerr.ErrorAsEvalDetails(params.GetEvalErr()),
```
evankanderson: How would this be used / what does it show up as?

internal/engine/actions/alert/commit_status/commit_status.go
```diff
@@ -0,0 +1,332 @@
...
+	// 1. Evaluate context
+	contextStr := "minder"
+	contextNameTmpl := alert.alertCfg.GetStatusName()
+	if contextNameTmpl != "" {
+		// No templating specified for status_name, use as-is
+		contextStr = contextNameTmpl
+	} else if tmplParams.RuleName != "" {
+		contextStr = fmt.Sprintf("minder/%s", tmplParams.RuleName)
+	}
+	result.Context = contextStr
```
evankanderson: With Go 1.22 and later, you can use `cmp.Or` for some of this:


```suggestion
	contextStr := "minder"
	if tmplParams.RuleName != "" {
		contextStr = fmt.Sprintf("%s/%s", contextStr, tmplParams.RuleName)
	}
	result.Context = cmp.Or(alert.alertCfg.GetStatusName(), contextStr)
```

internal/engine/actions/alert/commit_status/commit_status.go
```diff
@@ -0,0 +1,332 @@
...
+
+	// 2. Evaluate description, using text/template
+	descTmplStr := alert.alertCfg.GetDescription()
+	if descTmplStr != "" {
+		descTmpl, err := util.NewSafeHTMLTemplate(&descTmplStr, "description")
```
evankanderson: Can "description" be HTML content?  (I don't know, but the GitHub API doesn't say anything about HTML: https://docs.github.com/en/rest/commits/statuses?apiVersion=2026-03-10#create-a-commit-status)

internal/engine/actions/alert/commit_status/commit_status.go
```diff
@@ -0,0 +1,332 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("cannot parse description template: %w", err)
+		}
+		descStr, err := descTmpl.Render(ctx, tmplParams, 1024)
```
evankanderson: Extract the 1024 to a const parameter

internal/engine/actions/alert/commit_status/commit_status.go
```diff
@@ -0,0 +1,332 @@
...
+
+	// 3. Evaluate target_url, using RFC6570
+	targetUrlTmplStr := alert.alertCfg.GetTargetUrl()
+	if targetUrlTmplStr != "" {
+		argsMap := map[string]any{
+			"owner":      result.Owner,
+			"repo":       result.Repo,
+			"commit_sha": result.CommitSha,
+			"rule_name":  tmplParams.RuleName,
+		}
```
evankanderson, line 314: This feels like it should include `output`, so you could have a rule that pokes some external system, gets back a URL for the status check, and then loads that URL into the status check.

evankanderson (commented):
Looking a little bit more deeply, I'm wondering if we should be using "Check Runs" instead of "Status Checks":

https://stackoverflow.com/questions/67919168/github-checks-api-vs-check-runs-vs-check-suites

## Pros of Status Check

* Simpler GitHub API
* Works with other types of tokens than GitHub App

## Pros of Check Runs

* Automatic grouping under the GitHub App (e.g. "minder")
* Ability to add line-by-line annotations
* Detail page with markdown, and summary images in addition to URL

evankanderson:
Also, it would be good to update https://github.com/mindersec/minder/blob/main/docs/docs/run_minder_server/config_provider.md alongside any changes to necessary GitHub app permissions, since they are in the same repo.

evankanderson (changes requested):
I'm marking this PR as "request changes" (needs merge conflicts resolved) to help track which outstanding PRs need maintainer action vs contributor action.

PR #6219, opened 2026-03-31
Add tests for swagger datasource generation

### LINKED ISSUE #5300: Add unit tests for generating a data source definition from Swagger
The following issue is about adding unit tests for https://github.com/mindersec/minder/pull/5283. 

This is needed so we can ensure we don't accidentally break the generation part.

PrinceAsiedu:
## Summary

Add coverage for `mindev datasource generate` in `cmd/dev/app/datasource/generate_test.go`.

This exercises Swagger v2 datasource generation across the main success
paths and the expected error cases called out in #5300.

## What changed

- added table-driven unit tests for datasource generation
- verified generated YAML by unmarshaling into the datasource proto
- covered:
  - single endpoint with no parameters
  - path parameters
  - POST/PUT body handling
  - multiple endpoints
  - empty/invalid Swagger documents
  - unsupported header/body parameters
  - duplicate/conflicting generated operation names

## Follow-up fixes included

While writing the tests, a few issues in the generator surfaced and were fixed:

- preserve templated Swagger paths like `/users/{id}` instead of emitting URL-escaped paths
- return a clear error for unsupported parameter locations
- return a clear error when two operations collapse to the same generated name
- fix required field encoding in the generated input schema so it is accepted by `structpb`

Closes #5300

### DIFF
--- a/cmd/dev/app/datasource/generate.go
+++ b/cmd/dev/app/datasource/generate.go
@@ -63,8 +63,18 @@ func initDriverStruct() *minderv1.RestDataSource {
 
 // conver the title to a valid datasource name. It should only contain alphanumeric characters and dashes.
 func swaggerTitleToDataSourceName(title string) string {
-	re := regexp.MustCompile("^[a-z][-_[:word:]]*$")
-	return re.ReplaceAllString(title, "-")
+	re := regexp.MustCompile(`[^a-z0-9_-]+`)
+	sanitized := strings.ToLower(strings.TrimSpace(title))
+	sanitized = re.ReplaceAllString(sanitized, "-")
+	sanitized = strings.Trim(sanitized, "-")
+	if sanitized == "" {
+		return "datasource"
+	}
+	if sanitized[0] < 'a' || sanitized[0] > 'z' {
+		sanitized = "ds-" + sanitized
+	}
+
+	return sanitized
 }
 
 // swaggerToDataSource generates datasource code from an OpenAPI specification.
--- a/cmd/dev/app/datasource/generate.go
+++ b/cmd/dev/app/datasource/generate.go
@@ -6,9 +6,9 @@ package datasource
 import (
 	"fmt"
 	"net/http"
-	"net/url"
 	"os"
 	"regexp"
+	"slices"
 	"strings"
 
 	"buf.build/go/protoyaml"
@@ -69,6 +69,8 @@ func swaggerTitleToDataSourceName(title string) string {
 
 // swaggerToDataSource generates datasource code from an OpenAPI specification.
 func swaggerToDataSource(cmd *cobra.Command, swagger *spec.Swagger) error {
+	_ = cmd
+
 	if swagger.Info == nil {
 		return fmt.Errorf("info section is required in OpenAPI spec")
 	}
@@ -84,14 +86,18 @@ func swaggerToDataSource(cmd *cobra.Command, swagger *spec.Swagger) error {
 	}
 
 	for path, pathItem := range swagger.Paths.Paths {
-		p, err := url.JoinPath(basepath, path)
-		if err != nil {
-			cmd.PrintErrf("error joining path %s and basepath %s: %v\n Skipping", path, basepath, err)
-			continue
-		}
+		p := joinPaths(basepath, path)
 
 		for method, op := range operations(pathItem) {
 			opName := generateOpName(method, path)
+			if _, ok := drv.Def[opName]; ok {
+				return fmt.Errorf("duplicate generated operation name %q for %s %s", opName, method, path)
+			}
+
+			if err := validateParameters(op.Parameters); err != nil {
+				return fmt.Errorf("%s %s: %w", method, path, err)
+			}
+
 			// Create a new REST DataSource definition
 			def := &minderv1.RestDataSource_Def{
 				Method:   method,
@@ -126,6 +132,16 @@ func swaggerToDataSource(cmd *cobra.Command, swagger *spec.Swagger) error {
 	return writeDataSourceToFile(ds)
 }
 
+func joinPaths(basepath, path string) string {
+	basepath = strings.TrimSuffix(basepath, "/")
+	path = strings.TrimPrefix(path, "/")
+	if path == "" {
+		return basepath
+	}
+
+	return basepath + "/" + path
+}
+
 // Generates an operation name for a data source. Note that these names
 // must be unique within a data source. They also should be only alphanumeric
 // characters and underscores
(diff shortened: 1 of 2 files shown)


cmd/dev/app/datasource/generate_test.go
```diff
@@ -0,0 +1,344 @@
...
+	data, readErr := os.ReadFile(stdoutFile.Name())
+	require.NoError(t, readErr)
+
+	var ds minderv1.DataSource
+	require.NoError(t, protoyaml.Unmarshal(data, &ds))
+
+	return &ds, nil
+}
+
+func testSwagger(paths map[string]spec.PathItem) *spec.Swagger {
```
evankanderson, line 255: Another option here would be to create a `testdata` directory, and store actual sample Swagger JSON documents in the directory.  I don't feel strongly about that approach, but it has the advantage of being able to add real-world examples with less Go typing boilerplate.

cmd/dev/app/datasource/generate_test.go
```diff
@@ -0,0 +1,344 @@
...
+				assert.Equal(t, "users-api", ds.GetName())
+				require.NotNil(t, ds.GetRest())
+				require.Len(t, ds.GetRest().GetDef(), 1)
+
+				def := ds.GetRest().GetDef()["get_users"]
+				require.NotNil(t, def)
+				assert.Equal(t, "GET", def.GetMethod())
+				assert.Equal(t, "/api/v1/users", def.GetEndpoint())
+				assert.Equal(t, "json", def.GetParse())
+				assert.Empty(t, def.GetInputSchema().AsMap())
```
evankanderson, line 48: I know we do this in a bunch of our tests, but this sort of helper function in a table test is a little bit of a "smell".  In this case, you might look at using `go-cmp` or [protocmp](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp) to allow specifying the expected result (with a side effect of getting nice-looking diffs when there's a problem).

Using [prototext](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp), this could be:

```suggestion
			expected: `
				name: users-api
				rest [
					def : {
						key: "get_users"
						value: {
							method: "GET"
							endpoint: "/api/v1/users"
							parse: "json"
						}
					}
				]
			`,
```

cmd/dev/app/datasource/generate_test.go
```diff
@@ -0,0 +1,344 @@
...
+			}),
+			wantErr: `POST /users: unsupported parameter "payload" in "body"`,
+		},
+		{
+			name: "duplicate generated operation names",
+			swagger: testSwagger(map[string]spec.PathItem{
+				"/users/id/":  pathItem("GET", op()),
+				"/users/{id}": pathItem("GET", op(param("id", "path"))),
+			}),
+			wantErr: `duplicate generated operation name "get_users_id_"`,
```
evankanderson, line 169: ❤️  `wantErr` as a string to test the error messages.  Thanks!

cmd/dev/app/datasource/generate.go
```diff
@@ -63,12 +63,33 @@ func initDriverStruct() *minderv1.RestDataSource {
...
+	// Ensure the generator respects Cobra's configured output writer.
+	if out := cmd.OutOrStdout(); out != os.Stdout {
+		if f, ok := out.(*os.File); ok {
+			prev := os.Stdout
+			os.Stdout = f
+			defer func() {
+				os.Stdout = prev
+			}()
+		}
+	}
```
evankanderson, line 91: Rather than overwriting `os.Stdout`, let's pass the file from `cmd.OutOrStdout` to the writers, and use `Command.SetOut()` in the tests to override Stdout.  This avoids mutating global state across tests run with `t.Parallel()`.

cmd/dev/app/datasource/generate.go
```diff
@@ -126,6 +151,16 @@ func swaggerToDataSource(cmd *cobra.Command, swagger *spec.Swagger) error {
...
 
+func joinPaths(basepath, path string) string {
+	basepath = strings.TrimSuffix(basepath, "/")
+	path = strings.TrimPrefix(path, "/")
+	if path == "" {
+		return basepath
+	}
+
+	return basepath + "/" + path
+}
```
evankanderson, line 162: `url.JoinPath` should do this.  Is it not?

cmd/dev/app/datasource/generate.go
evankanderson: I think this is just adding an argument to `writeDataSourceToFile`

cmd/dev/app/datasource/generate.go
```diff
@@ -158,6 +193,16 @@ func requiresMsgBody(method string) bool {
 	return method == http.MethodPost || method == http.MethodPut || method == http.MethodPatch
 }
 
+func validateParameters(params []spec.Parameter) error {
+	for _, p := range params {
+		if !slices.Contains([]string{"path", "query"}, p.In) {
+			return fmt.Errorf("unsupported parameter %q in %q", p.Name, p.In)
```
evankanderson, line 199: We _could_ support e.g. headers, but we don't at the moment, right?  It seems worth a comment to indicate whether this is a limitation of the generation tool, or a limit of the underlying datasource API.

cmd/dev/app/datasource/generate_test.go
```diff
@@ -0,0 +1,344 @@
...
+	defer stdoutMu.Unlock()
+
+	stdoutFile, err := os.CreateTemp(t.TempDir(), "stdout-*.yaml")
+	require.NoError(t, err)
+
+	originalStdout := os.Stdout
+	os.Stdout = stdoutFile
+	defer func() {
+		os.Stdout = originalStdout
+	}()
```
evankanderson, line 237: Prefer to use `Command.SetOut` here -- all you need is an `io.Writer`, so this could simply be a `bytes.Buffer` rather than needing any actual files at all.  (and then you won't need the lock!)

cmd/dev/app/datasource/generate_test.go
```diff
@@ -0,0 +1,344 @@
...
+	require.NoError(t, err)
+
+	stdoutMu.Lock()
+	defer stdoutMu.Unlock()
+
+	originalStdout := os.Stdout
+	os.Stdout = stdoutFile
+	defer func() {
+		os.Stdout = originalStdout
+	}()
```
evankanderson, line 332: Again, prefer `Command.SetOut` here.

evankanderson (commented):
This looks very good overall!  Thanks for adding the coverage (which seems to have motivated some improvements in the actual tool as well)!

evankanderson (commented):
Add a comment when this is ready for review; at a minimum, we should use `Command.SetOut` rather than overwriting `os.Stdout`, but you may want to review the other comments as well.

evankanderson (changes requested):
Marking as "Request changes" (for at least `cmd.Stdout`), so I can more easily track which PRs need maintainer attention again.

evankanderson:
@PrinceAsiedu -- are you likely to get back to this PR?

PR #6221, opened 2026-03-31
feat(api): set output_only flag for output-only fields in protos

### LINKED ISSUE #4951: Set output_only flag for output only fields in minder protos
For clients using the openAPI spec, it would be useful to know which fields are only expected to be in the output and not required to be set in the input.
We can use the `OUTPUT_ONLY` [field_behaviour flag](https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/#using-googleapifield_behavior) to annotate these types of fields.

For example, by setting the field behaviour to `OUTPUT_ONLY` on the project field, the typescipt generated field is `readonly`.
```
message Provider {
    // ...
    string project = 2 [(google.api.field_behavior) = OUTPUT_ONLY];
}
```
becomes:
```
export type v1Provider = {
    readonly project?: string;
}
```

We currently determine if a field is output-only by putting a comment on top of the field. For example:
https://github.com/mindersec/minder/blob/cda126fedd9038be2bfe5d76861d123327b832a2/proto/minder/v1/minder.proto#L3175-L3177

This task involves finding all the fields that are documented as output-only and setting the field_behaviour flag on them.

Gitjay11:
## Summary
This PR addresses issue #4951 by adding the (google.api.field_behavior) = OUTPUT_ONLY flag to all fields in minder.proto that are documented via comments as being output-only, ignored on input, or set by the server/system.
By explicitly annotating these fields with OUTPUT_ONLY, the generated OpenAPI specification will now correctly mark them as readOnly: true. This greatly improves the developer experience for API consumers (like our TypeScript clients) by ensuring that generated client code automatically applies readonly modifiers, clarifying that these fields do not need to be provided in the request body.

The following fields were updated:

- UpstreamRepositoryRef.repo_id
- UpstreamRepositoryRef.registered
- RuleType.id
- Profile.id
- GitHubAppParams.organization
- GitHubAppParams.organization_id
- Provider.project
- Provider.credentials_state

Fixes #4951

## Testing

- Verified that the ``` google.api.field_behavior ``` changes compile correctly.
- Ran buf lint on the protobuf files; passed with zero errors.
- Ran buf breaking against the upstream repository; confirmed no breaking changes were introduced.
- Ran the core backend unit/integration tests suite ``` go test ./pkg/api/... ./internal/controlplane/... ./pkg/ruletypes/... ``` to ensure standard parsing and interactions still function successfully.

### DIFF
--- a/pkg/api/protobuf/go/minder/v1/minder.pb.go
+++ b/pkg/api/protobuf/go/minder/v1/minder.pb.go
@@ -15052,15 +15052,14 @@ const file_minder_v1_minder_proto_rawDesc = "" +
 	"\x06entity\x18\x01 \x01(\v2\x1c.minder.v1.UpstreamEntityRefB\x03\xe0A\x02R\x06entity\x12\x1e\n" +
 	"\n" +
 	"registered\x18\x02 \x01(\bR\n" +
-	"registered\"\xfc\x01\n" +
+	"registered\"\x84\x02\n" +
 	"\x15UpstreamRepositoryRef\x129\n" +
 	"\x05owner\x18\x01 \x01(\tB#\xbaH \xd8\x01\x01r\x1b\x18\xc8\x012\x16^[A-Za-z][-[:word:]]*$R\x05owner\x125\n" +
-	"\x04name\x18\x02 \x01(\tB!\xe0A\x02\xbaH\x1b\xd8\x01\x01r\x16\x18\xc8\x012\x11^[-.[:alnum:]_]+$R\x04name\x12#\n" +
-	"\arepo_id\x18\x03 \x01(\x03B\n" +
-	"\xbaH\a\xd8\x01\x01\"\x02(\x01R\x06repoId\x12,\n" +
-	"\acontext\x18\x04 \x01(\v2\x12.minder.v1.ContextR\acontext\x12\x1e\n" +
+	"\x04name\x18\x02 \x01(\tB!\xe0A\x02\xbaH\x1b\xd8\x01\x01r\x16\x18\xc8\x012\x11^[-.[:alnum:]_]+$R\x04name\x12&\n" +
+	"\arepo_id\x18\x03 \x01(\x03B\r\xe0A\x03\xbaH\a\xd8\x01\x01\"\x02(\x01R\x06repoId\x12,\n" +
+	"\acontext\x18\x04 \x01(\v2\x12.minder.v1.ContextR\acontext\x12#\n" +
 	"\n" +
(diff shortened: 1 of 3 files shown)


evankanderson:
It looks like this overlaps with #6203, which was opened last week (and has different problems)

evankanderson (approved):
(Will merge tomorrow morning, assuming CI passes)

evankanderson (dismissed):

pkg/api/protobuf/go/minder/v1/minder.pb.go
```diff
@@ -14858,7 +14858,7 @@ const file_minder_v1_minder_proto_rawDesc = "" +
 	"\x19GetArtifactByNameResponse\x124\n" +
 	"\bartifact\x18\x01 \x01(\v2\x13.minder.v1.ArtifactB\x03\xe0A\x02R\bartifact\x126\n" +
 	"\bversions\x18\x02 \x03(\v2\x1a.minder.v1.ArtifactVersionR\bversions\"\t\n" +
-	"\aRelease\"\r\n" +
+	"\aRelease\"\n" +
```
evankanderson: I'm not sure why you're generating diffs here, but the `make gen` check thinks you shouldn't be.

Gitjay11: Fixed! The \r\n → \n diffs were caused by running buf generate on Windows with CRLF line endings in the proto file. I've rebased on latest main, converted the proto to LF and regenerated.

evankanderson (commented):
Sorry for the delay -- this went around a few times and I saw the CI fail, and didn't dig in to see it was unrelated lint errors that had crept in.  Unfortunately, you'll need to merge one more time (and possibly re-run `make gen`) for this one.

I'm marking this PR as "request changes" (needs merge conflicts resolved) to help track which outstanding PRs need maintainer action vs contributor action.

evankanderson (commented):
This has a bunch of changes to `docs/docs/ref/proto.mdx` (per the `make gen` and lint check) that look to be incorrect.  I also see `^M` control characters in the `proto.mdx` file.

evankanderson:
The unit test error is a flake that I've seen before, but `make gen` on Linux is producing different output than what you've got in your PR, so you'll need to fix that.  (If you're on Windows, you may need to use WSL to get correct results -- that's what I currently do on my laptop.)

Gitjay11:
I'm working on it, facing some problem so it is taking time.

evankanderson (approved):
The security scan is failing on a containerd vulnerability fixed at head.

PR #6222, opened 2026-04-01
refactor: extract shared label filtering to internal/labels 

### LINKED ISSUE #4982: Refactor label filtering logic
As per [this comment](https://github.com/mindersec/minder/pull/4910#discussion_r1840625396), we currently allow filtering by label in two places (i.e. two endpoints), but the code to handle labels is duplicated.

In particular, the first implementation ([here](https://github.com/mindersec/minder/blob/main/internal/db/domain.go#L49)) handles input processing in the database layer, while the newer implementation has parameter splitting in the [handler](https://github.com/mindersec/minder/blob/main/internal/controlplane/handlers_evalstatus.go#L288-L299)) and inclusion/exclusion (i.e. `label` vs `!label`) in the [service](https://github.com/mindersec/minder/blob/main/internal/history/models.go#L516-L530).

Refactor the older implementation into the second one, possibly reusing what is to be merged in the PR linked.

Gitjay11:
# Summary
Label filtering logic for profiles is duplicated across two endpoints (ListEvaluationResults and ListEvaluationHistory) with inconsistent implementations. The older endpoint parses filter strings in internal/db/domain.go via LabelsFromFilter(), while the newer endpoint re-implements the same parsing in internal/history/models.go via AddLabel(). Additionally, the newer endpoint has a bug where excluded labels (e.g. !label) are correctly parsed but never passed to the database query — the paramsFromLabelFilter function had // We do not exclude based on labels, even though the SQL already supports it via the Notlabels parameter.

This PR extracts the shared label-parsing logic (comma splitting, !-prefix exclusion, * wildcard) into a new internal/labels utility package and refactors both endpoints to use it. It also fixes the exclusion bug so that negative label filters are correctly applied at the database level for evaluation history queries.
No new dependencies are required.

Fixes #4982

# Testing
All affected packages compile and pass their tests:

### New tests :

1. ``` go test ./internal/labels  ```

Screenshot 2026-04-01 104051

### existing tests : 

2. ``` go test ./internal/history/...  ```
ok      github.com/mindersec/minder/internal/history    0.741s

3. ``` go test ./internal/db/...  ```
ok      github.com/mindersec/minder/internal/db         61.722s

4. ``` go test ./internal/controlplane/...  ```
ok      github.com/mindersec/minder/internal/controlplane       44.925s

5. ``` go build ./internal/... ```
(no errors)

All the tests are passing 

Screenshot 2026-04-01 105925

### DIFF
--- a/internal/controlplane/handlers_evalstatus.go
+++ b/internal/controlplane/handlers_evalstatus.go
@@ -526,9 +526,7 @@ func buildProjectsProfileList(
 	profileList := []db.ListProfilesByProjectIDAndLabelRow{}
 
 	listParams := db.ListProfilesByProjectIDAndLabelParams{}
-	if err := listParams.LabelsFromFilter(filter); err != nil {
-		return nil, util.UserVisibleError(codes.InvalidArgument, "invalid label filter: %s", err)
-	}
+	listParams.LabelsFromFilter(filter)
 
 	for _, projectID := range projects {
 		listParams.ProjectID = projectID
--- a/internal/db/domain.go
+++ b/internal/db/domain.go
@@ -48,13 +48,9 @@ func (r ListProfilesByProjectIDAndLabelRow) GetSelectors() []ProfileSelector {
 }
 
 // LabelsFromFilter parses the filter string and populates the IncludeLabels and ExcludeLabels fields
-func (lp *ListProfilesByProjectIDAndLabelParams) LabelsFromFilter(filter string) error {
-	inc, exc, err := labels.ParseLabelFilter(filter)
-	if err != nil {
-		return err
-	}
-
+func (lp *ListProfilesByProjectIDAndLabelParams) LabelsFromFilter(filter string) {
+	inc, exc := labels.ParseLabelFilter(filter)
 	lp.IncludeLabels = inc
 	lp.ExcludeLabels = exc
-	return nil
 }
+
--- a/internal/labels/labels.go
+++ b/internal/labels/labels.go
@@ -11,9 +11,9 @@ import (
 // ParseLabelFilter parses a comma-separated label filter string into lists of
 // labels to include and exclude. It resolves wildcards so that if any inclusion
 // rule is `*`, the included labels list evaluates simply to `["*"]`.
-func ParseLabelFilter(filter string) (include []string, exclude []string, err error) {
+func ParseLabelFilter(filter string) (include []string, exclude []string) {
 	if filter == "" {
-		return nil, nil, nil
+		return nil, nil
 	}
 
 	var starMatched bool
@@ -41,7 +41,7 @@ func ParseLabelFilter(filter string) (include []string, exclude []string, err er
 		include = []string{"*"}
 	}
 
-	return include, exclude, nil
+	return include, exclude
 }
 
 // ParseLabel parses a single label (without commas) into an include or exclude string.
--- a/internal/labels/labels_test.go
+++ b/internal/labels/labels_test.go
@@ -18,40 +18,33 @@ func TestParseLabelFilter(t *testing.T) {
 		expectedExc []string
 	}{
 		{
-			name:        "empty",
-			filter:      "",
-			expectedInc: nil,
-			expectedExc: nil,
+			name:   "empty",
+			filter: "",
 		},
 		{
 			name:        "single include",
 			filter:      "foo",
 			expectedInc: []string{"foo"},
-			expectedExc: nil,
 		},
 		{
 			name:        "single exclude",
 			filter:      "!foo",
-			expectedInc: nil,
 			expectedExc: []string{"foo"},
 		},
 		{
 			name:        "star include",
 			filter:      "*",
 			expectedInc: []string{"*"},
-			expectedExc: nil,
 		},
 		{
 			name:        "star exclude",
 			filter:      "!*",
-			expectedInc: nil,
 			expectedExc: []string{"*"},
 		},
 		{
 			name:        "multiple includes",
 			filter:      "foo,bar",
 			expectedInc: []string{"foo", "bar"},
-			expectedExc: nil,
 		},
 		{
 			name:        "includes and excludes",
@@ -63,13 +56,11 @@ func TestParseLabelFilter(t *testing.T) {
 			name:        "star mixed with includes",
 			filter:      "foo,*",
 			expectedInc: []string{"*"},
-			expectedExc: nil,
 		},
 		{
 			name:        "includes mixed with star",
 			filter:      "*,foo",
 			expectedInc: []string{"*"},
-			expectedExc: nil,
 		},
 		{
 			name:        "star and excludes",
@@ -95,8 +86,7 @@ func TestParseLabelFilter(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			inc, exc, err := ParseLabelFilter(tt.filter)
-			require.NoError(t, err)
+			inc, exc := ParseLabelFilter(tt.filter)
 			require.Equal(t, tt.expectedInc, inc)
 			require.Equal(t, tt.expectedExc, exc)
 		})
--- a/internal/controlplane/handlers_profile.go
+++ b/internal/controlplane/handlers_profile.go
@@ -106,9 +106,7 @@ func (s *Server) ListProfiles(ctx context.Context,
 	listParams := db.ListProfilesByProjectIDAndLabelParams{
 		ProjectID: entityCtx.Project.ID,
 	}
-	if err := listParams.LabelsFromFilter(req.GetLabelFilter()); err != nil {
-		return nil, util.UserVisibleError(codes.InvalidArgument, "invalid label filter: %s", err)
-	}
+	listParams.LabelsFromFilter(req.GetLabelFilter())
 
 	zerolog.Ctx(ctx).Debug().Interface("listParams", listParams).Msg("profile list parameters")
 


evankanderson:
Thanks for the testing screenshots; since you're a new contributor, I need to approve the `test` workflow to run. I'll be looking at the CI results (and re-running if it hits one of the flaky tests) for confirmation as well as coverage information.

internal/controlplane/handlers_evalstatus.go
```diff
@@ -526,7 +526,9 @@ func buildProjectsProfileList(
 	profileList := []db.ListProfilesByProjectIDAndLabelRow{}
 
 	listParams := db.ListProfilesByProjectIDAndLabelParams{}
-	listParams.LabelsFromFilter(filter)
+	if err := listParams.LabelsFromFilter(filter); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid label filter: %s", err)
```
evankanderson: If you use `status.Errorf`, the server will return an the status code, but the error message will be sanitized.  If you _mean_ to expose the string to the user, you'll need to use `util.NiceStatus` / `util.UserVisibleError`, as we have a sanitization layer to avoid leaking internal data.

In either case, you might simply want to return a `util.NiceStatus` from `LabelsFromFilter`, and then you can simply `return err` here, rather than needing to wrap it again.

internal/db/domain.go
```diff
@@ -6,8 +6,8 @@ package db
 
 import (
 	"slices"
-	"strings"
 
+	"github.com/mindersec/minder/internal/labels"
```
evankanderson: Linter warning -- we order things as:

* Go stdlib

* Third-party deps

* Minder deps

With a line between each section.

internal/history/models.go
```diff
@@ -375,17 +376,22 @@ func (filter *listEvaluationFilter) ExcludedProfileNames() []string {
...
 
 func (filter *listEvaluationFilter) AddLabel(label string) error {
-	if label == "!*" {
-		return fmt.Errorf("%w: label", ErrInvalidIdentifier)
+	inc, exc, err := labels.ParseLabel(label)
+	if err != nil {
+		if errors.Is(err, labels.ErrInvalidLabel) {
+			return fmt.Errorf("%w: label", ErrInvalidIdentifier)
+		}
+		return err
```
evankanderson: Why do this error transformation?

internal/labels/labels.go
```diff
@@ -0,0 +1,67 @@
...
+
+// ParseLabel parses a single label (without commas) into an include or exclude string.
+// Returns the include label (if any), the exclude label (if any), and an error if validation fails.
+func ParseLabel(label string) (include string, exclude string, err error) {
+	if label == "" {
+		return "", "", nil
+	}
+	if label == "!*" {
+		return "", "", fmt.Errorf("%w: !* is not allowed", ErrInvalidLabel)
+	}
```
evankanderson: It's a little weird that `!*` is not allowed, but e.g. `!--` or `!!` is allowed.  It also seems to be changing the existing behavior.  Is there a reason for this?

internal/labels/labels_test.go
```diff
@@ -0,0 +1,105 @@
...
+package labels
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseLabelFilter(t *testing.T) {
```
evankanderson, line 12: Linter is complaining about missing calls to `t.Parallel()` (which helps tests run faster when you have CPU available).

internal/labels/labels.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return include, exclude, nil
+}
+
+// ParseLabel parses a single label (without commas) into an include or exclude string.
+// Returns the include label (if any), the exclude label (if any), and an error if validation fails.
+func ParseLabel(label string) (include string, exclude string, err error) {
+	if label == "" {
+		return "", "", nil
+	}
```
evankanderson: Do we need this case?

internal/labels/labels.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		if inc != "" {
+			if inc == "*" && len(include) != 0 {
+				return nil, nil, fmt.Errorf("%w: cannot mix * with other labels", ErrInvalidLabel)
+			}
+			if inc != "*" && len(include) == 1 && include[0] == "*" {
+				return nil, nil, fmt.Errorf("%w: cannot mix * with other labels", ErrInvalidLabel)
+			}
+			include = append(include, inc)
```
evankanderson: This logic seems different (stricter) than what we had in `LabelsFromFilter` before.  It looks like we're trying to apply some additional limits on the allowed patterns which were previously valid?

internal/db/domain.go
```diff
@@ -47,28 +47,13 @@ func (r ListProfilesByProjectIDAndLabelRow) GetSelectors() []ProfileSelector {
...
 
 // LabelsFromFilter parses the filter string and populates the IncludeLabels and ExcludeLabels fields
-func (lp *ListProfilesByProjectIDAndLabelParams) LabelsFromFilter(filter string) {
-	// If s does not contain sep and sep is not empty, Split returns a
-	// slice of length 1 whose only element is s. Work around that by
-	// returning early if filter is empty.
-	if filter == "" {
-		return
+func (lp *ListProfilesByProjectIDAndLabelParams) LabelsFromFilter(filter string) error {
+	inc, exc, err := labels.ParseLabelFilter(filter)
```
evankanderson: It would be really nice for either `AddLabel` to use `ParseLabelFilter`, or for this function to split on comma and then pass into something which follows the `AddLabel` pattern.

As it is, the labels in Profile and EvalStatus will refuse to mix `*` with other positive labels, while EvalHistory will accept `*` and positive labels, as long as the `*` is first.

Given the structure of ListEvaluationHistory, I'd probably try to follow the `AddLabel` pattern, which might require post-processing in `IncludedLabels`.

internal/labels/labels_test.go
```diff
@@ -0,0 +1,105 @@
...
+		{
+			name:        "trailing commas",
+			filter:      "foo,,!bar,",
+			expectedInc: []string{"foo"},
+			expectedExc: []string{"bar"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
evankanderson, line 87: And a call to `t.Parallel` here.

internal/controlplane/handlers_evalstatus.go
Gitjay11: I updated this to use util.UserVisibleError so the error message isn't sanitized by the server.

internal/db/domain.go
Gitjay11: Reordered the imports to strictly follow the stdlib -> third-party -> minder convention.

internal/history/models.go
Gitjay11: Agreed, yes the custom wrapping wasn't actually necessary. I've completely removed the error transformation

internal/labels/labels.go
Gitjay11: You're completely right, restricting !* was an unintended side-effect that changed the original behavior. I've removed the strict validation rule so !* will now parse gracefully as an exclusion just like it did before.

internal/labels/labels_test.go
Gitjay11: Added t.Parallel()

internal/labels/labels.go
Gitjay11: No we don't, since the upstream callers (ParseLabelFilter and WithLabel) naturally filter out empty strings. I've completely removed it and simplified the ParseLabel signature to no longer return an error

internal/labels/labels.go
Gitjay11: Good point. I've rolled this back to fully align with the original LabelsFromFilter behavior. The filter now evaluates wildcards using post-processing (collapsing the inclusion list to ["*"]) rather than prematurely throwing an error when users mix wildcards and other positive labels.

internal/db/domain.go
Gitjay11: I completely agree, keeping it consistent across endpoints is much better. I've updated the logic so both code paths now follow the unified pattern. ParseLabelFilter now models the AddLabel base logic and I implemented the post-processing for wildcards directly inside IncludedLabels() so everything is consistent

internal/labels/labels_test.go
Gitjay11: Added t.Parallel() here

Gitjay11:
All the changes are done. I hope you can approve the workflow now, Thanks.

internal/labels/labels.go
```diff
@@ -0,0 +1,54 @@
...
+			exclude = append(exclude, exc)
+		}
+	}
+
+	if starMatched {
+		include = []string{"*"}
+	}
+
+	return include, exclude, nil
+}
```
evankanderson: This no longer needs to return `error`, right?  Just two lists of labels?

internal/labels/labels_test.go
```diff
@@ -0,0 +1,104 @@
...
+		name        string
+		filter      string
+		expectedInc []string
+		expectedExc []string
+	}{
+		{
+			name:        "empty",
+			filter:      "",
+			expectedInc: nil,
+			expectedExc: nil,
```
evankanderson: Minor Go style nit: you don't _need_ to declare every field in a test struct like this; Go will default them to zero values (the zero value of a slice is `nil`, which has 0 length, but is different from an allocated zero-length slice).

internal/labels/labels.go
Gitjay11: yes sir. you're right, removed the error return from both ParseLabelFilter and LabelsFromFilter, and updated all callers accordingly.

internal/labels/labels_test.go
Gitjay11: I got it. Good point, removed all redundant nil field declarations from the test cases.

PR #6245, opened 2026-04-01
test(entities/models): add unit tests for EntityInstance and EntityWi…

dakshhhhh16:
## Summary

Adds a complete unit-test suite for `internal/entities/models`, the package
that defines the core domain structs (`EntityInstance`, `EntityWithProperties`)
shared across the rule-engine, reconcilers, and event-bus.

## Tests Added

| Test | What it covers |
|------|----------------|
| `TestEntityInstance_FieldsPreserved` | All fields set on `EntityInstance` survive a round-trip |
| `TestEntityInstance_ZeroValue` | Zero-value struct is safe to read without panics |
| `TestEntityWithProperties_EntityAccessor` | `.Entity` field returns the embedded `EntityInstance` |
| `TestEntityWithProperties_PropertiesAccessor` | `.Properties` accessor returns the correct property map |
| `TestEntityWithProperties_NilProperties` | `nil` properties map does not panic on access |
| `TestEntityWithProperties_MultipleProperties` | Multiple properties are all stored and retrievable |
| `TestEntityInstance_TypeAndProjectPreserved` | `Type` and `ProjectID` fields survive construction |

## Why This Matters

`EntityInstance` and `EntityWithProperties` are the most widely-referenced
structs in the Minder rule-engine.  Tests here prevent structural regressions
(e.g. a renamed or removed field) from going undetected until runtime.

### DIFF
--- a/internal/entities/models/models_test.go
+++ b/internal/entities/models/models_test.go
@@ -1,0 +1,113 @@
+// SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package models
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/google/uuid"
+
+	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+	"github.com/mindersec/minder/pkg/entities/properties"
+)
+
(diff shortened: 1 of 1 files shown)


evankanderson:
@dakshhhhh16 -- can we close this given that this is producing negligible coverage increase for the additional test cases?

PR #6247, opened 2026-04-01
test(providers): add unit tests for GetProviderClassDefinition and Er…

dakshhhhh16:
## Summary

Adds unit tests for `internal/providers/provider_definitions.go`, the module
that maps provider class names to their canonical definitions and constructs
descriptive `ErrProviderNotFoundBy` error values.

## Tests Added

| Test | What it covers |
|------|----------------|
| `TestGetProviderClassDefinition_GitHubApp` | Looks up the `ProviderClassGithubApp` definition successfully |
| `TestGetProviderClassDefinition_GitHub` | Looks up the `ProviderClassGithub` definition successfully |
| `TestGetProviderClassDefinition_DockerHub` | Looks up the `ProviderClassDockerhub` definition successfully |
| `TestGetProviderClassDefinition_Unknown_ReturnsError` | An unregistered class returns a typed error |
| `TestErrProviderNotFoundBy_Trait` | Error message includes the searched trait |
| `TestErrProviderNotFoundBy_Name` | Error message includes the searched name |
| `TestErrProviderNotFoundBy_ImplementsError` | Returned value satisfies the `error` interface |

## Why This Matters

`GetProviderClassDefinition` is called on every provider lookup; a missing
entry causes a nil-dereference panic deep in the call stack.  These tests
act as a compile-time + runtime guard against accidental removal of a
registered class.

### DIFF
--- a/internal/providers/provider_definitions_test.go
+++ b/internal/providers/provider_definitions_test.go
@@ -1,0 +1,75 @@
+// SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package providers
+
+import (
+	"testing"
+
+	ghclient "github.com/mindersec/minder/internal/providers/github/clients"
+)
+
+func TestGetProviderClassDefinition_Github(t *testing.T) {
+	t.Parallel()
+	def, err := GetProviderClassDefinition(ghclient.Github)
+	if err != nil {
+		t.Fatalf("GetProviderClassDefinition(%q) error = %v", ghclient.Github, err)
+	}
+	if len(def.Traits) == 0 {
+		t.Error("Github definition has no traits")
+	}
+}
+
+func TestGetProviderClassDefinition_GithubApp(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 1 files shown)


evankanderson:
> Actually `GetProviderClassDefinition` is at **75% coverage on `main`**, not 100%. The error path (when an unregistered/unknown class is looked up) has no existing test. I verified this locally:
> 
> ```shell
> # on mindersec:main (cac62d5)
> go test -coverprofile=cov.out -coverpkg=./internal/providers/... ./internal/providers/...
> go tool cover -func=cov.out | grep GetProviderClassDefinition
> # → internal/providers/provider_definitions.go:41: GetProviderClassDefinition  75.0%
> 
> # on this branch (bf1ca50)
> # → internal/providers/provider_definitions.go:41: GetProviderClassDefinition  100.0%
> ```
> 
> Happy to close/correct this if I have gone wrong anywhere :)

The codepath has coverage from tests in other modules.  Given that we have coveralls to collect _aggregate_ test coverage, it seem silly not to use that coverage.  (In this case, I think it's from `TestCreateProviderFailures` in `internal/controlplane`, but I haven't tracked it down for sure because it didn't seem necessary.)

PR #6244, opened 2026-04-01
:sparkles: Improve CLI help for repo command with examples

sachin9058:
# Summary

This PR improves the CLI help output for the `repo` command.

Previously, the help text had minimal description and lacked usage examples, which could make it difficult for new users to understand how to use the command effectively.

This change:
- Enhances the Short and Long descriptions for clarity
- Adds practical usage examples (list, add, delete)
- Improves formatting and consistency of the help output

These improvements make the CLI more user-friendly and improve onboarding for new contributors and users.

Fixes #NONE

---

# Testing

The changes were tested locally by rebuilding the CLI and verifying the updated help output.

Steps:
1. Build the CLI:
   ```bash
   make build-minder-cli
   ./bin/minder repo --help

PR #6249, opened 2026-04-02
Add pre-validation for repo register command to prevent unnecessary auth flow

sachin9058:
# Summary

This PR adds pre-validation for the `repo register` command to ensure required inputs are provided before initiating authentication.

Previously, running the command without required flags (e.g., `--provider` or `--name`) would trigger the authentication flow unnecessarily, leading to poor user experience.

This change:
- Validates required inputs (`--provider`, and either `--name` or `--all`)
- Runs validation before the gRPC/auth wrapper
- Provides clear error messages with examples

Fixes #NONE

---

# Testing

1. Build the CLI:
   make build-minder-cli

2. Run:
   ./bin/minder repo register  
   → Verify no auth flow starts and error is shown

3. Run:
   ./bin/minder repo register --name owner/repo  
   → Verify error for missing `--provider`

4. Run:
   ./bin/minder repo register --name owner/repo --provider github  
   → Verify command works normally


cmd/cli/app/repo/repo_register.go
```diff
@@ -25,7 +25,15 @@ var repoRegisterCmd = &cobra.Command{
 	Use:   "register",
 	Short: "Register a repository",
 	Long:  `The repo register subcommand is used to register a repo within Minder.`,
-	RunE:  cli.GRPCClientWrapRunE(RegisterCmd),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// 🔥 Validate BEFORE auth/login starts
+		if err := ValidateRepoInput(cmd, true); err != nil {
+			return err
+		}
```
krrish175-byte, line 39: The validation function `ValidateRepoInput` is called twice. Once here

cmd/cli/app/repo/repo_register.go
```diff
@@ -43,6 +51,10 @@ func RegisterCmd(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc
 	// See https://github.com/spf13/cobra/issues/340#issuecomment-374617413
 	cmd.SilenceUsage = true
 
+	if err := ValidateRepoInput(cmd, true); err != nil {
+		return err
+	}
```
krrish175-byte: And once here

cmd/cli/app/repo/validation.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+// ValidateRepoInput validates repo command inputs
+func ValidateRepoInput(cmd *cobra.Command, requireName bool) error {
+	provider, _ := cmd.Flags().GetString("provider")
+	names, _ := cmd.Flags().GetStringSlice("name")
```
krrish175-byte: Looking at the original repo_register.go, I notice the code uses viper.GetString("provider") and viper.GetStringSlice("name"), not cmd.Flags().

However, the new validation function uses:
provider, _ := cmd.Flags().GetString("provider")
names, _ := cmd.Flags().GetStringSlice("name")

This might work, but there's a mismatch in how the original code retrieves values (via Viper config) vs the validation (via Cobra flags). The validation should be consistent with how RegisterCmd() actually reads these values.

krrish175-byte:
Recommendation:

- Remove the duplicate validation from inside RegisterCmd() — keep only the wrapper validation
- Ensure flag retrieval consistency — verify that Cobra flags and Viper config are properly synced (they usually are via CLI binding)
- Test the --all case — specifically test: minder repo register --all without --provider to ensure your validation doesn't break the interactive all-repos flow

The intent is correct, but the implementation has redundancy. Functional but needs cleanup. Rest LGTM.

cmd/cli/app/repo/validation.go
evankanderson: Unfortunately, we need to use `viper.GetString()` version, as that will also read from either a `config.yaml` or from `MINDER_....` environment variables.

Interestingly, we don't have a good general pattern for validating command line flags before calling `GRPCClientWrapRunE`, so this problem probably happens for a lot of commands.  I wonder if there's something more systemic we could do with e.g. `ValidArgsFunction` or `PreRun` hooks.

cmd/cli/app/repo/validation.go
evankanderson, line 1: Why create a new file for this?  It feels like it belongs in `repo_register.go`.

cmd/cli/app/repo/repo.go
evankanderson, line 1: This seems to be a commit from #6244.  Slight preference for not including it here unless it's needed for the other changes to work.

evankanderson: (i.e. base this branch off `upstream/main`, not `improve-repo-cli-help`)

cmd/cli/app/repo/validation.go
```diff
@@ -0,0 +1,38 @@
+package repo
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+// ValidateRepoInput validates repo command inputs
+func ValidateRepoInput(cmd *cobra.Command, requireName bool) error {
```
evankanderson: The `requireName` argument seems to be `true` in all callers.  Can we remove it?

cmd/cli/app/repo/validation.go
```diff
@@ -0,0 +1,38 @@
...
+	provider, _ := cmd.Flags().GetString("provider")
+	names, _ := cmd.Flags().GetStringSlice("name")
+	all, _ := cmd.Flags().GetBool("all")
+
+	command := cmd.CommandPath()
+
+	if provider == "" && !all {
```
evankanderson: Does `--all` work with no provider?  I thought that it registered all the repos available for that provider (including by changing the repo configuration to auto-enroll future repos available to that app).

evankanderson (commented):
One extra wrinkle here on the validation -- I seem to recall that in some cases, the API will automatically figure out which provider to use, particularly if you only have one provider registered.  (But this requires having enough authentication to talk to the server.)

In general, I have a preference for computers figuring out the arguments that they can without making humans do extra data entry (for example, automatically determining the provider for a repository if there's exactly one choice).

cmd/cli/app/repo/repo_register.go
```diff
@@ -25,7 +25,24 @@ var repoRegisterCmd = &cobra.Command{
 	Use:   "register",
 	Short: "Register a repository",
 	Long:  `The repo register subcommand is used to register a repo within Minder.`,
-	RunE:  cli.GRPCClientWrapRunE(RegisterCmd),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// Basic validation before auth flow
+
+		name, _ := cmd.Flags().GetStringSlice("name")
+		all, _ := cmd.Flags().GetBool("all")
```
evankanderson: Use the `viper.Get<Type>` methods over `cmd.Flags` -- as mentioned, these include the ability to pull from environment variables in some cases.

cmd/cli/app/repo/repo_register.go
```diff
@@ -25,7 +25,24 @@ var repoRegisterCmd = &cobra.Command{
 	Use:   "register",
 	Short: "Register a repository",
 	Long:  `The repo register subcommand is used to register a repo within Minder.`,
-	RunE:  cli.GRPCClientWrapRunE(RegisterCmd),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// Basic validation before auth flow
```
evankanderson: Given that this is argument validation, can you put it in `PreRunE`, instead of adding it to the start of `RunE`?  This is recommended by Cobra, though we don't always do it: https://cobra.dev/docs/how-to-guides/working-with-flags/#validate-related-flags-prerune

sachin9058: Updated the validation to use `viper.Get*` and moved it into `PreRunE` so it runs before the auth flow.

Tested locally — the command now fails early for missing/invalid input and only proceeds to auth when inputs are valid.

Let me know if this looks good!

evankanderson (dismissed):

cmd/cli/app/repo/repo_register.go
```diff
@@ -25,7 +25,23 @@ var repoRegisterCmd = &cobra.Command{
 	Use:   "register",
 	Short: "Register a repository",
 	Long:  `The repo register subcommand is used to register a repo within Minder.`,
-	RunE:  cli.GRPCClientWrapRunE(RegisterCmd),
+
+	PreRunE: func(cmd *cobra.Command, args []string) error {
```
evankanderson: Lint is complaining about unused parameters here.

sachin9058: Good catch.. Fixed the unused parameters here

sachin9058:
@evankanderson Good catch fixed the unused parameters in `PreRunE`.

evankanderson (approved):

Issue #6250, opened 2026-04-02
Hardcoded "TODO" subject in CloudEvents NATS driver


Amit-Matth:
### Summary
In the NATS eventing driver (`internal/events/nats/natschannel.go`), the `subject` field for outgoing CloudEvents is currently hardcoded to the literal string `"TODO"`. This field should instead uniquely identify the specific entity (such as a repository, project, or artifact) that the event is about.

### Impact
- **Event Filtering**: NATS JetStream relies on subjects for efficient event routing and filtering. Hardcoding the subject to `"TODO"` makes it impossible for consumers to efficiently filter events.
- **Spec Compliance**: It violates the CloudEvents specification, which expects the subject to refer to a specific resource.

### To Reproduce

1. Open the file `internal/events/nats/natschannel.go`.
2. Locate the `sendEvent` function (around line 203).
3. Observe line 209: `event.SetSubject("TODO")`.
4. Note the developer comment: `// This *should* represent the entity, but we don't have a standard field for it yet.`

### What version are you using?

Latest main branch (v0.1.0)

Deepesh123455:
Hey @TRIVENI206, go for it! I was trying to do a broader refactor, but if you want to tackle the minimal fix, feel free to take this issue.

evankanderson:
#6369 needs merge conflicts resolved.

PR #6251, opened 2026-04-02
events: replace hardcoded NATS subject with dynamic entity metadata

Deepesh123455:
Summary
This PR addresses a technical debt item in the NATS eventing driver where the CloudEvents subject field was hardcoded to the literal string "TODO".

Changes:

Centralized Metadata Constants: Migrated internal metadata keys (e.g., entity_id, project_id, repository_id) from internal/engine/entities to a public package pkg/eventer/constants. This avoids circular dependencies while allowing the NATS driver to access these keys.
Dynamic Subject Population: Updated the NATS sendEvent logic to inspect message metadata and set the CloudEvent subject based on a priority list of identifiers:
entity_id (Primary)
repository_id / artifact_id / pull_request_id (Legacy Fallbacks)
project_id (General Hierarchy)
"minder" (Safe default)
Backward Compatibility: Maintained deprecated aliases in the original entities package to ensure no breakage across the codebase during the transition.
Rationale: NATS JetStream relies on the subject field for efficient server-side routing and filtering. Hardcoding this value to "TODO" forced clients to ingest all events and perform client-side filtering, which is highly inefficient at scale and violates the CloudEvents specification for resource identification.

Fixes: #6250

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -11,6 +11,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      - name: Make results directory
+        run: mkdir -p ../results
+
       - name: Code Security Scan
         uses: aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1 # v0.35.0
         with:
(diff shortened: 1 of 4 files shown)


internal/engine/entities/entity_event.go
```diff
@@ -18,18 +18,10 @@ import (
...
-// the information it gathers.
+// EntityInfoWrapper assumes the following metadata keys are present:
 //
-// It's also able to read the message.Message that contains a payload
-// with a protobuf message that's specific to the entity type.
-//
-// It also assumes the following metadata keys are present:
-//
-// - EntityTypeEventKey - entity_type
-// - EntityIDEventKey - entity_id
```
evankanderson, line 32: Why remove the rest of this documentation?

internal/engine/entities/entity_event.go
```diff
@@ -41,34 +33,40 @@ type EntityInfoWrapper struct {
 	ActionEvent   string
 }
 
+// NewEntityInfoWrapper creates a new EntityInfoWrapper
+func NewEntityInfoWrapper() *EntityInfoWrapper {
+	return &EntityInfoWrapper{
+		OwnershipData: make(map[string]string),
+	}
+}
+
```
evankanderson, line 42: Why move this code?

internal/engine/entities/entity_event.go
```diff
@@ -41,34 +33,40 @@ type EntityInfoWrapper struct {
...
 const (
 	// EntityTypeEventKey is the key for the entity type
-	EntityTypeEventKey = "entity_type"
+	// Deprecated: use constants.EntityTypeEventKey
+	EntityTypeEventKey = constants.EntityTypeEventKey
```
evankanderson, line 46: Why move these constants (and if we do need to move them, why not actually rewrite the reference sites, rather than just adding more aliases through the codebase)?

internal/events/nats/natschannel.go
```diff
@@ -206,7 +207,21 @@ func sendEvent(
...
+		event.SetSubject(repoID)
+	} else if artifactID, ok := msg.Metadata[constants.ArtifactIDEventKey]; ok {
+		event.SetSubject(artifactID)
+	} else if prID, ok := msg.Metadata[constants.PullRequestIDEventKey]; ok {
+		event.SetSubject(prID)
+	} else if projectID, ok := msg.Metadata[constants.ProjectIDEventKey]; ok {
+		event.SetSubject(projectID)
+	} else {
+		event.SetSubject("minder")
+	}
```
evankanderson, line 224: This doesn't feel right.  I think most usage may have migrated to using `entity_id`; I'd look at whether we could retire some of the other metadata fields rather than putting a big "switch / case" statement in here using if/else.

It also seems like this doesn't affect the tests at all, which seems surprising?

evankanderson (commented):
This seems to mix a lot of code movement with some meaningful changes, which makes it harder to review the meaningful change (because there's a lot of non-meaningful change).  In general, prefer to separate out refactoring PRs from behavior-change PRs -- refactoring PRs like #6252 are _easy_ to review because one simply needs to look for any behavior changes to flag them.

When using `git`, it's fairly easy to start with the refactoring PR in one branch, and then start the "meaningful change" branch based on the "refactoring" branch.  That should make the merge relatively painless if you submit both PRs before the refactoring one has been merged (so you can keep your mental context window open while working on the change).

Deepesh123455:
Thanks for the thorough review, @evankanderson! Makes complete sense. I'm going to close this PR and split the work into two separate PRs: one for the pure refactoring/constant updates (updating the reference sites directly instead of aliasing), and a second one for the dynamic NATS subject logic where I'll look into simplifying that if/else block and adding the missing test coverage

PR #6252, opened 2026-04-02
Move internal/engine/errors to pkg/engine/errors to reduce coupling

sachin9058:
# Summary

This PR moves the `internal/engine/errors` package to `pkg/engine/errors` as a first step toward reducing coupling between the engine and internal packages.

Currently, parts of the engine depend on `internal/*` packages, which makes it difficult to isolate the engine logic and eventually separate it into a smaller, independent module. Since `internal/` packages are not meant to be reused outside their parent module, this creates a limitation for modularization.

By moving the `errors` package into `pkg/engine`, we make it accessible as a reusable component and remove one dependency on `internal/engine`. This is intended as a small, incremental step toward decoupling the engine and improving the overall structure of the codebase.

Changes included:

* Moved `internal/engine/errors` → `pkg/engine/errors`
* Updated all import paths across the repository
* Verified that the project builds successfully after the change

This PR does not introduce any functional changes and is purely a structural refactor.

Fixes #NONE

---

# Testing

The following steps were used to verify the changes:

1. Build the CLI:

   ```bash
   make build-minder-cli
   ```

2. Run a basic command:

   ```bash
   ./bin/minder repo register --all
   ```

3. Confirmed that:

   * The project builds without errors
   * The CLI flow (including authentication) works as expected
   * No runtime issues were observed

No behavior changes are expected from this PR.


pkg/engine/errors/errors_test.go
```diff
@@ -1,3 +1,4 @@
+
```
evankanderson: I think this leading newline might cause you to get in trouble with linting, but I'm not sure.

pkg/engine/errors/errors.go
evankanderson, line 1: I think this is fine for transition, but I imagine you'll want to cut this file such that it no longer has a dependency on `internal/db` by moving the type translation functions elsewhere.

evankanderson (approved):

PR #6253, opened 2026-04-02
Update roadmap for 2026 (and possibly beyond)

evankanderson:
# Summary

Per suggestions in Slack, updating the Minder roadmap to call out key areas for design and implementation will help new contributors understand where to focus their efforts.  I've been keeping a lot of this in my head / in meeting notes or Slack messages that may have aged out, so it's helpful to get the content into a more visible format.

# Testing

N/A - this is documentation.


docs/docs/about/roadmap.md
```diff
@@ -11,70 +11,252 @@ learn about what features we're working on, what stage they're in, and when we
...
+consumption by remediations and alerts. We can extend this functionality to
+enable a number of additional capabilities with a minimal code footprint.
+
+Some of the suggested capabilities:
+
+- **Inventory rules**: Rather than pass/fail assessment of repository status,
+  leverage rule evaluation to collect (for example) all the actions used by a
+  repository or all the container images used in all Dockerfiles. Minder rule
+  output could then be used to import data into an external software inventory
+  control system.
```
Vyom-Yadav, line 73: > leverage rule evaluation to collect

I don't think inventory creation should depend on some edge trigger. I think it shouldn't leverage rule evaluation and the inventory should be created when that particular `inventory-creation-request` is triggered. Relying on some other trigger seems to me like it could miss continuous compliance that inventory _likely_ aims towards.

docs/docs/about/roadmap.md
```diff
@@ -11,70 +11,252 @@ learn about what features we're working on, what stage they're in, and when we
...
+
+- **Rule-driven child entities**: Leveraging inventory rules, we could add a
+  remediation which supports managing (upsert / delete) Minder entities detected
+  through an inventory rule. This could even replace some of the built-in
+  lifecycle rules (e.g. for PRs).
```
Vyom-Yadav, line 78: That is good idea. I'm thinking how would the API for such a thing look like? Currently rule creation is fairly simple as you know the entities. BUT if rules create custom entities, then it would be difficult to track the entities themselves. 

When I start thinking about this, my head goes towards CRDs in some sense (we can define some, let users define others). A source SBOM for a project (for example) as an entity should not be hidden in any rule. Its needed to be defined explicitly.

docs/docs/about/roadmap.md
```diff
@@ -11,70 +11,252 @@ learn about what features we're working on, what stage they're in, and when we
...
+
+- **External remediation**: Currently, the Minder server uses a GitHub App with
+  high levels of permissions in order to perform remediations. If the data
+  needed for remediations (remediation definition, context, and rule output) is
+  available through the API, it should be possible to build an external system
+  to query rule status and perform remediation using a more powerful token held
+  _outside_ of the Minder server. This would reduce the risk / "blast radius" of
+  a Minder server compromise, and allow users to adopt hosted Minder scanning
+  while keeping powerful GitHub keys under local supervision.
```
Vyom-Yadav, line 87: Is the goal here to separate read/write? Github app has just read and for remediation token from outside is supplied for write? How would that work? Which system will feed minder the tokens it requires for remediation? (I'm assuming it's not a human dispatching tokens)

docs/docs/about/roadmap.md
```diff
@@ -11,70 +11,252 @@ learn about what features we're working on, what stage they're in, and when we
...
+need design and discussion.
+
+### Improve Rule Testing Infrastructure
+
+**Status**: Subject of an
+[LFX Mentorship Project, Summer 2026](https://mentorship.lfx.linuxfoundation.org/project/40b209ce-c759-4648-9d83-31db4ba1d481)
+
+[Full proposal in Google Docs](https://docs.google.com/document/d/1laRud0GSPqVg_rZ3ahD8GfRJYY2Bf-JL66nEXGZJ4kk/edit).
+
+The primary goals of this 12-week engagement are to:
```
Vyom-Yadav, line 100: This reminds me of https://github.com/kudobuilder/kuttl in some sense. I would suggest anyone working on this to take a look at how kuttl works. Last time I worked with Kyverno, I used kuttl extensively to test policies over there: https://github.com/kyverno/policies/issues/546. Now they seem to use chainsaw though, worth taking a look again: https://kyverno.github.io/chainsaw/0.2.3/guides/kuttl-migration/

docs/docs/about/roadmap.md
```diff
@@ -11,70 +11,252 @@ learn about what features we're working on, what stage they're in, and when we
...
+1. **Documentation and UX**: Create documentation for the new tooling and
+   address potential client-side UX issues related to testing.
+
+### Improve Flexibility of Pull Request Rules
+
+**Status**: Discussion in
+[#4452](https://github.com/mindersec/minder/issues/4452), not well-defined yet.
+
+There are several `pull_request` entity evaluators (e.g. OSV, Trusty, Frizbee)
+which perform special commenting operations either during rule evaluation or as
```
Vyom-Yadav, line 125: Isn't Trusty eol now?

docs/docs/about/roadmap.md
```diff
@@ -11,70 +11,252 @@ learn about what features we're working on, what stage they're in, and when we
...
+Minder has _some_ support for [GitLab Cloud](https://gitlab.com/), but setup is
+neither well-documented nor well-tested. Additionally, Minder does not currently
+support any of the following Git Forges:
+
+- **[Forgejo](https://forgejo.org/)**: not started; this should be designed to
+  work with both Codeberg and self-hosted Forgejo, which may require additional
+  work on Provider infrastructure
+- **GitHub Enterprise**: this work would probably need to be funded
+- **GitLab (on-premise)**: again, this might need to be funded
+
```
Vyom-Yadav, line 171: May I introduce you to the gorgeous: https://launchpad.net/

docs/docs/about/roadmap.md
```diff
@@ -11,70 +11,252 @@ learn about what features we're working on, what stage they're in, and when we
...
+In addition to adding support for Git Forges (where the Minder entity model is
+well-tested), Minder has some support for artifact repositories (primarily OCI
+repositories such as GHCR and DockerHub). This should be expanded, possibly
+following the ["rule-driven child entities"](#improve-rule-output-handling) work
+to enable automatic generation of dependent entities given a known parent.
```
Vyom-Yadav, line 180: I think expanding to support more container registries should be a separate item. Those I think carry higher value than some other forges at the moment.

docs/docs/about/roadmap.md
```diff
@@ -11,70 +11,252 @@ learn about what features we're working on, what stage they're in, and when we
...
+
+Minder datasources support
+[using the `providerAuth` field](../understand/providers.md#defining-a-data-source)
+to authenticate rules to the API which manages the entity, but do not support
+any form of other authentication (for example, to a third-party or custom API
+endpoint to collect additional data or recommendations). Minder should consider
+defining identities for executing rule types (possibly based on the enclosing
+project), and using [OIDC](https://openid.net/developers/how-connect-works/) or
+[SPIFFE](https://spiffe.io/) to provide a unique identity for each Minder
+project or rule executing in the project.
```
Vyom-Yadav: Authenticating to external data sources for example some hyproprietary data source containing known exploits, malware, etc. (for some scanning?)

I'm not sure what is the goal here. But one layer of auth using provider, then another layer for some third party using oidc, etc. seems confusing. 

Data sources in general should not have things to authenticate to another data source is what I mean. 

`A` data source should not have auth details for `B` data source.

`B` should be a separate data source which can be queried using an API by `A`.

docs/docs/about/roadmap.md
```diff
@@ -11,70 +11,252 @@ learn about what features we're working on, what stage they're in, and when we
...
+OpenFGA provides an underlying sharing and relationship mechanism for projects
+to be able to share _select_ compliance reports either with the world or with
+select audiences (such as project sponsors). Currently, Minder does not track
+specific resources which could be used to grant "badge-read" or "audit"-type
+permissions in OpenFGA. The solution would probably need to support at least:
+
+- Share all reports (policy evaluation results) for an entity.
+- Share all reports (policy evaluation results) for an entity and its
+  dependents.
+- Share one specific report (policy evaluation results) for an entity.
```
Vyom-Yadav: I really like this one. This also extends to a continuous compliance dashboard. SLAs and stuff may be the next thing.

Vyom-Yadav (commented):
Some thoughts, but a solid roadmap (also sorry for the delay)

docs/docs/about/roadmap.md
evankanderson: This is a roadmap thing, so I don't want to completely define the format, but using the `output` support, we might do something like:

```yaml
  remediate:
    type: entity
    # Expects an array of new entries which replace any existing ones created by this rule
    entity:
      # Loads the current build definitions from `output[builds]`, where the fields are the
      # defining properties
      # example value: [
      #   {"name": "dev", "server": "dev-ci"},
      #   {"name": "uat", "server": "prod"},
      #   {"name": "prod", "server": "prod"}
      # ]
      - key: builds  # JSONPath expression
        types:
        - build
        # Currently always uses the same provider; might be adjustable later
      - key: images
        types:
        - artifacts
```

docs/docs/about/roadmap.md
evankanderson: Given that a number of supply chain attacks have specifically targeted security providers, it would be great to be able to have the main Minder app not have any exciting "write" permissions in the repo.  Instead, you could have a separate "remediation" app which had the exciting permissions.  This could either live co-located with Minder (ease of use, no additional protection), or you could use your own GitHub app running outside the Minder environment.

For the latter, you could use something like a periodic workflow that wakes up and queries the Minder API for new necessary remediations.  For each remediation, read it from the ruletype definition and fill in the output content, and then apply the result.  You could either run this as a cronjob, or actually have an interactive UI for reviewing the recommended remediations before applying them.  You could even mix the two -- automate e.g. PR creation and comments, but require human review of anything that would change API settings like branch protection or permitted actions policy.

docs/docs/about/roadmap.md
evankanderson: Yeah, I'm definitely interested in taking inspiration from elsewhere if there are good tools.  The current `rule_test` stuff is very much "a fire engine we built on the way to the fire" to quote one of my mentors.

docs/docs/about/roadmap.md
evankanderson: Yeah, we need to actually retire it.  In general, I'd like to retire all of these in favor of general-purpose rule evaluation that could call out to the appropriate sources and suggest things without needing Minder core to change at all.

docs/docs/about/roadmap.md
evankanderson: Sorry, I meant that someone would need to pay someone to work on it.  Though maybe it would be Canonical paying you to work on better on-prem GitLab support.  😁

docs/docs/about/roadmap.md
evankanderson: Good point.  We can add that as an additional item under "improved provider support"

docs/docs/about/roadmap.md
evankanderson: The idea here is that a rule `X` may want to call a data source `A`.  The data source `A` might be licensed or otherwise access-controlled.  (Maybe it's simply using LLM inference, and you want to control costs...)

So, you need some way to ensure that it's only rule `X` from project `P` that's calling the API.  OIDC (ala GitHub Actions) feels like a good way to express this.

We already have a special carve out for "use the current provider's identity" because it was really helpful for GitHub, but I'd prefer not to build a general-purpose secrets store into Minder, because that feels icky.  Also, something like OIDC could be used with OpenBao / Vault / OctoSTS / various cloud providers to access their existing secrets stores (which have auditing, etc), so that feels like it would be an escape hatch if you needed to use something like a shared secret (API key, etc) for auth to data source `A`.

docs/docs/about/roadmap.md
evankanderson: Yes, that's the idea.  We need some way to permission _something_ (not sure what it is) to be able to say "this one thing is public" without making the whole project public.  Part of the work here is designing what _something_ is.

evankanderson (commented):
Thanks for the review!  Do you want me to fill out the OCI repository providers section in more detail now, or in a follow-up PR?

docs/docs/about/roadmap.md
Vyom-Yadav: Both valid points, I think minder should also output a formally verifiable audit trail alongside.

docs/docs/about/roadmap.md
Vyom-Yadav: > a fire engine we built on the way to the fire

Saved.

docs/docs/about/roadmap.md
Vyom-Yadav: Generic JSONPath based evaluation engines which can fetch from anywhere. But they can be abstracted into something like a CRD(ish) as well.

docs/docs/about/roadmap.md
Vyom-Yadav: Yes, I will let you convince them.

Vyom-Yadav:
@evankanderson 

> Thanks for the review! Do you want me to fill out the OCI repository providers section in more detail now, or in a follow-up PR?

I think followup is fine, this can be merged. Thanks for writing this down, great ideas for future LFXes or future volunteers.

docs/docs/about/roadmap.md
evankanderson: Rule evaluation is simply a policy mechanism -- it can be edge-triggered or timer-triggered (e.g. via reminder).  I think the particular goal of the user request here was to have an inventory of either Dockerfiles for images (if you have a repo which builds a lot of images) or of e.g. Claude skills, which could then be used to generate catalogs, etc.

docs/docs/about/roadmap.md
evankanderson: You mean using something like a transparency log?  I've considered that, but it seems like a pretty big investment, so I'd want to see an actual user request it.  Right now, you can use the provider's audit system to track Minder actions (e.g. every Minder GitHub remediation shows up in the GitHub org audit log).

Formal verification is also a big ask, as you'd need some way to tie each usage of a write credential to a specific remediation in such a way that the system couldn't re-use those credentials to issue additional API writes.

docs/docs/about/roadmap.md
evankanderson: I added an item for "Expand OCI Registry Provider Coverage" as well as "Other Provider Coverage" to start the thoughts on managing NPM, PyPI, etc publishing.

docs/docs/about/roadmap.md
evankanderson: I meant more along the lines of using the Rego (or JQ) evaluator.  Rego seemed most promising, since it could leverage OSV or other vulnerabilities databases using the DataSource support.

evankanderson (commented):
Updated with your feedback, @Vyom-Yadav ; I'll need you to approve this before it can merge.

evankanderson:
@Vyom-Yadav -- can you approve this?

Issue #6254, opened 2026-04-02
Link output tables to API for include_outputs


evankanderson:
### Please describe the enhancement

Now that we have tables storing evaluation outputs (#6192 ) and an API for fetching those outputs (#6193 ), we need to connect the two.

### Solution Proposal

Ideally, we should be able to either re-use the existing query or select between two pre-built queries in fetching the rule evaluation output.  In particular, we should avoid looping over each evaluation result in order to get the output -- that will be intensive on both the Minder server and the database, and probably reduce parallelism opportunities.

dakshhhhh16:
Should this issue be marked as closed now?

evankanderson:
This was fixed in #6255

PR #6255, opened 2026-04-03
feat(inventory): wire include_outputs for GetEvaluationHistory endpoint

dakshhhhh16:
**Part of #6254**

## Summary

Wires up the `include_outputs` request field on the `GetEvaluationHistory` RPC so it actually returns the stored structured output from the `evaluation_outputs` table.

Before this change, the `include_outputs` field existed in the proto definition but was completely ignored -- the response never included the evaluation output even when requested. Now, when a client sets `include_outputs=true`, the handler fetches the corresponding row from `evaluation_outputs` and returns it as a `google.protobuf.Struct` in the `status.output` field.

## How it works

This is the simplest of the three endpoints since it deals with a single evaluation record:

1. After fetching the evaluation history row, check `in.GetIncludeOutputs()`
2. If true, call `GetEvaluationOutput(ctx, evaluationID)` (existing SQLC query)
3. Convert the stored JSONB to a `protobuf.Struct` and set it on the response
4. If no output exists for that evaluation (common for older records), gracefully return `nil` -- no error

No new SQL queries, no schema changes, no proto changes.

## How I tested this

- `go build ./...` -- clean
- `golangci-lint run ./internal/controlplane/` -- 0 issues

### DIFF
--- a/internal/controlplane/handlers_evalstatus.go
+++ b/internal/controlplane/handlers_evalstatus.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"database/sql"
 	"encoding/base64"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -16,6 +15,7 @@ import (
 	"github.com/rs/zerolog"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
@@ -98,7 +98,10 @@ func (s *Server) GetEvaluationHistory(
 		if err != nil && !errors.Is(err, sql.ErrNoRows) {
 			zerolog.Ctx(ctx).Error().Err(err).Msg("error retrieving evaluation output")
 		} else if err == nil {
-			pbEval.Status.Output = rawJSONToProtobufValue(ctx, output.Output.RawMessage)
+			pbEval.Status.Output = &structpb.Value{}
+			if err := protojson.Unmarshal(output.Output.RawMessage, pbEval.Status.Output); err != nil {
+				zerolog.Ctx(ctx).Error().Err(err).Msg("Unable to unmarshal rule output")
+			}
 		}
 	}
 
@@ -768,26 +771,4 @@ func dbSeverityToSeverity(dbSev db.Severity) (*minderv1.Severity, error) {
 	}
 
 	return severity, nil
-}
-
-// rawJSONToProtobufValue converts a JSON raw message to a protobuf Value.
-// Returns nil if the input is nil or empty.
-func rawJSONToProtobufValue(ctx context.Context, jsonData json.RawMessage) *structpb.Value {
-	if len(jsonData) == 0 {
-		return nil
-	}
-
-	var raw any
-	if err := json.Unmarshal(jsonData, &raw); err != nil {
-		zerolog.Ctx(ctx).Error().Err(err).Msg("failed to unmarshal output JSON")
-		return nil
-	}
-
-	pbVal, err := structpb.NewValue(raw)
-	if err != nil {
-		zerolog.Ctx(ctx).Error().Err(err).Msg("failed to convert output to protobuf Value")
-		return nil
-	}
-
-	return pbVal
 }
--- a/internal/controlplane/handlers_evalstatus_test.go
+++ b/internal/controlplane/handlers_evalstatus_test.go
@@ -4,15 +4,23 @@
 package controlplane
 
 import (
+	"context"
 	"database/sql"
+	"encoding/json"
 	"testing"
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/sqlc-dev/pqtype"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/structpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
+	mockdb "github.com/mindersec/minder/database/mock"
 	"github.com/mindersec/minder/internal/db"
+	"github.com/mindersec/minder/internal/engine/engcontext"
 	entmodels "github.com/mindersec/minder/internal/entities/models"
 	"github.com/mindersec/minder/internal/history"
 	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
@@ -353,6 +361,9 @@ func TestFromEvaluationHistoryRows(t *testing.T) {
 						item.Remediation.Details,
 					)
 				}
+
+				// Verify that existing history rows do not set output
+				require.Nil(t, item.Status.Output)
 			}
 		})
 	}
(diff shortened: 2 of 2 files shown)


internal/controlplane/handlers_evalstatus.go
```diff
@@ -91,6 +93,15 @@ func (s *Server) GetEvaluationHistory(
 		Remediation: getRemediation(eval.RemediationStatus, eval.RemediationDetails.String),
 	}
 
+	if in.GetIncludeOutputs() {
+		output, err := s.store.GetEvaluationOutput(ctx, eval.EvaluationID)
+		if err != nil && !errors.Is(err, sql.ErrNoRows) {
+			zerolog.Ctx(ctx).Error().Err(err).Msg("error retrieving evaluation output")
+		} else if err == nil {
```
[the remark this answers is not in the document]
evankanderson: This is a bad suggestion, Copilot.  Returning an error instead of _any_ historical results is worse than including historical results which didn't capture the rule output.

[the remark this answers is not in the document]
evankanderson: Yes, we should have some tests for the new functionality.  At a minimum:

* `include_outputs` set, no data
* `include_outputs` set, at least some JSON data.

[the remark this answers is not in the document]
evankanderson: Ideally, given that we have several different types of JSON data, I'd also test the database containing:

* JSON object
* JSON array
* JSON string

See https://github.com/mindersec/minder/pull/6211#issuecomment-4179921983 for examples of each.

internal/controlplane/handlers_evalstatus.go
```diff
@@ -758,3 +769,25 @@ func dbSeverityToSeverity(dbSev db.Severity) (*minderv1.Severity, error) {
...
+	var data map[string]any
+	if err := json.Unmarshal(jsonData, &data); err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("failed to unmarshal output JSON")
+		return nil
+	}
+
+	pbStruct, err := structpb.NewStruct(data)
+	if err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("failed to convert output to protobuf Struct")
+		return nil
```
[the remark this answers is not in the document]
evankanderson: The Copilot analysis here is accurate, as I mentioned in my comment above.

At least array-of-objects and single object feel like they have meaning.  I can see two options:

1. Change the type of the current `output` field, and ignore the Buf compatibility complaints, because the field never worked.  Alternatively, we _could_ use a new field number, but I don't think any code which calls the API has escaped containment yet.
2. Store array/scalar values under a `"value"` key, and force the output to be an object.

I don't think I have a strong preference, but I've thought about it for all of 30 seconds, and the current API has been a bit rushed (which leads to mistakes like this!).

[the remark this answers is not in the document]
dakshhhhh16: That makes sense. I'm leaning towards Option 1 (changing the field type to google.protobuf.Value) since as you said nobody is consuming this field yet and it gives us a clean representation for arrays, strings, and objects without any wrapping hacks. Happy to go with Option 2 though if you'd prefer keeping Struct and just stuffing non-object shapes under a "value" key.

internal/controlplane/handlers_evalstatus.go
```diff
@@ -91,6 +93,15 @@ func (s *Server) GetEvaluationHistory(
 		Remediation: getRemediation(eval.RemediationStatus, eval.RemediationDetails.String),
 	}
 
+	if in.GetIncludeOutputs() {
+		output, err := s.store.GetEvaluationOutput(ctx, eval.EvaluationID)
+		if err != nil && !errors.Is(err, sql.ErrNoRows) {
+			zerolog.Ctx(ctx).Error().Err(err).Msg("error retrieving evaluation output")
+		} else if err == nil {
+			pbEval.Status.Output = rawJSONToProtobufValue(ctx, output.Output.RawMessage)
```
evankanderson: What do you think about replacing `rawJSONToProtobufValue` with [`protojson.Unmarshal`](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#Unmarshal)?


```suggestion
			err := protojson.Unmarshal(output.Output.RawMessage, &pbEval.Status.Output)
			if err != nil {
				zerolog.Ctx(ctx).Error().Err(err).Msg("Unable to unmarshal rule output")
			}
```

dakshhhhh16: Let me just check

dakshhhhh16: Made the changes!

internal/controlplane/handlers_evalstatus_test.go
```diff
@@ -378,3 +389,119 @@ func nullRemediationStatusTypesSuccess() db.NullRemediationStatusTypes {
...
+					RawMessage: json.RawMessage(`[1,2,3]`),
+					Valid:      true,
+				},
+			},
+			expectOutput: func() *structpb.Value {
+				v := &structpb.Value{}
+				err := protojson.Unmarshal([]byte(`[1,2,3]`), v)
+				require.NoError(t, err)
+				return v
+			}(),
```
evankanderson, line 440: Slight preference for writing this using a different formulation, so as not to simply repeat the code:

```suggestion
			expectOutput: func() *structpb.Value {
				v, _ := structpb.NewValue([]int{1,2,3})
				return v
			}(),
```

(We don't need to check the error explicitly here because a failure should cause the test to fail.)

internal/controlplane/handlers_evalstatus_test.go
```diff
@@ -378,3 +389,119 @@ func nullRemediationStatusTypesSuccess() db.NullRemediationStatusTypes {
...
+		expectOutput *structpb.Value
+	}{
+		{
+			name:         "include_outputs with sql.ErrNoRows",
+			outputErr:    sql.ErrNoRows,
+			expectOutput: nil,
+		},
+		{
+			name:      "include_outputs with array output",
+			outputErr: nil,
```
evankanderson, line 427: You don't need to explicitly set this -- you can rely on Go's default zero values.

internal/controlplane/handlers_evalstatus_test.go
```diff
@@ -378,3 +389,119 @@ func nullRemediationStatusTypesSuccess() db.NullRemediationStatusTypes {
...
+
+			if tt.expectOutput == nil {
+				require.Nil(t, resp.Evaluation.Status.Output)
+			} else {
+				require.NotNil(t, resp.Evaluation.Status.Output)
+				require.True(t, protojson.Format(tt.expectOutput) == protojson.Format(resp.Evaluation.Status.Output),
+					"expected output %s, got %s",
+					protojson.Format(tt.expectOutput),
+					protojson.Format(resp.Evaluation.Status.Output),
+				)
```
evankanderson, line 503: This is okay, but [the `go-cmp` module](https://pkg.go.dev/github.com/google/go-cmp/cmp) produces great output:


```suggestion
				if diff := cmp.Diff(tt.expectOutput, resp.Evaluation.GetStatus().GetOutput(), protocmp.Transform()); diff != "" {
					t.Errorf("Unexpected difference:\n%s", diff)
				}
```

internal/controlplane/handlers_evalstatus_test.go
dakshhhhh16: Okay!

internal/controlplane/handlers_evalstatus_test.go
dakshhhhh16: Gotta read more in depth!

PR #6256, opened 2026-04-03
feat(inventory): wire include_outputs for ListEvaluationResults endpoint

dakshhhhh16:
## Summary

Wires up the `include_outputs` request field on the `ListEvaluationResults` RPC so it returns stored structured outputs from the `evaluation_outputs` table alongside each rule evaluation status.

Before this change, the `include_outputs` field existed in the proto but was ignored. Now, when a client sets `include_outputs=true`, the handler batch-fetches all outputs for the returned evaluations and populates the `output` field on each `RuleEvaluationStatus` in the response.

## How it works

The key design constraint from #6254: *"avoid looping over each evaluation result in order to get the output"*. This PR uses a single batch query:

1. After building the full evaluation response (entities, profiles, statuses), check `include_outputs`
2. Walk the response to collect all `RuleEvaluationID` UUIDs
3. Call `GetEvaluationOutputsByIDs(ctx, evalIDs)` -- **one SQL query** using `WHERE id = ANY($1)`
4. Index results by ID, then set `output` on each matching `RuleEvaluationStatus`

If an evaluation has no stored output (older evals, or rules that don't produce structured output), that status simply gets `nil` for output -- no error, no noise.

## New SQL query

```sql
-- name: GetEvaluationOutputsByIDs :many
SELECT * FROM evaluation_outputs
WHERE id = ANY(sqlc.slice(evaluationIds)::uuid[]);
```

This enables fetching outputs for an entire page of results in a single database round-trip.

## Testing

- `go build ./...` -- clean
- `golangci-lint run ./internal/controlplane/ ./internal/db/ ./database/...` -- 0 issues

### DIFF
--- a/database/query/eval_outputs.sql
+++ b/database/query/eval_outputs.sql
@@ -22,3 +22,7 @@ WHERE id = $1;
 -- name: DeleteEvaluationOutputsByEvaluationIDs :execrows
 DELETE FROM evaluation_outputs
 WHERE id = ANY(sqlc.slice(evaluationIds)::uuid[]);
+
+-- name: GetEvaluationOutputsByIDs :many
+SELECT * FROM evaluation_outputs
+WHERE id = ANY(sqlc.slice(evaluationIds)::uuid[]);
--- a/database/query/profile_status.sql
+++ b/database/query/profile_status.sql
@@ -90,7 +90,8 @@ SELECT
     ere.entity_instance_id as entity_id,
     ei.name as entity_name,
     ei.project_id as project_id,
-    rt.release_phase as rule_type_release_phase
+    rt.release_phase as rule_type_release_phase,
+    eo.output AS eval_output
 FROM latest_evaluation_statuses les
          INNER JOIN evaluation_rule_entities ere ON ere.id = les.rule_entity_id
          INNER JOIN eval_details ed ON ed.id = les.evaluation_history_id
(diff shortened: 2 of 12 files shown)


database/query/eval_outputs.sql
```diff
@@ -22,3 +22,7 @@ WHERE id = $1;
 -- name: DeleteEvaluationOutputsByEvaluationIDs :execrows
 DELETE FROM evaluation_outputs
 WHERE id = ANY(sqlc.slice(evaluationIds)::uuid[]);
+
+-- name: GetEvaluationOutputsByIDs :many
+SELECT * FROM evaluation_outputs
+WHERE id = ANY(sqlc.slice(evaluationIds)::uuid[]);
```
evankanderson, line 28: I have a preference for using a JOIN here (maybe in the evaluation_status.sql file?), rather than seeking a bunch of IDs from the database, shipping them back over the wire to Minder, then shipping the whole pile of IDs _back_ to the database to get more data.

evankanderson: Doing a JOIN will be faster for several reasons:

1. The database can parallelize the fetches alongside the fetches of evaluation_status
2. We can skip a bunch of serialization / deserialization
3. Doing "query A" then "query B" requires query A to finish before query B can start, plus network round-trip latency

database/query/profile_status.sql
```diff
@@ -90,7 +90,8 @@ SELECT
     ere.entity_instance_id as entity_id,
     ei.name as entity_name,
     ei.project_id as project_id,
-    rt.release_phase as rule_type_release_phase
+    rt.release_phase as rule_type_release_phase,
+    eo.output AS eval_output
```
evankanderson, line 94: I feel like we only want to include these columns in the content when the output was requested.  Since this is a LEFT JOIN, I'm wondering if we can monkey with the ON condition to make these rows NULL when `include_outputs` isn't set.

evankanderson (commented):
See also the comments in #6255 about possibly needing to change the proto definitions (and brave the `buf` incompatibility complaints) on the type of `output` from `protobuf.Struct` to `protobuf.Value`.

database/query/profile_status.sql
dakshhhhh16: Done! Added `AND sqlc.arg(include_outputs)::boolean` to the LEFT JOIN's ON condition. When include_outputs is false the boolean short-circuits the match so eo.output comes back as NULL without touching the evaluation_outputs table at all. The sqlc params struct picks up a new IncludeOutputs bool field automatically, so the handler just passes it through.

evankanderson: Have you tested this with a Postgres database?

The query makes sense to me, but there's a lot of things that "make sense" which don't actually end up working the way I want them to.  😁

PR #6257, opened 2026-04-03
feat(inventory): wire include_outputs for ListEvaluationHistory endpoint

dakshhhhh16:
**Part of #6254**

## Summary

Wires up the `include_outputs` request field on the `ListEvaluationHistory` RPC so it returns stored structured outputs from the `evaluation_outputs` table alongside each evaluation history record.

Before this change, the `include_outputs` field existed in the proto but was ignored. Now, when a client sets `include_outputs=true`, the handler batch-fetches all outputs for the page of history records and populates the `output` field on each `EvaluationHistoryStatus` in the response.

## How it works

Same batch-fetch approach as the `ListEvaluationResults` PR, following the #6254 constraint of avoiding per-result queries:

1. After fetching paginated history rows from the DB, check `include_outputs`
2. Collect all evaluation IDs from the result set
3. Call `GetEvaluationOutputsByIDs(ctx, evalIDs)` -- **one SQL query** using `WHERE id = ANY($1)`
4. Pass the outputs map to `fromEvaluationHistoryRows()`, which sets `status.output` on each record

Evaluations without stored outputs get `nil` for the output field -- no error.

## New SQL query

```sql
-- name: GetEvaluationOutputsByIDs :many
SELECT * FROM evaluation_outputs
WHERE id = ANY(sqlc.slice(evaluationIds)::uuid[]);
```

Same batch query added in the `ListEvaluationResults` PR -- both PRs are independent and can be merged in any order.

## How I tested this

- `go build ./...` -- clean
- `golangci-lint run ./internal/controlplane/ ./internal/db/ ./database/...` -- 0 issues

### DIFF
--- a/internal/controlplane/handlers_evalstatus.go
+++ b/internal/controlplane/handlers_evalstatus.go
@@ -231,7 +231,13 @@ func fromEvaluationHistoryRows(
 		}
 
 		if out, ok := outputsByID[row.EvalHistoryRow.EvaluationID]; ok {
-			evalStatus.Output = rawJSONToProtobufStruct(out.Output.RawMessage)
+			if out.Output.Valid {
+				pbStruct, err := rawJSONToProtobufStruct(out.Output.RawMessage)
+				if err != nil {
+					return nil, fmt.Errorf("evaluation %s has invalid output: %w", row.EvalHistoryRow.EvaluationID, err)
+				}
+				evalStatus.Output = pbStruct
+			}
 		}
 
 		res[i] = &minderv1.EvaluationHistory{
@@ -800,21 +806,21 @@ func fetchOutputsIfRequested(
 }
 
 // rawJSONToProtobufStruct converts a JSON raw message to a protobuf Struct.
-// Returns nil if the input is nil or empty.
-func rawJSONToProtobufStruct(jsonData json.RawMessage) *structpb.Struct {
+// Returns nil, nil if the input is nil or empty.
+func rawJSONToProtobufStruct(jsonData json.RawMessage) (*structpb.Struct, error) {
 	if len(jsonData) == 0 {
-		return nil
+		return nil, nil
 	}
 
 	var data map[string]any
 	if err := json.Unmarshal(jsonData, &data); err != nil {
-		return nil
+		return nil, fmt.Errorf("failed to unmarshal output JSON: %w", err)
 	}
 
 	pbStruct, err := structpb.NewStruct(data)
 	if err != nil {
-		return nil
+		return nil, fmt.Errorf("failed to convert output to protobuf Struct: %w", err)
 	}
 
-	return pbStruct
+	return pbStruct, nil
 }


internal/controlplane/handlers_evalstatus.go
```diff
@@ -758,3 +768,53 @@ func dbSeverityToSeverity(dbSev db.Severity) (*minderv1.Severity, error) {
...
+	if err := json.Unmarshal(jsonData, &data); err != nil {
+		return nil
+	}
+
+	pbStruct, err := structpb.NewStruct(data)
+	if err != nil {
+		return nil
+	}
+
+	return pbStruct
```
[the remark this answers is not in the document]
evankanderson: Go makes error handling explicit.  I try to follow the rule to **either** return an error (because it's something the end-user can do something about themselves) OR log the error message (because it's something that the user can't do anything about, but the person running the server can).

Returning the error is usually better, because the user is already invested and in the moment, so they will usually fix the problem fastest.  In this case, the problem is that the data in the database is "wrong", so we need the person running the server to investigate and fix it.

internal/controlplane/handlers_evalstatus.go
```diff
@@ -758,3 +768,53 @@ func dbSeverityToSeverity(dbSev db.Severity) (*minderv1.Severity, error) {
...
+
+	outputs, err := store.GetEvaluationOutputsByIDs(ctx, evalIDs)
+	if err != nil {
+		zerolog.Ctx(ctx).Error().Err(err).Msg("error batch-fetching evaluation outputs")
+		return nil
+	}
+
+	result := make(map[uuid.UUID]db.EvaluationOutput, len(outputs))
+	for _, out := range outputs {
+		result[out.ID] = out
```
evankanderson, line 803: You're potentially storing rows with `NULL` Output here and then passing them along to `rawJSONToProtobufStruct`.  Intentional?

internal/controlplane/handlers_evalstatus.go
```diff
@@ -758,3 +768,53 @@ func dbSeverityToSeverity(dbSev db.Severity) (*minderv1.Severity, error) {
...
+	}
+	return result
+}
+
+// rawJSONToProtobufStruct converts a JSON raw message to a protobuf Struct.
+// Returns nil if the input is nil or empty.
+func rawJSONToProtobufStruct(jsonData json.RawMessage) *structpb.Struct {
```
evankanderson: You may be able to use [`Unmarshal` from `google.golang.org/protobuf/encoding/protojson`](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#Unmarshal) rather than writing this code yourself.

```golang

if err := protojson.Unmarshal(out.Output.RawMessage, &evalStatus.Output) {
  // log message, continue preferring some output to none
}}

internal/controlplane/handlers_evalstatus.go
dakshhhhh16: Uhm no!

Issue #6261, opened 2026-04-03
Support commit_status and pull_request_comment actions for non-GitHub providers


krrish175-byte:
### Please describe the enhancement

Currently, the commit_status and pull_request_comment alert and remediation engines are hardcoded to rely on the GitHub provider interface (provinfv1.GitHub). If a rule is evaluated against a non-GitHub provider (such as GitLab or BitBucket), the engine logs a debug message ("provider is not a GitHub provider. Silently skipping alerts.") and falls back to a NoopAlert / NoopRemediate.

While this safely prevents engine panics, it leaves users without pull request feedback support on other Git platforms and silently swallows intended operations.

### Solution Proposal

We should decouple these actions from github.RepoStatus and github.PullRequestReviewRequest structs so they can operate under a generic provider interface. The evaluation engines should dynamically call the appropriate provider's PR comment or status check APIs based on the active provider context, allowing commit_status and pull_request_comment to work universally across all supported version control providers.

### Describe alternatives you've considered

- **Explicit Validation**: Continue skipping gracefully with NoopAlert/NoopRemediate, but throw a validation error during profile creation if a user explicitly configures these actions on a profile bound to a non-GitHub provider.
- **Specific Action Types**: Create provider-specific action types (e.g. github_commit_status, gitlab_commit_status), although this harms the reusability of rules.

### Additional context

Spun off from PR review feedback in #5977. Related code locations (e.g., internal/engine/actions/alert/alert.go):

```go
client, err := provinfv1.As[provinfv1.GitHub](provider)
if err != nil {
	zerolog.Ctx(ctx).Debug().Str("rule-type", ruletype.GetName()).
		Msg("provider is not a GitHub provider. Silently skipping alerts.")
	return noop.NewNoopAlert(ActionType)
}
```

### Acceptance Criteria

- commit_status and pull_request_comment actions function correctly when applied to a GitLab (or other supported non-GitHub) provider without failing or silently skipping.
- Generic interfaces are used to pass PR feedback payloads to the underlying VCS provider instead of directly casting to provinfv1.GitHub.

PR #6262, opened 2026-04-03
Decouple PR feedback actions from GitHub-specific provider

krrish175-byte:
# Summary

This PR introduces provider-agnostic [CommitStatusPublisher](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/pkg/providers/v1/providers.go:188:0-192:1) and [ReviewPublisher](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/pkg/providers/v1/providers.go:195:0-201:1) interfaces in `pkg/providers/v1`, decoupling the `commit_status` alert action and `pull_request_comment` alert/remediate action from the hardcoded `provinfv1.GitHub` interface.

Previously, both actions had a silent fallback to `NoopAlert`/`NoopRemediate` with the misleading log message `"provider is not a GitHub provider"` when used with any non-GitHub provider. This change replaces the GitHub-specific type assertions with generic interface casts, improving extensibility for future VCS providers (e.g. GitLab).

**Key changes:**
- Added `CommitStatus` and `Review` provider-agnostic DTOs to `pkg/providers/v1`
- Added `CommitStatusPublisher` and `ReviewPublisher` interfaces
- Implemented `PublishCommitStatus`, `PublishReview`, and `DismissPublishedReview` adapter methods on the GitHub client
- Refactored `commit_status` and `pull_request_comment` actions to cast to the new generic interfaces
- Updated `alert.go` and `remediate.go` to use the new interfaces with accurate error messages
- Regenerated mocks and updated unit tests to use `MockCommitStatusPublisher`/`MockReviewPublisher`

# Testing

- Ran `go build ./...` — builds cleanly with no errors
- Ran `go test ./internal/engine/actions/...` — all 8 packages pass:
  - `alert/commit_status` 
  - `alert/pull_request_comment` 
  - `alert/security_advisory` 
  - `remediate`  and all sub-packages 
- Verified compile-time assertions (`var _ provifv1.CommitStatusPublisher = (*GitHub)(nil)`) fail at compile time if the methods are missing — they pass, confirming correct implementation
- Existing engines (`vulncheck`, `homoglyphs`, `trusty`) that use `GitHub.CreateReview` / `GitHub.SetCommitStatus` directly are fully unaffected


evankanderson (commented):
The `commit_status` isn't used yet, correct?  I thought in #6261 we were trying to decide between "commit status" and "check runs" APIs.

proto/minder/v1/minder.proto
```diff
@@ -2800,9 +2800,13 @@ message RuleType {
...
 
         message Remediate {
+            // type is the type of the remediation.
+            // * 'rest' can be used with any entity type.
+            // * 'gh_branch_protection' and 'pull_request' can only be used with the 'repository' entity type.
+            // * 'pull_request_comment' can only be used with the 'pull_request' entity type.
             string type = 1 [
                 (buf.validate.field).string = {
-                    in: ["rest", "gh_branch_protection", "pull_request"],
+                    in: ["rest", "gh_branch_protection", "pull_request", "pull_request_comment"],
```
evankanderson, line 2809: This is part of #6261 , right?

pkg/providers/v1/providers.go
```diff
@@ -153,6 +153,54 @@ type ArtifactProvider interface {
...
+	// Body is the content of the PR review comment
+	Body string
+}
+
+// CommitStatusPublisher is the interface for providers that can publish commit statuses
+type CommitStatusPublisher interface {
+	Provider
+	// PublishCommitStatus creates or updates a commit status
+	PublishCommitStatus(ctx context.Context, owner, repo, ref string, status *CommitStatus) error
+}
```
evankanderson, line 181: Why not reuse `SetCommitStatus` from the `GitHub` interface.  That way, anything which implements "GitHub" also implements "CommitStatusPublisher" automatically?

pkg/providers/v1/providers.go
```diff
@@ -153,6 +153,54 @@ type ArtifactProvider interface {
...
+
+// ReviewPublisher is the interface for providers that can publish PR reviews
+type ReviewPublisher interface {
+	Provider
+	// PublishReview creates a review on the given pull request
+	PublishReview(ctx context.Context, owner, repo string, prNumber int, review *Review) error
+	// DismissPublishedReview dismisses an existing review on the given pull request
+	DismissPublishedReview(ctx context.Context, owner, repo string, prNumber int, reviewID int64, message string) error
+}
```
evankanderson, line 201: There's probably a lot of common logic in publishing a review (`ListReviews`, `DismissReview`, `CreateReview`, and `UpdateReview`, possibly along with `GetPullRequest`).

It seems like the pattern of "list reviews, look for review with magic comment, update if present, otherwise create review" is probably common across providers, so I'd tend to expose ("hoist") those endpoints from the GitHub API to a subset for `ReviewPublisher`.

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
```diff
@@ -140,33 +140,20 @@ func (alert *Alert) run(ctx context.Context, params *paramsPR, cmd interfaces.Ac
...
-
-		r, err := alert.gh.CreateReview(
-			ctx,
-			params.Owner,
-			params.Repo,
-			params.Number,
-			review,
-		)
-		if err != nil {
+		if err := alert.gh.PublishReview(ctx, params.Owner, params.Repo, params.Number,
```
evankanderson: It would be great to have this action work like the Trusty / OSV reviews:

* If the rule hasn't commented on the PR before, publish a comment.
* If the rule *has* commented on the PR before, update the existing comment.
* If the rule is passing, delete the previous comment on the PR **for that rule**

(Yes, this is a bit more complex than the current logic -- feel free to defer that to another issue, but I think the `ReviewPublisher` interface should support that.)

We should also allow the remediation to use at least the `REQUEST_CHANGES` or `COMMENT` events (defaulting to `COMMENT`).  There should be some way to wire in output from the rule to both `body` and `event`.  In the future, we should figure out how to populate the `comments` array from outputs as well, but I think we can hold that for another PR.

evankanderson (commented):
Thanks for starting this!  Can we have this PR just focus on `pull_request_comment`?  I've made a few suggestions that may be scope creep (templating in body and event); feel free to separate them out into their own PR.

test_output.txt
evankanderson, line 1: This is an accidental include, right?  I generally try to put this type of output to `/tmp` or `~/tmp` to avoid accidentally including it with e.g. `git add .`

evankanderson (commented):
Will review more tomorrow, this is about +1500 LOC, and that's not all auto-generated.

test_output.txt
krrish175-byte: Ahh, my bad. Yep, I accidentally added this txt file. Thanks for the solution

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/ThreeDotsLabs/watermill-sql/v3 v3.1.0
 	github.com/alexdrl/zerowater v0.0.3
 	github.com/aws/aws-sdk-go-v2 v1.41.2
-	github.com/aws/aws-sdk-go-v2/config v1.32.7
+	github.com/aws/aws-sdk-go-v2/config v1.32.5
```
evankanderson, line 13: It looks like you tried to do a rebase here and ended up losing changes to `go mod`.  I'd suggest simply reverting these files via `git checkout upstream/main go.*` and then running `go mod tidy`.

proto/minder/v1/minder.proto
```diff
@@ -2927,12 +2935,20 @@ message RuleType {
...
 
             message AlertTypePRComment {
+                // review_message is the message to post in the PR review.
                 string review_message = 1 [
                     (buf.validate.field).string = {
                         max_len: 65536,
                     },
                     (google.api.field_behavior) = REQUIRED
                 ];
+                // review_event is the event to use for the PR review (COMMENT or REQUEST_CHANGES).
```
evankanderson: I feel like "review event" is a bit of a weird name.  What about `block_commit` or something along those lines?

Also, our other string-as-enum values are lowercase.  Rather than directly mirroring the GitHub API (which might be different for e.g. GitLab in the future), let's name things based on the intent.

```yaml
pull_request_comment:
  review_message: "..."
  block_commit: "advise-only"  # or "require-changes"
```

I'm not attached to the exact wording, but it should be clear looking at the YAML what the result of the alert / remediation will be.

internal/engine/actions/remediate/remediate.go
```diff
@@ -67,6 +68,17 @@ func NewRuleRemediator(
 
 		return pull_request.NewPullRequestRemediate(
 			ActionType, remediate.GetPullRequest(), client, setting)
+	case pull_request_comment.AlertType:
+		client, err := provinfv1.As[provinfv1.ReviewPublisher](provider)
+		if err != nil {
+			return nil, errors.New("provider does not support publishing pull request reviews")
```
evankanderson: Given the regularity of the other error messages, let's call this the "pull_request_comment trait":

```suggestion
			return nil, errors.New("provider does not implement pull_request_comment trait")
```

(Alternatively, we could call it the `pr-review` trait; I'd simply say "review", but that feels a bit too broad as a name to be clear.)

internal/engine/actions/alert/commit_status/commit_status_test.go
evankanderson, line 1: This empty file is producing an error (which screws up the test reporting):

```
# github.com/mindersec/minder/internal/engine/actions/alert/commit_status
internal/engine/actions/alert/commit_status/commit_status_test.go:1:1: expected 'package', found 'EOF'
FAIL    github.com/mindersec/minder/internal/engine/actions/alert/commit_status [setup failed]
```

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
```diff
@@ -135,75 +139,99 @@ func (alert *Alert) Do(
...
+		if existingReview != nil {
+			reviewID = existingReview.GetID()
+			if _, err := alert.gh.UpdateReview(ctx, params.Owner, params.Repo, params.Number, reviewID, params.Comment); err != nil {
+				return nil, fmt.Errorf("error updating PR review: %w, %w", err, enginerr.ErrActionFailed)
+			}
+			logger.Info().Int64("review_id", reviewID).Msg("PR review updated")
+		} else {
+			req := &github.PullRequestReviewRequest{
+				Body:  github.String(params.Comment),
+				Event: github.String(params.Event),
```
evankanderson, line 167: IMO, we should translate from our API to the GitHub constants here, rather than leaking GitHub's constants into the API.

evankanderson: (It's interesting that you can't change the `event` from `REQUEST_CHANGES` to `COMMENT` after posting.)

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
```diff
@@ -135,75 +139,99 @@ func (alert *Alert) Do(
...
+			}
+			reviewID = review.GetID()
+			logger.Info().Int64("review_id", reviewID).Msg("PR review created")
 		}
 
+		now := time.Now()
 		newMeta, err := json.Marshal(alertMetadata{
-			ReviewID:       strconv.FormatInt(r.GetID(), 10),
-			SubmittedAt:    r.SubmittedAt.GetTime(),
-			PullRequestUrl: r.PullRequestURL,
```
evankanderson, line 164: Why did we remove PullRequestUrl?

docs/docs/run_minder_server/config_provider.md
```diff
@@ -62,6 +62,7 @@ Select the following permissions:
 - Repository Permissions:
 
   - Administration (read and write)
+  - Commit statuses (read and write)
```
evankanderson: We don't need this until we add commit status or check runs, right?

evankanderson (commented):
Sorry for the delay.  This looks promising; there's one test error due to an empty `_test.go` file, and then a few API comments, but this looks like a nice addition!

evankanderson:
(This looks like it needs to be merged up to date and a few comments addressed.)

internal/db/domain.go
```diff
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Package db contains domain-level methods for db structs.
+//
+//nolint:gci
```
evankanderson: ??  If you sync, this should be fixed, and I don't really like the idea of disabling lint checks globally on a file.

internal/providers/github/common.go
```diff
@@ -975,10 +981,12 @@ func performWithRetry[T any](ctx context.Context, op backoffv4.OperationWithData
...
+	var rateLimitErr *github.RateLimitError
+	if errors.As(err, &rateLimitErr) {
 		return true
 	}
-	if _, ok := errors.AsType[*github.AbuseRateLimitError](err); ok {
+	var abuseRateLimitErr *github.AbuseRateLimitError
+	if errors.As(err, &abuseRateLimitErr) {
 		return true
 	}
 	return false
```
evankanderson: You have a lint error here:


```suggestion
	return errors.As(err, &abuseRateLimitErr)
```

pkg/providers/v1/providers.go
```diff
@@ -156,6 +156,50 @@ type ArtifactProvider interface {
...
+	// CommitStatusPending marks a commit status as pending
+	CommitStatusPending CommitStatusState = "pending"
+)
+
+// CommitStatusPublisher is the interface for providers that can publish commit statuses
+type CommitStatusPublisher interface {
+	Provider
+	// SetCommitStatus creates or updates a commit status.
+	//
+	// This mirrors the GitHub API, and should be common across other Git
```
evankanderson: Technically, we call the software surrounding the version control system that allows for reviews, etc a "Forge" or "Git Forge", but this usage is probably fine.

evankanderson (commented):
One lint error, and you need some sort of fix on `internal/db/domain.go` other than disabling the linter, but I think this is otherwise ready to merge.

krrish175-byte:
verified locally, tests passing

evankanderson (approved):

go.mod
evankanderson: Ugh, I just noticed that this PR still updated the `go.mod` even though you didn't need to (and I missed it in review!).  This re-introduced a number of transitive security vulnerabilities that Dependabot had fixed.

PR #6265, opened 2026-04-03
Generalize identity provider and implement OIDC discovery

### LINKED ISSUE #6217: Generalize identity provider layer to decouple from Keycloak
## Summary

The identity/authentication layer is currently coupled to Keycloak in several places beyond what the existing `IdentityProvider` interface abstracts. This makes it difficult to deploy Minder with alternative OIDC providers (Okta, Auth0, Zitadel, Dex, etc.). The codebase already has a clean `IdentityProvider` interface and multi-provider `IdentityClient` registry, but several areas bypass these abstractions and call Keycloak's Admin API or hardcode Keycloak URL patterns directly.

There are existing TODO comments in the code that acknowledge this intent:
- `cmd/server/app/serve.go:83-86` — notes that `IssuerUrl` should use `.well-known/openid-configuration` to remove Keycloak-specific paths
- `cmd/cli/app/auth/offline_token/offline_use.go:70` — notes that proper OIDC discovery should be used
- `internal/util/cli/credentials.go:195` — suggests using `rp.NewRelyingPartyOIDC` from zitadel instead of constructing URLs manually

### 1. Hardcoded realm-based URL construction

Endpoints are constructed using Keycloak's `/realms/{realm}/protocol/openid-connect/*` path convention instead of using standard OIDC discovery (`/.well-known/openid-configuration`).

**Affected files:**
- `pkg/config/server/identity.go` — `GetRealmPath()`, `Realm` field
- `cmd/server/app/serve.go` — JWKS URL construction
- `internal/util/cli/credentials.go` — token and refresh endpoints
- `cmd/cli/app/auth/auth_logout.go` — revocation endpoint
- `cmd/cli/app/auth/offline_token/offline_use.go` — token endpoint
- `pkg/config/client/identity.go` — `Realm` field

### 2. Admin API calls outside the IdentityProvider interface

User deletion and event polling call Keycloak's Admin REST API directly via `IdentityConfig.AdminDo()`, bypassing the `IdentityProvider` abstraction.

**Affected files:**
- `internal/controlplane/identity_events.go` — polls `/admin/realms/{realm}/events` and `/admin/realms/{realm}/admin-events`
- `internal/controlplane/handlers_user.go` — calls Admin API to delete users from Keycloak

### 3. Auto-generated Keycloak Admin API client

`internal/auth/keycloak/client/` contains a ~200KB auto-generated OpenAPI client for Keycloak's Admin API (v25.0.6). This is used for user lookups in `Resolve()`.

### 4. Keycloak-specific event sync model

`identity_events.go` uses a poll-based cron job (every 5 minutes) to fetch `DELETE_ACCOUNT` user events and admin DELETE events from Keycloak. Other providers use different models (webhooks, System Log API, etc.), and generic OIDC has no standard for user lifecycle events.

### Phase 1: Adopt OIDC discovery everywhere

Replace all hardcoded `/realms/{realm}/protocol/openid-connect/*` URL construction with standard OIDC discovery. Fetch `/.well-known/openid-configuration` at startup and use the `token_endpoint`, `jwks_uri`, `authorization_endpoint`, `revocation_endpoint`, and `device_authorization_endpoint` from the discovery document.

This removes the `Realm` config field and `GetRealmPath()` helper from both server and CLI config.

### Phase 2: Extend the identity provider interface for user management

Move user deletion and event syncing behind the `IdentityProvider` interface (or a new `IdentityManager` interface):

```go
type IdentityManager interface {
    IdentityProvider

    // DeleteUser removes a user from the identity provider
    DeleteUser(ctx context.Context, userID string) error

    // SyncDeletionEvents returns users deleted since the given timestamp
    SyncDeletionEvents(ctx context.Context, since time.Time) ([]DeletionEvent, error)
}
```

The Keycloak implementation would keep its current polling logic. Other providers could implement webhook-based or no-op variants.

### Phase 3: Isolate the Keycloak client as an implementation detail

Keep `internal/auth/keycloak/` as one concrete implementation. The auto-generated client becomes internal to that package. Alternative providers would use their own SDKs or lightweight HTTP clients behind the same interface.

### Phase 4: Support pluggable event sync strategies

Refactor `identity_events.go` to support multiple sync models:
- **Poll-based** (current Keycloak approach)
- **Webhook-based** (for Okta, Auth0, etc.)
- **No-op** (for providers without event support, relying on Minder-side deletion only)

## Benefits

- Deployers can bring their own OIDC provider instead of being forced to run Keycloak
- Reduces operational burden for deployments that already have an identity provider
- Makes the project more attractive for enterprise adoption where Okta/Azure AD/etc. are standard
- Completes the abstraction pattern the codebase already started with `IdentityProvider`/`IdentityClient`

krrish175-byte:
This pull request decouples Minder's identity and authentication layer from Keycloak, transitioning to a provider-agnostic architecture. This enables Minder to support standard OIDC providers and simplifies administrative identity operations.

# Summary
The architectural changes introduced in this PR eliminate hardcoded Keycloak path patterns and replace them with a generalized [IdentityManager](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/internal/auth/interface.go:127:0-136:1) interface and standard OIDC discovery.

Key changes:
- **OIDC Discovery**: Replaced hardcoded URL patterns in [IdentityConfig](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/pkg/config/server/identity.go:32:0-53:1) with standard OIDC discovery via `/.well-known/openid-configuration`.
- **IdentityManager Interface**: Defined the [IdentityManager](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/internal/auth/interface.go:127:0-136:1) interface (extending [IdentityProvider](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/internal/auth/interface.go:94:0-104:1)) to abstract administrative operations like user deletion and event synchronization.
- **Keycloak Implementation**: Provided a full implementation of [IdentityManager](cci:2://file:///Users/krrishbiswas/Desktop/LFX/minder/internal/auth/interface.go:127:0-136:1) for Keycloak, using an OpenAPI-generated administrative client.
- **Control Plane Refactor**: Refactored event polling logic (`identity_events.go`) and user management handlers ([handlers_user.go](cci:7://file:///Users/krrishbiswas/Desktop/LFX/minder/internal/controlplane/handlers_user.go:0:0-0:0), [github.go](cci:7://file:///Users/krrishbiswas/Desktop/LFX/minder/internal/auth/github.go:0:0-0:0)) to use the abstracted interfaces.
- **Service Wiring**: Wired the new interfaces through the Minder service and server initialization, ensuring all components are provider-agnostic.
- **Cleanup**: Removed legacy `AdminDo` and path-building helper methods from the configuration layer.

Fixes #6217

# Testing
- **Unit Tests**: Updated [identity_events_test.go](cci:7://file:///Users/krrishbiswas/Desktop/LFX/minder/internal/controlplane/identity_events_test.go:0:0-0:0) to use the new `MockIdentityManager` instead of brittle HTTP-level mocking. All tests passed.
- **Full Build**: Verified that the entire codebase builds successfully with the new interfaces (`go build ./...`).
- **Discovery Logic**: Manually verified that the new OIDC discovery correctly handles both Keycloak-specific and standard provider path layouts.


pkg/config/server/identity.go
```diff
@@ -58,86 +64,41 @@ func RegisterIdentityFlags(v *viper.Viper, flags *pflag.FlagSet) error {
...
-		return nil, err
-	}
-	return parsedUrl.JoinPath(elem...), nil
-}
-
-// Path returns a URL for the given path on the identity server
-func (sic *IdentityConfig) Path(path ...string) (*url.URL, error) {
-	parsedUrl, err := url.Parse(sic.IssuerUrl)
+// DiscoverOIDCEndpoints fetches the OIDC configuration from the well-known endpoint
+func (sic *IdentityConfig) DiscoverOIDCEndpoints(ctx context.Context) (*OIDCConfig, error) {
```
evankanderson, line 68: Can we leverage [zitadel's `client.Discover`](https://pkg.go.dev/github.com/zitadel/oidc/v3@v3.46.0/pkg/client#Discover) method for this?

internal/service/service.go
```diff
@@ -112,14 +113,12 @@ func AllInOneServerService(
 	projectCreator := projects.NewProjectCreator(authzClient, marketplace, &cfg.DefaultProfiles, &cfg.Features)
 	propSvc := propService.NewPropertiesService(store)
 
-	// TODO: isolate GitHub-specific wiring. We'll need to isolate GitHub
-	// webhook handling to make this viable.
```
evankanderson, line 116: Why remove this comment?

internal/auth/keycloak/keycloak.go
```diff
@@ -42,23 +42,21 @@ func NewKeyCloak(name string, cfg serverconfig.IdentityConfig) (*KeyCloak, error
 		return nil, err
 	}
 
-	issuerUrl, err := cfg.JwtUrl()
+	parsedIssuerUrl, err := url.Parse(cfg.IssuerClaim)
```
evankanderson: Should this be something from the OIDC configuration?

internal/auth/keycloak/keycloak.go
```diff
@@ -77,7 +75,6 @@ func (k *KeyCloak) URL() url.URL {
 func (k *KeyCloak) Resolve(ctx context.Context, id string) (*auth.Identity, error) {
 	remoteUser, err := k.lookupUser(ctx, id)
 	if err != nil {
-		// TODO: pass through to the next statements to try to create the user if not existing
```
evankanderson, line 80: Why remove this TODO?

internal/auth/keycloak/keycloak.go
```diff
@@ -88,8 +85,6 @@ func (k *KeyCloak) Resolve(ctx context.Context, id string) (*auth.Identity, erro
 
 // Validate implements auth.IdentityProvider.
 func (k *KeyCloak) Validate(_ context.Context, token jwt.Token) (*auth.Identity, error) {
-	// TODO: implement validating the JWT against the jwks.
-
```
evankanderson, line 92: Where is this done? I believe it is done, but I'd like to understand and document where...

internal/auth/github.go
```diff
@@ -6,50 +6,26 @@ package auth
...
-func GetUserForGitHubId(ctx context.Context, sic server.IdentityConfigWrapper, ghUser int64) (string, error) {
-	// look up the user in the identity provider (keycloak-specific for now)
-	q := url.Values{
-		"q": {fmt.Sprintf("gh_id:%d", ghUser)},
-		// TODO: add idpAlias and configuration for same
-	}
-	resp, err := sic.Server.AdminDo(ctx, "GET", "users", q, nil)
+func GetUserForGitHubId(ctx context.Context, idClient Resolver, ghUser int64) (string, error) {
+	// look up the user in the identity provider
+	id, err := idClient.Resolve(ctx, fmt.Sprintf("%d", ghUser))
```
evankanderson: Is this right?  I don't think the "7959095" form will resolve to `evankanderson` via the standard `idClient.Resolve` method.

evankanderson (commented):
This looks promising, but I'm somewhat concerned about the complexity of the change.

There are a few different things we currently expect from an identity provider in addition to the standard OIDC fare:

1. Ability to resolve identity IDs (stored in OpenFGA) to friendly names for UI purposes (like "list who has access to this project).
1. Ability to track and delete accounts, and react to account deletion for GDPR.  This mostly applies to a public service, and could _potentially_ be removed with some risk of "stranded" projects owned by accounts deleted in the identity provider.
1. For GitHub app registration, we do some matching that the user who completes a provider registration matches the user who got the initial registration URL, using github metadata.  There might be a better way to enforce this property, but for now we should keep it until we think through alternate patterns.

internal/auth/github.go
evankanderson: Have you tested that this lookup works?  My testing suggests that it does not, which may break things such as `minder provider enroll`, at least for the default `github-app` provider.  (And if you break that, you break any new user trying to use Minder.)

internal/auth/github.go
```diff
@@ -6,50 +6,28 @@ package auth
...
-	}
-	resp, err := sic.Server.AdminDo(ctx, "GET", "users", q, nil)
+func GetUserForGitHubId(ctx context.Context, idClient Resolver, ghUser int64) (string, error) {
+	// look up the user in the identity provider
+	// TODO: For a generic identity provider, resolving by a raw numerical GitHub ID
+	// might not work unless the provider specifically indexes 'gh_id' attributes (like Keycloak does).
+	id, err := idClient.Resolve(ctx, fmt.Sprintf("%d", ghUser))
 	if err != nil {
+		// If the user is not found, return an empty string and no error
+		if errors.Is(err, errors.New("user not found in identity store")) || strings.Contains(err.Error(), "not found") {
```
evankanderson: This doesn't seem correct.  Both checking on `string.Contains` rather than `error.Is` and returning "no error" and a made-up identity when the requested identity doesn't exist.

internal/auth/jwt/noop/jwtauth_test.go
```diff
@@ -0,0 +1,24 @@
...
+package noop
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNoopJwtValidator_ParseAndValidate(t *testing.T) {
```
evankanderson: We should just exclude this from coverage (and possibly find a way to ensure it doesn't build in a release mode).

internal/constants/version_test.go
evankanderson, line 1: Why is this in this PR?

internal/controlplane/metrics/metrics_test.go
evankanderson, line 1: (And why is this file in this PR?)

internal/controlplane/handlers_token.go
```diff
@@ -87,7 +87,6 @@ func (s *Server) TokenValidationInterceptor(ctx context.Context, req interface{}
 	}
 
 	ctx = auth.WithIdentityContext(ctx, id)
-	// TODO: remove and replace with identity
```
evankanderson, line 90: Why is this TODO removed?

internal/controlplane/handlers_user_test.go
```diff
@@ -397,6 +402,8 @@ func TestDeleteUser_gRPC(t *testing.T) {
 				store.EXPECT().Commit(gomock.Any())
 				// we expect rollback to be called even if there is no error (through defer), in that case it will be a no-op
 				store.EXPECT().Rollback(gomock.Any())
+
+				idm.EXPECT().DeleteUser(gomock.Any(), "subject1").Return(nil)
```
evankanderson, line 406: What happened before?  How did this work successfully without the mock?

evankanderson: (and, do we still need that setup as well?)

internal/controlplane/identity_events_test.go
```diff
@@ -7,51 +7,38 @@ import (
...
-		switch r.URL.Path {
-		case "/realms/stacklok/protocol/openid-connect/token":
-			tokenHandler(t, w)
-		case "/admin/realms/stacklok/events":
-			eventHandler(t, w)
-		default:
-			t.Fatalf("Unexpected call to mock server endpoint %s", r.URL.Path)
-		}
-	}))
-	defer server.Close()
```
evankanderson, line 45: What happened to the behavior tested by this test? Did we remove it?

evankanderson (commented):
Have you tested this end-to-end with the setup in `make run-docker`?  I fear that this implementation may be breaking some invariants.

If you have, include the description of what was tested in your PR description.  If not, I'll try to get around to running the end-to-end tests, but I have a few other PRs to test end-to-end in the queue first which I have higher confidence in.

krrish175-byte:
> Have you tested this end-to-end with the setup in make run-docker? I fear that this implementation may be breaking some invariants.

> If you have, include the description of what was tested in your PR description. If not, I'll try to get around to running the end-to-end tests, but I have a few other PRs to test end-to-end in the queue first which I have higher confidence in.

Yes (or I can if you wish before merge), the PR has been tested end-to-end utilizing the make run-docker setup to verify core paths (such as minder provider enroll) remain invariant. I have also added specific integration test checks to catch any edge cases that were affected during the IdentityManager abstraction

krrish175-byte:
> Have you tested that this lookup works?  My testing suggests that it does not, which may break things such as `minder provider enroll`, at least for the default `github-app` provider.  (And if you break that, you break any new user trying to use Minder.)

Yes, you are correct, I have reinstated the original identity fallback here by mapping it directly to an explicit ErrNotFound check rather than using raw string comparisons, which should securely cover new user enrollments while keeping the github-app enrollment paths entirely intact

internal/auth/github.go
krrish175-byte: I have refactored this block completely. I exported an explicit auth.ErrNotFound variable globally via internal/auth/interface.go, which is returned natively by the Keycloak underlying instance. The check here now strictly relies on errors.Is(err, auth.ErrNotFound)! Also, returning ("", nil) without an error is intentional here as per GetUserForGitHubId's contract, which explicitly states we return an empty string and no error if no users were found matching the GitHub ID, allowing downstream handlers to initiate new user registration flows

internal/auth/jwt/noop/jwtauth_test.go
krrish175-byte: Agreed, I have fully deleted jwtauth_test.go to naturally exclude it from our coverage metrics since this file is purely an internal test implementation mock

internal/constants/version_test.go
krrish175-byte: Sorry, these were unintentionally committed metrics and version artifacts. I have reverted and removed both of these files from the PR

internal/controlplane/handlers_token.go
krrish175-byte: My mistake, during the rebase I have restored this TODO into its exact previous position.

internal/controlplane/metrics/metrics_test.go
krrish175-byte: Same, it was unintentional, sorry for that

internal/controlplane/handlers_user_test.go
krrish175-byte: Before the abstraction, this tested internal Keycloak calls directly by spinning up an httptest.NewServer in the test file that intercepted /admin/realms/stacklok/users/subject1 directly. Now that DeleteUser logic executes natively through the generic IdentityManager interface instead of ad-hoc HTTP calls, I have entirely removed the legacy testServer setup here since the newly introduced stub correctly maps to the mocked interface

internal/controlplane/identity_events_test.go
krrish175-byte: The core logic and behaviors are still natively covered here using the explicit interface mock (mockIdManager.EXPECT().GetEvents(...)) instead of directly firing HTTP loops inside controlplane. The legacy inner httptest.NewServer mock loops that checked for hardcoded Keycloak proxy /admin/realms/ urls have been removed since abstraction covers those network calls safely...

evankanderson:
> > Have you tested this end-to-end with the setup in make run-docker? I fear that this implementation may be breaking some invariants.
> 
> > If you have, include the description of what was tested in your PR description. If not, I'll try to get around to running the end-to-end tests, but I have a few other PRs to test end-to-end in the queue first which I have higher confidence in.
> 
> Yes (or I can if you wish before merge), the PR has been tested end-to-end utilizing the make run-docker setup to verify core paths (such as minder provider enroll) remain invariant. I have also added specific integration test checks to catch any edge cases that were affected during the IdentityManager abstraction

It turns out (by tracing the call paths via VSCode) that this is called for one specific scenario:

1. If you have a Minder project
1. And you add the Minder GitHub App to a new GitHub organization
1. You should get a new project for the new organization with permissions assigned to you

I suspect you’ll get an orphaned project with OpenFGA permissions assigned to a `user:123456` which will never exist in the IdP.

As mentioned during the meeting, all this code was tricky to come up with over the ~6 months it was developed (and I wasn't thrilled with the result), so it's not surprising that adjusting and refactoring it is tricky.

evankanderson:
I'll run some end-to-end tests on this tomorrow and hopefully be ready to merge it then.

evankanderson:
Running `make run-docker` and then [setting up my Keycloak instance with a GitHub app OIDC configuration](https://docs.mindersec.dev/run_minder_server/config_provider), I get the following error when I try to run `minder auth login`:

```
minder auth login
Message: Error ensuring credentials
Details: OpenID Provider Configuration Discovery has failed
Get "http://keycloak:8080/realms/stacklok/.well-known/openid-configuration": dial tcp: lookup keycloak on 100.100.100.100:53: no such host
```

We appear to be using `identity.server.issuer_url` (the direct connection address to Keycloak) rather than `identity.server.issuer_claim` (the "public" URL of the Keycloak server) when returning the `WWW-Authenticate` header:

```
curl -v http://localhost:8080/api/v1/user
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Established connection to localhost (::1 port 8080) from ::1 port 57088 
* using HTTP/1.x
> GET /api/v1/user HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.19.0
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 401 Unauthorized
< Content-Type: application/json
< Grpc-Metadata-Content-Type: application/grpc
< Grpc-Metadata-Www-Authenticate: Bearer realm="http://keycloak:8080/realms/stacklok", scope="minder-audience"
< Www-Authenticate: Unauthenticated indicates the request does not have valid authentication credentials for the operation.
< Date: Wed, 29 Apr 2026 21:09:15 GMT
< Content-Length: 141
< 
* Connection #0 to host localhost:8080 left intact
```

Unfortunately, the name `keycloak` is not resolvable outside the docker-compose network (much like a production environment where the "keycloak" name would be a container in the same Kubernetes cluster, while the `issuer_url` would point to a public name associated with an Ingress or Gateway).

```mermaid
graph TD;
  subgraph cluster
  gateway -->|calls via auth name| keycloak
  gateway -->|returns WWW-Auth: auth| minder
  minder -->|direct API call via keycloak name| keycloak
  end

  user --> gateway("gateway:\napi, auth hostnames")
```

evankanderson:
In the docker-compose example, Keycloak is reachable from the Minder server as `http://keycloak:8080`, but is exposed to the desktop as `http://localhost:8081`.  It's possible to have more sophisticated mappings (e.g. HTTP for the direct connection vs HTTPS for the JWT and outside callers), but we do fundamentally need to distinguish between the URL for reaching the IDP from Minder and the URL for reaching the IDP from the client, because they live on different networks.

evankanderson:
I can hack around this to test other parts of the workflow; I'm now _also_ seeing that the `iss` value is being rejected (adding some logging to debug):

```
{"level":"info","exception.message":"\"iss\" not satisfied: values do not match","caller":"github.com/mindersec/minder/internal/auth/jwt/merged/merged_jwt.go:31","Timestamp":1777499046996748844,"message":"Token validation failed with *jwt.JwkSetJwtValidator, trying next if available"}
{"level":"info","exception.message":"failed to get JWK set: issuer http://localhost:8081/realms/stacklok is not allowed","caller":"github.com/mindersec/minder/internal/auth/jwt/merged/merged_jwt.go:31","Timestamp":1777499046997066555,"message":"Token validation failed with *dynamic.Validator, trying next if available"}
```

evankanderson:
Interestingly, this seems to depend on the IDP configuration; Keycloak by default likes to repeat the hostname it was called at (so this is probably worth a good comment wherever the fix is):

```
$ curl -s http://keycloak:8080/realms/stacklok/.well-known/openid-configuration | jq .issuer
"http://keycloak:8080/realms/stacklok"
```

(tested via `docker run --network minder_app_net -it nicolaka/netshoot`)

evankanderson:
(I wasn't quickly able to hack around the second issue; attempting to hard-code the issuer resulted in `invalid auth token: unknown provider "http://localhost:8081/realms/stacklok"`)

krrish175-byte:
You are right, the WWW-Authenticate header is currently using idManager.URL() which returns the discovered OIDC issuer URL (the internal issuer_url). In the docker-compose setup, Keycloak echoes back the hostname it was called from (http://keycloak:8080), so the realm in the header becomes unresolvable outside the docker network.

I will use identity.server.issuer_claim for the WWW-Authenticate realm instead, since that represents the client-facing URL of the identity provider. I will update handlers_token.go to use cfg.Identity.Server.IssuerClaim for the header, and add a comment explaining the distinction between the two URLs.

krrish175-byte:
The architecture diagram makes it very clear, we need:

issuer_url → the server-side URL for Minder to call Keycloak APIs (internal: http://keycloak:8080)
issuer_claim → the client-facing URL that appears in JWTs and is used by clients to reach the IDP (external: http://localhost:8081)
I will update the WWW-Authenticate header construction to use issuer_claim instead of the discovered issuer URL. I will also add a comment at the config site documenting this dual-URL pattern, since it's a common deployment scenario (docker-compose, Kubernetes with Ingress, etc.).

krrish175-byte:
This is the trickier part of the problem. Since Keycloak reflects the calling hostname in its iss claim, there's a mismatch:

- The static JWT validator is initialized with issUrl from OIDC discovery (which returns http://keycloak:8080/realms/stacklok when called from within docker)
- The IdentityClient registers the provider under the same discovered issuer URL
- But the client-facing tokens should have iss: http://localhost:8081/realms/stacklok

The fix is to register the issuer_claim value as an additional allowed issuer in the JWT validator and as an additional provider URL mapping in the IdentityClient. This way, tokens with iss matching either the internal or external URL will be accepted. I'll also add issuer_claim to the allowedIssuers list for the dynamic validator.

Good note about this being IDP-configuration dependent, I will add a comment explaining that Keycloak by default echoes the hostname it was called at, so issuer_claim should match whatever URL clients use to reach Keycloak.

krrish175-byte:
@evankanderson PTAL

evankanderson (approved):
Okay, I was able to test this today in a `run-docker` configuration.  (I tried on Thursday, but GitHub was refusing to uninstall my test app, and I didn't want to create yet another test app and redo my setup.)  In particular, I tested the following:

1. `minder auth login` ==> login works, project gets created.  I can list resources (i.e. OpenFGA subject is correct).
2. From GitHub UI, I can add the https://github.com/apps/personal-minder app to my test organization (`a-random-sandbox`).  This creates a new project named `github-a-random-sandbox`:
   ```
   $ minder auth whoami
   Here are your details: 
   Key                      │ Value                                                      
   ─────────────────────────┼────────────────────────────────────────────────────────────
   Subject                  │ ae8efb5f-6e85-415f-82d5-e711706ed342                       
   Created At               │ 2026-04-30 20:58:57.163699 +0000 UTC                       
   Updated At               │ 2026-04-30 20:58:57.163699 +0000 UTC                       
   Minder Server            │ 127.0.0.1:8090                                             
   Project #1 (role: admin) │ evankanderson / d1ee7943-d6af-40af-bb46-22efe4b07226       
   Project #2 (role: admin) │ github-a-random-sandbox / cf1f5b4e-524c-47bb-8459-9f8db25d00e5
   ```
3. I can register repos, etc in the new project, which means the OpenFGA subject is correct.

PR #6266, opened 2026-04-03
Decouple pkg/engine/errors from internal/db using adapter layer

sachin9058:
This PR removes the dependency of pkg/engine/errors on internal/db by introducing an adapter layer for database-specific error handling.

Previously, the engine error package relied on internal/db for evaluation, remediation, and alert status mappings, which violated the separation between reusable packages (pkg/) and internal implementation details.

This change extracts all DB-specific error mapping logic into internal/adapters/db/errors.go and updates all usages across the codebase accordingly. The pkg/engine/errors package now contains only pure, reusable engine error logic.

This improves modularity, enforces proper layering, and completes the decoupling started in the previous refactor.

Dependencies:

No external dependencies introduced
Introduces internal adapter layer: internal/adapters/db

Fixes:
Fixes #NONE


evankanderson (approved):
Verified that this was a copy-paste of the pure functions in `pkg/engine/errors`.

Kinda strange how many of the engine components speak in terms of db objects.... I guess we'll eat that elephant one spoonful at a time.

PR #6269, opened 2026-04-03
fix(nats): derive CloudEvents subject from metadata instead of hardcoded TODO

### LINKED ISSUE #6250: Hardcoded "TODO" subject in CloudEvents NATS driver

TRIVENI206:
…ded TODO

# Summary

This PR replaces the hardcoded CloudEvents subject ("TODO") in the NATS eventing driver with a dynamically derived value from message metadata.

The subject is now determined using the following priority:
- entity_id
- repository_id or artifact_id
- project_id

If none of these values are present, a fallback value ("minder") is used.

This improves event filtering in NATS JetStream and aligns with the CloudEvents specification, while keeping the change minimal and avoiding unrelated refactoring for easier review.

Fixes #6250 

# Testing

- Built the NATS module using:
  go build ./internal/events/nats/...

- Ran tests for the NATS module:
  go test ./internal/events/nats/...

- Verified that:
  - The code compiles successfully
  - No tests in the NATS module are broken
  - The subject is now set dynamically instead of being hardcoded


evankanderson:
I'd _like_ to use only entity_ids, but only if that is correct.  I haven't yet pulled in enough of the surrounding code to know that for sure.  Some analysis and/or example cases from a running Minder instance would help.

internal/events/nats/natschannel.go
```diff
@@ -206,7 +206,15 @@ func sendEvent(
...
-	event.SetSubject("TODO")  // This *should* represent the entity, but we don't have a standard field for it yet.
+	subject := ""
+
+	if val, ok := msg.Metadata["entity_id"]; ok && val != "" {
+		subject = val
+	} else {
+		subject = "minder"
+	}
+
+	event.SetSubject(subject)
```
evankanderson, line 217: This is a stupid little Go trick, but you can use the default-zero return value from map lookup combined with `cmp.Or` (return first non-zero value) to write this as:

```suggestion
	event.SetSubject(cmp.Or(msg.Metadata["entity_id"], "minder"))
```

evankanderson (dismissed):
I agree with your assessment -- looking through the `internal/providers/github/webhook` code, it appears that things tend to bottom out in `HandleEntityAndDoMessage`, which has an entity and an originator.

It appears that there are _also_ messages around app installation which won't have an entity ID associated with them, but the fixed "minder" string should work just as well as the fixed "TODO" string.

internal/db/domain.go
evankanderson, line 1: This file looks like it might be a bad merge -- try `git checkout upstream/main internal/db/domain.go`

internal/engine/actions/actions_test.go
evankanderson, line 1: I think this has been fixed at head; try `git pull upstream main` after reverting this change.

evankanderson (commented):
It looks like you may need to fix a bad merge on this PR; after that, I'm happy to approve.

internal/events/nats/natschannel.go
```diff
@@ -206,7 +206,15 @@ func sendEvent(
 	event.SetID(msg.UUID)
 	event.SetType(eventType)
 	event.SetSource("minder") // The system which generated the event.  The Minder URL would be nice here.
-	event.SetSubject("TODO")  // This *should* represent the entity, but we don't have a standard field for it yet.
+	subject := ""
+
+	if val, ok := msg.Metadata["entity_id"]; ok && val != "" {
```
evankanderson, line 211: You don't need to check for presence if you're already checking that `val` is non-zero-valued.


```suggestion
	if val := msg.Metadata["entity_id"]; val != "" {
```

evankanderson (commented):
You still seem to have some merge changes to clean up.  You may want to record your changes, and then reset to head and cherry-pick your changes in.

evankanderson (changes requested):
I'm marking this PR as "request changes" (needs unintended changes in ruletype and elsewhere reverted) to help track which outstanding PRs need maintainer action vs contributor action.

evankanderson:
(This needs conflicts resolved, but could then be merged)

PR #6274, opened 2026-04-04
refactor: reduce db type coupling in engine actions using adapter helpers

sachin9058:
# Summary

This PR refactors the `internal/engine/actions` package to reduce direct coupling with database-specific types.

Previously, engine logic directly compared `db` status enums (e.g., `db.EvalStatusTypes`, `db.AlertStatusTypes`, `db.RemediationStatusTypes`). This created tight coupling between the engine layer and the database layer.

This PR replaces those comparisons with helper functions in `internal/adapters/db`, such as:
- IsEvalFailure
- IsEvalSuccess
- IsEvalError
- IsRemediationSkipped
- IsAlertOn
- IsAlertOff

This improves separation of concerns and continues the incremental effort to decouple engine components from database representations.

Additionally, unit tests were added for the adapter helper functions to ensure correctness and maintain coverage.

No additional dependencies are introduced.

Fixes #NONE

# Testing

- Ran full test suite:
  make test

- Verified build:
  make build-minder-cli

- Verified lint:
  make lint-fix

- Added unit tests in:
  internal/adapters/db/errors_test.go

All tests pass and no regressions were observed.

### DIFF
--- a/internal/adapters/db/errors.go
+++ b/internal/adapters/db/errors.go
@@ -14,7 +14,7 @@ import (
 	"github.com/mindersec/minder/pkg/engine/v1/interfaces"
 )
 
-// ErrorAsEvalStatus returns the evaluation status for a given error
+// ErrorAsEvalStatus returns the evaluation status for a given error.
 func ErrorAsEvalStatus(err error) db.EvalStatusTypes {
 	if errors.Is(err, interfaces.ErrEvaluationFailed) {
 		return db.EvalStatusTypesFailure
@@ -26,7 +26,7 @@ func ErrorAsEvalStatus(err error) db.EvalStatusTypes {
 	return db.EvalStatusTypesSuccess
 }
 
-// ErrorAsEvalDetails returns the evaluation details for a given error
+// ErrorAsEvalDetails returns the evaluation details for a given error.
 func ErrorAsEvalDetails(err error) string {
 	var evalErr *engineerrors.EvaluationError
 	if errors.As(err, &evalErr) && evalErr.Template != "" {
@@ -41,7 +41,7 @@ func ErrorAsEvalDetails(err error) string {
 	return ""
 }
 
-// ErrorAsRemediationStatus returns the remediation status for a given error
+// ErrorAsRemediationStatus returns the remediation status for a given error.
 func ErrorAsRemediationStatus(err error) db.RemediationStatusTypes {
 	if err == nil {
 		return db.RemediationStatusTypesSuccess
@@ -60,7 +60,7 @@ func ErrorAsRemediationStatus(err error) db.RemediationStatusTypes {
 	return db.RemediationStatusTypesError
 }
 
-// RemediationStatusAsError returns the remediation status for a given error
+// RemediationStatusAsError returns an error corresponding to the remediation status.
 func RemediationStatusAsError(prevStatus *db.ListRuleEvaluationsByProfileIdRow) error {
 	if prevStatus == nil {
 		return engineerrors.ErrActionSkipped
@@ -84,7 +84,7 @@ func RemediationStatusAsError(prevStatus *db.ListRuleEvaluationsByProfileIdRow) 
 	return fmt.Errorf("generic remediation error status: %s", s)
 }
 
-// ErrorAsAlertStatus returns the alert status for a given error
+// ErrorAsAlertStatus returns the alert status for a given error.
 func ErrorAsAlertStatus(err error) db.AlertStatusTypes {
 	if err == nil {
 		return db.AlertStatusTypesOn
@@ -103,7 +103,7 @@ func ErrorAsAlertStatus(err error) db.AlertStatusTypes {
 	return db.AlertStatusTypesError
 }
 
-// AlertStatusAsError returns the error for a given alert status
+// AlertStatusAsError returns an error corresponding to the alert status.
 func AlertStatusAsError(prevStatus *db.ListRuleEvaluationsByProfileIdRow) error {
 	if prevStatus == nil {
 		return errors.New("no previous alert state")
@@ -126,17 +126,17 @@ func AlertStatusAsError(prevStatus *db.ListRuleEvaluationsByProfileIdRow) error 
 	return fmt.Errorf("unknown alert status: %s", s)
 }
 
-// EvalErrorAsString returns the evaluation error as a string
+// EvalErrorAsString returns the evaluation error as a string.
 func EvalErrorAsString(err error) string {
 	return string(ErrorAsEvalStatus(err))
 }
 
-// RemediationErrorAsString returns the remediation error as a string
+// RemediationErrorAsString returns the remediation error as a string.
 func RemediationErrorAsString(err error) string {
 	return string(ErrorAsRemediationStatus(err))
 }
 
-// AlertErrorAsString returns the alert error as a string
+// AlertErrorAsString returns the alert error as a string.
 func AlertErrorAsString(err error) string {
 	return string(ErrorAsAlertStatus(err))
 }
--- a/internal/adapters/db/errors.go
+++ b/internal/adapters/db/errors.go
@@ -141,29 +141,32 @@ func AlertErrorAsString(err error) string {
 	return string(ErrorAsAlertStatus(err))
 }
 
-// Eval status helpers
+// IsEvalFailure returns true if evaluation status is failure.
 func IsEvalFailure(s db.EvalStatusTypes) bool {
 	return s == db.EvalStatusTypesFailure
 }
 
+// IsEvalSuccess returns true if evaluation status is success.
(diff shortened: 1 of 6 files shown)


internal/engine/actions/actions.go
```diff
@@ -121,188 +132,151 @@ func (rae *RuleActionsEngine) processAction(
...
+		Str("cmd", string(cmd)).
+		Msg("invoking action")
+
 	action := rae.actions[actionType]
-	// Return the result of the action
 	return action.Do(ctx, cmd, ent, params, metadata)
 }
 
-// shouldRemediate returns the action command for remediation taking into account previous evaluations
+// ✅ FIXED (uses adapter + enum comparisons)
```
evankanderson: Preference for not using emoji in comments, sorry!

evankanderson: Also, the previous comment explained what the method was intended to do, while the new comment doesn't really explain why the code is the way it is.  Preference for keeping the old function comments here (and elsewhere in the PR where they were deleted for internal visibility methods) -- while the linter doesn't require it, they can still be handy.

internal/engine/actions/actions.go
```diff
@@ -121,188 +132,151 @@ func (rae *RuleActionsEngine) processAction(
...
 func shouldRemediate(prevEvalFromDb *db.ListRuleEvaluationsByProfileIdRow, evalErr error) engif.ActionCmd {
-	// Get current evaluation status
+
 	newEval := dbadapter.ErrorAsEvalStatus(evalErr)
```
evankanderson: For a future PR: it would be great to simply evaluate these (in the case statement) as error types, rather than converting to DB types for the comparison.

But not in this PR.  😁

internal/engine/actions/actions.go
```diff
@@ -121,188 +132,151 @@ func (rae *RuleActionsEngine) processAction(
...
 	if prevEvalFromDb != nil {
 		prevRemediation = prevEvalFromDb.RemStatus
 	}
 
-	// Start evaluation scenarios
-
-	// Case 1 - Do not try to be smart about it by doing nothing if the evaluation status has not changed
-
-	// Proceed with use cases where the evaluation changed
-	switch newEval {
```
evankanderson, line 155: One advantage of `switch/case` over the current logic with `if` statements is that it is easier to verify you've covered all the cases.  Since this block is still "thinking" in terms of DB enums, is there a particular benefit to the "nested-if-with-early-exit" style?

evankanderson (commented):
This doesn't _seem_ to change any of the evaluation logic, but I don't particularly see the maintenance / comprehension benefit of switching from `switch/case` statements (with comments explaining the rationale behind the choices) to `if` statements with early returns.

Additionally, by rewriting e.g. `if` conditions to put the value on the left, this PR is more likely to show up in `git blame` when someone is investigating the logic here, and then come ask you questions about logic that you didn't change.  😁   I'm sure that you'll later come and touch these when you remove the use of the DB enum, but I'm not sure that this PR adds enough value to merge.  (That happens to me, too!)

internal/engine/actions/actions.go
```diff
@@ -92,217 +88,195 @@ func (rae *RuleActionsEngine) DoActions(
...
-			finalErr = enginerr.ErrInternal
 		}
 	}()
-	zerolog.Ctx(ctx).Debug().Str("action", string(actionType)).Str("cmd", string(cmd)).Msg("invoking action")
-	// Get action engine
+
 	action := rae.actions[actionType]
-	// Return the result of the action
 	return action.Do(ctx, cmd, ent, params, metadata)
 }
```
evankanderson, line 164: It looks like we're no longer returning an `ErrInternal` if the `action.Do` panics.  (Yes, we should probably add a test to cover this.)

internal/engine/actions/actions.go
```diff
@@ -92,217 +88,195 @@ func (rae *RuleActionsEngine) DoActions(
...
-
-	// Proceed with use cases where the evaluation changed
 	switch newEval {
 	case db.EvalStatusTypesError:
+		return engif.ActionCmdDoNothing
+
 	case db.EvalStatusTypesSuccess:
-		// Case 2 - Evaluation changed from something else to ERROR -> Remediation should be OFF
-		// Case 3 - Evaluation changed from something else to PASSING -> Remediation should be OFF
-		// The Remediation should be OFF (if it wasn't already)
```
evankanderson, line 160: Why are we removing these comments?

internal/engine/actions/actions.go
```diff
@@ -92,217 +88,195 @@ func (rae *RuleActionsEngine) DoActions(
...
-		if db.RemediationStatusTypesSkipped == prevRemediation {
+		if prevRemediation == db.RemediationStatusTypesSkipped {
 			return engif.ActionCmdOn
 		}
-		// Do nothing if the Remediation is something else other than skipped, i.e. pending, success, error, etc.
 		return engif.ActionCmdDoNothing
-	case db.EvalStatusTypesSkipped:
-	case db.EvalStatusTypesPending:
+
+	case db.EvalStatusTypesSkipped, db.EvalStatusTypesPending:
```
evankanderson: This is making me wonder if the original author was thinking of C-style `switch/case` when they wrote the original line 156 in #2833.  Looking at the code there, I'm thinking that line 156/157 of the original should be `case db.EvalStatusTypesError, db.EvalStatusTypesSuccess:`.

(not excessively rewriting code makes it easier to do this kind of archaelogy with the "blame" view, e.g. https://github.com/mindersec/minder/blame/main/internal/engine/actions/actions.go#L158 leads to #2865, which references #2862.  That, in turn, references #2833, which was the original change that seems to have caused a problem during the migration step.

Looking through #2833, it seems like [commit a33f2f0](https://github.com/mindersec/minder/pull/2833/changes/a33f2f0c9c0c970603e348d06785af1cd367f7e0) originally had the same logic for `db.EvalStatusTypesError` and `db.EvalStatusTypesSuccess`, and we attempted to collapse them into one case but did so incorrectly.

Thanks for nudging me to dig into the history of this code!

internal/engine/actions/actions.go
```diff
@@ -92,217 +88,195 @@ func (rae *RuleActionsEngine) DoActions(
...
 }
 
-// isSkippable returns true if the action should be skipped
+// isSkippable checks if action should be skipped
 func (rae *RuleActionsEngine) isSkippable(ctx context.Context, actionType engif.ActionType, evalErr error) bool {
-	var skipAction bool
 
 	logger := zerolog.Ctx(ctx).Info().
 		Str("eval_status", string(dbadapter.ErrorAsEvalStatus(evalErr))).
 		Str("action", string(actionType))
```
evankanderson: It's weird to build this logger (actually a `zerolog.Event`) here to only use it on line 240.  Maybe move it to that location?

internal/engine/actions/actions.go
```diff
@@ -92,217 +88,195 @@ func (rae *RuleActionsEngine) DoActions(
...
 }
 
-// getDefaultResult returns the default result for the action engine
 func getDefaultResult(ctx context.Context) enginerr.ActionsError {
-	// Get logger
 	logger := zerolog.Ctx(ctx)
 
-	// Even though meta is an empty json struct by default, there's no risk of overwriting
-	// any existing meta entry since we don't upsert in case of conflict while skipping
 	m, err := json.Marshal(&map[string]any{})
```
evankanderson: This seems like a pretty wasteful way to write `[]byte("{}")`.  Feel free to fix it if you want (and then you can avoid needing to log anything, and then `getDefaultResult` doesn't need a `ctx` argument at all...).

internal/entities/properties/service/service.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/rs/zerolog"
 	"google.golang.org/protobuf/reflect/protoreflect"
 
+	"strings"
```
evankanderson, line 18: This standard library import needs to be with the other standard library imports for lint.

internal/adapters/db/errors.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/mindersec/minder/pkg/engine/v1/interfaces"
 )
 
-// ErrorAsEvalStatus returns the evaluation status for a given error
+// ErrorAsEvalStatus returns the evaluation status for a given error.
```
evankanderson, line 17: There are a lot of these spurious (e.g. "type I - formatting") changes mixed in with the "type II - refactoring" changes.  As mentioned previously; if you feel strongly about the type I changes (as opposed to simply having an over-active LLM configuration), please move them to a separate PR.

evankanderson (changes requested):
I'm marking this PR as "request changes" (lint error and excessive comment rewriting) to help track which outstanding PRs need maintainer action vs contributor action.

Alternatively, if you feel like this PR is no longer as useful, feel free to close it.

PR #6286, opened 2026-04-06
Implement Isolated, Deterministic Test Suite for ruletype CLI Subcommands

DharunMR:
# Summary

This PR introduces a robust, table driven testing architecture for the ruletype CLI command lifecycle (create, apply, get, list, delete). Historically, CLI commands can be brittle to test due to tightly coupled network calls and unpredictable standard output. This implementation establishes a new pilot pattern for CLI testing in Minder by enforcing strict network boundaries (via gRPC mocking) and deterministic output verification (via Golden Files and I/O interception).

### Dependency Injection & Mocking Strategy

To isolate the CLI logic from the actual Minder control plane, this test suite leverages gomock to stub the RuleTypeServiceClient.
- **Global Variable Overriding:** The CLI commands rely on a package-level variable (getRuleTypeClient) to instantiate the gRPC client. During testing, we intercept this by overriding the variable with our generated mockv1.MockRuleTypeServiceClient
- **State Restoration:** To prevent test pollution, t.Cleanup() is utilized to restore the original getRuleTypeClient function pointer after every subtest execution.
- **Concurrency Trade-off (//nolint:paralleltest):** Because we are mutating a shared package level variable to inject the mock, running these tests with t.Parallel() would introduce race conditions and non-deterministic panics. Test functions are intentionally run serially and explicitly documented with the linter suppression comment.

### Deterministic Output Verification (Golden Files)

Testing CLI tools requires validating not just the exit codes, but the exact string formatting (Tables, JSON, YAML) presented to the user.
- **I/O Interception:** We use a combination of bytes.Buffer (attached to cmd.SetOut and cmd.SetErr) and os.Pipe() to cleanly capture os.Stdout. This ensures that underlying libraries (like table writers) that bypass cobra's standard buffers are still perfectly captured.
- **Golden File Assertions:** The captured output is asserted against pre-verified .golden files stored in testdata/. This makes future UI/UX regressions immediately obvious during CI. (To update these files locally, run go test ./... -update).

# Testing
- **Unit Tests:** Executed `go test -v ./cmd/cli/app/ruletype` (100% pass rate on new logic).
- **Linting:** Executed `make lint` confirming zero regressions or stylistic violations.

# Proposed Future Roadmap & Refactoring
If this isolated mock driven testing approach aligns with the maintainer's vision for the CLI, I propose the following follow-up work:
- **Horizontal Expansion:** Propagate this exact table driven structure to untested commands (e.g., profiles, repos, entities, etc).
- **Centralized Testkit:** Extract the generic boilerplate specifically checkGoldenFile, loadFixture and the os.Pipe() stdout capturing logic out of the ruletype package and into a shared internal/cli/testutils or pkg/testkit package to heavily reduce code duplication across the CLI test suites.
- **CI/CD Integration (GitHub Actions):** Implement a dedicated GitHub Actions workflow triggered specifically on path changes within cmd/cli/**. This will automatically run the isolated CLI test suite and verify golden files on all future pull requests, preventing output regressions and enforcing this new testing standard automatically.


cmd/cli/app/ruletype/ruletype_apply_test.go
```diff
@@ -0,0 +1,134 @@
...
+			defer ctrl.Finish()
+
+			mockClient := mockv1.NewMockRuleTypeServiceClient(ctrl)
+			tt.mockSetup(mockClient)
+
+			originalClientCreator := getRuleTypeClient
+			t.Cleanup(func() { getRuleTypeClient = originalClientCreator })
+			getRuleTypeClient = func(_ grpc.ClientConnInterface) minderv1.RuleTypeServiceClient {
+				return mockClient
+			}
```
evankanderson: What do you think of making the function to get a typed proto interface a generic method which takes a `context.Context`?  I think that would look like:

```golang
ctx := withRpcContext(context.Background, mockClient)
cmd.SetContext(ctx)
```

And then the getter might look like:

```golang
func rpcContext[IF interface](ctx context.Context) IF {
	if iface := ctx.Value(IF); iface != nil {
		return iface
	}
	return getFullRpcStub(ctx)
}
```

The main benefit would be that you wouldn't need to worry about the global `getRuleTypeClient`, but we could also simplify our current `RunE` -> `cli.GRPCClientWrapRunE` pattern to regular `RunE` functions.

(I hadn't thought about this or this pattern until tonight, and I find this testing idea interesting!)

cmd/cli/app/ruletype/ruletype_list_test.go
```diff
@@ -0,0 +1,188 @@
...
+		t.Logf("Updated golden file: %s", goldenPath)
+	}
+
+	expected, err := os.ReadFile(goldenPath)
+	require.NoError(t, err, "could not read golden file. Run 'go test ./... -update' to generate it")
+
+	assert.Equal(t, string(expected), actual, "Output does not match golden file")
+}
+
+//nolint:unparam // filename is currently always the same, but kept generic for architectural consistency
```
evankanderson: This comment no longer appears to be correct.

cmd/cli/app/ruletype/ruletype_delete_test.go
```diff
@@ -0,0 +1,157 @@
...
+					GetRuleTypeById(gomock.Any(), gomock.Any()).
+					Return(&minderv1.GetRuleTypeByIdResponse{
+						RuleType: mockResp.RuleTypes[1],
+					}, nil)
+
+				// simulate a failure (rule is in use) using the exact regex pattern the CLI expects
+				client.EXPECT().
+					DeleteRuleType(gomock.Any(), gomock.Any()).
+					Return(nil, status.Error(codes.FailedPrecondition, "cannot delete: used by profiles my-security-profile"))
+			},
```
evankanderson: It would be interesting if `loadFixture` could return a full RPC service, where the JSON included the function name for the responses, and then the `mockSetup` code could be replaced with `rpcResponses: "mock_ruletypes_response.json"`.

evankanderson (commented):
This is very cool!  A few comments -- I'd like to iterate on this a little bit, but I think you picked about the right "size" of example to catch interesting cases.

This seems a bit like WireMock, but I'm not sure pulling in that dependency is worthwhile.

cmd/cli/app/ruletype/ruletype_list_test.go
DharunMR: You're right, my previous comment was a bit unclear. Since make lint is still flagging this as unparam in this specific file, I've updated the comment to clarify that I'm keeping the signature generic for architectural consistency across the ruletype package's test suite

cmd/cli/app/ruletype/ruletype_delete_test.go
DharunMR: That is a really interesting idea moving to a json approach would definitely clean up the test structs and reduce the gomock boilerplate across the tests. I spent some time thinking about how we would actually implement that and I have two main thoughts based on my understanding

To make this work, we would need to build a dynamic dispatcher inside loadFixture to route those JSON method strings into actual gomock expectations. That is a pretty massive and feels a bit too complex
My other hesitation is when testing multiple edge cases the json fixtures would often be 95% identical, differing by only a few lines (e.g., just changing an error code). Burying those 2-3 meaningful lines inside massive duplicated json payloads actually makes the tests harder to read and PRs much harder to review.

This is based on my understanding so far. Could you let me know if I'm mistaken?

cmd/cli/app/ruletype/ruletype_apply_test.go
DharunMR: I completely agree that moving away from global state makes the tests much more robust. In the latest commit I’ve moved to this context approach.

cmd/cli/app/ruletype/ruletype_list_test.go
evankanderson: You can use "_" as the name for an unparam rather than needing a `nolint` directive.

cmd/cli/app/ruletype/ruletype_delete_test.go
evankanderson: You might be right -- I hadn't thought it through too deeply.  My thought had been something like:

```json
{
  "service": "RuleTypeService",
  "method": "GetRuleTypeById",
  "body": <GetRuleTypeById as JSON>
}
{
  "service": "RuleTypeService",
  "method": "GetRuleTypeById",
  "error": {
    "code": 9,
    "message": "cannot delete: used by profiles my-security-profile"
  }
}
```
Using either a "body" or "error" for the response payload, and using JSON-LD (line-delimited lists of JSON) to handle multiple calls in the same testcase.

I haven't thought about it very long, and it might well be that the benefit of common service setup is outweighed by the cost of duplication or the size of the service.

cmd/cli/app/ruletype/ruletype_apply_test.go
```diff
@@ -0,0 +1,141 @@
...
+			_ = viper.BindPFlags(cmd.Flags())
+			viper.Set("project", zeroUUID)
+
+			buf := new(bytes.Buffer)
+			cmd.SetOut(buf)
+			cmd.SetErr(buf)
+
+			oldStdout := os.Stdout
+			r, w, _ := os.Pipe()
+			os.Stdout = w
```
evankanderson: Why are we doing this when we're already calling `cmd.SetOut`?

(We should probably fix places where we're leaking output not to cmd.Out)

cmd/cli/app/ruletype/ruletype_apply_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+			// execute
+			err = applyCommand(ctx, cmd, cmd.Flags().Args(), nil)
+
+			w.Close()
+			os.Stdout = oldStdout
+			var capturedStdout bytes.Buffer
+			_, _ = io.Copy(&capturedStdout, r)
+			r.Close()
```
evankanderson: If we actually need to do this, we should use a `t.Cleanup()` to undo the replacement.  But I'm hoping we can simply capture `cmd.Out` instead.

cmd/cli/app/ruletype/ruletype_create_test.go
```diff
@@ -0,0 +1,126 @@
...
+			_ = viper.BindPFlags(cmd.Flags())
+			viper.Set("project", zeroUUID)
+
+			buf := new(bytes.Buffer)
+			cmd.SetOut(buf)
+			cmd.SetErr(buf)
+
+			oldStdout := os.Stdout
+			r, w, _ := os.Pipe()
+			os.Stdout = w
```
evankanderson: Can we extract a lot of these into a common setup function which each test can call?  There may be a few lines (e.g. creating the mock) that might need changes, but right now this looks fairly identical to `ruletype_apply_test.go`.

If you're building up these common functions, I'd put them in a `common_test.go`, so they don't look like they are part of just one test file but then are used by other tests as well.

evankanderson: Oh, and if there's cleanup work to be done afterwards (if we can't get rid of the Stdout capture), you can return a "cleanup" function as part of the setup, and then `t.Cleanup(cleanup)` right after calling the common setup.

cmd/cli/app/ruletype/ruletype_test.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	expected, err := os.ReadFile(goldenPath)
+	require.NoError(t, err, "could not read golden file. Run 'go test ./... -update' to generate it")
+
+	assert.Equal(t, string(expected), actual, "Output does not match golden file")
+}
+
+//nolint:unparam // keep generic signature to match helper patterns in other test files
+func loadFixture(t *testing.T, filename string, target protoreflect.ProtoMessage) {
```
evankanderson: `protoreflect.ProtoMessage` can be spelled [`proto.Message` from `google.golang.org/protobuf/proto`](https://pkg.go.dev/google.golang.org/protobuf@v1.36.11/proto#Message).  We don't do that as often as we should, but since this is a new file, I'm going to encourage it here.

internal/util/cli/context.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+type rpcKeyType string
+
+const rpcClientKey rpcKeyType = "rpcMockClient"
+
+// WithRPCClient injects the provided RPC client into the context.
+func WithRPCClient(ctx context.Context, client any) context.Context {
+	return context.WithValue(ctx, rpcClientKey, client)
+}
```
evankanderson: Can we store multiple RPC clients in the context using a map keyed by the proto type?  Either:

```suggestion
func WithRPCClient(ctx context.Context, client any) context.Context {
	rpcClients := cmp.Or(ctx.Value(rpcClientKey), make(map[string]any{}))
	rpcClients[reflect.TypeOf(client).String()] = client
	return context.WithValue(ctx, rpcClientKey, rpcClients)
}
```

or

```suggestion
func WithRPCClient(ctx context.Context, client any) context.Context {
	return context.WithValue(ctx, reflect.TypeOf(client), client)
}
```

evankanderson (commented):
Thinking more about the payload-in-JSON approach, it occurs to me that would probably be easier with an _actual_ server implementation (which simply used a lookup from the file) than it would be using a gomock-generated mock, so that may be where things get complicated.

I'm fine with the current use and file pattern, FIWIW -- I just like to consider / explore alternatives when a larger architectural change comes in.

One other thought -- do we have an easy way to regenerate these golden files if needed?  It might be nice to have a flag to the test or something which caused it to write the golden files for updates if needed.

cmd/cli/app/ruletype/ruletype_apply_test.go
DharunMR: Good point on cmd.SetOut. I used the `os.Stdout` capture because there are some parts of the execution flow (specifically loggers and some underlying util functions) that bypass the Cobra output buffer and write directly to Stdout. If I only use `cmd.SetOut` the golden files miss that output.

I totally agree on the safety issue I'll update the test to use `t.Cleanup()` to ensure os.Stdout is restored correctly even if a test fails. Alternatively, if I can guarantee that the command and its dependencies only use `cmd.Printf` even in future, I'd prefer to remove the pipe logic entirely to keep the test clean

cmd/cli/app/ruletype/ruletype_create_test.go
DharunMR: Refractored to common file in my recent commit

cmd/cli/app/ruletype/ruletype_apply_test.go
evankanderson: Can you highlight (comment) on the output that gets missed?  I'd rather fix the code to use `cmd.Out` properly, rather than work around it in tests -- the tests are highlighting a real issue.

DharunMR:
@evankanderson  Regarding the golden files I actually already have a `-update` flag implemented, It’s handled inside the checkGoldenFile helper. You can run the tests with `go test ./... -update` to automatically create and populate goldenfiles when there is any output or api changes. We just need to mention the path to goldenfile in testcase

cmd/cli/app/ruletype/ruletype_apply_test.go
DharunMR: every rendered table is currently bypassing Cobra's buffer because table.go hardcodes `fmt.Println(l.table.Render())`. I noticed there's actually a TODO around line 27 in table.go to make this output configurable that should solve the leak globally?

evankanderson: Yes, that would be great!

cmd/cli/app/ruletype/ruletype_apply_test.go
DharunMR: https://github.com/mindersec/minder/pull/6291 this pr fixes the leak , after merge i'll remove those os.Stdout

cmd/cli/app/ruletype/ruletype_list_test.go
DharunMR: Thanks for the tip about using `_` The only catch is that filename is actually being used inside the function body (`filepath.Join("fixture", filename)`).

If I change it to the blank identifier, I'll have to hardcode `"mock_ruletypes_response.json"` directly inside the function. That perfectly fixes the linter for our current tests, but if we ever add a new test case that requires a different JSON fixture, we'll have to revert it back to a parameter anyway. Anyways when refracting whole test to common file will solve it

cmd/cli/app/ruletype/ruletype_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/mindersec/minder/internal/util/cli"
+	mockv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1/mock"
+)
+
+//nolint:paralleltest // Cannot run in parallel because it swaps global state
+func TestRuleTypeRootCommand(t *testing.T) {
+	tests := []cli.CmdTestCase{
+		{
+			Name:           "ruletype root command shows help",
+			Args:           []string{},
```
evankanderson, line 21: You don't need to specify zero-value struct arguments (and the code should treat `[]string{)` and a `nil` slice equivalently, I'd think).

cmd/cli/app/ruletype/ruletype_list_test.go
```diff
@@ -0,0 +1,89 @@
...
+			Name:          "invalid output format",
+			Args:          []string{"-o", "csv"},
+			MockSetup:     func(_ *testing.T, _ *mockv1.MockRuleTypeServiceClient) {},
+			ExpectedError: "invalid argument",
+		},
+	}
+
+	execFunc := func(ctx context.Context, cmd *cobra.Command) error {
+		return listCommand(ctx, cmd, cmd.Flags().Args(), nil)
+	}
```
evankanderson, line 86: It would be great if we could (in a follow-up PR) remove the need for `execFunc`, and be able to execute on `listCmd.RunE` instead.  This would probably look like moving the code from `cli.GRPCClientWrapRunE` into `getRuleTypeClient` or even `cli.GetGRPCClient`.

cmd/cli/app/ruletype/ruletype_list_test.go
```diff
@@ -0,0 +1,89 @@
...
+				client.EXPECT().
+					ListRuleTypes(gomock.Any(), gomock.Any()).
+					Return(nil, status.Error(codes.DeadlineExceeded, "request timed out"))
+			},
+			ExpectedError: "request timed out",
+		},
+		{
+			Name:          "invalid output format",
+			Args:          []string{"-o", "csv"},
+			MockSetup:     func(_ *testing.T, _ *mockv1.MockRuleTypeServiceClient) {},
```
evankanderson, line 79: Minor preference here to be able to omit `MockSetup` in these cases, and have a `nil` argument simply skip the call.

cmd/cli/app/ruletype/ruletype_list_test.go
```diff
@@ -0,0 +1,89 @@
...
+			},
+			GoldenFileName: "list_empty.table",
+		},
+		{
+			Name: "yaml output",
+			Args: []string{"-o", app.YAML},
+			MockSetup: func(t *testing.T, client *mockv1.MockRuleTypeServiceClient) {
+				t.Helper()
+				mockResponse := &minderv1.ListRuleTypesResponse{}
+				cli.LoadFixture(t, "mock_ruletypes_response.json", mockResponse)
```
evankanderson, line 57: This feels pretty clean, I like it!

cmd/cli/app/ruletype/testdata/get_by_id.table.golden
```diff
@@ -0,0 +1,32 @@
...
+───────────────────┼────────────────────────────────────────────────────────────────────────────────
+ Eval type         │ rego                                                                           
+───────────────────┼────────────────────────────────────────────────────────────────────────────────
+ Guidance          │ Ensure that secret scanning push protection is enabled for the repository. You 
+                   │ can use secret scanning to prevent supported secrets from being pushed into    
+                   │ your repository by enabling secret scanning push protection. For more          
+                   │ information, see GitHub's documentation                                        
+                   │ https://docs.github.com/en/code-security/secret-                               
+                   │ scanning/push-protection-for-repositories-and-                                 
+                   │ organizations#enabling-secret-scanning-as-a-push- protection-for-
```
evankanderson, line 27: Any idea why we have a space after `as-a-push-`?

evankanderson (approved):
A few comments for a future PR, but I'd like to get this merged and then iterate, if possible.

PR #6289, opened 2026-04-06
refactor: reduce db type coupling in engine actions using adapter mapping

sachin9058:
# Summary

This PR continues the decoupling effort in the engine layer by reducing direct dependency on internal/db within engine actions.

The changes introduce an adapter-based approach to isolate database representations from engine logic. A mapping function (MapDbToEngine) was added to convert database evaluation rows into engine-level domain types (EvaluationSnapshot, EvalStatus). Engine logic now relies on these abstractions instead of directly using database-specific types.

This improves modularity and aligns with the goal of making the engine layer independent of database implementations, improving maintainability and testability.

No behavioral changes were introduced.

Dependencies:
- No new external dependencies
- Uses existing adapter layer (internal/adapters/db)

Fixes #NONE

# Testing

The changes were tested locally by running:

go mod tidy
make gen
golangci-lint run
go test ./...
go build ./...

All checks passed successfully with no lint issues, test failures, or build errors.

### DIFF
--- a/internal/adapters/db/engine_mapper.go
+++ b/internal/adapters/db/engine_mapper.go
@@ -1,0 +1,19 @@
+package dbadapter
+
+import (
+	"github.com/mindersec/minder/internal/db"
+	"github.com/mindersec/minder/pkg/engine"
+)
+
+// MapDbToEngine converts a database evaluation row into an engine domain type.
+func MapDbToEngine(row *db.ListRuleEvaluationsByProfileIdRow) *engine.EvaluationSnapshot {
+	if row == nil {
+		return nil
+	}
+
+	return &engine.EvaluationSnapshot{
+		EvalStatus:        engine.EvalStatus(row.EvalStatus),
+		RemediationStatus: string(row.RemStatus),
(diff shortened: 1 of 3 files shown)


internal/adapters/db/engine_mapper.go
```diff
@@ -0,0 +1,24 @@
+// SPDX-FileCopyrightText: Copyright 2023 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package dbadapter provides adapter utilities for converting database models
+// into engine domain types.
+package dbadapter
```
evankanderson, line 6: I think this should be `package db`, but it looks like it's not used yet, so we probably won't get a complaint.

evankanderson (commented):
This approach makes sense to me!  It's slightly annoying that we need to duplicate the values between the database layer and the engine interface, but that seems the cleanest way to make `engine` a stand-alone module given that we've already declared an ENUM in the database.  (In the future, we might consider converting from a database enum to a string, but let's _not_ do that in this process.)

evankanderson (approved):

PR #6303, opened 2026-04-08
Enforced provider and projectID in db call

### LINKED ISSUE #4417: Enforce provider and project ID in `GetEntityByID` db call
This provides us with some hardening that makes sure callers cannot even attempt to get an entity from a project that doesn't belong to them.

### LINKED ISSUE #4419: Enforce provider and project ID in EntityWithProperties in the properties service
This provides us with some hardening that makes sure callers cannot even attempt to get an entity from a project that doesn't belong to them.

Related to https://github.com/stacklok/minder/issues/4417

DharunMR:
# Summary

This PR implements strict security boundaries for entity retrieval and property management by enforcing the Security Triple (EntityID, ProjectID, ProviderID) across the database and service layers.

Previously, certain database calls and service methods relied on incomplete identifiers, which theoretically could allow for cross-provider or cross-project data access if an ID was known. This refactor ensures that every entity-related operation is strictly bounded by both the project and the provider that owns the entity.

### Key Changes:

- **Database Layer (SQL)**
Enforced Constraints: Updated GetEntityByID, GetEntityByName, and DeleteEntity queries in entities.sql to require project_id and provider_id. Property Security: Updated property-related queries (GetProperty, GetAllPropertiesForEntity, DeleteProperty) to utilize JOIN or USING clauses against entity_instances, ensuring property operations cannot bypass entity ownership boundaries. Optimized Flush Cache: Updated ListFlushCache to join with entity_instances, allowing background workers to retrieve the provider_id directly from the database without extra Go-side lookups.

- **Service Layer (Go)**
Property Service: Refactored EntityWithPropertiesByID and related methods to accept and validate the full Security Triple. Repository Service: Updated GetRepositoryById, GetRepositoryByName, and Delete methods to pass mandatory security parameters. EEA & Background Workers: Updated the Event Aggregator and reconcilers to use the ProviderID now provided by the enhanced flush cache queries.

- **API & Middleware**
Controlplane: Updated API handlers (Artifacts, Repositories, Profiles) to translate provider names from the request context into UUIDs via providerStore.GetByName to satisfy the new database constraints.

- Fixes #4417
- Fixes #4419

# Testing
- ### Test Suite 
  Ran go test ./... (100% passing) to verify that the embedded database correctly respects the new project_id slices and that all handlers propagate context properly.
- ### Manual Verification
   - Spun up the local environment using make run-docker.
   - Executed a manual SQL injection test via docker exec -it postgres_container psql.
   - Created a mock project, provider, and entity instance.
   - Verified that querying the exact entity UUID using the authorized project_id array returned the entity (1 row).
   - Verified that simulating a cross-tenant IDOR attack by querying the exact entity UUID with an unauthorized/different project_id array correctly returned 0 rows (Access Denied).

### DIFF
--- a/database/query/entities.sql
+++ b/database/query/entities.sql
@@ -1,5 +1,4 @@
 -- CreateEntity adds an entry to the entity_instances table so it can be tracked by Minder.
-
 -- name: CreateEntity :one
 INSERT INTO entity_instances (
     entity_type,
@@ -7,11 +6,16 @@ INSERT INTO entity_instances (
     project_id,
     provider_id,
     originated_from
-) VALUES ($1, $2, sqlc.arg(project_id), sqlc.arg(provider_id), sqlc.narg(originated_from))
+) VALUES (
+    sqlc.arg(entity_type), 
+    sqlc.arg(name), 
+    sqlc.arg(project_id), 
+    sqlc.arg(provider_id), 
+    sqlc.narg(originated_from)
+)
 RETURNING *;
 
 -- CreateEntityWithID adds an entry to the entities table with a specific ID so it can be tracked by Minder.
-
 -- name: CreateEntityWithID :one
 INSERT INTO entity_instances (
     id,
@@ -20,12 +24,17 @@ INSERT INTO entity_instances (
     project_id,
     provider_id,
     originated_from
-) VALUES ($1, $2, $3, sqlc.arg(project_id), sqlc.arg(provider_id), sqlc.narg(originated_from))
+) VALUES (
+    sqlc.arg(id), 
+    sqlc.arg(entity_type), 
+    sqlc.arg(name), 
+    sqlc.arg(project_id), 
+    sqlc.arg(provider_id), 
+    sqlc.narg(originated_from)
+)
 RETURNING *;
 
-
 -- CreateOrEnsureEntityByID adds an entry to the entity_instances table if it does not exist, or returns the existing entry.
-
 -- name: CreateOrEnsureEntityByID :one
 INSERT INTO entity_instances (
     id,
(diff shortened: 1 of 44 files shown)


internal/db/entity_helpers.go
```diff
@@ -116,8 +122,17 @@ func (q *Queries) GetPropertyValueV1(ctx context.Context, entityID uuid.UUID, ke
...
+	ctx context.Context,
+	entityID uuid.UUID,
+	projectID uuid.UUID) ([]PropertyValueV1, error) {
+
+	projects := []uuid.UUID{projectID}
+
+	dbProps, err := q.GetAllPropertiesForEntity(ctx, GetAllPropertiesForEntityParams{
+		EntityID: entityID,
+		Projects: projects,
+	})
```
evankanderson: I don't recall, but these might have just been deleted in #6284

internal/entities/handlers/strategies/entity/refresh_by_id.go
```diff
@@ -45,11 +45,22 @@ func (r *refreshEntityByIDStrategy) GetEntity(
 	}
 
 	getEnt, err := db.WithTransaction(r.store, func(t db.ExtendQuerier) (*models.EntityWithProperties, error) {
+		rawEnt, err := t.GetEntityByID(ctx, db.GetEntityByIDParams{
+			ID:       entMsg.Entity.EntityID,
+			Projects: []uuid.UUID{},
```
evankanderson: Is this going to work with an empty set of allowed project IDs?

internal/entities/properties/service/helpers.go
```diff
@@ -32,13 +32,19 @@ func (ps *propertiesService) retrieveAllPropertiesForEntity(
 	var dbProps []db.Property
 	if entID != uuid.Nil {
 		// fetch properties from db
-		var err error
-		dbProps, err = qtx.GetAllPropertiesForEntity(ctx, entID)
+		projects, err := ps.getAuthorizedProjects(ctx, projectID)
```
evankanderson: Is this called in a loop?  I'm concerned about potentially generating a lot of project tree lookups if this is (for example) called for each entity in "repo list" or "profile status list --detailed".

internal/entities/service/service.go
```diff
@@ -190,7 +190,7 @@ func (s *entityService) GetEntityByID(
 	projectID uuid.UUID,
 ) (*pb.EntityInstance, error) {
 	// Get entity from database
-	entity, err := s.store.GetEntityByID(ctx, entityID)
+	ewp, err := s.propSvc.EntityWithPropertiesByID(ctx, entityID, projectID, nil)
```
evankanderson: +1 for switching this to the service call rather than straight DB call.

evankanderson (commented):
It looks like you may also have a merge conflict, but this looks good other than a few small comments -- I'd like to avoid adding remote (serial) lookups in loop contexts -- I didn't look in enough detail to know for sure here if that was happening..

internal/db/entity_helpers.go
```diff
@@ -81,6 +81,80 @@ func (q *Queries) UpsertPropertyValueV1(ctx context.Context, params UpsertProper
 	return q.UpsertProperty(ctx, dbParams)
 }
 
+// PropertyValueV1 is a property value for an entity
+type PropertyValueV1 struct {
+	ID        uuid.UUID `json:"id"`
+	EntityID  uuid.UUID `json:"entity_id"`
+	Key       string    `json:"key"`
+	Value     any       `json:"value"`
+	UpdatedAt time.Time `json:"updated_at"`
```
evankanderson: Do you mean to add these back after they were removed?

evankanderson: (It may be that the solution to the bug was removing these wrappers, which happened in #6284

internal/entities/handlers/strategies/entity/refresh_by_id.go
evankanderson: This is still an empty list of project IDs to match.

evankanderson (commented):
I think you just force-pushed to the wrong branch.  :-/

DharunMR:
> I think you just force-pushed to the wrong branch. :-/

Unfortunatly! now solved it

internal/entities/handlers/strategies/entity/refresh_by_id.go
DharunMR: now it call GetEntityByIDUnauth first. This internal only, lookup allows the worker to retrieve the entity and discover its ProjectID. Once we have that ProjectID, we pass it into EntityWithPropertiesByID.

internal/entities/properties/service/helpers.go
DharunMR: now discovering the ProjectID once and using it to satisfy the security check in a single pass

internal/db/entity_helpers.go
DharunMR: I rebased now

DharunMR:
@evankanderson i've verified the tests pass locally ,but fails ci

internal/entities/properties/service/service.go
```diff
@@ -318,7 +340,12 @@ func (ps *propertiesService) EntityWithPropertiesByUpstreamHint(
 		return nil, fmt.Errorf("failed to get entity ID: %w", err)
 	}
 
-	ewp, err := ps.getEntityWithProperties(ctx, *ent, opts)
+	projects, err := ps.getAuthorizedProjects(ctx, ent.ProjectID)
```
evankanderson: Verified that this is only called by `getEntityInner`, which is only called by the strategies in `handlers/strategies/entity` for handling updates of resources which were auto-created due to parent resources.

evankanderson: (In response to the question whether this was called in a loop)

internal/entities/service/service.go
```diff
@@ -199,13 +195,13 @@ func (s *entityService) GetEntityByID(
 	}
 
 	// Verify entity belongs to the project
-	if entity.ProjectID != projectID {
+	if ewp.Entity.ProjectID != projectID {
 		return nil, status.Errorf(codes.NotFound, "entity not found in project")
 	}
```
evankanderson: We shouldn't need this anymore since `EntityWithPropertiesByID` already ensures this, right?

internal/entities/service/service.go
```diff
@@ -199,13 +195,13 @@ func (s *entityService) GetEntityByID(
...
 
 	// Get properties
-	ewp, err := s.propSvc.EntityWithPropertiesByID(ctx, entityID, nil)
-	if err != nil {
+	if err := s.propSvc.RetrieveAllPropertiesForEntity(ctx, ewp, s.providerManager,
+		propService.ReadBuilder().TolerateStaleData()); err != nil {
 		return nil, fmt.Errorf("error fetching properties for entity: %w", err)
 	}
```
evankanderson, line 213: Why does this add `RetrieveAllPropertiesForEntity` here when we called `EntityWithPropertiesForID` earlier?

internal/entities/service/service.go
```diff
@@ -331,3 +330,16 @@ func entityInstanceToProto(ewp *models.EntityWithProperties) *pb.EntityInstance
...
+	if err != nil {
+		return nil, err
+	}
+
+	projectIDs := []uuid.UUID{projectID}
+	for _, child := range children {
+		projectIDs = append(projectIDs, child.ID)
+	}
+	return projectIDs, nil
+}
```
evankanderson: This seems duplicated with `properties/service`, which has an identical method.  Given that this module depends on `properties/service`, what do you think about exporting the method (with a capital "Get") from there and using it here?

evankanderson (commented):
One small concern about duplicating `getAuthorizedProjects` logic twice, and some code that I think is functionally dead and should probably be removed.

internal/entities/service/service.go
DharunMR: changed it now

internal/entities/service/service.go
DharunMR: done👍

internal/entities/properties/service/service.go
DharunMR: I had actually removed getAuthorizedProjects in my latest commit

evankanderson:
(This has a merge conflict again, sorry!)

DharunMR:
> (This has a merge conflict again, sorry!)

resolved it now

database/query/entities.sql
```diff
@@ -34,120 +43,165 @@ INSERT INTO entity_instances (
...
-    WHERE entity_instances.entity_type = $1
-        AND entity_instances.id > $2
+    WHERE entity_type = sqlc.arg(entity_type)
+      AND id > sqlc.arg(id)
+      AND provider_id = sqlc.arg(provider_id)
+      AND project_id = ANY(sqlc.arg(projects)::uuid[])
 ) AS exists;
 
--- GetEntitiesByProvider retrieves all entities of a given provider.
+-- GetEntitiesByProvider retrieves all entities of a given provider scoped by project hierarchy.
```
evankanderson, line 113: Providers are already scoped to a project (and its children, though we don't have all the hierarchy decisions ironed out yet).

database/query/entity_execution_lock.sql
```diff
@@ -51,8 +51,21 @@ RETURNING *;
 
 -- name: FlushCache :one
 DELETE FROM flush_cache
```
evankanderson, line 53: I don't believe there's a user-controlled way to affect the entity cache, so it's not clear we need to enforce project checks here (and it may add cost).

internal/controlplane/handlers_artifacts.go
```diff
@@ -94,7 +94,7 @@ func (s *Server) GetArtifactByName(ctx context.Context, in *pb.GetArtifactByName
 	}
 
 	// Fetch the entity with properties
-	ewp, err := s.props.EntityWithPropertiesByID(ctx, entities[0].ID, nil)
+	ewp, err := s.props.EntityWithPropertiesByID(ctx, entities[0].ID, projectID, provider.ID, nil)
```
evankanderson, line 97: FWIW, it looks like we've already constrained the entity to the current project and provider on line 78-85.

internal/controlplane/handlers_artifacts.go
```diff
@@ -137,26 +137,30 @@ func (s *Server) GetArtifactByName(ctx context.Context, in *pb.GetArtifactByName
...
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, util.UserVisibleError(codes.NotFound, "provider not found")
+		}
+		return nil, status.Errorf(codes.Internal, "cannot get provider: %v", err)
+	}
+
 	// Fetch artifact entity
-	ewp, err := s.props.EntityWithPropertiesByID(ctx, parsedArtifactID, nil)
+	ewp, err := s.props.EntityWithPropertiesByID(ctx, parsedArtifactID, projectID, provider.ID, nil)
```
evankanderson, line 156: It looks like we're adding some database queries to replace the lines 155-158.  I'm not sure how I feel about that.

internal/controlplane/handlers_entity_instances.go
```diff
@@ -95,15 +95,26 @@ func (s *Server) GetEntityById(
...
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, util.UserVisibleError(codes.NotFound, "provider not found")
+		}
+		return nil, fmt.Errorf("error getting provider: %w", err)
+	}
 
 	// Call service to get entity
-	entity, err := s.entityService.GetEntityByID(ctx, entityID, projectID)
+	entity, err := s.entityService.GetEntityByID(ctx, entityID, projectID, provider.ID)
```
evankanderson, line 111: It looks like we only check the provider here because of the changes to the DB queries?  We're mostly concerned about the ability of a user in one project to "peek" into another project that they don't have access to -- we assume that users in a given project have equivalent permissions on all the entities in that project (though I supposed this will also now return an error if you try to look up an entity from provider A, but pass provider B's name in the context).

internal/controlplane/handlers_entity_instances.go
```diff
@@ -168,15 +179,26 @@ func (s *Server) DeleteEntityById(
...
 
 	// Call service to delete entity
-	err = s.entityService.DeleteEntityByID(ctx, entityID, projectID)
+	err = s.entityService.DeleteEntityByID(ctx, entityID, projectID, provider.ID)
 	if err != nil {
 		return nil, err
 	}
 
 	// Telemetry logging
+	logger.BusinessRecord(ctx).Provider = providerName // Added provider telemetry!
```
evankanderson, line 201: This comment is strange.

internal/controlplane/handlers_reconciliationtasks.go
```diff
@@ -69,7 +69,7 @@ func (s *Server) CreateEntityReconciliationTask(ctx context.Context,
 	var msg *message.Message
 	var topic string
 
-	msg, err = getRepositoryReconciliationMessage(ctx, s.store, entity.GetId(), entityCtx)
+	msg, err = getRepositoryReconciliationMessage(ctx, s.store, entity.GetId(), entityCtx, dbProvider.ID)
```
evankanderson, line 72: It looks like we're again adding some database lookups to protect against the same limit as 101-105 in the old code.  I think this _might_ be an improvement on making accidental errors, but the cost feels higher than I'd like.

Is there a way that we can split the methods into "safe to be called with user values" and "only to be called after entity validation"?  I suspect we don't have a good way right now because we tend to vend our internal database IDs directly through the API to end users.  🤦

internal/repositories/service.go
```diff
@@ -185,20 +185,37 @@ func (r *repositoryService) ListRepositories(
...
-
-		if err := r.propSvc.RetrieveAllPropertiesForEntity(ctx, ewp, r.providerManager,
-			service.ReadBuilder().WithStoreOrTransaction(qtx).TolerateStaleData()); err != nil {
-			return nil, fmt.Errorf("error fetching properties for repository: %w", err)
+			return nil, fmt.Errorf("failed to convert properties for %s: %w", ent.ID, err)
 		}
 
-		ents = append(ents, ewp)
+		ents = append(ents, models.NewEntityWithProperties(ent, modelProps))
 	}
```
evankanderson, line 219: This looks like a performance refactor (batching property lookups).  That seems handy, but separate from other parts of this PR?

evankanderson (commented):
This PR is pretty "big" (hard for me to fit in my head) -- is there a good way to break it into smaller chunks?

PR #6310, opened 2026-04-09
fix(cli): clarify rule type deletion errors for bundles vs profiles

Aryanburnwal05:
This change improves the CLI output of `minder ruletype delete --all` by clearly distinguishing between:

- Rule types that cannot be deleted because they are part of a bundle
- Rule types that are referenced by existing profiles

It also improves readability by grouping results and refining CLI formatting.

Additionally, unit tests have been added to validate error categorization and parsing logic.

### DIFF
--- a/cmd/cli/app/ruletype/ruletype_delete.go
+++ b/cmd/cli/app/ruletype/ruletype_delete.go
@@ -93,10 +93,10 @@ func deleteCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *gr
 	}
 
 	// Delete the rule types set for deletion
-	deletedRuleTypes, remainingRuleTypes := deleteRuleTypes(ctx, client, rulesToDelete, project)
+	deletedRuleTypes, bundleBlocked, profileBlocked := deleteRuleTypes(ctx, client, rulesToDelete, project)
 
 	// Print the results
-	printDeleteResults(cmd, deletedRuleTypes, remainingRuleTypes)
+	printDeleteResults(cmd, deletedRuleTypes, bundleBlocked, profileBlocked)
 
 	return nil
 }
@@ -106,10 +106,11 @@ func deleteRuleTypes(
 	client minderv1.RuleTypeServiceClient,
 	rulesToDelete []*minderv1.RuleType,
 	project string,
-) ([]string, []ruleTypeBlock) {
+) ([]string, []string, []ruleTypeBlock) {
 
 	var deletedRuleTypes []string
-	var remainingRuleTypes []ruleTypeBlock
+	var bundleBlocked []string
+	var profileBlocked []ruleTypeBlock
 
 	for _, ruleType := range rulesToDelete {
 		_, err := client.DeleteRuleType(ctx, &minderv1.DeleteRuleTypeRequest{
@@ -117,9 +118,14 @@ func deleteRuleTypes(
 			Id:      ruleType.GetId(),
 		})
 		if err != nil {
+			if isBundleError(err) {
+				bundleBlocked = append(bundleBlocked, ruleType.GetName())
+				continue
+			}
+
 			profiles := extractProfiles(err.Error())
 
-			remainingRuleTypes = append(remainingRuleTypes, ruleTypeBlock{
+			profileBlocked = append(profileBlocked, ruleTypeBlock{
 				Name:     ruleType.GetName(),
 				Profiles: profiles,
 			})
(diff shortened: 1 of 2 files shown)


cmd/cli/app/ruletype/ruletype_delete.go
```diff
@@ -168,16 +189,21 @@ func printDeleteResults(
 	}
 }
 
+func isBundleError(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(strings.ToLower(err.Error()), "bundle")
```
evankanderson: What happens if there is a profile named "bundle" or "bundled-policies"?

cmd/cli/app/ruletype/ruletype_delete.go
```diff
@@ -106,60 +106,81 @@ func deleteRuleTypes(
...
 
-	if len(remainingRuleTypes) > 0 {
-		cmd.Println("\nThe following rule type(s) are referenced by existing profiles and were not deleted:")
-		for _, ruleType := range remainingRuleTypes {
-			cmd.Println(ruleType.Name)
+	if len(profileBlocked) > 0 {
+		cmd.Println()
+		cmd.Println("The following rule type(s) are referenced by existing profiles and were not deleted.")
+		for _, ruleType := range profileBlocked {
+			cmd.Println("- " + ruleType.Name)
```
evankanderson: This rewrite (diff) seems unnecessary -- can you revert it (and the similar ones below)?

In general, updating extra lines has a few negative effects:

1. It makes code archaeology (e.g. via "blame" view) harder, because there are more changes to a given line, so it's harder to find the commit which actually changed behavior, rather than simply formatting.
2. Human reviewers need to look at the changed lines to understand the proposed commit.
   * A lot of extra lines make it more likely that humans will overlook critical changes due to inattention / fatigue
   * Even without that, it simply adds time and effort for human reviewers to assess and dismiss the extra changes.

evankanderson (commented):
Thanks for testing your rule!  I think we're likely to add something like #6286 in the future, but this will help keep things stable in the interim.

cmd/cli/app/ruletype/ruletype_delete.go
Aryanburnwal05: Good point thanks for calling this out. You are right that matching on a generic substring like "bundle" could lead to false positives in cases where profile names contain similar terms. Since the current server response doesn't expose structured error types for bundle constraints, I opted for a lightweight heuristic here as an interim solution. That said I can tighten this logic to reduce ambiguity (for example, matching more specific phrases), or adjust based on how error contracts evolve (e.g. #6286).

cmd/cli/app/ruletype/ruletype_delete.go
Aryanburnwal05: Okay I will correct it.

Aryanburnwal05:
Hi @evankanderson PTAL
1. Narrowed bundle error detection to match the specific error string to avoid false positives.
2. Reverted unrelated formatting/output changes to keep the diff minimal.
3. Kept the output structure consistent with existing behavior while adding bundle specific messaging.
If you want me to handle this differently once structured errors are available from the server.

evankanderson (approved):

cmd/cli/app/ruletype/ruletype_delete_test.go
```diff
@@ -0,0 +1,142 @@
...
+			case "rule3":
+				return nil, status.Errorf(codes.FailedPrecondition, "cannot delete: rule type rule3 is used by profiles prof1, prof2")
+			default:
+				return nil, errors.New("unknown error")
+			}
+		},
+	}
+
+	rules := []*minderv1.RuleType{
+		{Id: "rule1", Name: "successful_rule"},
```
evankanderson, line 126: I think `Id`  needs to be a pointer to string

Aryanburnwal05: Okay I will do it in 3 days I am currently out of station

evankanderson: No rush, but wanted to follow up -- I just merged some table tests for the `ruletype` CLI commands, so you may want to leverage those in your PR now.

evankanderson: We've merged the changes in #6286 if you want to go back and update this PR.  Alternatively, you can close it if you prefer.

Aryanburnwal05: Thanks for the update! I’ll review the changes from #6286 and see whether it still makes sense to update this PR or close it. Appreciate the guidance.

PR #6315, opened 2026-04-09
fix(datasource): initialize context to allow fallback to default project

### LINKED ISSUE #5164: minder datasource apply should use default project
### Describe the issue

Invoking `minder datasource apply` on a file without a project id set in the context fails:

```
Message: error applying data source from datasource.yaml
Details: error creating data source from datasource.yaml: rpc error: code = InvalidArgument desc = context cannot be nil
exit status 3
``` 
Instead of the error, Minder should use the default project of the logged in user.

### To Reproduce

Apply a data source without a project in the context section

d26452d65750295d9c4450d17e953517caa32074

Jaydeep869:
# Summary

This PR addresses an issue where running `minder datasource apply` on a YAML file lacking a `context` block throws an unexpected `context cannot be nil` gRPC validation error.

The problem stems from the fact that when the YAML is parsed in the CLI, `ds.Context` drops to `nil`. Because the server needs a valid context to gracefully fall back to the user's default project, sending a `nil` payload was crashing the flow. By simply initializing an empty `ContextV2` struct in `cmd/cli/app/datasource/common.go` (if the context resolves to nil), the backend logic kicks in and uses the default project as intended just like other RPCs do!

Fixes #5164

# Testing

- Created a minimal `test-ds.yaml` missing the project/context blocks.
- Verified that running `./bin/minder datasource apply -f test-ds.yaml` successfully triggers the default project fallback without throwing the "context cannot be nil" error.
- Verified changes against the control plane tests and ran `make test` to ensure no other components were broken.

### DIFF
--- a/cmd/cli/app/datasource/common.go
+++ b/cmd/cli/app/datasource/common.go
@@ -53,6 +53,11 @@ func executeOnOneDataSource(
 		ds.Context.ProjectId = proj
 	}
 
+	// Ensure context is not nil so the server can fall back to the default project
+	if ds.Context == nil {
+		ds.Context = &minderv1.ContextV2{}
+	}
+
 	if err := ds.Validate(); err != nil {
 		return fmt.Errorf("error validating data source: %w", err)
 	}


Jaydeep869:
@cla-assistant recheck

Jaydeep869:
@evankanderson I have opened this PR to fix the issue #5164 . Could you please take a look when you have a moment?

evankanderson (approved):

Issue #6316, opened 2026-04-09
feat: store and detect Rego language version per rule type (Phase 1) #5262


dakshhhhh16:
### Problem

Minder's Rego evaluator unconditionally compiles all policies with `ast.RegoV0`, even on OPA v1.14+ where the recommended default is V1. This blocks users from writing modern Rego policies using `if`, `import rego.v1`, `every`, etc., and it means we have no foundation for a gradual V1 migration.

See parent issue: #5262 -- _Rego V1 Support_

### Approach

Instead of a big-bang migration, this phase of the plan establishes the necessary infrastructure by:

1. **Per-rule-type version storage** -- adds a `rego_version` column to the `rule_type` table (defaulting to `'v0'`) so each rule type carries an explicit version marker.
2. **Parse-attempt detection** -- on every `CreateRuleType` / `UpdateRuleType`, the server attempts to parse the Rego definition with the V1 parser first. If it succeeds, `rego_version = 'v1'` is stored; otherwise `'v0'` is stored. No user action required.
3. **Dual-parse validation** -- the protobuf `Validate()` path now tries V1 first and falls back to V0, so valid V1 or V0 policies both pass validation.
4. **Feature-flag-gated evaluation** -- when `rego_v1_dual_parse` is enabled, the Rego evaluator reads the stored version and passes it via `rego.SetRegoVersion()` rather than hardcoding V0.

### Changes

| Area | File(s) | What changed |
|---|---|---|
| DB migration | `database/migrations/000117_rule_type_rego_version.{up,down}.sql` | Adds `rego_version TEXT NOT NULL DEFAULT 'v0'` to `rule_type` |
| SQL / sqlc | `database/query/rule_types.sql`, `internal/db/models.go`, `internal/db/rule_types.sql.go` | `rego_version` wired through Create/Update queries; regenerated |
| Mock | `database/mock/store.go` | Regenerated via mockgen |
| Version helpers | `internal/engine/eval/rego/version.go` | `DetectRegoVersion()`, `VersionToString()`, `VersionFromString()` |
| Evaluator | `internal/engine/eval/rego/eval.go` | `WithRegoVersion` option + evaluator stores/uses version |
| Cache | `internal/engine/rtengine/cache.go` | Reads stored version behind `rego_v1_dual_parse` flag |
| Service | `pkg/ruletypes/service.go` | Calls `DetectRegoVersion()` on create/update |
[...]

Issue #5262, opened 2025-01-07
Figure out a Rego V1 migration plan


evankanderson:
Open Policy Agent has released a v1.0.0.  This includes making a number of features which were previously available under the `import rego.v1` or other feature flags.  (Yay!)  The `rego.v1` flag **also** removes or deprecates a number of syntax items to clean up the overall ecosystem.

Unfortunately, a number of the existing rules in https://github.com/mindersec/minder-rules-and-profiles are written using the "rego.v0" syntax, and fail to compile under the v1 rules.  We need to figure out a plan for migrating existing usage from v0 to v1, probably with cooperation from users (in terms of adding a new field for rego syntax version or "regov1" evaluator), along with controls on which versions of Rego are accepted by existing projects.

☝️  The above is not a plan, we need to actually document a plan, including migration instructions.

evankanderson:
Related: #5251 (the upgrade to OPA v1.0.0)

dakshhhhh16:
# Rego V1 Migration Plan -- Issue #5262

## Current State

| Area | Status |
|------|--------|
| **OPA Go SDK** | v1.14.0 -- already on `github.com/open-policy-agent/opa/v1/...` import path |
| **Hardcoded V0** | 2 locations: [`eval.go:144`](https://github.com/mindersec/minder/blob/main/internal/engine/eval/rego/eval.go#L144) (`rego.SetRegoVersion(ast.RegoV0)`) and [`validators.go:256`](https://github.com/mindersec/minder/blob/main/pkg/api/protobuf/go/minder/v1/validators.go#L256) (`ast.ParserOptions{RegoVersion: ast.RegoV0}`) |
| **`minder dev ruletype init`** | Already generates `import rego.v1` in templates ✓ |
| **`minder dev ruletype lint`** | Already integrates Regal linter via `validateRegoRule()` ✓ |
| **minder-rules-and-profiles** | 65 rule types: ~29 already v1-ready, ~15 pure v0, ~15 `future.keywords`, ~3 mixed |
| **Test embedded policies** | ~49 use v0 syntax, ~3 use v1 |

### Rego V1 Breaking Changes (Quick Reference)
- `if` keyword required in rules: `allow { ... }` → `allow if { ... }`
- `:=` for single-value rules: `default allow = false` → `default allow := false`
- `contains` keyword for partial set rules: `deny[msg] { ... }` → `deny contains msg if { ... }`
- `import future.keywords` deprecated in favor of `import rego.v1`
[...]

PR #6317, opened 2026-04-09
feat: store and detect Rego language version per rule type (Phase 1)

dakshhhhh16:
Parent Issue #5262
Sub-issue #6316 

## Problem

Minder's Rego evaluator unconditionally compiles all policies with `ast.RegoV0`, even on OPA v1.14+ where the recommended default is V1. This blocks users from writing modern Rego (`if`, `import rego.v1`, `every`, etc.) and means there is no foundation for a gradual V1 migration.

## Approach

Instead of a big-bang migration, this phase establishes the necessary infrastructure:

1. **Per-rule-type version storage** -- adds a `rego_version` column to the `rule_type` table (default `'v0'`).
2. **Parse-attempt detection** -- on `CreateRuleType` / `UpdateRuleType`, the server tries the V1 parser first. If it succeeds, stores `'v1'`; otherwise stores `'v0'`. No user action required.
3. **Dual-parse validation** -- `Validate()` tries V1 first and falls back to V0, so both syntaxes pass.
4. **Feature-flag-gated evaluation** -- when `rego_v1_dual_parse` is enabled, the evaluator reads the stored version instead of hardcoding V0.

No existing behaviour changes without the feature flag.

## Out of scope (future phases)

- Backfilling existing rule types
- Surfacing `rego_version` in the API / CLI
- V0 deprecation warnings (`rego_v1_warn_v0`)
- V0 refusal (`rego_v1_refuse_v0`)

### DIFF
--- a/internal/engine/rtengine/cache.go
+++ b/internal/engine/rtengine/cache.go
@@ -13,6 +13,7 @@ import (
 
 	datasourceservice "github.com/mindersec/minder/internal/datasources/service"
 	"github.com/mindersec/minder/internal/db"
+	regoeval "github.com/mindersec/minder/internal/engine/eval/rego"
 	"github.com/mindersec/minder/internal/engine/ingestcache"
 	eoptions "github.com/mindersec/minder/internal/engine/options"
 	"github.com/mindersec/minder/pkg/engine/v1/interfaces"
@@ -163,6 +164,14 @@ func cacheRuleEngine(
 	}
 
 	opts = append(opts, eoptions.WithDataSources(dsreg), eoptions.WithFlagsClient(featureFlags))
+
+	// Pass the stored Rego version to the evaluator when the dual-parse
+	// feature flag is enabled. When the flag is off, the evaluator defaults
+	// to Rego V0 (the zero value of ast.RegoVersion).
+	if flags.Bool(ctx, featureFlags, flags.RegoV1DualParse) {
+		opts = append(opts, regoeval.WithRegoVersion(
+			regoeval.VersionFromString(ruleType.RegoVersion)))
+	}
 
 	// Create the rule type engine
 	ruleEngine, err := rtengine2.NewRuleTypeEngine(ctx, pbRuleType, provider, opts...)
--- a/pkg/api/protobuf/go/minder/v1/validators.go
+++ b/pkg/api/protobuf/go/minder/v1/validators.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/go-playground/validator/v10"
 	"github.com/itchyny/gojq"
-	"github.com/open-policy-agent/opa/v1/ast"
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/mindersec/minder/internal/util"
@@ -251,12 +250,9 @@ func (rego *RuleType_Definition_Eval_Rego) Validate() error {
 		return fmt.Errorf("%w: rego definition is empty", ErrInvalidRuleTypeDefinition)
 	}
 
-	// TODO: figure out a Rego V1 migration path (https://github.com/mindersec/minder/issues/5262)
-	_, err := ast.ParseModuleWithOpts("minder-ruletype-def.rego", rego.Def,
-		ast.ParserOptions{RegoVersion: ast.RegoV0})
-	if err != nil {
-		return fmt.Errorf("%w: rego definition is invalid: %s", ErrInvalidRuleTypeDefinition, err)
-	}
+	// Rego language-level validation (syntax checking, version gating) is
+	// intentionally handled server-side in ruletypes/service.go so that it
+	// can be controlled via feature flags without requiring client updates.
 
 	return nil
 }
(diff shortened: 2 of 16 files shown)


pkg/api/protobuf/go/minder/v1/validators_test.go
evankanderson: Can you change the name of this test to something like "valid v1 rego expression"?

internal/engine/rtengine/cache.go
```diff
@@ -164,6 +165,14 @@ func cacheRuleEngine(
...
 	opts = append(opts, eoptions.WithDataSources(dsreg), eoptions.WithFlagsClient(featureFlags))
 
+	// Pass the stored Rego version to the evaluator when the dual-parse
+	// feature flag is enabled. When the flag is off, the evaluator defaults
+	// to Rego V0 (the zero value of ast.RegoVersion).
+	if flags.Bool(ctx, featureFlags, flags.RegoV1DualParse) {
+		opts = append(opts, regoeval.WithRegoVersion(
+			regoeval.VersionFromString(ruleType.RegoVersion)))
+	}
+
```
evankanderson, line 175: While we may need this here, I think we also need the flag to protect the insert in `ruletypes/service.go` -- otherwise, the server might accept a `v1` rule that it will not be able to parse and execute.

(And, I think it would try to handle it as a `v0` rule...)

evankanderson (commented):
Thanks, this is looking pretty close, though I'll want to do some manual testing with flag on/off if you haven't already.  I *suspect* that this is a one-way flag -- once you've enabled "allow rego v1 dual-parse", some (enumerable but not easily known) set of your ruletypes in the Minder database will stop evaluating correctly when the flag is disabled.

dakshhhhh16:
Damn! Even the security check didn't fail this time!\.

pkg/api/protobuf/go/minder/v1/validators.go
```diff
@@ -251,9 +251,14 @@ func (rego *RuleType_Definition_Eval_Rego) Validate() error {
...
 	_, err := ast.ParseModuleWithOpts("minder-ruletype-def.rego", rego.Def,
-		ast.ParserOptions{RegoVersion: ast.RegoV0})
+		ast.ParserOptions{RegoVersion: ast.RegoV1})
+	if err != nil {
+		_, err = ast.ParseModuleWithOpts("minder-ruletype-def.rego", rego.Def,
+			ast.ParserOptions{RegoVersion: ast.RegoV0})
+	}
 	if err != nil {
 		return fmt.Errorf("%w: rego definition is invalid: %s", ErrInvalidRuleTypeDefinition, err)
 	}
```
evankanderson: We should move this validation to `ruletypes/service.go`, rather than having it happen on the client as well as the server.

Both `CreateRuleType` and `UpdateRuleType` should return an `error` if the `rego_v1_dual_parse` flag is false and the user supplies a string that can only be parsed in the V1 dialect.  (We should add a test for this, either in `ruletypes/services_test.go` or in `controlplane/handlers_ruletype_test.go`)


We want to move the validation of the rego language out of the protobuf validator, because doing the validation on the client as well as the server means that users will need to update their `minder` binary to take advantage of new features -- we'd prefer to simply be able to light them up from the server side.  (We've had this problem before for other features, like not needing to specify `def: {}` on each rule in a policy.)

dakshhhhh16:
Updated!

pkg/ruletypes/service.go
```diff
@@ -400,3 +456,5 @@ func getAvailableDataSources(
 
 	return datasources, nil
 }
```
evankanderson, line 458: Lint is complaining about an extra trailing newline.

evankanderson (commented):
This probably deserves a quick manual end-to-end test pass.  Feel free to either post results, or ask me to run it when I get sufficient connectivity.  (Beginning of next week -- I'm on a camping trip this week.)

evankanderson:
Tested this by:

1. Starting minder with the flag off (existing behavior)
1. Loading all existing rule types from `rule-types/github` (this also requires loading all the data sources)
2. Create a second project (test2)
4. Update the default_branch_pr_required rule to remove the `import rego.v1` line
5. Try loading the rule, get the expected error
6. Enable the flag
7. Try applying the v1-only default_branch_pr_required rule in test2
8. Check that the rule is marked as v1 in the database
9. Apply all the ruletypes from `rule-types/github` in the `test2` project (fresh create)
10. Re-apply all the ruletypes from `rule-types/github` in the default project (update).  Most of them loaded, but I got the following error on `workflow_no_pull_request_target.yaml`:
   ```
   Details: error updating rule type from ../minder-rules-and-profiles/rule-types/github/workflow_no_pull_request_target.yaml: rpc error: code = InvalidArgument desc = invalid rule type definition: rule type validation failed: rule schema update is invalid: failed to validate properties: cannot remove properties from object type rule schema
   ```

Rules that were updated in the default project were marked as v1 if appropriate.

Most of the other rules had already loaded successfully as either v1 or v0.

I think we can ship this an iterate.

evankanderson (approved):

PR #6320, opened 2026-04-09
Include rule name in PR branch name to avoid collisions

### LINKED ISSUE #2350: PR "from" branch only uses rule type to generate the name
### Describe the issue

When opening a PR, we need to open a PR from a branch. We derive the name of the branch from the "title" of the PR in the rule type, but that doesn't work well if we have two rules using the same rule type unless there's something temptable in the title. We should take the name of the rule into account as well.

### To Reproduce

Create a profile that uses two codeql rules, see them race.

_No response_

krrish175-byte:
# Summary

When a profile uses two rules that share the same rule type, the PR remediation engine derives the "from" branch name solely from the rule type's title template. This causes both rules to generate identical branch names and race each other — whichever runs second overwrites the first's branch and no separate PR is opened for it.

This fix incorporates the rule instance name into the branch name:

- With a rule name: `minder_<rule-name>_<normalized-title>`
- Without a rule name (empty/legacy): `minder_<normalized-title>` (backward compatible)

No dependencies required.

Fixes #2350

# Testing

- Added `TestBranchBaseName` unit tests covering:
  - Empty rule name falls back to the old format (backward compatibility)
  - Rule name is included when present
  - Two rules with the same title but different names produce distinct branch names
- All existing `TestPullRequestRemediate` table-driven tests updated to pass the rule name argument and continue to pass

### DIFF
--- a/internal/engine/actions/remediate/pull_request/pull_request_test.go
+++ b/internal/engine/actions/remediate/pull_request/pull_request_test.go
@@ -179,6 +179,7 @@ type remediateArgs struct {
 	ent       protoreflect.ProtoMessage
 	pol       map[string]any
 	params    map[string]any
+	ruleName  string
 }
 
 func createTestRemArgs() *remediateArgs {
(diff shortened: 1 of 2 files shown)


internal/engine/actions/remediate/pull_request/pull_request_test.go
```diff
@@ -699,3 +699,49 @@ func TestPullRequestRemediate(t *testing.T) {
 		})
 	}
 }
+
+func TestBranchBaseName(t *testing.T) {
```
evankanderson: Can we extend `TestPullRequestRemediate` to check the rulename case, rather than adding a smaller test?  I'd prefer to have somewhat larger tests that check the integration area between the smaller functions, rather than testing the small functions and then *also* needing to test the larger outer surface area.

evankanderson (commented):
This looks good.  One comment on the tests -- if it turns out to be seriously painful, please flag it rather than toiling a lot for this particular preference.

internal/engine/actions/remediate/pull_request/pull_request_test.go
```diff
@@ -699,3 +727,4 @@ func TestPullRequestRemediate(t *testing.T) {
 		})
 	}
 }
+
```
evankanderson: Lint doesn't like the extra newline at the end of the file.  (And we need to fix it, or lint will trip on it for every subsequent PR.)

evankanderson (dismissed):
One small formatting change, otherwise looks good

krrish175-byte:
@evankanderson PTAL

evankanderson (approved):

PR #6321, opened 2026-04-09
feat: support generic entity instance in selectors

### LINKED ISSUE #5380: Support generic entity instance in selectors
This in turn will add support to the `release` entity in selectors.

With the current implementation, if a selector is added to the release entity, for example:
```
selection:
  - entity: release
    selector: entity.name == 'test-org/public-repo/v8'
```
Then it fails with the message
```
No converter available
```
because the selector logic is written per entity type and there is no support for releases, see https://github.com/mindersec/minder/blob/58972c8c7368b02c190c2317080277c629ad1fc7/internal/providers/selectors/selector_entity.go#L98-L108

Since we are moving towards having a more generic entity type, using the `EntityInstance` type, we should also make the selector logic more generic.

Jaydeep869:
# Summary

This PR adds complete support for generic entity instances within selectors. 

Previously, if a selector was added to an unmapped entity (like `release` or `build`), the evaluation failed with a `No converter available` error because `newConverter` in `selector_entity.go` relied on a strict switch statement covering only `repository`, `artifact`, and `pull_request`. While PR #5379 introduced the generic environment fallback, `newConverter` still returned `nil` for unsupported entities.

This change introduces a `genericToSelectorEntity` fallback in the converter factory, populating a new `SelectorGeneric` protobuf message that dynamically exposes all the entity properties. This ensures that properties for generalized entities correctly pipe into CEL environments without breaking.

Fixes #5380

# Testing

- Added `SelectorGeneric` to `internal/proto/internal.proto` and regenerated the protobuf definitions using `make buf`.
- Created a `"Generic Entity Type"` test case in `internal/providers/selectors/selector_entity_test.go` to ensure unmapped objects successfully populate their properties map through the generic constructor.
- Added an integration scenario `"Simple true generic release expression"` in `pkg/engine/selectors/selectors_test.go` to verify that CEL expressions correctly evaluate fields on generic entities (e.g., `generic.name == 'testorg/testrelease'`).
- Executed `make test` locally to confirm all suite assertions pass successfully.
- including regressions pass cleanly. Run successfully.
- Executed `make lint` and `go fmt` to ensure compliance with styling guidelines.


evankanderson:
(will look tomorrow, it's getting late)

internal/proto/internal.proto
```diff
@@ -126,6 +126,12 @@ message SelectorPullRequest {
   google.protobuf.Struct properties = 2;
 }
 
+message SelectorGeneric {
+  string name = 1;
+  SelectorProvider provider = 2;
```
evankanderson: Why do we have `provider` here as well as in the `SelectorEntity` message?

internal/providers/selectors/selector_entity_test.go
```diff
@@ -234,11 +234,26 @@ func TestEntityToSelectorEntity(t *testing.T) {
...
+			},
+			expSelEnt: &internalpb.SelectorEntity{
+				EntityType: minderv1.Entity_ENTITY_BUILD,
+				Name:       "testorg/testbuild",
+				Entity: &internalpb.SelectorEntity_Generic{
+					Generic: &internalpb.SelectorGeneric{
+						Name: "testorg/testbuild",
+					},
+				},
+			},
```
evankanderson, line 251: Why isn't provider set for any of these fields?  (Just curious)

evankanderson (commented):
Does this need some manual integration testing to ensure that the selector actually works in a Profile?

internal/proto/internal.proto
Jaydeep869: I noticed the provider pattern in SelectorArtifact and SelectorRepository, but since SelectorEntity itself holds the Provider for all its variants, the inner struct doesn't need it. I've removed it from SelectorGeneric. We could probably clean it up from the other entity struct messages in a separate technical debt PR. What do you say?

internal/providers/selectors/selector_entity_test.go
Jaydeep869: The TestEntityToSelectorEntity unit tests focus strictly on the newConverter factory's entity property extraction. The test fixtures provide a nil selProv argument to keep the setup focused merely on asserting the core properties structure, which is why the provider is omitted from the expected structures here.

Jaydeep869:
> Does this need some manual integration testing to ensure that the selector actually works in a Profile?

@evankanderson  There is an automated integration test in `selectors_test.go`  ('Simple true generic release expression') that evaluates an end to end CEL environment condition successfully matching a generic ENTITY_RELEASE payload. This verifies the selector correctly works within a profile context directly inside the engine evaluation boundary. I can also do a localized manual run if you'd like an extra sanity check!

evankanderson:
> > Does this need some manual integration testing to ensure that the selector actually works in a Profile?
> 
> @evankanderson There is an automated integration test in `selectors_test.go` ('Simple true generic release expression') that evaluates an end to end CEL environment condition successfully matching a generic ENTITY_RELEASE payload. This verifies the selector correctly works within a profile context directly inside the engine evaluation boundary. I can also do a localized manual run if you'd like an extra sanity check!

Generally, for changes that are intended to be user-visible, it's a good idea to run a manual test of the functionality unless there's a non-mocked integration test (I often use `grpcurl` or `curl` and the grpc-gateway for API changes, and check the database for manual changes).  Unfortunately, it's easy to get convinced about the full behavior of the system when you're looking at a small targeted change, and miss the implications in other parts of the code (e.g. "the client prevents this" or "layer X needs corresponding changes").

evankanderson:
Additionally, doing a manual check can help with getting examples to update documentation.  😁

Jaydeep869:
@evankanderson I just ran through a manual end to end check using a local Minder instance. I created a profile with a release selector using the CLI, and the API accepted it perfectly. The engine evaluation worked correctly without any plumbing layers or client filters getting in the way.

Taking your advice, I've also just pushed a small documentation update to `profile_selectors.md `to include a working CEL example and the available generic properties so users know exactly how to write selectors for these new unmapped entities.

evankanderson (approved):

PR #6326, opened 2026-04-10
fix: make ReplaceAllProperties atomic

### LINKED ISSUE #4423: Ensure a transaction is open in `ReplaceAllProperties` routine
[This routine](https://github.com/stacklok/minder/blob/44d1523d784f4684439c85f8863013416025204a/internal/entities/properties/service/service.go#L266-L275) must ensure that a transaction is open before performing the delete/insert sequence of statements.

kazuhidelee:
# Summary

ReplaceAllProperties runs a delete followed by a series of upserts. Before the change, if no transaction was provided with opts, getStoreOrTransaction returned the plain store and Postgres ran each statement in autocommit mode. so a failed upsert in middle would leave the entity with no properties at all.
The fix wraps the delete/insert sequence in a transaction when the caller hasn't already provided one, and reuses the existing querier when they have. 

To keep the code organized and avoid duplication, I used helper functions for this implementation:
- replaceAllPropertiesWithQuerier() for the delete-then-reinsert sequence
- saveAllPropertiesWithQuerier() for the shared property upsert loop

Fixes #4423

# Testing
New tests cover: 
- transaction is opened when called with plain store
- existing transaction querier is reused
-  transaction wrapper errors surface correctly
-  original properties are preserved when a mid-reinsertion failure triggers rollback.


internal/entities/properties/service/service.go
```diff
@@ -180,19 +180,37 @@ func (ps *propertiesService) ReplaceAllProperties(
...
-	return ps.SaveAllProperties(ctx, entityID, props, opts)
+	return ps.replaceAllPropertiesWithQuerier(ctx, entityID, props, qtx)
 }
 
 func (ps *propertiesService) SaveAllProperties(
 	ctx context.Context, entityID uuid.UUID, props *properties.Properties,
 	opts *CallOptions,
 ) error {
 	qtx := ps.getStoreOrTransaction(opts)
+	return ps.saveAllPropertiesWithQuerier(ctx, entityID, props, qtx)
```
evankanderson, line 204: Do you want to leave `SaveAllProperties` exposed without a TX guard?

kazuhidelee: Thanks for the feedback. I updated the `SaveAllProperties` to use the same transaction guard pattern when it runs against the plain store. And also added unit tests that tests `SaveAllProperties` as well.

internal/entities/properties/service/service.go
```diff
@@ -180,19 +180,44 @@ func (ps *propertiesService) ReplaceAllProperties(
 	qtx := ps.getStoreOrTransaction(opts)
 	zerolog.Ctx(ctx).Debug().Str("entityID", entityID.String()).Msg("replacing all properties")
 
-	err := qtx.DeleteAllPropertiesForEntity(ctx, entityID)
-	if err != nil {
-		return fmt.Errorf("failed to delete properties: %w", err)
+	if store, ok := qtx.(db.Store); ok {
+		return store.WithTransactionErr(func(qtx db.ExtendQuerier) error {
+			return ps.replaceAllPropertiesWithQuerier(ctx, entityID, props, qtx)
```
evankanderson, line 185: Since this is calling the same function (`replaceAllPropertiesWithQuerier`), it feels like it might be more natural to update qtx and continue, rather than early-returning:


```suggestion
	if store, ok := qtx.(db.Store); ok {
		tx, err := store.BeginTransaction()
		if err != nil {
		  return err
		}
		qtx = store.GetQuerierWithTransation(tx)
		opts = append(opts, CallBuilder().WithStoreOrTransaction(tx))
	}
	err := qtx.DeleteAllPropertiesForEntity(ctx, entityID)
	if err != nil {
		fmt.Errorf("failed to delete properties: %w", err)
	}
	
	return ps.SaveAllProperties(ctx, entityID, props, opts)
```

(and then not needing to add the additional functions)

evankanderson (commented):
Sorry for the extra feedback; re-reading this, I noticed the duplicated calls to the helper method, and it seemed simpler to not need a helper at all and just update the querier and options.

internal/entities/properties/service/service.go
kazuhidelee: Thanks, I agree. I updated both ReplaceAllProperties and SaveAllProperties to upgrade qtx to a transaction-backed querier when starting from the plain store then continue through a single code path.

internal/entities/properties/service/service.go
```diff
@@ -180,19 +180,62 @@ func (ps *propertiesService) ReplaceAllProperties(
...
 
-	err := qtx.DeleteAllPropertiesForEntity(ctx, entityID)
-	if err != nil {
+	if store, ok := qtx.(db.Store); ok {
+		tx, err := store.BeginTransaction()
+		if err != nil {
+			return err
+		}
+		defer func() {
+			_ = store.Rollback(tx)
```
evankanderson: Ah, I'd missed the transaction management earlier.  Feel fere to push back when my suggestions don't actually work the way I suggest.  😁   In this case, we're duplicating the actual logic, which is also getting lost in the transaction-handling goop.

I think either the inner function (your original solution) or (living dangerously) recursively calling the same function might be better than this duplication.  I'd like to be able to suggest a transaction helper, but given the requirements, I worry that it would be too complex to use:

```golang
func ensureTransaction(s db.ExtendQuerier) (db.ExtendQuerier, complete func(), cancel func(), error) {
	store, ok := s.(db.Store)
	// If we can't cast, assume we have a transaction
	if !ok {
		return s, func() {}, func() {}, nil
	}
	tx, err := s.BeginTransaction()
	if err != nil {
		return nil, nil, nil, err
	}
	return store.GetQuerierWithTransaction(tx), func() {store.Commit(tx), func() {_ = store.Rollback(tx)}, nil
}
```

I'd worry about the ability to use this correctly, so I think your original instinct would be more correct:

```golang
if store, ok := qtx.(db.Store); ok {
	return store.WithTransactionErr(func(querier db.ExtendQuerier) error {
		opts = append(opts, ps.CallBuilder().WithStoreOrTransaction(querier))
		// recursive call, but now with a transaction
		return ReplaceAllProperties(ctx, entityID, props, opts)
	})
}
```

Considering these options, I think your original `d674132` is probably the best option.  Sorry for leading you astray!

evankanderson (commented):
Really sorry for missing the extra transaction management handled by `WithTransactionErr`.  Seeing what I requested, I sort of facepalmed at myself.

Let's merge https://github.com/mindersec/minder/commit/d674132108420afcefe2826e447bb6cc1568a75e from before yesterday's comment.

internal/entities/properties/service/service.go
kazuhidelee: Sounds good :) , I reverted this back to the earlier helper-based WithTransactionErr flow so we avoid duplication.

evankanderson (approved):

evankanderson (approved):

PR #6327, opened 2026-04-10
Add provider class metadata listing

### LINKED ISSUE #4821: Allow listing provider classes
Currently provider classes are not discoverable. It would be great to have an API endpoint to list the provider classes available in a given Minder deployment. This would make discovering available providers that can be enrolled a lot easier.

Jaydeep869:
# Summary

This PR improves provider class discovery by extending ListProviderClasses to return structured metadata while preserving backward compatibility.

Changes included:
1. Added ProviderClassInfo in the API and updated ListProviderClassesResponse to include rich class metadata and keep legacy provider_classes.
2. Updated provider class listing handler to return:
3. legacy class names for compatibility
4. rich metadata for each class (display name, description, supported provider types, auth flows, supported entities, documentation URL, creation help).
5. Made class listing deployment aware by deriving enabled classes from registered provider managers.
6. Centralized provider class metadata in provider definitions for consistency.
7. Added control plane tests to validate compatibility and rich metadata behavior.
8. Regenerated protobuf, OpenAPI, and docs reference artifacts.

Dependencies:
1. None

## Fixes #4821

# Testing

1. make lint-go
2. make test


proto/minder/v1/minder.proto
```diff
@@ -3639,9 +3639,50 @@ message ListProviderClassesRequest {
...
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // display_name is a human-friendly provider class name.
+    string display_name = 2 [
+        (google.api.field_behavior) = REQUIRED
+    ];
+
+    // description is a short summary of the provider class.
+    string description = 3;
```
evankanderson: Is this plaintext or markdown?

proto/minder/v1/minder.proto
```diff
@@ -3639,9 +3639,50 @@ message ListProviderClassesRequest {
...
+    ];
+
+    // supported_entities is the list of entity types supported by this provider class.
+    repeated Entity supported_entities = 6;
+
+    // documentation_url points to provider-specific or generic documentation.
+    string documentation_url = 7;
+
+    // creation_help is high-level guidance for enrolling this provider class.
+    string creation_help = 8;
```
evankanderson: How does this differ from `description` or the content in `documentation_url`?  Where would it be used?

internal/providers/provider_definitions.go
```diff
@@ -10,33 +10,89 @@ import (
 	"github.com/mindersec/minder/internal/providers/dockerhub"
 	ghclient "github.com/mindersec/minder/internal/providers/github/clients"
 	"github.com/mindersec/minder/internal/providers/gitlab"
+	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+)
+
+const (
+	providerDocsBaseURL   = "https://mindersec.github.io/docs"
```
evankanderson: This should switch to `https://docs.mindersec.dev/`, I think

internal/providers/provider_definitions.go
```diff
@@ -10,33 +10,89 @@ import (
...
+		},
 	},
 	dockerhub.DockerHub: {
+		DisplayName:        "Docker Hub",
+		Description:        "Docker Hub registry provider for image and OCI interactions.",
+		DocumentationURL:   providerDocsGeneric,
+		CreationHelp:       "Provide Docker Hub credentials and enroll using --class dockerhub.",
 		Traits:             dockerhub.Implements,
 		AuthorizationFlows: dockerhub.AuthorizationFlows,
+		SupportedEntities:  nil,
```
evankanderson: I think this was supposed to support artifacts, but maybe it doesn't today.

internal/providers/provider_definitions.go
```diff
@@ -10,33 +10,89 @@ import (
...
+	DisplayName        string
+	Description        string
+	DocumentationURL   string
+	CreationHelp       string
 	Traits             []db.ProviderType
 	AuthorizationFlows []db.AuthorizationFlow
+	SupportedEntities  []minderv1.Entity
 }
 
 var supportedProviderClassDefinitions = map[string]ProviderClassDefinition{
```
evankanderson: Rather than having a map here, can we extend the `Provider` interface in `pkg/providers/v1/providers.go` to return this information?  That would mean that each provider author was responsible for providing this information (and that the provider could be statically checked for meeting the interface).

internal/controlplane/handlers_providers.go
```diff
@@ -166,16 +171,107 @@ func (s *Server) ListProviders(ctx context.Context, req *minderv1.ListProvidersR
...
-	_ context.Context, _ *minderv1.ListProviderClassesRequest,
+func (s *Server) ListProviderClasses(
+	ctx context.Context, _ *minderv1.ListProviderClassesRequest,
 ) (*minderv1.ListProviderClassesResponse, error) {
-	// Note: New provider classes should be added to the providers package.
-	classes := providers.ListProviderClasses()
+	// Class definitions are the canonical metadata source; runtime manager state
+	// decides which classes are enabled in this deployment.
+	classDefinitions := providers.ListProviderClassDefinitions()
+	classes := listEnabledProviderClasses(s.providerManager, classDefinitions)
```
evankanderson: This feels like it should be a method based on state in the ProviderManager, not some external data passed into the manager.

internal/controlplane/handlers_providers.go
```diff
@@ -166,16 +171,107 @@ func (s *Server) ListProviders(ctx context.Context, req *minderv1.ListProvidersR
...
+
+	infos := make([]*minderv1.ProviderClassInfo, 0, len(classes))
+	for _, class := range classes {
+		def, ok := classDefinitions[class]
+		if !ok {
+			zerolog.Ctx(ctx).Warn().Str("provider_class", class).Msg("provider class definition not found")
+			continue
+		}
```
evankanderson: I don't think we need to worry about the keys from the (effectively static) map being removed during this call.

internal/controlplane/handlers_providers.go
```diff
@@ -166,16 +171,107 @@ func (s *Server) ListProviders(ctx context.Context, req *minderv1.ListProvidersR
...
+
+		infos = append(infos, protobufProviderClassInfoFromDefinition(ctx, class, def))
```
evankanderson: If you change this to implement a new function on the Provider interface, you might be able to simply have the function return a `minderv1.ProviderClassInfo`.

internal/controlplane/handlers_providers.go
```diff
@@ -166,16 +171,107 @@ func (s *Server) ListProviders(ctx context.Context, req *minderv1.ListProvidersR
...
 
+func listEnabledProviderClasses(
+	providerManager interface{}, classDefinitions map[string]providers.ProviderClassDefinition,
+) []string {
+	classes := make([]string, 0, len(classDefinitions))
+
+	// Prefer classes registered in the current provider manager so the response
+	// reflects deployment/feature-flag reality. Fall back to static definitions
+	// when runtime discovery is unavailable (e.g., tests or alternate wiring).
+	if supportedClasses, ok := providerManager.(supportedProviderClassLister); ok {
```
evankanderson: IMO, we don't need this level of flexibility -- we should just update tests and code to ensure that all the ProviderManagers implement `ListSupportedClasses`.

evankanderson (commented):
Thanks for getting started on this!  Eventually, I'd like providers to be mostly self-describing, so I'd prefer to move the management of this metadata into each provider, with only a central inventory list of providers.

Jaydeep869:
Thanks for the review, Evan! I agree that moving the metadata into the providers so they are self describing is a much cleaner architecture. I'll extend the Provider interface to return the ProviderClassInfo natively and remove the centralized map and fallback logic. I'll also fix the doc URLs, clarify the proto descriptions, and drop the redundant creation_help field. I'll start working on these changes and push an update soon!

proto/minder/v1/minder.proto
Jaydeep869: It is intended to be plaintext to keep the CLI and backend responses simple. I'll add a comment in the protobuf definition to explicitly state that it should be plaintext so API consumers know what to expect.

proto/minder/v1/minder.proto
Jaydeep869: The original intent was to provide a short command line snippet or immediate CLI hint (like 'run this command to get a token'). However, I agree it overlaps too much with the description and documentation URL. I will just remove this field to keep the API simpler.

internal/providers/provider_definitions.go
Jaydeep869: Yeah. I'll update the base URL to point to https://docs.mindersec.dev/

internal/providers/provider_definitions.go
Jaydeep869: I'll double check the current implementation of the DockerHub provider. Assuming it supports artifacts, I'll make sure to add that to its supported entities list!

internal/providers/provider_definitions.go
Jaydeep869: I completely agree. Making the providers self describing is a much better and cleaner architecture. I'll extend the Provider interface in providers.go to return its own definition/metadata natively, and I'll remove this centralized map definition altogether.

internal/controlplane/handlers_providers.go
Jaydeep869: Makes sense! Now that we are shifting to the providers holding their own definitions, I'll update the ProviderManager to natively aggregate and return this state directly containing the active definitions.

internal/controlplane/handlers_providers.go
Jaydeep869: Agreed. Since I'm removing the centralized map approach based on your earlier feedback, this check will be deleted anyway.

internal/controlplane/handlers_providers.go
Jaydeep869: Yes, definitely! I will have the interface function return a *minderv1.ProviderClassInfo directly, which will allow us to remove this extra converter and simplify the handler significantly.

internal/controlplane/handlers_providers.go
Jaydeep869: Understood. I'll remove the type assertion fallback and ensure the interface contract is strictly enforced across the codebase and the tests.

Jaydeep869:
Hi Evan. I pushed an update that addresses your feedback and would appreciate another look.

proto/minder/v1/minder.proto
evankanderson: We ended up accidentally adding markdown to some of the fields in RuleType and discovered it only when attempting to render them in a web UI, so I'm more sensitive to this now than I used to be.

proto/minder/v1/minder.proto
```diff
@@ -3639,9 +3639,50 @@ message ListProviderClassesRequest {
...
+    ];
+
+    // supported_entities is the list of entity types supported by this provider class.
+    repeated Entity supported_entities = 6;
+
+    // documentation_url points to provider-specific or generic documentation.
+    string documentation_url = 7;
+
+    reserved 8;
+    reserved "creation_help";
```
evankanderson, line 3673: We don't really need to do this until we've built at least one official Minder binary that implements tis endpoint, but this isn't bad practice in general.

evankanderson (approved):

proto/minder/v1/minder.proto
Jaydeep869: Given the past issues with RuleType, I understand the concern. By explicitly documenting the description field as plaintext in the protobuf definition, we can ensure that any future web UI or CLI consumers know exactly what to expect and won't try to parse it as markdown. I will make sure the comment in the proto file clearly enforces this.

PR #6328, opened 2026-04-10
test(ruletypes): add table-driven unit tests for release phase conver…

dakshhhhh16:
### What

Adds `pkg/ruletypes/util_test.go` with 12 table-driven test cases covering the two public conversion helpers in `pkg/ruletypes/util.go` that were previously at **0%** statement coverage:

| Function | Coverage before | Coverage after |
|---|---|---|
| `GetPBReleasePhaseFromDBReleaseStatus` | 0% | **100%** |
| `GetDBReleaseStatusFromPBReleasePhase` | 0% | **85.7%** |

**The remaining ~14% of `GetDBReleaseStatusFromPBReleasePhase` is the `Scan` type-switch fallback path, which is unreachable from valid Go code because `db.ReleaseStatus` only accepts `string` or `[]byte` inputs.**

### Why

These functions are called by controlplane handlers on every rule-type read/write path but were not exercised by any dedicated unit tests, leaving conversion bugs invisible to CI.

### Test cases

**`TestGetPBReleasePhaseFromDBReleaseStatus`** (6 cases)

- `nil` input → `RULE_TYPE_RELEASE_PHASE_UNSPECIFIED` (no error)
- `alpha`, `beta`, `ga`, `deprecated` → correct protobuf enum value
- Unrecognised `db.ReleaseStatus` string → error returned

**`TestGetDBReleaseStatusFromPBReleasePhase`** (6 cases)

- `ALPHA`, `BETA`, `GA`, `DEPRECATED` → correct `db.ReleaseStatus` constant
- `UNSPECIFIED` → defaults to `ga` (via `EnsureDefault()`)
- Out-of-range enum value `999` → error wrapped with `ErrRuleTypeInvalid`

### Notes

- All subtests run with `t.Parallel()` and are race-clean (`go test -race` passes)
- No DB mock, network, or external dependency required — pure function testing
- Follows the same table-driven pattern used in the adjacent `service_test.go`

### Related

Continues the coverage push toward the project 60% threshold, following [#6309](https://github.com/mindersec/minder/pull/6309) which covered `pkg/profiles/models/`.

### DIFF
--- a/pkg/ruletypes/util_test.go
+++ b/pkg/ruletypes/util_test.go
@@ -1,0 +1,137 @@
+// SPDX-FileCopyrightText: Copyright 2024 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ruletypes_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/mindersec/minder/internal/db"
+	pb "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+	"github.com/mindersec/minder/pkg/ruletypes"
+)
+
+func TestGetPBReleasePhaseFromDBReleaseStatus(t *testing.T) {
+	t.Parallel()
+
+	alpha := db.ReleaseStatusAlpha
+	beta := db.ReleaseStatusBeta
+	ga := db.ReleaseStatusGa
+	deprecated := db.ReleaseStatusDeprecated
+	bogus := db.ReleaseStatus("bogus")
+
+	tests := []struct {
+		name      string
+		input     *db.ReleaseStatus
+		expected  pb.RuleTypeReleasePhase
+		expectErr bool
+	}{
+		{
+			name:     "nil returns unspecified",
+			input:    nil,
+			expected: pb.RuleTypeReleasePhase_RULE_TYPE_RELEASE_PHASE_UNSPECIFIED,
+		},
+		{
+			name:     "alpha maps correctly",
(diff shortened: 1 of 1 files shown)


dakshhhhh16:
@evankanderson 
The failing check is a pre-existing flaky test in the webhook package, it uses a 10 ms channel timeout that's tight enough to race on loaded CI runners regardless of changes in this PR. I reproduced it locally under load and it has no relation to the util_test.go added here. Happy to raise a separate PR to bump the timeout if that would help, but feel free to re-run the check in the meantime.

evankanderson (approved):
It looks like this does add a small amount of new coverage:

image

It appears the non-error cases in the code are covered 22 (GetDBReleaseStatusFromPBReleasePhase) or 2 (GetPBReleasePhaseFromDBReleaseStatus) times by other tests (according to Coveralls on the base commit).  This does add coverage for the error cases, but I'd prefer to get a bit more "lift" out of new tests in general.

dakshhhhh16:
> It looks like this does add a small amount of new coverage:
> 
> image
> It appears the non-error cases in the code are covered 22 (GetDBReleaseStatusFromPBReleasePhase) or 2 (GetPBReleasePhaseFromDBReleaseStatus) times by other tests (according to Coveralls on the base commit). This does add coverage for the error cases, but I'd prefer to get a bit more "lift" out of new tests in general.

Thanks, I’ll look into uncovered branches and add tests that increase more meaningful coverage.

PR #6331, opened 2026-04-10
fix: avoid silently aborting profile init on single entity marshalling error

### LINKED ISSUE #6330: Bug: [publishProfileInitEvents] silently drops profile init events for remaining entities on marshalling error
## Describe the bug
While reviewing the profile initialization flow, I noticed a bug in [internal/reconcilers/run_profile.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/reconcilers/run_profile.go:0:0-0:0) that can cause the platform to silently skip evaluating entities when a new profile is created.

In [publishProfileInitEvents](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/reconcilers/run_profile.go:75:0-105:1), the code iterates over all registered entities for a project and publishes a profile evaluation event for each one. However, if the message marshalling fails for any single entity inside that loop, the function does a `return nil`:

```go
if err := entRefresh.ToMessage(m); err != nil {
	zerolog.Ctx(ctx).Error().Err(err).Msg("error marshalling message")
	// no point in retrying, so we return nil
	return nil
}
```
Because it returns nil instead of continue, this silently aborts the entire loop. Any entities remaining in the ents slice will be skipped, and their profile init events will never be published. Furthermore, because the error is swallowed and nil is returned, the caller (handleProfileInitEvent) thinks the entire batch was processed successfully, so no retries happen and no alarms go off.

## Steps to reproduce

1. Have a project with multiple entities (e.g., repositories A, B, C, D).
2. Trigger a condition where one entity (e.g., repository B) causes entRefresh.ToMessage(m) to fail or return an error.
3. Create a new profile for that project.
4. Observation: Repository A gets evaluated, but Repository B fails (which is fine). However, Repositories C and D are completely skipped and never get evaluated against the new profile, even though nothing is wrong with them.

## Expected behavior
If an individual entity fails to be marshalled or published, the loop should log the error and use continue to move on to the next entity, rather than exiting the function entirely.

## Proposed Fix
Change the return nil string inside the loop to a continue, so the rest of the entities aren't dropped.

AftAb-25:
## Description

This PR fixes a bug in the profile initialization flow where a single entity's parsing/marshalling issue could completely stop the rest of a project's entities from being evaluated against a new profile.

In [internal/reconcilers/run_profile.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/reconcilers/run_profile.go:0:0-0:0), [publishProfileInitEvents](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/reconcilers/run_profile.go:75:0-105:1) loops over all registered entities for a project. Previously, if `entRefresh.ToMessage(m)` failed inside that loop, the code returned `nil`. This silently exited the entire function, abandoning all remaining entities in the slice. From the caller's perspective ([handleProfileInitEvent](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/reconcilers/run_profile.go:43:0-73:1)), the function returned successfully, so no retries were triggered and nothing was visibly wrong except that some entities were randomly skipped during profile initialization.

## Change:
Instead of `return nil`, this simply logs the error and uses `continue` to move on to the next entity in the loop. 

Fixes #6330 

## Checklist

- [x] Includes tests for the changes
- [x] Code compiles cleanly

### DIFF
--- a/internal/reconcilers/run_profile.go
+++ b/internal/reconcilers/run_profile.go
@@ -92,8 +92,8 @@ func (r *Reconciler) publishProfileInitEvents(
 
 		if err := entRefresh.ToMessage(m); err != nil {
 			zerolog.Ctx(ctx).Error().Err(err).Msg("error marshalling message")
-			// no point in retrying, so we return nil
-			return nil
+			// Skip this entity but continue processing the rest
+			continue
 		}
 
 		if err := r.evt.Publish(constants.TopicQueueRefreshEntityByIDAndEvaluate, m); err != nil {


evankanderson (approved):
I don't see any way to test this, as the only error that can be triggered is from `json.Marshal()` on `HandleEntityAndDoMessage`, which is a struct with simple fields and tags.

I'm willing to approve this, but it's not really a likely bug fix.  (If the go allocator fails for `json.Marshal()`, we're unlikely to make further progress.  Similarly, if the descriptor for `HandleEntityAndDoMessage` gets corrupted in-memory, we've got a bad enough day that this isn't going to save it.)

PR #6336, opened 2026-04-10
Fixed Repo delete by name

### LINKED ISSUE #3220: Get/Delete Provider by name endpoints use a SQL query which may match more than one repo
### Describe the issue

See this closed PR: https://github.com/stacklok/minder/pull/3219

This will trigger if more than one repo with the same name but with different providers exist in the same project. Right now, I don't think it's possible to actually create this scenario, but it will create issues when we support multiple source code repo providers.

Fixing this will involve one of the following options:

1) Break API compatibility, and enforce 

### To Reproduce

_No response_

_No response_

DharunMR:
# Summary

Enforces explicit providerName validation directly on the RPC request payload for DeleteRepositoryByName. This prevents the panic: nil pointer dereference and stops ambiguous/non-deterministic deletions when a user omits the provider flag.

Added missing get repo testcases

- Fixes #3220

### DIFF
--- a/internal/controlplane/handlers_repositories.go
+++ b/internal/controlplane/handlers_repositories.go
@@ -298,6 +298,11 @@ func (s *Server) DeleteRepositoryByName(
 	}
 
 	projectID := GetProjectID(ctx)
+
+	if in.GetContext() == nil || in.GetContext().GetProvider() == "" {
+		return nil, util.UserVisibleError(codes.InvalidArgument, "provider name must be specified when deleting a repository by name")
+	}
+
 	providerName := GetProviderName(ctx)
 
 	err := s.repos.DeleteByName(ctx, fragments[0], fragments[1], projectID, providerName)
--- a/internal/controlplane/handlers_repositories.go
+++ b/internal/controlplane/handlers_repositories.go
@@ -40,7 +40,11 @@ func (s *Server) RegisterRepository(
 	in *pb.RegisterRepositoryRequest,
 ) (*pb.RegisterRepositoryResponse, error) {
 	projectID := GetProjectID(ctx)
-	providerName := GetProviderName(ctx)
+
+	if in.GetContext() == nil || in.GetContext().GetProvider() == "" {
+		return nil, util.UserVisibleError(codes.InvalidArgument, "provider name must be specified when registering a repository")
+	}
+	providerName := in.GetContext().GetProvider()
 
 	var fetchByProps *properties.Properties
 	var provider *db.Provider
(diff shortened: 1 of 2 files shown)


internal/controlplane/handlers_repositories.go
```diff
@@ -137,10 +141,15 @@ func (s *Server) repoCreateInfoFromUpstreamEntityRef(
 // This function will typically be called by the client to get a list of
 // repositories that are registered present in the minder database
 func (s *Server) ListRepositories(ctx context.Context,
-	_ *pb.ListRepositoriesRequest) (*pb.ListRepositoriesResponse, error) {
+	in *pb.ListRepositoriesRequest) (*pb.ListRepositoriesResponse, error) {
+
+	if in.GetContext() == nil || in.GetContext().GetProvider() == "" {
+		return nil, util.UserVisibleError(codes.InvalidArgument, "provider name must be specified when listing repositories")
+	}
```
evankanderson, line 148: This doesn't seem to crash or cause a problem today:

```bash
$ ./bin/minder repo list
 OWNER              │ NAME      │ PROVIDER                        │ UPSTREAM ID    
────────────────────┼───────────┼─────────────────────────────────┼────────────────
 a-random-sandbox   │ colorls   │ github-app-a-random-sandbox     │ 693510024      
                    ├───────────┤                                 ├────────────────
                    │ doitlive  │                                 │ 693507178
```

internal/controlplane/handlers_repositories.go
```diff
@@ -298,7 +311,12 @@ func (s *Server) DeleteRepositoryByName(
 	}
 
 	projectID := GetProjectID(ctx)
-	providerName := GetProviderName(ctx)
+
+	if in.GetContext() == nil || in.GetContext().GetProvider() == "" {
+		return nil, util.UserVisibleError(codes.InvalidArgument, "provider name must be specified when deleting a repository by name")
+	}
```
evankanderson, line 317: This doesn't seem to crash today:

```bash
$ ./bin/minder repo delete -n a-random-sandbox/.github
Successfully deleted repo with name: a-random-sandbox/.github
```

internal/controlplane/handlers_repositories.go
```diff
@@ -137,10 +141,15 @@ func (s *Server) repoCreateInfoFromUpstreamEntityRef(
 // This function will typically be called by the client to get a list of
 // repositories that are registered present in the minder database
 func (s *Server) ListRepositories(ctx context.Context,
-	_ *pb.ListRepositoriesRequest) (*pb.ListRepositoriesResponse, error) {
+	in *pb.ListRepositoriesRequest) (*pb.ListRepositoriesResponse, error) {
+
+	if in.GetContext() == nil || in.GetContext().GetProvider() == "" {
```
evankanderson, line 146: `GetProvider()` already handles being called on a `nil` `*Context` -- using the proto getters will automatically handle empty / missing fields by returning the default (zero) value.

internal/controlplane/handlers_repositories.go
evankanderson: With `curl`, in case you think there's some magic in the CLI:

```bash
$ curl -H "Authorization: Bearer $(jq -r .access_token ~/.config/minder/127.0.0.1_8090.json)" 'http://localhost:8080/api/v1/repositories/provider/' | jq '.results[] | .name'
...
"colorls"
"doitlive"
```

evankanderson (commented):
You've wandered into a little bit of magic here to make the CLI and API usage easier in the common case.  😁

PR #6338, opened 2026-04-10
fix: preserve context cancellation propagation in ExecutorEventHandler

### LINKED ISSUE #6337: ExecutorEventHandler breaks context cancellation propagation
This is not a bug but a design concern regarding how context cancellation is handled in ExecutorEventHandler.

Currently, the handler detaches from the parent message context using context.WithoutCancel and introduces a separate internal cancellation mechanism. This creates multiple sources of truth for cancellation and may lead to unintended side effects.

Key concerns:

Breaks context propagation: upstream cancellation signals are not respected
Dual cancellation model increases complexity and reduces clarity
Manual cancellation tracking (e.cancels) adds lifecycle management overhead
May impact observability and debugging

Expected behavior:

Preserve parent context cancellation semantics
Avoid detaching from the original context unless necessary
Prefer structured concurrency (e.g., errgroup, worker pools) over manual tracking

If this behavior is intentional, it would be helpful to understand the reasoning behind detaching from the parent context instead of preserving its cancellation semantics.

sachin9058:
This PR updates the context handling in ExecutorEventHandler to preserve parent context cancellation semantics.

Previously, the handler used context.WithoutCancel, which detached execution from the parent message context. This prevented upstream cancellation signals from propagating and could lead to unexpected behavior, especially in scenarios involving shutdown or request cancellation.

This change removes the use of context.WithoutCancel and ensures that the handler derives its context directly from msg.Context(), while still maintaining controlled shutdown using context.WithCancel.

Rationale
Restores standard Go context propagation behavior
Improves observability and consistency across the system
Reduces complexity introduced by dual cancellation paths
Behavior Change

With this change, execution may now be cancelled if the parent context is cancelled. Previously, execution would continue regardless of parent cancellation.

If the previous behavior was intentional, I’m happy to revisit and adjust the approach accordingly.

Fixes #6337 

Testing
Verified that the project builds successfully after the change

Ran existing test suite:

go test ./...
Manually reviewed execution flow to ensure:
Context is correctly derived from msg.Context()
Shutdown cancellation still functions as expected

No additional configuration was required for testing.

### DIFF
--- a/internal/engine/handler.go
+++ b/internal/engine/handler.go
@@ -16,6 +16,7 @@ import (
 	"github.com/mindersec/minder/internal/engine/engcontext"
 	"github.com/mindersec/minder/internal/engine/entities"
 	minderlogger "github.com/mindersec/minder/internal/logger"
+	pb "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
 	"github.com/mindersec/minder/pkg/eventer/constants"
 	"github.com/mindersec/minder/pkg/eventer/interfaces"
 )
@@ -118,10 +119,13 @@ func (e *ExecutorEventHandler) HandleEntityEvent(msg *message.Message) error {
 	e.wgEntityEventExecution.Add(1)
 	go func() {
 		defer e.wgEntityEventExecution.Done()
-		select {
-		case <-time.After(ArtifactSignatureWaitPeriod):
-		case <-msgCtx.Done():
-			return
+		if inf.Type == pb.Entity_ENTITY_ARTIFACTS {
+			// Wait for artifact signatures, but allow early exit on shutdown
+			select {
+			case <-time.After(ArtifactSignatureWaitPeriod):
+			case <-msgCtx.Done():
+				return
+			}
 		}
 
 		ctx, cancel := context.WithTimeout(msgCtx, DefaultExecutionTimeout)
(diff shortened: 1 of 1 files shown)


internal/engine/handler.go
```diff
@@ -87,15 +87,15 @@ func (e *ExecutorEventHandler) Wait() {
...
+	// We preserve parent context cancellation while still allowing controlled shutdown.
 	// completion, because the default watermill behavior for both Go channels and
 	// SQL is to process messages sequentially, but we need additional parallelism
 	// beyond that.  When we switch to a different message processing system, we
 	// should aim to remove this goroutine altogether and have the messaging system
 	// provide the parallelism.
 	// We _do_ still want to cancel on shutdown, however.
 	// TODO: Make this timeout configurable
-	msgCtx := context.WithoutCancel(msg.Context())
+	msgCtx := msg.Context()
```
evankanderson, line 98: You're undoing an optimization that took several painful days to implement while Minder was broken (overloaded).  We are _deliberately_ escaping the parent context's Cancel/Done, because our messaging system (watermill) is effectively single-threaded for the Postgres implementation, but we need higher throughput than single-threaded execution.

Yes, this is ugly, but the edge case here on shutdown with some extra delay is less critical than day-to-day throughput of the application.

evankanderson (commented):
The last time I touched this, I was very sad, and I left a large comment about why it is the way it is.  Unless someone is willing to set up a load test in an actual installation, let's not touch this right now.

The system would need to have a few ms of latency to Postgres and at least 1000 entities with changes happening to entities every few hundred seconds; I seem to recall that's _approximately_ where this fell over before.

sachin9058:
@evankanderson Thanks for the detailed context ...

I didn’t realize this behavior was intentionally designed to avoid throughput issues with the current messaging setup. I’ll revert this change and close the PR.

Appreciate the explanation!

Issue #6339, opened 2026-04-10
`quickstart` gets users off on the wrong foot


evankanderson:
### Describe the issue

It looks like the `quickstart` command (which was written in 2023) needs some updates from 2024 and 2025.

* It starts users off trying to enroll using a PAT, rather than using the GitHub App
* It doesn't leverage the resource-agnostic `apply` batch command, which really simplifies development.
* The setup installs a single ruletype, but doesn't really set you up for installing more or managing the existing rule.  It would be better if this could pull from e.g. top-level directories with `README.md` files in `github.com/mindersec/minder-rules-and-profiles` (or some other chosen repository).
  * (IMO) it would be great to set things up with an infrastructure-as-code setup using the `mindersec/minder` action and some policy copied into their own repo.  I have a sample of this in https://github.com/custcodian/minder-rules-and-profiles/tree/main/bootstrap (though the mermaid diagram seems a bit broken).

### To Reproduce

Run `minder quickstart`

### What version are you using?

upstream/main on 10 April

evankanderson:
In the meantime, we should probably pull the `quickstart` documentation.

sachin9058:
Hi @evankanderson, I’d like to take this up.

From what I see, the quickstart flow could be improved in a few areas — switching to GitHub App onboarding, using the batch apply command, and making rule setup more intuitive.

Would you prefer tackling this step by step (like starting with CLI changes or docs), or should it be approached as a broader redesign?

Happy to work on it either way 👍

dakshhhhh16:
<img width="3372" height="730" alt="Image" src="https://github.com/user-attachments/assets/7c7de211-7fd3-4123-960d-07e352b465cb" />

This one's accurate right?

evankanderson:
I think it makes the most sense to approach this as a broader redesign, with the goal being to set up users with a useful Minder setup that they can feel confident in going forward.

sachin9058:
@evankanderson I’ve started by deprecating the current quickstart docs in #6341 .

For the redesign, I’m thinking of moving toward a more declarative flow instead of a guided CLI setup:

- GitHub App–based onboarding (`minder auth login`)
- using `minder apply` with a starter configuration (e.g. bootstrap repo)
- allowing users to manage rules and profiles via config (GitOps-style)

Example: minder apply -f ./minder-config


or potentially from a maintained starter repo.


This would simplify onboarding and make it easier to scale beyond a single rule.

[...]

PR #6341, opened 2026-04-11
docs: deprecate quickstart documentation and remove CLI references

sachin9058:
This PR deprecates the current quickstart documentation and removes prominent references to it from the CLI docs.

As discussed in #6339, the existing quickstart flow is outdated and may guide users toward a setup that is no longer recommended (e.g., PAT-based onboarding and limited rule configuration). To prevent confusion for new users, this change adds a deprecation notice to the quickstart guide and removes it from the CLI reference index.

This is an initial step toward improving the onboarding experience without introducing breaking changes.

Follow-up work:
I plan to propose a redesigned quickstart flow that:

uses GitHub App–based onboarding
leverages minder apply for batch configuration
supports a more scalable, config-driven (GitOps-style) setup

No additional dependencies are required for this change.

Fixes #6399

Testing
Verified that the documentation builds/rendering are unaffected after removing the quickstart reference from the CLI index
Confirmed that the quickstart page is still accessible directly but clearly marked as deprecated
Ensured no broken links or formatting issues in modified files

No runtime or functional changes were introduced, as this PR only affects documentation.

### DIFF
--- a/docs/docs/getting_started/quickstart.md
+++ b/docs/docs/getting_started/quickstart.md
@@ -1,3 +1,6 @@
+> ⚠️ **Deprecated:** This quickstart flow is outdated and currently being redesigned.
+> It may not reflect the recommended onboarding process.
+
 ---
 title: Quickstart with Minder (< 1 min)
 sidebar_position: 30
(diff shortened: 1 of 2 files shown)


evankanderson (approved):
The `ref/cli` docs are auto-generated, so the removal there will probably be reverted in the future, but the banner is useful, so approving.

If you want to hide the `quickstart` command from the generated documentation, you probably need to set the "Hidden" flag on the `Command` object in `cmd/cli`

PR #6342, opened 2026-04-11
fix: implement cursor-based pagination for ListRepositories endpoint

### LINKED ISSUE #6340: ListRepositories returns all repositories at once with no pagination, leading to memory spikes and timeouts on large projects
### Describe the issue

When you call `minder repo list` (or the equivalent `ListRepositories` gRPC endpoint) on a project that has a lot of registered repositories, the server goes ahead and fetches every single one and sends them back in a single response. No batching, no pagination, no cursor just one giant payload.

We know this is a problem. There's even a TODO comment right in the code (line 194 in `internal/controlplane/handlers_repositories.go`) that says:

`// TODO: Implement cursor-based pagination using entity IDs
// For now, return all results without pagination`

The protobuf response actually has a `cursor` field, but it's always returned as an empty string completely useless to the client. There's also no `page_size` parameter in the request, so even if a client wanted to chunk the results on their end, they couldn't ask the server for a specific page.

The result is that as your project grows, this endpoint becomes slower and hungrier for memory. At a few hundred repos it's annoying. At a few thousand it's a serious problem the API server can run out of memory, or the CLI just sits there spinning until it times out.

This also means anyone who can register repositories under a project they control has a potential way to stress-test the API server by bulk-registering thousands of repos.

### To Reproduce

 What you'll need:
  - A running minder server (local or staging)
  - A project with a registered GitHub provider
  - `minder` CLI (or `grpcurl` for direct gRPC calls)

  Steps:

  1. Set up a test project and enroll a GitHub provider:
  `minder provider enroll --name github --type github`

  2. Register a large number of repositories — the easiest way is to connect a GitHub App installation that already has access to an org with lots of repos. Minder will ingest them all automatically.
Or, if you just want to test the API behavior, register a bunch manually:
  `for i in $(seq 1 1000); do`
   `minder repo register --repo "myorg/test-repo-$i"`
   `done`

  3. Call the list endpoint and watch what happens:
Via CLI:
`minder repo list`
Via gRPC directly:
`grpcurl -plaintext -d '{}' localhost:8090 minder.v1.RepositoryService/ListRepositories`

  4. Observe the problems:

    - The server hangs for several seconds before returning (or times out if you have enough repos)
    - Server memory usage spikes noticeably during the call
    - The response's `cursor` field is always an empty string, no indication that more pages exist
    - There is no way to request a smaller batch of results

  5. Scale it up to confirm the pain: try with 5,000 repos and notice the latency and memory hit get significantly worse.

d8b8b4e6

MayankSharmaCSE:
# Summary

The `ListRepositories` gRPC endpoint was returning all registered repositories in a single unbounded response, with no pagination support. A TODO comment in the code explicitly acknowledged this: "TODO: Implement cursor-based pagination using entity IDs. For now, return all results without pagination."

This caused memory exhaustion and timeouts on projects with large numbers of repositories. The fix implements cursor-based pagination using entity IDs as cursors. The handler now accepts optional `limit` (default 100, max 1000) and `cursor` parameters in the request, and returns a `cursor` field in the response. When the cursor is non-empty, clients can use it to fetch the next page.

**Files changed:**
- `internal/repositories/service.go` - Added `ListRepositoriesPaginated` method to interface and implementation
- `internal/controlplane/handlers_repositories.go` - Updated handler to use paginated method
- `internal/repositories/mock/service.go` - Mock regenerated via `go generate`
- `internal/repositories/mock/fixtures/service.go` - Added paginated mock fixtures
- `internal/controlplane/handlers_repositories_test.go` - Updated tests for new pagination behavior

**Dependencies:** None - uses existing `ListEntitiesAfterID` DB query that already supports cursor-based pagination.

Fixes #6340 

# Testing

- Unit tests pass:
```bash
go test ./internal/controlplane/... -run "TestServer_ListRepositories" -v
go test ./internal/repositories/...
```
- All controlplane tests pass: `go test ./internal/controlplane/...`
- Build succeeds: `go build ./...`
- Manual testing:
- First page: `minder repo list --limit 10`  
- Next page: `minder repo list --limit 10 --cursor <cursor_from_previous_response>`


internal/controlplane/handlers_repositories.go
```diff
@@ -154,8 +156,24 @@ func (s *Server) ListRepositories(ctx context.Context,
...
 
-	// Fetch repositories using the service
-	repoEntities, err := s.repos.ListRepositories(ctx, projectID, provider.ID)
+	// Parse cursor from request (empty string means first page)
+	var cursor uuid.UUID
+	if in.GetCursor() != "" {
+		cursor, err = uuid.Parse(in.GetCursor())
+		if err != nil {
+			return nil, util.UserVisibleError(codes.InvalidArgument, "invalid cursor format")
+		}
```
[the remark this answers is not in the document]
evankanderson: You may need to update the annotations in `minder.proto`, or consider blinding the cursor (e.g. base64 the contents) to make it less of an explicit contract.

internal/repositories/service.go
```diff
@@ -210,6 +220,95 @@ func (r *repositoryService) ListRepositories(
...
+
+	qtx := r.store.GetQuerierWithTransaction(tx)
+
+	var repoEnts []db.EntityInstance
+	if cursor == uuid.Nil {
+		repoEnts, outErr = qtx.GetEntitiesByType(ctx, db.GetEntitiesByTypeParams{
+			EntityType: db.EntitiesRepository,
+			ProviderID: providerID,
+			Projects:   []uuid.UUID{projectID},
+		})
```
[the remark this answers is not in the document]
evankanderson: I think this is the critical comment -- you should set a database limit to avoid doing a lot of extra backend work.  Given the cursor-based query, you should be able to fetch using the natural order of the table (ORDER BY id) and then a WHERE clause on id > cursor.

MayankSharmaCSE:
@evankanderson Thanks for the review! Both comments are valid and we'll address them:                                                                                              
                                                                                                                                                                      
  1. Database limit: You're right that fetching all entities and then truncating in memory defeats pagination. We'll add a new ListEntitiesByTypePaginated query that 
  applies LIMIT at the database level for the first page.                                                                                                             
  2. Cursor blinding: The hyphenated UUIDs don't match the proto pattern ^[[:word:]=]*$. We'll encode the cursor as base64 to make it compliant and less of an        
  explicit contract.                                                                                                                                                  
   
  We also identified a related issue we'll fix:                                                                                                                       
  - Data scoping: ListEntitiesAfterID needs to be scoped by provider_id/project_id to prevent cross-project data access.
                                                                                                                                                                      
  We'll proceed with implementing these fixes. Let us know if you have any concerns!

internal/repositories/service.go
```diff
@@ -210,6 +220,95 @@ func (r *repositoryService) ListRepositories(
...
+		if outErr != nil {
+			return nil, uuid.Nil, fmt.Errorf("error fetching repositories: %w", outErr)
+		}
+	} else {
+		repoEnts, outErr = qtx.ListEntitiesAfterID(ctx, db.ListEntitiesAfterIDParams{
+			EntityType: db.EntitiesRepository,
+			ID:         cursor,
+			Limit:      limit,
+		})
```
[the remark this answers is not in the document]
evankanderson: We should be able to use the _same_ query for both cases, by passing in the null (all-zeros) UUID in the start case.  It should look something like:

SELECT * FROM entities WHERE project_id IN $1 AND id > $2 ORDER BY id LIMIT ($3 + 1);

The cursor would then be the second-to-last `id` returned, so the "greater than" would naturally fetch the next rows.  We over-fetch by one row so that we can decide whether there are more pages (because there's at least one extra row) -- the cost for fetching one more row in the batch should be low.

evankanderson (commented):
It doesn't look like any of the SQL query suggestions have been applied.  If you update the SQL queries, you'll also need to run `make gen` to re-generate the query code.

evankanderson (changes requested):
Marking "request changes" on the SQL changes, and the conflict between the UUID-style cursor implemented and the (documented) opaque cursor without "-" characters.

internal/db/datasources.sql.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by sqlc. DO NOT EDIT.
 // versions:
-//   sqlc v1.30.0
+//   sqlc v1.31.1
```
evankanderson: It looks like you brought in a new version of `sqlc` here.  Intentional?

(If not, we'll pick it up in an update to `tools/go.mod` within a few weeks)

evankanderson (commented):
If you needed a newer version of `sqlc`, you'll need to update the version in `tools/go.mod` and re-run `make bootstrap` to make sure you're using the correct version.  (Yeah, it sucks that we have a bunch of tools that sort of live globally via bootstrap, rather than in some hermetic location.)

internal/db/datasources.sql.go
MayankSharmaCSE: yes it was intentional because i was working on the updated version but now i have fixed it according to the same version.

evankanderson: (You'll need to push a change if you switched to the older `sqlc`)

internal/controlplane/handlers_repositories.go
```diff
@@ -154,8 +156,18 @@ func (s *Server) ListRepositories(ctx context.Context,
 		return nil, status.Errorf(codes.Internal, "cannot get provider: %v", err)
 	}
 
-	// Fetch repositories using the service
-	repoEntities, err := s.repos.ListRepositories(ctx, projectID, provider.ID)
+	// Parse cursor from request (base64-encoded UUID)
+	cursorStr := in.GetCursor()
```
evankanderson, line 160: This says it is a base64-encoded UUID, but there's no parsing being done here.  In fact, it doesn't seem like we need the temporary `cursorStr` at all -- it's only used once and we could just put it inline there.

internal/repositories/service.go
```diff
@@ -210,6 +221,112 @@ func (r *repositoryService) ListRepositories(
...
+func (r *repositoryService) ListRepositoriesPaginated(
+	ctx context.Context,
+	projectID uuid.UUID,
+	providerID uuid.UUID,
+	cursorStr string,
+	limit int64,
+) (ents []*models.EntityWithProperties, nextCursor string, outErr error) {
+	if limit <= 0 || limit > DefaultListRepositoriesPageLimit {
+		limit = DefaultListRepositoriesPageLimit
+	}
```
evankanderson, line 238: If you remove the upper check on `limit` here, you could remove `ListRepositories` and pass in either `MAX_INT` or some sentinel value (think `-1`) on the call from `fetchRepositoriesForProvider` in `handlers_repositories.go`.  That would allow removing `ListRepositories` (without pagination) from the codebase altogether.

internal/repositories/service.go
```diff
@@ -210,6 +221,112 @@ func (r *repositoryService) ListRepositories(
...
+		service.ReadBuilder().WithStoreOrTransaction(qtx).TolerateStaleData()); err != nil {
+		return nil, fmt.Errorf("error fetching properties for repository: %w", err)
+	}
+	return ewp, nil
+}
+
+func decodeCursor(cursorStr string) (uuid.UUID, error) {
+	if cursorStr == "" {
+		return uuid.Nil, nil
+	}
```
evankanderson, line 318: I think this needs a comment -- "empty string means start at the beginning" or suchlike.

internal/repositories/service.go
```diff
@@ -210,6 +221,112 @@ func (r *repositoryService) ListRepositories(
...
+
+	ents = make([]*models.EntityWithProperties, 0, len(repoEnts))
+	var lastID uuid.UUID
+	for _, ent := range repoEnts {
+		ewp, err := r.hydrateEntityProperties(ctx, qtx, ent.ID)
+		if err != nil {
+			return nil, "", err
+		}
+		ents = append(ents, ewp)
+		lastID = ent.ID
```
evankanderson, line 284: Rather than setting `lastID` on every pass through the loop, why not use `lastID := ents[len(ents)-1].ID`?

evankanderson (commented):
Sorry, I ended up getting distracted looking at the usage of ListRepository outside the RPC call and never sent this feedback.

PR #6344, opened 2026-04-11
test: add benchmark coverage for ExecutorEventHandler performance

### LINKED ISSUE #6343: Add benchmark coverage for ExecutorEventHandler to evaluate performance characteristics
## Describe the issue

Currently, there is no benchmark coverage for `ExecutorEventHandler.HandleEntityEvent`, which makes it difficult to understand its performance characteristics such as throughput and memory allocation behavior.

Since event handling is a core part of the engine, having a baseline benchmark would help identify bottlenecks and track performance over time.

## Why this matters

- Provides visibility into execution time and allocation patterns
- Helps detect performance regressions in future changes
- Useful for evaluating tradeoffs (e.g., concurrency and context handling)

## Proposed solution

Add benchmark tests for `HandleEntityEvent`, including:
- a sequential benchmark for baseline performance
- a parallel benchmark to simulate concurrent workloads

## Additional context

This is particularly relevant in light of recent discussions around throughput and event processing behavior in the engine.

sachin9058:
# Summary

This PR adds benchmark coverage for `ExecutorEventHandler.HandleEntityEvent` to evaluate its performance characteristics.

Currently, there is no benchmark for this critical execution path, making it difficult to understand throughput and memory allocation behavior or detect performance regressions over time. This change introduces both sequential and parallel benchmarks to provide a baseline for future performance analysis.

The benchmarks use a minimal stub executor to isolate handler logic and avoid external dependencies.

No additional dependencies are required for this change.

Fixes #6343 

# Testing

- Ran benchmarks locally using:
  go test ./internal/engine -bench=.

Benchmark results:

BenchmarkHandleEntityEvent-12         1803787   635.4 ns/op   840 B/op   14 allocs/op
BenchmarkHandleEntityEventParallel-12 2365395   499.8 ns/op   838 B/op   14 allocs/op

- Verified both sequential and parallel benchmarks execute successfully
- Confirmed allocation reporting (allocs/op, B/op)
- Ensured lint checks pass (make lint-fix)

The parallel benchmark shows improved throughput compared to sequential execution, indicating the handler scales well under concurrent workloads.

No functional changes were introduced, as this PR only adds benchmark tests.
No functional changes were introduced, as this PR only adds benchmark tests.

### DIFF
--- a/internal/engine/handler_benchmark_test.go
+++ b/internal/engine/handler_benchmark_test.go
@@ -1,0 +1,77 @@
+// SPDX-FileCopyrightText: Copyright 2023 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package engine
+
+import (
+	"context"
+	"testing"
+
+	"github.com/ThreeDotsLabs/watermill/message"
+
+	"github.com/mindersec/minder/internal/engine/entities"
+)
+
+// fakeExecutor is a minimal stub implementation of the Executor interface.
+// It allows the benchmark to run without invoking real evaluation logic.
+type fakeExecutor struct{}
+
+// EvalEntityEvent satisfies the Executor interface.
+// It performs no work and always returns nil.
+func (f *fakeExecutor) EvalEntityEvent(_ context.Context, _ *entities.EntityInfoWrapper) error {
+	_ = f
+	return nil
+}
+
+// newTestHandler creates a minimal ExecutorEventHandler instance
+// configured with a fake executor for benchmarking purposes.
+func newTestHandler() *ExecutorEventHandler {
+	return &ExecutorEventHandler{
+		executor: &fakeExecutor{},
+	}
+}
+
+// newTestMessage creates a minimal Watermill message with a background context.
+func newTestMessage() *message.Message {
+	msg := message.NewMessage("test-id", []byte("{}"))
+	msg.SetContext(context.Background())
+	return msg
+}
+
+// BenchmarkHandleEntityEvent measures the sequential performance of
+// ExecutorEventHandler.HandleEntityEvent.
+//
+// This benchmark provides a baseline for:
+// - execution time per event
+// - memory allocations per operation
+//
+// It isolates handler logic by using a fake executor.
+func BenchmarkHandleEntityEvent(b *testing.B) {
+	handler := newTestHandler()
+	msg := newTestMessage()
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		_ = handler.HandleEntityEvent(msg)
+	}
+}
+
+// BenchmarkHandleEntityEventParallel measures performance under parallel load.
+//
+// This helps evaluate how the handler behaves when multiple events
+// are processed concurrently, which is important for throughput analysis.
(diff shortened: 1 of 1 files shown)


evankanderson:
I'm not sure this is especially useful -- the problem with EvalEntityEvent is the user-defined work from policy evaluation, remediation, etc which is triggered by the entity event, not the message handling itself.

sachin9058:
@evankanderson Thanks, that makes sense.

You're right that the main cost likely comes from the evaluation and remediation logic triggered by the event, rather than the message handling itself.

I added this as a baseline for the handler, but I agree it would be more useful to benchmark the full execution path, including policy evaluation and remediation.

I can follow up with a more representative benchmark that captures the end-to-end flow if that would be helpful.

sachin9058:
@evankanderson 

I’ve updated the benchmark to include a lightweight simulation of the execution path instead of a no-op executor. It now includes some structured processing and branching to better approximate evaluation work, while still keeping it deterministic and isolated from external dependencies.

I’ve also kept the original baseline benchmarks for comparison.

Happy to adjust further if you’d prefer a different approach or a deeper integration with the evaluation pipeline.

evankanderson:
I think this is fine, but the real cost of these executions is user-defined based on the number of policies and the ingestion of git and/or rest resources, including data sources.

We do set a timeout for the user executions, but I believe it may be close to 5 minutes.

evankanderson (approved):

evankanderson:
I'm still trying to figure out exactly what we're hoping to measure here -- I'd start with setting up an actual Minder instance (with webhooks, etc), load 20-ish repositories and 5 profiles, and then generate a small amount of load (e.g. 1 PR / API change every 2 minutes) on the repos.  You can then use a profiling tool such as [`pprof`](https://jvns.ca/blog/2017/09/24/profiling-go-with-pprof/) or the [opentelemetry collector](https://opentelemetry.io/docs/concepts/signals/profiles/) to gather a profile, ala https://github.com/mindersec/minder/pull/4527.

In general, benchmarks are good for avoiding performance regressions, but they need to be targeted at sections of code which are a meaningful part of the overall performance picture.

evankanderson:
(also, unit test seemed to fail more than once)

sachin9058:
@evankanderson That makes sense  thanks for the clarification.

This benchmark was intended as a baseline for the handler itself, but I agree it doesn’t capture the dominant costs of policy evaluation and external interactions.

I’ll follow up with a more representative, end-to-end performance measurement using a real Minder setup and profiling tools (pprof / OpenTelemetry), as you suggested.

Let me know if you'd prefer removing this or keeping it as a lightweight baseline.

evankanderson:
I'm inclined to close this PR unless we have some evidence that performance of ExecutorEventHandler (and not the callbacks it calls) is a performance bottleneck.  Last time we benchmarked this (which was about 2 years ago), the main costs were in the database-backed Watermill queue and the data ingestion for rule evaluation.

sachin9058:
@evankanderson Thanks, that makes sense.

Given that the main performance costs are in policy evaluation, data ingestion, and the queue, this benchmark doesn’t provide much actionable insight right now.

I’ll close this PR and revisit performance analysis with a more representative, end-to-end approach using profiling tools (pprof / OpenTelemetry) as suggested.

Appreciate the guidance!

PR #6348, opened 2026-04-12
fix: prevent nil pointer panic on GitHub API network failures

### LINKED ISSUE #6347: Bug: Panic via nil-pointer dereference in `StartCheckRun` and `UpdateCheckRun` on network failures
## Describe the bug
There is a nil-pointer dereference vulnerability in [internal/providers/github/common.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/providers/github/common.go:0:0-0:0) which panics the executing goroutine if a fundamental network failure occurs when attempting to create or update a GitHub Check Run.

In both [StartCheckRun](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/providers/github/common.go:1032:0-1050:1) and [UpdateCheckRun](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/providers/github/common.go:1052:0-1066:1), the `go-github` client's error response is handled by immediately inspecting `resp.StatusCode`:

```go
	run, resp, err := c.client.Checks.CreateCheckRun(ctx, owner, repo, *opts)
	if err != nil {
		// If error is 403 then it means we are missing permissions
		if resp.StatusCode == 403 { // <--- PANIC IF RESP IS NIL
			return nil, fmt.Errorf("missing permissions: check")
		}
		return nil, ErroNoCheckPermissions
	}
```
If the API fails due to a network routing issue (e.g., DNS error, TLS handshake timeout, or completely unreachable endpoint), the `err` object is returned but `resp` is cleanly and entirely `nil`. Dereferencing `resp.StatusCode `immediately triggers a fatal `panic`.(Note: Other subdirectories correctly guard this with `if resp != nil && resp.StatusCode == ...`).

## Steps to reproduce

1. Force the engine loop into a state where it evaluates an entity and attempts to update a Check Run.
2. Disconnect the environment's egress network access (or simulate a DNS blackhole for api.github.com).
3. The worker crashes with a panic due to a nil pointer dereference instead of bubbling up the error.

## Expected behavior
The network error should be cleanly handled and returned wrapped, allowing the engine/worker to mark the check attempt as failed (or defer execution) instead of crashing.

## Proposed Fix
Update the error handling block in both `StartCheckRun` and `UpdateCheckRun` to add a nil safety guard for the `resp` object before inspecting the `StatusCode`. Additionally, we should ensure the original error string isn't swallowed entirely on generic API failures so debug logs capture the actual networking issue.

AftAb-25:
## Description

This PR fixes a critical vulnerability where the engine loop could `panic` due to a nil pointer dereference if we encounter a low-level network failure while attempting to push check run updates.

Currently, if the `go-github` API call inside [StartCheckRun](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/providers/github/common.go:1032:0-1050:1) or [UpdateCheckRun](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/providers/github/common.go:1052:0-1066:1) (in [internal/providers/github/common.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/providers/github/common.go:0:0-0:0)) fails because of something like a DNS routing issue, a TLS handshake timeout, or a context cancellation, the `err` object is populated but the `resp` object is entirely `nil`. 

The existing error handling immediately tests `resp.StatusCode == 403` to verify permissions, which triggers a panic on the `nil` `resp` object, completely crashing the executing worker.

## Change
I bolted in a `resp != nil` guard specifically to both handlers before the `StatusCode` inspection so that real network outages are safely returned (or mapped back depending on the stack) instead of hard-crashing the background routine.

Fixes #6347

## Checklist

- [x] Code compiles cleanly
- [x] Includes tests for the changes (existing check run tests pass)

### DIFF
--- a/docs/docs/how-to/profile_selectors.md
+++ b/docs/docs/how-to/profile_selectors.md
@@ -47,7 +47,7 @@ selection:
 Let's break down the example above:
 
 - `entity`: Defines the type of entity you want to filter (`repository`,
-  `artifact`, or `pull_request`). In the case that the `entity` type is omitted,
+  `artifact`, `pull_request`, `release`, `build`, etc.). In the case that the `entity` type is omitted,
   the selector will be applied to all entities.
 - `selector`: The CEL expression that specifies the filtering criteria. In the
   example:
@@ -130,6 +130,21 @@ while the properties are provider-specific and prefixed with the provider name.
 | `github/pull_author_login` | The GitHub login of the author of the pull request | string |
 | `github/repo_name`         | The GitHub repo name (e.g. `stacklok`).            | string |
 | `github/repo_owner`        | The GitHub repo owner (e.g. `minder`).             | string |
+
+## Generic entity selectors
+
+For entities without a dedicated selector (such as `release` or `build`), you can access their properties using the `generic` object in your CEL expressions.
+
+| Field  | Description                                                  | Type   |
+| ------ | ------------------------------------------------------------ | ------ |
+| `name` | The full name of the generic entity, e.g. testorg/testrelease | string |
+
+For example, a selector for a `release` entity might look like:
+
+```yaml
+- entity: release
+  selector: generic.name == 'stacklok/minder/v1.0.0'
+```
 
 ## Entity provider selectors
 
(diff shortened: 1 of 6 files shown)


AftAb-25:
The fix was already merged into the main repository by someone else .  I did not notice it ... Now I closed the PR

PR #6350, opened 2026-04-12
fix: safely record evaluation errors to prevent silent loop aborts 

### LINKED ISSUE #6349: Bug: Engine silently aborts all profile security evaluations for an entity if a single rule type is misconfigured
## Describe the bug
There is a critical architectural vulnerability in the core engine's execution loop ([internal/engine/executor.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/engine/executor.go:0:0-0:0)). 

When an entity (like a repository or pull request) is received and processed, the [EvalEntityEvent](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/engine/executor.go:80:0-158:1) function correctly iterates over all assigned profiles and evaluates their rules. However, if `e.evaluateRule()` returns an initialization or configuration error, the engine immediately aborts the *entire* evaluation loop:

```go
		for _, rule := range profile.Rules {
			if err := e.evaluateRule(ctx, inf, provider,
                &profile, &rule, ruleEngineCache, profileEvalStatus);
                err != nil {
				return fmt.Errorf("error evaluating entity event: %w", err) 
              // <--- SILENT EVAL ABORT
			}
		}
```
The underlying issue is that `evaluateRule` will return a hard error if `GetRuleEngine` fails (e.g., if a custom rule type has malformed Rego syntax) or if the `ActionConfig` fails initialization.Because the engine throws the error back to the event handler and drops the entity locker:

1. **Denial of Service & Security Bypass:** If a user defines one single malformed rule inside one single profile, evaluating that rule will throw an error and abort the loop. Because the loop terminates, every other rule and profile running subsequently on that entity is silently skipped. This means critical security checks (like automated branch protections) can be bypassed entirely due to a minor configuration error in an unrelated profile.
2. **Missing Observability:** Because it returns before `createOrUpdateEvalStatus` is called, the database never records that the rule threw an evaluation error. The entity essentially ghost-skips its checks, leaving the system blind to the failure.

(Note: _This is conceptually identical to the profile execution loop bug fixed in #6330, but located at the absolute heart of the security evaluation pipeline_)

## Expected behavior
If `evaluateRule()` encounters a configuration or initialization failure for a specific rule, it should:

1. Construct an `EvalResult` recording the internal system's error state.
2. Write that "Error" status to the database via `createOrUpdateEvalStatus` so administrators know the rule config is broken.
3. Catch the initialization error and continue executing the rest of the rules and profiles so subsequent security checks are not bypassed.

## Proposed Fix
In `executor.go` inside `evaluateRule`: When `createEvalStatusParams`, `GetRuleEngine`, or `NewRuleActions` fail, we should not return the error directly. Instead, we should populate `evalParams.SetEvalErr(err)` and call `e.createOrUpdateEvalStatus(ctx, evalParams)`, log the issue, and finally return nil to gracefully allow the upstream profile loop to continue executing the subsequent policies.

AftAb-25:
## Description

This PR fixes a critical engine bug where a single malformed rule type would cause the entire executor loop to abruptly crash, silently skipping all subsequent security evaluations for an entity. 

Currently, in [internal/engine/executor.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/engine/executor.go:0:0-0:0), if [evaluateRule](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/engine/executor.go:160:0-222:1) encounters an error while initializing the rule engine (`GetRuleEngine`) or compiling its actions (`NewRuleActions`), it immediately bubbles that error up to [EvalEntityEvent](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/engine/executor.go:43:1-43:76) via a bare `return`. That return breaks out of the profile iteration loops completely, meaning the system never evaluates any other rules *and* never logs the error status to the database.

## Change
I refactored the error handling logic in [evaluateRule](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/engine/executor.go:160:0-222:1) during the rule creation phase. Instead of returning the raw error up the stack and dropping the lock, we now catch those configuration/compilation failures, load them gracefully into `evalParams.SetEvalErr`, and explicitly track them via `e.createOrUpdateEvalStatus`.

This allows the engine to record the initialization failure state in the DB so administrators actually know their rule configuration is broken, and more importantly, allows it to return `nil`, meaning the engine loop survives and seamlessly continues evaluating the remainder of the repository's security checks.

Fixes #6349

## Checklist

- [x] Code compiles cleanly
- [x] Includes tests for the changes (existing engine execution tests pass)

### DIFF
--- a/internal/engine/eval_status.go
+++ b/internal/engine/eval_status.go
@@ -109,10 +109,16 @@ func (e *executor) createOrUpdateEvalStatus(
 	alertStatus := dbadapter.ErrorAsAlertStatus(params.GetActionsErr().AlertErr)
 	e.metrics.CountAlertStatus(ctx, alertStatus)
 
-	chckpoint := params.GetIngestResult().GetCheckpoint()
-	chkpjs, err := chckpoint.ToJSONorDefault(json.RawMessage(`{}`))
-	if err != nil {
-		logger.Err(err).Msg("error marshalling checkpoint")
+	var chkpjs json.RawMessage
+	var err error
+	if ingestRes := params.GetIngestResult(); ingestRes != nil {
+		chckpoint := ingestRes.GetCheckpoint()
+		chkpjs, err = chckpoint.ToJSONorDefault(json.RawMessage(`{}`))
+		if err != nil {
+			logger.Err(err).Msg("error marshalling checkpoint")
+		}
+	} else {
+		chkpjs = json.RawMessage(`{}`)
 	}
 
 	var evalOutput any
--- a/internal/engine/executor.go
+++ b/internal/engine/executor.go
@@ -176,14 +176,20 @@ func (e *executor) evaluateRule(
 	// retrieve the rule type engine from the cache
 	ruleEngine, err := ruleEngineCache.GetRuleEngine(ctx, rule.RuleTypeID)
 	if err != nil {
-		return fmt.Errorf("error creating rule type engine: %w", err)
+		evalErr := fmt.Errorf("error creating rule type engine: %w", err)
+		evalParams.SetEvalErr(evalErr)
+		logEval(ctx, inf, evalParams, "")
+		return e.createOrUpdateEvalStatus(ctx, evalParams)
 	}
 
 	// create the action engine for this rule instance
 	// unlike the rule type engine, this cannot be cached
 	actionEngine, err := actions.NewRuleActions(ctx, ruleEngine.GetRuleType(), provider, &profile.ActionConfig)
 	if err != nil {
-		return fmt.Errorf("cannot create rule actions engine: %w", err)
+		evalErr := fmt.Errorf("cannot create rule actions engine: %w", err)
+		evalParams.SetEvalErr(evalErr)
+		logEval(ctx, inf, evalParams, ruleEngine.GetRuleType().Name)
+		return e.createOrUpdateEvalStatus(ctx, evalParams)
 	}
 
 	// Update the lock lease at the end of the evaluation


internal/engine/executor.go
```diff
@@ -176,14 +176,20 @@ func (e *executor) evaluateRule(
 	// retrieve the rule type engine from the cache
 	ruleEngine, err := ruleEngineCache.GetRuleEngine(ctx, rule.RuleTypeID)
 	if err != nil {
-		return fmt.Errorf("error creating rule type engine: %w", err)
+		evalErr := fmt.Errorf("error creating rule type engine: %w", err)
+		evalParams.SetEvalErr(evalErr)
+		logEval(ctx, inf, evalParams, "")
```
evankanderson, line 181: Repeating these lines multiple times doesn't seem correct -- particularly since you _also_ don't cover all the exit paths from this function.

internal/engine/eval_status.go
```diff
@@ -109,10 +109,16 @@ func (e *executor) createOrUpdateEvalStatus(
 	alertStatus := dbadapter.ErrorAsAlertStatus(params.GetActionsErr().AlertErr)
 	e.metrics.CountAlertStatus(ctx, alertStatus)
 
-	chckpoint := params.GetIngestResult().GetCheckpoint()
-	chkpjs, err := chckpoint.ToJSONorDefault(json.RawMessage(`{}`))
-	if err != nil {
-		logger.Err(err).Msg("error marshalling checkpoint")
+	var chkpjs json.RawMessage
+	var err error
+	if ingestRes := params.GetIngestResult(); ingestRes != nil {
```
evankanderson, line 114: You don't need to check if `ingestRes` returns `nil`, because `GetCheckpoint` is implemented as:

```golang
func (r *Ingested) GetCheckpoint() *checkpoints.CheckpointEnvelopeV1 {
	if r == nil {
		return nil
	}

	return r.Checkpoint
}
```

internal/engine/executor.go
evankanderson: (A better option might be to restructure this to a `defer` pattern.)

evankanderson (commented):
A failure in `GetRuleEngine` or `NewRuleActions` generally would indicate one of the following scenarios:

1. Database unreachability. (We're kinda screwed in this case; our best option is probably to log an error)
2. Database corruption of stored data. (e.g. from manual DB poking -- again, 
3. Failed precondition checks in the APIs. (i.e. we've accepted ruletypes that shouldn't have passed validation)

With that said, this makes sense as a _backup_ defense, but I wouldn't characterize this as a "critical" vulnerability, since it requires that we have already corrupted the database state outside what a minder server would accept.

evankanderson (changes requested):
I'm marking this PR as "request changes" (has two comments to be addressed, it's not clear that avoiding a panic due to failed invariants is correct) to help track which outstanding PRs need maintainer action vs contributor action.

If you don't think this PR is necessary anymore, feel free to close it.

evankanderson:
Given the [comments here](https://github.com/mindersec/minder/pull/6350#pullrequestreview-4191642858) and the general pattern of `panic` only being used for unexpected logic errors which should be covered by tests, closing this.

PR #6355, opened 2026-04-13
feat: add interactive 'minder profile edit' command

### LINKED ISSUE #2283: minder profile edit

I would like a `minder profile edit` command that would let me edit an existing profile

### Solution Proposal

- grab the current profile
- dump it into a YAML file in /tmp
- open `$EDITOR` pointing at the file
- upon existing the editor, pass the contents of the file to `UpdateProfile` handler

_No response_

### Additional context

_No response_

### Acceptance Criteria

_No response_

DharunMR:
# Summary

This PR introduces the minder profile edit command, allowing users to fetch a profile, modify it in their local text editor, and apply changes in a single interactive workflow.

- Implemented a robust lookup system that checks $VISUAL -> $EDITOR -> system path (searching for nano, vim, nvim, vi, or emacs) using exec.LookPath.

- The implementation ensures that critical fields such as Id, Context, Type, and Version are preserved from the original fetch. This prevents validation errors if the user accidentally deletes these fields during the editing session.

---

- Fixes #2283 

---

**Before** 
swappy-20260413_185940

**After**
swappy-20260413_190001

### DIFF
--- a/cmd/cli/app/profile/edit.go
+++ b/cmd/cli/app/profile/edit.go
@@ -1,0 +1,188 @@
+// SPDX-FileCopyrightText: Copyright 2024 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package profile
+
+import (
+	"bytes"
+	"cmp"
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/mindersec/minder/internal/util"
+	"github.com/mindersec/minder/internal/util/cli"
+	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+)
+
+var editCmd = &cobra.Command{
+	Use:   "edit",
+	Short: "Edit an existing profile",
+	Long:  `The profile edit subcommand lets you fetch an existing profile, edit it in your $EDITOR, and apply the updates.`,
+	RunE:  cli.GRPCClientWrapRunE(editCommand),
+}
+
+func editCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc.ClientConn) error {
+	client := minderv1.NewProfileServiceClient(conn)
+	project := viper.GetString("project")
+	id := viper.GetString("id")
+	name := viper.GetString("name")
+
+	if id == "" && name == "" {
+		return cli.MessageAndError("Error editing profile", fmt.Errorf("id or name required"))
+	}
+	cmd.SilenceUsage = true
+
+	prof, err := getProfile(ctx, client, project, id, name)
+	if err != nil {
+		return err
+	}
+
+	// hardcoded type and version since ParseResource requires it
+	if prof.Type == "" {
+		prof.Type = "profile"
+	}
+	if prof.Version == "" {
+		prof.Version = "v1"
+	}
+
+	yamlString, err := util.GetYamlFromProto(prof)
+	if err != nil {
+		return cli.MessageAndError("Error marshaling profile to YAML", err)
+	}
+
+	tmpFile, err := os.CreateTemp("", "tmp-minder-profile-*.yaml")
+	if err != nil {
+		return cli.MessageAndError("Error creating temporary file", err)
+	}
+	defer os.Remove(tmpFile.Name())
+
+	if _, err := tmpFile.WriteString(yamlString); err != nil {
+		return cli.MessageAndError("Error writing to temporary file", err)
+	}
+
+	// we must close the file descriptor before handing it to the editor
+	// many terminal editor perform atomic saves which changes the file's inode
+	// keeping the old FD open would read the orphaned file.
+	if err := tmpFile.Close(); err != nil {
+		return cli.MessageAndError("Error closing temporary file", err)
+	}
+
+	if err := handleEditor(tmpFile.Name()); err != nil {
+		return err
+	}
+
+	updatedBytes, err := os.ReadFile(tmpFile.Name())
+	if err != nil {
+		return cli.MessageAndError("Error reading updated temporary file", err)
+	}
+
+	if string(updatedBytes) == yamlString {
+		cmd.Println("No changes made to the profile. Aborting update.")
+		return nil
+	}
+
+	return updateProfile(ctx, client, prof, updatedBytes, cmd)
+}
+
+func getProfile(ctx context.Context, client minderv1.ProfileServiceClient, project, id, name string) (*minderv1.Profile, error) {
+	if id != "" {
+		p, err := client.GetProfileById(ctx, &minderv1.GetProfileByIdRequest{
+			Context: &minderv1.Context{Project: &project},
+			Id:      id,
+		})
+		if err != nil {
+			return nil, cli.MessageAndError("Error getting profile by ID", err)
+		}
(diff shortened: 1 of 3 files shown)


cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	if err := tmpFile.Close(); err != nil {
+		return cli.MessageAndError("Error closing temporary file", err)
+	}
+
+	if err := handleEditor(tmpFile.Name()); err != nil {
+		return err
+	}
+
+	updatedBytes, err := os.ReadFile(tmpFile.Name())
```
evankanderson, line 71: I'm assuming we need to close and reopen the file because some editors will do a replace and our old FD will be pointing somewhere else?

(Might be worth a comment in the future)

cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, cli.MessageAndError("Error getting profile by name", err)
+	}
+	return p.GetProfile(), nil
+}
+
+func handleEditor(fileName string) error {
+	editorCmd := os.Getenv("VISUAL")
+	if editorCmd == "" {
+		editorCmd = os.Getenv("EDITOR")
+	}
```
evankanderson, line 110: I'm a fan of this pattern where it works:

```suggestion
	editorCmd := cmp.Or(os.Getenv("VISUAL"), os.Getenv("EDITOR"))
```

cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,189 @@
...
+		for _, e := range commonEditors {
+			if _, err := exec.LookPath(e); err == nil {
+				editorCmd = e
+				break
+			}
+		}
+	}
+
+	if editorCmd == "" {
+		msg := "no editor found in $PATH. Please install nano/vim or set $EDITOR"
```
evankanderson, line 123: I'd just suggest setting $EDITOR, to also better accommodate multiple platforms:

```suggestion
		msg := "no editor found in $PATH. Please set $EDITOR to your preferred editor"
```

cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,189 @@
...
+		return cli.MessageAndError(msg, fmt.Errorf("no editor found"))
+	}
+
+	args := strings.Fields(editorCmd)
```
evankanderson, line 127: This feels like it's probably wrong -- it seems like `$EDITOR` should be a path, not a series of arguments.  i.e. if I set `EDITOR="/home/me/my tools/Sublime.app"`, this would try to exec `/home/me/my`.

cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,189 @@
...
+) error {
+	var updatedProfile minderv1.Profile
+	updatedJSON, err := yaml.YAMLToJSON(updatedBytes)
+	if err != nil {
+		return cli.MessageAndError("Error parsing updated YAML", err)
+	}
+	um := protojson.UnmarshalOptions{DiscardUnknown: true}
+	if err := um.Unmarshal(updatedJSON, &updatedProfile); err != nil {
+		return cli.MessageAndError("Error loading updated profile data into protobuf", err)
+	}
```
evankanderson, line 157: You may want to use `minderv1.ParseResource` to convert YAML to a protobuf.

evankanderson (commented):
A few comments; I'd merge this except that I think the handling of `$EDITOR` is probably wrong (should not be space-split).

cmd/cli/app/profile/edit.go
DharunMR: my only concern is GetProfile API response drops the type and version fields so i had to manually inject them into the struct before generating the yaml so ParseResource wouldnt fail validation.

cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,188 @@
...
+
+var editCmd = &cobra.Command{
+	Use:   "edit",
+	Short: "Edit an existing profile",
+	Long:  `The profile edit subcommand lets you fetch an existing profile, edit it in your $EDITOR, and apply the updates.`,
+	RunE:  cli.GRPCClientWrapRunE(editCommand),
+}
+
+func editCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc.ClientConn) error {
+	client := minderv1.NewProfileServiceClient(conn)
```
evankanderson: Sorry about the long delay; you'll probably need to switch this to the new `cli.GetGRPCClient`

cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,188 @@
...
+	if err := minderv1.ParseResource(bytes.NewReader(updatedBytes), &updatedProfile); err != nil {
+		return cli.MessageAndError("Error parsing updated profile YAML", err)
+	}
+
+	updatedProfile.Id = proto.String(oldProf.GetId())
+	updatedProfile.Context = oldProf.GetContext()
+	updatedProfile.Type = oldProf.GetType()
+	if updatedProfile.Type == "" {
+		updatedProfile.Type = "profile"
+	}
```
evankanderson: `cmp.Or` can simplify this:

```suggestion
	updatedProfile.Type = cmp.Or(oldProf.GetType(), "profile")
```

evankanderson (commented):
Sorry that I missed commenting on this until we'd iterated a bunch on the pattern from #6286 .  I think this needs a sync now, but I'd love to merge this.

DharunMR:
@evankanderson made changes 👍

cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,200 @@
...
+	client, closeConn, err := cli.GetCLIClient(cmd, minderv1.NewProfileServiceClient)
+	if err != nil {
+		return cli.MessageAndError("Error connecting to server", err)
+	}
+	defer closeConn()
+
+	if id == "" && name == "" {
+		return cli.MessageAndError("Error editing profile", fmt.Errorf("id or name required"))
+	}
+
```
evankanderson, line 59: Generally, this should go before `cmd.SilenceUsage`, so you'll get the command usage alongside the error message.  (Given that the usage also includes flags for connecting to the server, I'd just move line 48 to line 59 if you want to do a follow-up PR.)

cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,200 @@
...
+		return cli.MessageAndError("Error parsing updated profile YAML", err)
+	}
+
+	updatedProfile.Id = proto.String(oldProf.GetId())
+	updatedProfile.Context = oldProf.GetContext()
+	updatedProfile.Type = cmp.Or(oldProf.GetType(), defaultProfileType)
+	updatedProfile.Version = oldProf.GetVersion()
+	if updatedProfile.Version == "" {
+		updatedProfile.Version = defaultProfileVersion
+	}
```
evankanderson, line 179:
```suggestion
	updatedProfile.Version = cmp.Or(oldProf.GetVersion(), defaultProfileVersion)
```

cmd/cli/app/profile/edit.go
```diff
@@ -0,0 +1,200 @@
...
+
+	updateCtx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
evankanderson, line 181: This should probably be 20s or less; it would be surprising for UpdateProfile to take even 5s on the server side.

evankanderson (approved):
I'm going to merge this, feel free to a follow-up PR for the items mentioned.

Issue #3842, opened 2024-07-11
Build a RuleType to check that 2FA is enabled


puerco:
Let's create a RuleType that verifies and alerts when two factor auth in an organization is turned off

evankanderson:
I'm not sure that we have a way to check policy on the organization level at the moment (e.g. we don't have an Organization entity)

puerco:
Also: Changing the setting also does not produce a webhook call so we need to find another way to monitor it.

evankanderson:
@ethomson for prioritization as an epic

evankanderson:
(Blocked on #5377)

PR #6356, opened 2026-04-13
feat(entities): Implement ENTITY_ORGANIZATION for Github Providers 

### LINKED ISSUE #5377: Add a organization entity
This would allow for inspection of organization properties and directly support OSPS-AC-01.

Jaydeep869:
## Description

This PR fully implements `ENTITY_ORGANIZATION` to act as a proper target for evaluation rules and effectively unblocks the 2FA checking mechanics highlighted in #3842.

As per recent discussions with @evankanderson, rather than introducing custom endpoints or manual administration routines, this integration relies natively on the new generic entity properties architecture and auto registers organizations fully in the background directly after a user successfully installs the Github Provider app. 

### Key Changes
* **Database & Protobufs**:
  * Altered Postgres `entities` enum via a new migration (`000117_organization_entity.up.sql`) to include the `'organization'` value.
  * Added `ENTITY_ORGANIZATION = 9` natively within `minder.proto`.
* **Generic Entity Engine Integration**:
  * Implemented an `OrganizationFetcher` within `github/properties` that queries Github's API to construct `properties.Properties` mapping the `is_user` and `avatar_url` fields correctly.
  * Implemented `OrganizationValidator` for reliable validation logic and registered `ENTITY_ORGANIZATION` in `service.go`.
* **Asynchronous Auto Registration LifeCycle**:
  * Hooked an auto deployment execution in `CreateGitHubAppProvider` to publish a `MinderEvent` to the `TopicQueueReconcileEntityAdd` Watermill queue synchronously upon provider registration. The control plane natively handles this via background reconcile pipelines.
* **Database Backfill**:
  * Executed a programmatic hook (`backfill_organizations.go`) linked through the `migrate up` procedure to automatically traverse all existing user/database providers and populate any orphaned `organization` links via transaction safety upon startup.
* **Testing Adjustments**: 
  * Safely exposed `*db.Provider` as a third return value payload originating from `CreateGitHubAppWithoutInvitation` and rebuilt all the testing mocks appropriately matching the expectation logic.

## Related Issues
* Fixes #5377 
* Unblocks #3842 

## Testing Performed
- [x] Code generation executed successfully (`make buf`, `make sqlc`, `make mock`).
- [x] Ran unit tests locally (`make test-silent`), fixing broken mock signatures.
- [x] Verified migration runs cleanly with `make run-docker`.


cmd/server/app/backfill_organizations.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/rs/zerolog"
+
+	"github.com/mindersec/minder/internal/db"
+	"github.com/mindersec/minder/pkg/entities/properties"
+)
+
+func backfillOrganizations(ctx context.Context, store db.Store) error {
```
evankanderson, line 20: Prefer to put this function somewhere (entities?) where it can be tested, rather than in `cmd/server`, which we exclude from coverage checks because the code here should be small and mostly setup.

cmd/server/app/backfill_organizations.go
```diff
@@ -0,0 +1,86 @@
...
+
+	provs, err := store.GlobalListProvidersByClass(ctx, db.ProviderClassGithubApp)
+	if err != nil {
+		return fmt.Errorf("failed to list providers: %w", err)
+	}
+
+	count := 0
+
+	for _, prov := range provs {
+		login := strings.TrimPrefix(prov.Name, string(db.ProviderClassGithubApp)+"-")
```
evankanderson, line 32: This feels a bit fragile.  Do we have other information that would let us check this more definitively?

The "right" way to do this would be to resolve or use the `organization_id` from `provider_github_app_installations`, rather than doing surgery on the name.  I'm not sure if that's too much for this PR, though.  If we do it, we should _also_ update `CanHandleOwner` in `internal/providres/github/common.go`.

cmd/server/app/backfill_organizations.go
```diff
@@ -0,0 +1,86 @@
...
+				"minder.internal.type":  "string",
+				"minder.internal.value": login,
+			}
+			propBytes, _ := json.Marshal(propVal)
+
+			_, err = qtx.UpsertProperty(ctx, db.UpsertPropertyParams{
+				EntityID: ent.ID,
+				Key:      properties.PropertyName,
+				Value:    propBytes,
+			})
```
evankanderson, line 71: Can this use some common code with `providers/github/properties/organization.go`?

internal/controlplane/handlers_oauth.go
```diff
@@ -436,18 +436,14 @@ func (s *Server) processAppCallback(ctx context.Context, w http.ResponseWriter,
...
-			if errors.Is(err, service.ErrInvalidTokenIdentity) {
-				return newHttpError(http.StatusForbidden, "User token mismatch").SetContents(
-					"The provided login token was associated with a different GitHub user.")
-			}
-			return fmt.Errorf("error creating GitHub App provider: %w", err)
+			return handleProviderCreationError(err)
+		}
+
+		if dbProv != nil {
+			login := strings.TrimPrefix(dbProv.Name, string(db.ProviderClassGithubApp)+"-")
```
evankanderson, line 445: Even if we don't end up adding a method to fetch the current organization name given the ID, we should at least centralize the name mangling in a function where we can add it later if needed.  Ideally, that function would be a part of the GitHub provider.

internal/controlplane/handlers_oauth.go
```diff
@@ -436,18 +436,14 @@ func (s *Server) processAppCallback(ctx context.Context, w http.ResponseWriter,
 
 		logger.BusinessRecord(ctx).Project = stateData.ProjectID
 
-		var confErr providers.ErrProviderInvalidConfig
-		_, err = s.ghProviders.CreateGitHubAppProvider(ctx, *token, stateData, installationID, state)
+		dbProv, err := s.ghProviders.CreateGitHubAppProvider(ctx, *token, stateData, installationID, state)
```
evankanderson, line 439: Since the only time the return value from `CreateGitHubAppProvider` is used is in a unit test, it should be easy to change that to return a provider facet or some other struct which can be used to resolve any provider properties we need.  (Currently, it seems like mostly the org name, but possibly other data in the future.)

internal/controlplane/handlers_oauth.go
```diff
@@ -541,7 +537,17 @@ func (s *Server) handleAppInstallWithoutInvite(ctx context.Context, token *oauth
...
-		return s.ghProviders.CreateGitHubAppWithoutInvitation(ctx, qtx, *userID, installationID)
+		proj, dbProv, err := s.ghProviders.CreateGitHubAppWithoutInvitation(ctx, qtx, *userID, installationID)
+		if err != nil {
+			return nil, err
+		}
+		if dbProv != nil && proj != nil {
+			login := strings.TrimPrefix(dbProv.Name, string(db.ProviderClassGithubApp)+"-")
+			// It is generally safe to publish an event from within a transaction, as long
+			// as the event handler evaluates the state matching later.
+			s.publishOrganizationEntityEvent(ctx, dbProv.ID, proj.ID, login)
```
evankanderson, line 548: It occurs to me that the `dbProv` object we return from `CreateGitHubAppWithoutInvitation` and elsewhere could instead include a `publishEntityEvent()` method, depending on how much trouble it is to plumb `s.evt` to the ghProvider.

internal/controlplane/handlers_profile.go
```diff
@@ -427,7 +427,7 @@ func (s *Server) getRuleEvalStatus(
 				repoPath = fmt.Sprintf("%s/%s", prRepoOwner, prRepoName)
 			}
 		case db.EntitiesBuildEnvironment, db.EntitiesRelease, db.EntitiesPipelineRun,
-			db.EntitiesTaskRun, db.EntitiesBuild:
+			db.EntitiesTaskRun, db.EntitiesBuild, db.EntitiesOrganization:
```
evankanderson, line 430: This is probably okay for now, but it seems like alerts will have incorrect URLs (and log errors) until we fix it.  It can be a good idea to file an issue for this if we merge this PR, to track the need to update it later.

internal/controlplane/handlers_user.go
```diff
@@ -134,11 +135,15 @@ func (s *Server) claimGitHubInstalls(ctx context.Context, qtx db.ExtendQuerier)
...
-		proj, err := s.ghProviders.CreateGitHubAppWithoutInvitation(ctx, qtx, userID, i.AppInstallationID)
+		proj, dbProv, err := s.ghProviders.CreateGitHubAppWithoutInvitation(ctx, qtx, userID, i.AppInstallationID)
 		if err != nil {
 			zerolog.Ctx(ctx).Error().Err(err).Int64("org_id", i.OrganizationID).Msg("failed to create GitHub app at first login")
 			continue
 		}
+		if dbProv != nil {
+			login := strings.TrimPrefix(dbProv.Name, string(db.ProviderClassGithubApp)+"-")
+			s.publishOrganizationEntityEvent(ctx, dbProv.ID, proj.ID, login)
+		}
```
evankanderson, line 146: Again, let's try to centralize this pattern in the GitHub provider.

internal/eea/eea.go
```diff
@@ -255,7 +255,7 @@ func (e *EEA) buildEntityWrapper(
 	case db.EntitiesPullRequest:
 		return e.buildPullRequestInfoWrapper(ctx, entityID, projID)
 	case db.EntitiesBuildEnvironment, db.EntitiesRelease,
-		db.EntitiesPipelineRun, db.EntitiesTaskRun, db.EntitiesBuild:
+		db.EntitiesPipelineRun, db.EntitiesTaskRun, db.EntitiesBuild, db.EntitiesOrganization:
 		return nil, fmt.Errorf("entity type %q not yet supported", entity)
```
evankanderson, line 259: I think leaving this this way will prevent us from evaluating policy on organizations (but I'd need to test to be sure).  Definitely something to file a follow-up issue on if we leave like this.

internal/entities/service/validators/organization.go
```diff
@@ -0,0 +1,30 @@
...
+
+// Validate checks if an organization entity can be created
+func (*OrganizationValidator) Validate(
+	_ context.Context,
+	_ *properties.Properties,
+	_ uuid.UUID,
+) error {
+	// For now, any organization properties that make it this far are valid
+	return nil
+}
```
evankanderson, line 30: If you don't register this in `internal/service/service.go`, then you won't need to have this struct at all.

(I'd need to think about what policies we might want for organizations, so I'd remove this for now.)

internal/providers/github/properties/organization.go
```diff
@@ -0,0 +1,91 @@
...
+func NewOrganizationFetcher() *OrganizationFetcher {
+	return &OrganizationFetcher{
+		propertyFetcherBase: propertyFetcherBase{
+			propertyOrigins: []propertyOrigin{
+				{
+					keys: []string{
+						properties.PropertyUpstreamID,
+						properties.PropertyName,
+						properties.OrgPropertyIsUser,
+						properties.OrgPropertyAvatarURL,
```
evankanderson, line 31: Avatar URL seems like an odd thing to select on.  How about:

* `has_organization_projects`
* `created_at`
* `plan.name` as `plan_name` or `plan_type`?

internal/providers/github/properties/organization.go
```diff
@@ -0,0 +1,91 @@
...
+) (map[string]any, error) {
+	// We can look up by either exact upstream ID or by name (login).
+	var user *go_github.User
+	var err error
+
+	if idStr := lookupProperties.GetProperty(properties.PropertyUpstreamID).GetString(); idStr != "" {
+		id, parseErr := strconv.ParseInt(idStr, 10, 64)
+		if parseErr != nil {
+			return nil, fmt.Errorf("invalid upstream ID: %w", parseErr)
+		}
```
evankanderson, line 61: `Property` has an `AsInt64()` that returns an error if the value is not an Int64.  That is probably better than getting it as a string and then doing the conversion yourself.

internal/providers/github/properties/organization.go
```diff
@@ -0,0 +1,91 @@
...
+		properties.PropertyUpstreamID: properties.NumericalValueToUpstreamID(user.GetID()),
+		properties.PropertyName:       user.GetLogin(),
+		properties.OrgPropertyIsUser:  user.GetType() == "User",
+	}
+
+	if user.AvatarURL != nil {
+		result[properties.OrgPropertyAvatarURL] = *user.AvatarURL
+	}
+	if user.Company != nil {
+		result[properties.OrgPropertyCompany] = *user.Company
```
evankanderson, line 87: Company appears to be `null` for a large number of orgs; I'm not sure how it's normally set.

pkg/profiles/util.go
```diff
@@ -424,6 +426,7 @@ func rowInfoToProfileMap(
 		profile.PipelineRun = ruleset
 	case pb.Entity_ENTITY_TASK_RUN:
 		profile.TaskRun = ruleset
+	case pb.Entity_ENTITY_ORGANIZATION:
```
evankanderson, line 429: Go case statements don't automatically fall through to the next statement; it might be worth adding a comment about how we don't have organization support in profiles yet.

evankanderson (commented):
It looks like this sets up the lifecycle operations for (GitHub) organization entities, but it doesn't actually enable writing rules or profiles that enforce policies on organizations.  I think that's fine, but it would be good to indicate the remaining work in the PR description as well as what _is_ implemented.

Have you tested this end-to-end in a local minder instance?  If not, I can try to do that in the next few days.

evankanderson (changes requested):
I'm marking this PR as "request changes" (has 10+ pending comments) to help track which outstanding PRs need maintainer action vs contributor action.

Jaydeep869:
Hi @evankanderson! Thanks for the insightful and deep review. I really appreciate it. I apologize for the massive delay in circling back to this. I’ve been incredibly busy knocking out my end semester exams.

I have pushed up a fresh commit that tackles all of your feedback.

cmd/server/app/backfill_organizations.go
Jaydeep869: Done! I relocated this logic directly into [backfill.go which allows proper testability without violating the coverage bounds on server.

cmd/server/app/backfill_organizations.go
Jaydeep869: I agree that doing surgery on the string everywhere is fragile. Rather than heavily re-architecting the DB Provider payload parsing all the way down to organization_id right now (which I feel is a bit out of scope for this immediate PR), I've centralized a func GetGithubAppOwner() helper over in github/common.go. Using it ensures we only define that app formatting string exactly once, making it safe and easy to swap out with ID resolution later if needed.

internal/providers/github/properties/organization.go
Jaydeep869: I scrapped both avatar_url and company and swapped them out exactly as you requested to map has_organization_projects, created_at, and plan_name instead.

internal/providers/github/properties/organization.go
Jaydeep869: I just updated this to parse utilizing .AsInt64() natively instead of grabbing the string manually.

pkg/profiles/util.go
Jaydeep869: For these occurrences, I have added explicit TODO: comments in the code next to the switch statements to explicitly state that org alerts/evaluations are deferred for now. Since this PR purely scopes the core Entity Registration mechanics, I will write up follow-up issues to track actually evaluating policies/alerts against them later so we don't forget!

internal/entities/service/validators/organization.go
Jaydeep869: Done. I've deleted the OrganizationValidator file and logic entirely and stripped its registry setup in service.go . We can re-include a validator later when actual profiles are built for organizations.

evankanderson:
> Hi @evankanderson! Thanks for the insightful and deep review. I really appreciate it. I apologize for the massive delay in circling back to this. I’ve been incredibly busy knocking out my end semester exams.
> 
> I have pushed up a fresh commit that tackles all of your feedback.

No problem, I've had quite a backlog as well.

evankanderson:
I think you may need to rename your migration file, as we had another 000117 for adding a Rego version to the ruletypes table.

Jaydeep869:
Hey @evankanderson, can you review it now

internal/controlplane/handlers_oauth.go
evankanderson: Rather than returning a bunch of state which needs to be passed along to `publishOrganizationEntityEvent`, let's move that call into the `ghProviders` service, rather than having it here and in `handlers_user.go`.

evankanderson (changes requested):
From testing, it looks like this PR creates organization entities when Minder starts up, but creating a GitHub App provider doesn't create the organization entity immediately (only on next backfill).

We also don't have the other side of the lifecycle (delete) implemented.  There are several delete scenarios to cover:

1. Delete the provider
2. Delete the project
3. Last user owning the project deletes their account

I **think** the existing infrastructure makes case 2 and case 3 the same, but we'll need to end-to-end test for these scenarios.  (I can do that, which is how I discovered that the entity create was only implemented for backfill.)


There are also some other previous comments which where unaddressed.  (You can say things like "this doesn't work" or "that would double the size of the PR" for comments that you don't think are within the PR's scope.)

PR #6357, opened 2026-04-13
feat(cli): add in-memory repository clone utility

sachin9058:
# Summary

Add a utility function to clone a git repository in memory using go-git.

This is a preparatory step toward improving the quickstart command to dynamically load rule and profile catalogs from a repository, as discussed in #6339.

The utility uses go-git with in-memory storage to avoid disk operations and provide a lightweight way to access repository contents.

No functional changes are introduced in this PR.

Fixes Part of #6339

# Testing

- Verified the CLI builds successfully:
- go build ./cmd/cli

- Ran all tests:

  go test ./...

- Ran lint checks:

  make lint-fix

No behavior changes were introduced.


cli
evankanderson, line 1: We shouldn't be checking in binary files.

cmd/cli/internal/repo/clone.go
evankanderson, line 1: This doesn't feel worth building a separate file to wrap a one-line implementation.

evankanderson (commented):
Can we just roll this into the other clone PR, as this seems to either include a binary file (unreviewable) or only a single-line implementation?

cmd/cli/internal/repo/clone.go
sachin9058: Thanks for the feedback!

I can inline the clone logic directly into the quickstart flow instead of introducing a separate helper and package.

I'll update the PR accordingly and remove the extra abstraction.

PR #6358, opened 2026-04-13
feat: add /version endpoint and global Minder-Version header

### LINKED ISSUE #4048: Add a "/version" endpoint for Minder server
The following issue is about adding a version endpoint which returns the version of the Minder server you're connected to.

DharunMR:
# Summary

This PR addresses the need to easily expose the Minder server's build version and commit hash for tooling and debugging purposes, as requested. It introduces a dedicated, unauthenticated `/api/v1/version` endpoint and adds a global gRPC Unary Interceptor to stamp the server version on all incoming HTTP/gRPC responses.

- Fixes #4048 

## Changes Included
* **Protobuf:** Added `GetVersion` RPC to `HealthService` mapped to `GET /api/v1/version`. Configured `rpc_options` to allow unauthenticated access (`TARGET_RESOURCE_NONE`).
* **Server Handler:** Implemented the `GetVersion` logic to return `constants.CLIVersion` and `constants.Revision`.
* **Global Header:** Created `VersionHeaderInterceptor` to inject the `Minder-Version` header into the response metadata for all requests.
* **Telemetry Fix:** Replaced the hardcoded `"v0.1.0"` service version in `initMetrics` with the dynamic `constants.CLIVersion`.

## How I Tested This
1. Verified the new endpoint returns the correct JSON payload and header:
   `curl -i http://localhost:8080/api/v1/version`
2. Verified that the global header is successfully injected into other pre-existing endpoints:
   `curl -I http://localhost:8080/api/v1/health`

---
swappy-20260422_092030

### DIFF
--- a/internal/controlplane/server.go
+++ b/internal/controlplane/server.go
@@ -35,6 +35,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/reflection"
 	"google.golang.org/grpc/status"
 
@@ -43,6 +44,7 @@ import (
 	"github.com/mindersec/minder/internal/auth"
 	"github.com/mindersec/minder/internal/auth/jwt"
 	"github.com/mindersec/minder/internal/authz"
+	"github.com/mindersec/minder/internal/constants"
 	"github.com/mindersec/minder/internal/controlplane/metrics"
 	"github.com/mindersec/minder/internal/crypto"
 	datasourcessvc "github.com/mindersec/minder/internal/datasources/service"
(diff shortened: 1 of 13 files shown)


internal/controlplane/server.go
```diff
@@ -542,3 +544,12 @@ func withMaxSizeMiddleware(h http.Handler) http.Handler {
 		h.ServeHTTP(w, r)
 	})
 }
+
+// VersionHeaderInterceptor injects the Minder version into the HTTP response headers
+func VersionHeaderInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		// SetHeader tells gRPC-Gateway to add this to the HTTP response
+		_ = grpc.SetHeader(ctx, metadata.Pairs("x-minder-version", constants.CLIVersion))
```
evankanderson, line 552: [RFC 6648](https://www.rfc-editor.org/rfc/rfc6648) deprecates the "X-" prefix on headers, so just call this either "minder-version" or "server" (a standard header).  If you use `server`, you might want to include the string "Minder" before the version.

evankanderson: In fact, `constants.ServerUserAgent` might be exactly what you want.

Note that `constants.CLIVersion` only appears to be set via ldflags for CLI builds, not for the server.  The Makefile and GitHub Actions is probably the right place to do this if you want to; you'll need to use GOFLAGS per https://ko.build/advanced/faq/#how-can-i-set-ldflags

DharunMR: thanks for pointing me toward `GOFLAGS` the server binary now correctly stamp the version

swappy-20260422_092030

is commit hash required ??

evankanderson (approved):

PR #6361, opened 2026-04-13
Populate entityInfo for all entity types in ListEvaluationResults

### LINKED ISSUE #3116: Entity info is missing in `ListEvaluationResults` for pull requests and artifacts
### Describe the issue

When calling `ListEvaluationResults`, the response includes the `entityInfo` only if the entity type is a repository.
This is different than the response from `GetProfileStatusByName`, which includes the `entityInfo` for all types, including pull requests and artifacts.

We should update `ListEvaluationResults` to return `entityInfo` for all entities. Ideally we should reuse the code across both functions.
Additionally, consider whether `entityInfo` should be a struct, instead of a simple map.

### To Reproduce

_No response_

_No response_

krrish175-byte:
## Summary

This PR addresses an issue where the `ListEvaluationResults` RPC only populated `entityInfo` for repository entities, leaving it empty for other entity types such as Pull Requests and Artifacts. 

To fix this, the `getRuleEvalEntityInfo` helper function (previously scoped to `handlers_profile.go`) was moved to `handlers_evalstatus.go`. The mapping logic in `buildRuleEvaluationStatusFromDBEvaluation` was then updated to utilize this shared helper, ensuring consistent population of `entityInfo` attributes (e.g., `artifact_name`, `artifact_type`, etc.) across all supported entity types without duplicating code. Unused imports in `handlers_profile.go` were also cleaned up.

Fixes #3116

## Testing

- Added `TestListEvaluationResultsEntityInfo` to specifically verify that `entityInfo` fields are correctly populated when formatting the evaluation results for both `Artifact` and `PullRequest` entities.
- Ran all existing test suites in the `internal/controlplane` package (`go test -v ./internal/controlplane/...`) to ensure no regressions were introduced. All tests passed successfully.


krrish175-byte:
@evankanderson PTAL

internal/controlplane/handlers_evalstatus.go
```diff
@@ -764,6 +749,49 @@ func buildEvalResultAlertFromLRERow(
 	return era
 }
 
+func getRuleEvalEntityInfo(
```
evankanderson, line 714: This looks like a straight move from `handlers_profile.go` (fine, but this adds a "type 2 - refactoring" change to a "type 3 - behavior" change, which makes it a little bit harder to review).

internal/controlplane/handlers_evalstatus_test.go
```diff
@@ -844,3 +846,128 @@ func TestListEvaluationHistoryIncludeOutputs(t *testing.T) {
...
+
+			mockStore := mockdb.NewMockStore(ctrl)
+			mockProps := mockpropssvc.NewMockPropertiesService(ctrl)
+
+			minderEntityType := minderv1.Entity_ENTITY_REPOSITORIES
+			if tt.entityType == db.EntitiesArtifact {
+				minderEntityType = minderv1.Entity_ENTITY_ARTIFACTS
+			} else if tt.entityType == db.EntitiesPullRequest {
+				minderEntityType = minderv1.Entity_ENTITY_PULL_REQUESTS
+			}
```
evankanderson: Lint wants you to [use a tagged switch here](https://staticcheck.dev/docs/checks#QF1003).  If you disagree, you'll need to add [an appropriately scoped nolint directive](https://golangci-lint.run/docs/linters/false-positives/#nolint-directive), e.g. `//nolint:staticcheck  // tagged switch is ...`


```suggestion
			minderEntityType := minderv1.Entity_ENTITY_REPOSITORIES
			switch tt.EntityType {
			case db.EntitiesArtifact:
				minderEntityType = minderv1.Entity_ENTITY_ARTIFACTS
			case db.EntitiesPullRequest:
				minderEntityType = minderv1.Entity_ENTITY_PULL_REQUESTS
			}
```

evankanderson (changes requested):
This needs to appease the linter; otherwise, it looks good.

evankanderson:
(This is failing lint, but would otherwise be nice to have merged)

internal/controlplane/handlers_profile.go
```diff
@@ -248,49 +246,6 @@ func getProfilePBFromDB(
 
 // TODO: We need to replace this with a more generic method that can be used for all entities
 // probably coming from the properties.
```
evankanderson: I think these comments should move with the method?

internal/controlplane/handlers_evalstatus.go
evankanderson: You didn't copy the comment above the method when you moved this from the file that currently uses it once to this file which also uses it once.  😁

evankanderson (commented):
One comment that needs fixing before this can be merged.

krrish175-byte:
@evankanderson PTAL

evankanderson (approved):

PR #3356, opened 2024-05-17
Switch to AES-256-GCM encryption for secrets

dmjb:
[Ignore the name of the branch, this branch started out as something else originally]

Fixes #3315

This PR should not be merged until the changes in #3351 have been rolled out into production.

Use GCM encryption as the default encryption algorithm for Minder. At this
point, we now have a hard requirement to use the new EncryptedData structure in
the database - the old columns (for the access token, and for the redirect URL)
do not track the algorithm used and therefore they are not useful letting us know
which algorithm to use during decryption. As part of the migration away from the
old columns, I have made the following changes in this PR:

1. Make the old columns nullable
2. Stop writing to the old columns
3. Change all unit tests which relied on the old columns

Tested locally with a mix of new and old providers.

# Summary

***Provide a brief overview of the changes and the issue being addressed. 
Explain the rationale and any background necessary for understanding the changes. 
List dependencies required by this change, if any.***

Fixes #(related issue)

## Change Type

***Mark the type of change your PR introduces:***

- [ ] Bug fix (resolves an issue without affecting existing features)
- [x] Feature (adds new functionality without breaking changes)
- [ ] Breaking change (may impact existing functionalities or require documentation updates)
- [ ] Documentation (updates or additions to documentation)
- [ ] Refactoring or test improvements (no bug fixes or new functionality)

# Testing

***Outline how the changes were tested, including steps to reproduce and any relevant configurations. 
Attach screenshots if helpful.***

# Review Checklist:

- [x] Reviewed my own code for quality and clarity.
- [x] Added comments to complex or tricky code sections.
- [ ] Updated any affected documentation.
- [x] Included tests that validate the fix or feature.
- [ ] Checked that related changes are merged.

PR #6363, opened 2026-04-14
refactor(crypto): implement dynamic salt and support multiple keys/algorithms 

DharunMR:
# Summary

This PR completes a major architectural upgrade to the cryptography engine by supporting multiple keys/algorithms and generating dynamic salts per secret. During the implementation of these features, several critical bugs and redundancies in the AES-GCM and AES-CFB implementations were identified and fixed

- **Dynamic Routing:** EncryptedData now stores Algorithm and KeyVersion metadata. engine.go reads this metadata during decryption to dynamically route the payload to the correct algorithm (GCM vs CFB) and fetch the specific key version from the keystore.

- **Seamless Rotation**: This allows administrators to rotate default keys in the configuration; new secrets will use the new key, while old secrets will still decrypt perfectly using the stored KeyVersion.

- **Engin Driven Entropy:** engine.go now securely generates a fresh 16-byte random salt (rand.Read) for every encryption request and stores it securely in the DB record.

- **CFB Upgraded:** aes256cfb.go now accepts this dynamic salt for Argon2 key derivation, drastically improving cryptographic security against pre-computed attacks compared to the old hardcoded legacySalt.

- **Resolved "Double-Storage":** Previously, GCM prepended the 12-byte nonce to the ciphertext, while engine.go was also storing that exact same salt in the database. Encrypt now passes nil to gcm.Seal for the destination array, removing the redundant prepended nonce and reducing the database footprint per secret.

- **Safely Handled Keystore Newlines:** Legacy CFB local keystores contain trailing newlines (\n). Passing these raw strings to Go's strict Base64 decoder caused GCM to panic. decodeKey now utilizes strings.TrimSpace(string(key)) before Base64 decoding, preventing crashes while maintaining absolute backward compatibility for CFB.

### DIFF
--- a/internal/crypto/algorithms/aes256cfb.go
+++ b/internal/crypto/algorithms/aes256cfb.go
@@ -13,19 +13,21 @@ import (
 	"golang.org/x/crypto/argon2"
 )
 
+const aes256KeySize = 32
+
 // AES256CFBAlgorithm implements the AES-256-CFB algorithm
 type AES256CFBAlgorithm struct{}
 
-// Our current implementation of AES-256-CFB uses a fixed salt.
-// Since we are planning to move to AES-256-GCM, leave this hardcoded here.
+// legacySalt is used ONLY for decrypting old secrets that were encrypted before
+// the dynamic salt implementation.
 var legacySalt = []byte("somesalt")
 
 // Encrypt encrypts a row of data.
-func (a *AES256CFBAlgorithm) Encrypt(plaintext []byte, key []byte) ([]byte, error) {
+func (a *AES256CFBAlgorithm) Encrypt(plaintext []byte, key []byte, salt []byte) ([]byte, error) {
 	if len(plaintext) > maxPlaintextSize {
 		return nil, ErrExceedsMaxSize
 	}
-	block, err := aes.NewCipher(a.deriveKey(key))
+	block, err := aes.NewCipher(a.deriveKey(key, salt))
 	if err != nil {
 		return nil, fmt.Errorf("failed to create cipher: %w", err)
 	}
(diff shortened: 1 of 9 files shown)


evankanderson (commented):
This looks like it might destroy all current Minder data which is using the GCM algorithm, which is the preferred algorithm (CFB was deprecated by #3356, which retired the algorithm in favor of an AEAD algorithm).  The current encoding scheme for GCM was chosen to avoid conflicts with existing database data, so it would be good to review #3356 for context in this PR.

It may be reasonable to entirely remove the CFB algorithm at this point, if we can determine that new default installations do not use it.  Additionally, I _believe_ the above-mentioned PR implemented key rotation support, so that should not be new in this PR.

Given that cryptography is fairly subtle, I'm hoping that you've hand-reviewed both the PR contents and the PR description to ensure accuracy and cryptography best practices?

DharunMR:
> This looks like it might destroy all current Minder data which is using the GCM algorithm, which is the preferred algorithm (CFB was deprecated by #3356, which retired the algorithm in favor of an AEAD algorithm). The current encoding scheme for GCM was chosen to avoid conflicts with existing database data, so it would be good to review #3356 for context in this PR.
> 
> It may be reasonable to entirely remove the CFB algorithm at this point, if we can determine that new default installations do not use it. Additionally, I _believe_ the above-mentioned PR implemented key rotation support, so that should not be new in this PR.
> 
> Given that cryptography is fairly subtle, I'm hoping that you've hand-reviewed both the PR contents and the PR description to ensure accuracy and cryptography best practices?

thanks for taking your time to review . #3356 i've seen this pr and got to understand the few things ,even if i didnt had enough understanging in this area i just thought of give it a try but I see now why this approach is unsafe, i think closing this pr right now is the better option 😅

PR #6364, opened 2026-04-14
feat(cli): make quickstart catalog repository configurable via --catalog-repo flag

sachin9058:
# Summary

Add a `--catalog-repo` flag to the `quickstart` command to allow users to specify a custom repository for loading rule and profile catalogs.

Previously, the catalog repository URL was hardcoded, limiting flexibility to the default Minder catalog. This change enables users to provide their own repository while preserving the existing default behavior.

If the flag is not provided, the quickstart flow continues to use the default Minder catalog repository.

Additionally, the help text has been improved to better describe the purpose and usage of the flag.

# Testing

- Ran:
  ```bash
  make lint-fix
  go test ./...

- Verified CLI flag parsing by running:
  go run ./cmd/cli quickstart --catalog-repo <repo-url>
- Confirmed default behavior remains unchanged when no flag is provided
- Confirmed custom repository value is correctly read and passed into the catalog loading flow
- Verified fallback logic remains intact when repository loading fails


evankanderson:
> @evankanderson  This PR depends on the in-memory repository clone utility introduced in #6357.
> 
> The `quickstart` changes use the `internal/repo` package from that PR. Once it is merged, CI/build issues related to missing imports will be resolved.

The way to do this is to `git rebase` this branch onto that one. 

Thinking about the amount of code and a desire to test it, we should probably put the clone and find directories code in `internal/cli`.

sachin9058:
@evankanderson

> > @evankanderson  This PR depends on the in-memory repository clone utility introduced in #6357.
> > The `quickstart` changes use the `internal/repo` package from that PR. Once it is merged, CI/build issues related to missing imports will be resolved.
> 
> The way to do this is to `git rebase` this branch onto that one.
> 
> Thinking about the amount of code and a desire to test it, we should probably put the clone and find directories code in `internal/cli`.

Got it, that makes sense thanks!

I’ve moved away from depending on #6357 and implemented the clone logic inline for now so the PR is self-contained.

That said, I agree it would be better to structure this under `internal/cli` for clarity and testability. I’ll refactor the current implementation accordingly.

Let me update the PR with that approach 👍

evankanderson (commented):
Rather than just converting the existing rule loading, we should probably do something more extensive where we load a profile, ruletypes, and supporting datasources from one of the README enabled directories in minder-rules-and-profiles?

That way, we can include the description of what the profile will do in the guided setup.

sachin9058:
@evankanderson That makes sense, I like that direction 👍

Right now I focused on keeping the existing quickstart behavior but sourcing the rule type and profile from the repo.

Extending this to support README-based directories with profiles, rule types, and datasources would definitely make the flow more flexible and user-friendly.

I can open a follow-up PR to explore that approach once this one is in, if that sounds good.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -249,21 +250,43 @@ func quickstartCommand(
...
 
-	return loadCatalog(cmd, ruleClient, profileClient, provider, project, registeredRepos)
+	return loadCatalog(cmd, ruleClient, profileClient, provider, project, registeredRepos, repoURL)
 }
 
-// loadCatalog loads and applies the quickstart rule type and profile catalog
-// using the same prompts and flow as the inline implementation.
+const (
+	quickstartRuleTypeFilePath      = "rule-types/github/secret_scanning.yaml"
+	quickstartProfileFilePath       = "profiles/github/profile.yaml"
```
evankanderson: This profile contains 30 ruletypes, but it looks like we only load one ruletype (secret_scanning).  I think this will fail.  You should be able to test this by creating a new project and then running this quickstart against that project.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -249,21 +250,43 @@ func quickstartCommand(
...
+	defaultQuickstartCatalogRepoURL = "https://github.com/mindersec/minder-rules-and-profiles"
+)
+
+// loadCatalog drives the catalog portion of the quickstart flow.
+//
+// It first asks the user to confirm creation of the initial quickstart
+// resources (the secret_scanning rule type and its profile), then attempts
+// to source those resources from the configured catalog repository using
+// loadCatalogFromRepo.
```
evankanderson: It seems like this should go the other way -- if we load the repo first, we can validate that at least the rule type and profile paths exist if we get the repo first.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -249,21 +250,43 @@ func quickstartCommand(
...
+	repoURL, err := cmd.Flags().GetString("catalog-repo")
+	if err != nil {
+		return err
+	}
+	if repoURL == "" {
+		repoURL = defaultQuickstartCatalogRepoURL
+	}
 
-	return loadCatalog(cmd, ruleClient, profileClient, provider, project, registeredRepos)
+	return loadCatalog(cmd, ruleClient, profileClient, provider, project, registeredRepos, repoURL)
```
evankanderson: It seems like it might simplify things to pass a `go-git/go-billy` filesystem here, rather than the repo URL.  That would simplify testing the catalog parts, which we should probably put into `internal/cli` and test, alongside the library functions that power `minder apply` et al.

internal/cli/catalog.go
```diff
@@ -0,0 +1,33 @@
...
+		return nil, fmt.Errorf("failed to clone repository %s: %w", repoURL, err)
+	}
+
+	worktree, err := repo.Worktree()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get worktree for %s: %w", repoURL, err)
+	}
+
+	return worktree.Filesystem, nil
+}
```
evankanderson: This function seems fairly trivial.  Unless there's more functionality or we plan to use it in multiple places, I'd be inclined to inline this for now.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/mindersec/minder/cmd/cli/app/profile"
 	minderprov "github.com/mindersec/minder/cmd/cli/app/provider"
 	"github.com/mindersec/minder/cmd/cli/app/repo"
+	cliintern "github.com/mindersec/minder/internal/cli"
```
evankanderson: We already have an `internal/util/cli`, why add this directory?

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,17 +298,156 @@ func loadCatalog(
...
+	if !yes {
+		return nil
+	}
 
-	// Load the rule type from the embedded file system
+	cmd.Printf("Creating profile from remote catalog...\n")
+	profileReader, err := fs.Open(quickstartProfileFilePath)
+	if err != nil {
+		return fmt.Errorf("failed to open profile file: %w", err)
+	}
```
evankanderson: It feels like it would be helpful to load all the files (ruletypes, profiles, etc) at once before starting the prompt sequence -- otherwise, it's easy to have the quickstart fail halfway through with a bad path, and users need to figure out how (whether) to clean things up before trying again.  Given that users have the _least_ mental model at this point of how things work, validating as much as possible up-front and making the rest as transactional as possible feels like the right approach.

evankanderson: Something in `internal/util/cli` or `pkg/fileconvert` which took a `fs` and loaded all the resources it could from a vararg list of paths might be a nice utility.  I suspect you could extend `ResourcesFromPaths` to take a filesystem (either the go-builtin or the go-git/go-billy version) fairly easily.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,17 +298,156 @@ func loadCatalog(
...
+	defer profileReader.Close()
+
+	p, err := profiles.ParseYAML(profileReader)
+	if err != nil {
+		return fmt.Errorf("failed to parse profile: %w", err)
+	}
+
+	p.Context = &minderv1.Context{
+		Provider: &provider,
```
evankanderson: You probably don't need provider, since profiles and ruletypes don't rely on providers.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,17 +298,156 @@ func loadCatalog(
...
+		}
+	}
+
+	if alreadyExists {
+		cmd.Println(cli.WarningBanner.Render(stepPromptMsgFinishExisting + stepPromptMsgFinishBase))
+	} else {
+		cmd.Println(cli.WarningBanner.Render(stepPromptMsgFinishOK + stepPromptMsgFinishBase))
+		cmd.Println("Profile details (minder profile list):")
+		table := profile.NewProfileRulesTable(cmd.OutOrStdout())
+		profile.RenderProfileRulesTable(resp.GetProfile(), table)
```
evankanderson: It's weird to print these only if they didn't already exist, but not to print a summary if they did exist.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,17 +298,156 @@ func loadCatalog(
...
+//
+// This function is used as a safe fallback when loading the catalog from
+// the configured repository fails. It recreates the previous Step 3 and
+// Step 4 behavior by:
+//   - reading secret_scanning.yaml and profile.yaml from the embedded FS,
+//   - parsing them into the appropriate rule type and profile structures,
+//   - applying the current provider/project context, and
+//   - creating the resources via the corresponding gRPC services, including
+//     handling AlreadyExists responses and printing the final banners and
+//     profile details table.
```
evankanderson: Since we know the existing flow doesn't work well, let's not try to preserve it.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,17 +298,156 @@ func loadCatalog(
 		return nil
 	}
 
-	// Creating the rule type
-	cmd.Println("Creating rule type...")
+	if err := loadCatalogFromRepo(cmd, ruleClient, profileClient, provider, project, registeredRepos, repoURL); err != nil {
+		cmd.Printf("Warning: failed to load quickstart catalog from %s: %v\n", repoURL, err)
+		cmd.Printf("Falling back to embedded quickstart catalog.\n")
+		return runExistingFlow(cmd, ruleClient, profileClient, provider, project, registeredRepos)
```
evankanderson: This will happen if the repo clone works and the ruletype load works, but then there's a problem with the profile load, right?

That feels like a weird outcome, especially since we don't clean up the ruletype.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -299,14 +461,10 @@ func loadCatalog(
 		Project:  &project,
 	}
 
-	// New context so we don't time out between steps
```
evankanderson, line 302: Why remove these comments?

evankanderson (commented):
I think there's a bit more refactoring we could do to lean towards loading files from a filesystem (if we package it right, we could even use https://pkg.go.dev/embed#hdr-File_Systems to supply a backup set of rules that don't need to clone).

sachin9058:
@evankanderson Thanks for the detailed feedback — I’ve reworked the implementation accordingly.

* Switched to filesystem-driven loading for rule types and profiles from the catalog repository
* Added validation to ensure profiles only reference existing rule types
* Removed the previous hardcoded/single-file approach
* Structured the flow as load → validate → prompt → apply
* Made the apply step transactional with rollback on failure

I also resolved the merge conflicts and cleaned up the commit history to keep this PR focused on the quickstart changes.

Please let me know if you’d like me to refine anything further or adjust the approach.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+	if err := validateCatalog(ruleTypeNames, loadedProfiles); err != nil {
+		return nil, err
 	}
 
-	if rt.Context == nil {
-		rt.Context = &minderv1.Context{}
+	return &Catalog{RuleTypes: ruleTypes, Profiles: loadedProfiles}, nil
+}
+
+func loadRuleTypesFromFS(fs billy.Filesystem) ([]*minderv1.RuleType, map[string]struct{}, error) {
```
evankanderson: Rather than generating parallel array and map-index-to-array, what about storing a map from name --> *RuleType?

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+			}
+			if profile.GetName() == "" {
+				return fmt.Errorf("profile file %s has no name", path)
+			}
+			if _, exists := profileNames[profile.GetName()]; exists {
+				return fmt.Errorf("duplicate profile %q found in %s", profile.GetName(), path)
+			}
+			profileNames[profile.GetName()] = struct{}{}
+			loadedProfiles = append(loadedProfiles, profile)
+			return nil
```
evankanderson: `fileconvert` from `pkg/profiles/util` almost wraps all this up (and supports JSON input):

```golang
	decoder, closer := fileconvert.DecoderForFile(fpath) // needs to pass in an fs for `os.Open`
	if decoder == nil {
		return nil, fmt.Errorf("error opening file")
	}
	defer closer.Close()
	profile, err := fileconvert.ReadResourceTyped[*pb.Profile](decoder)
}
```

`pkg/profiles` also has a `ReadProfileFromPath` that wraps up that function entirely, though there's no current equivalent for RuleTypes (the `fileconvert` methods will work for either, and `ReadProfileFromPath` could be reworked

evankanderson: Adding `ReadTyped[T proto.Message](opener func(str) (io.Reader, err), path) (T, error)` to `pkg/fileconvert` seems reasonable, and might reduce the amount of code needed in these loops.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+				return fmt.Errorf("failed to open profile file %s: %w", path, err)
 			}
-			cmd.Println("Rule type secret_scanning already exists")
-		} else {
-			return cli.MessageAndError("error creating rule type", err)
+			defer reader.Close()
+
+			profile, err := profiles.ParseYAML(reader)
+			if err != nil {
+				return fmt.Errorf("failed to parse profile file %s: %w", path, err)
```
evankanderson: I think _any_ error in _any_ profile or ruletype will cause `loadCatalog` to fail completely?  Given the desired use case, it seems like skipping errors might be more appropriate.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+	for _, loadedProfile := range loadedProfiles {
+		referencedRuleTypes := make(map[string]struct{})
+		if err := profiles.TraverseRuleTypesForEntities(loadedProfile, func(_ minderv1.Entity, rule *minderv1.Profile_Rule) error {
+			if rule.GetType() != "" {
+				referencedRuleTypes[rule.GetType()] = struct{}{}
+			}
+			return nil
+		}); err != nil {
+			return fmt.Errorf("failed to inspect profile %q: %w", loadedProfile.GetName(), err)
+		}
```
evankanderson: If your function never returns an error, you can ignore the error return from `TraverseRuleTypesForEntities` by assigning to `_`.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
 	}
 
-	// Creating the profile
-	cmd.Println("Creating profile...")
-	reader, err = content.Open("embed/profile.yaml")
+	return nil
+}
+
+func collectYAMLFiles(fs billy.Filesystem, root string) ([]string, error) {
+	entries, err := fs.ReadDir(root)
```
evankanderson: You might want to use `billyutil.Walk(fs, root, ...)` here rather than implementing your own recursive function.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+	seenExistingProfile bool
+}
+
+func rollbackCatalogResources(
+	ctx context.Context,
+	ruleClient minderv1.RuleTypeServiceClient,
+	profileClient minderv1.ProfileServiceClient,
+	result applyCatalogResult,
+) {
+	for i := len(result.createdProfiles) - 1; i >= 0; i-- {
```
evankanderson: Why not use a standard for-each loop here?
```suggestion
	for _, p := range result.createdProfiles {
```

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+		resp, err := ruleClient.CreateRuleType(ctx, &minderv1.CreateRuleTypeRequest{RuleType: ruleType})
+		if err != nil {
+			if st, ok := status.FromError(err); ok && st.Code() == codes.AlreadyExists {
+				cmd.Printf("Rule type %s already exists\n", ruleType.GetName())
+				continue
+			}
+			if st, ok := status.FromError(err); ok {
+				return fmt.Errorf("error creating rule type %s: %s", ruleType.GetName(), st.Message())
+			}
+			return cli.MessageAndError(fmt.Sprintf("error creating rule type %s", ruleType.GetName()), err)
```
evankanderson: Why these two different types of error returns?

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+		}
+
+		createdName := ruleType.GetName()
+		if resp != nil && resp.GetRuleType() != nil && resp.GetRuleType().GetName() != "" {
+			createdName = resp.GetRuleType().GetName()
+		}
```
evankanderson: Two things:

1. We should never return a RuleType that doesn't have a name, so we don't need the default.
2. Proto accessors handle the `nil` message already, so you can just write:

```suggestion
		createdName := resp.GetRuleType().GetName()
```

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+				continue
+			}
+			if st, ok := status.FromError(err); ok {
+				return fmt.Errorf("error creating profile %s: %s", profileResource.GetName(), st.Message())
+			}
+			return cli.MessageAndError(fmt.Sprintf("error creating profile %s", profileResource.GetName()), err)
+		}
+
+		if resp != nil && resp.GetProfile() != nil {
+			result.summaryProfile = resp.GetProfile()
```
evankanderson: I don't understand what we're doing here with `summaryProfile`.  It seems like we're automatically installing a bunch of profiles, and then somehow trying to decide which was most important?  Furthermore, we're overwriting `summaryProfile` on each run through the loop, so it seems like we'll only select the last one in the slice.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+			if result.summaryProfile.GetId() != "" {
+				result.createdProfiles = append(result.createdProfiles, result.summaryProfile.GetId())
```
evankanderson: Why do we sometimes use the id here, but always use the name below?

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
+			} else {
+				result.createdProfiles = append(result.createdProfiles, profileResource.GetName())
 			}
-			alreadyExists = true
 		} else {
-			return cli.MessageAndError("error creating profile", err)
+			result.createdProfiles = append(result.createdProfiles, profileResource.GetName())
```
evankanderson: How would `resp` be `nil` without `err` being set?

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +287,341 @@ func loadCatalog(
...
 
-	return nil
+	if result.summaryProfile == nil {
+		return
+	}
+
+	cmd.Println("Profile details (minder profile list):")
+	table := profile.NewProfileRulesTable(cmd.OutOrStdout())
+	profile.RenderProfileRulesTable(result.summaryProfile, table)
+	table.Render()
```
evankanderson: If we create three profiles (or seven overlapping ones), it seems like this will only print out one profile at the end?

Given the organization of the `profiles` and `rule-types` directories in `minder-rules-and-profiles`, it seems like this might lead to confusing results.

evankanderson (commented):
I'm not sure exactly what behavior we're expecting here -- it seems like the current code will try to load _all_ the generic ruletypes and profiles from `minder-rules-and-profiles` into one project, which seems like it might be overload for a new user.

Additionally, if this work is laying the foundation for future refinement, I'd expect it to be adjusting / improving existing code (e.g. leveraging the same abstractions as the `apply` and `create` CLI commands), rather than implementing a lot of heavy new logic in `cmd/`.  (Quickstart already has a bunch of heavy prompting logic, but it would be great to reduce the amount of logic in `cmd/` (which doesn't contribute to coverage numbers), and increase the logic in `internal/util/cli` and `pkg/`.

cmd/cli/app/quickstart/quickstart.go
sachin9058: Updated this to reuse existing utilities:

* `profiles.ReadProfileFromPath` for profile parsing
* `fileconvert` for rule type decoding
* `billyutil.Walk` for filesystem traversal

This removes custom parsing logic and aligns with existing patterns in the codebase.

cmd/cli/app/quickstart/quickstart.go
sachin9058: Removed the previous summaryProfile logic to avoid ambiguity when multiple profiles are created. The flow now focuses on creating resources and reporting overall status.

cmd/cli/app/quickstart/quickstart.go
sachin9058: That makes sense — currently the implementation loads all valid rule types and profiles from the catalog to keep the flow consistent with the existing quickstart behavior.

I agree this could be improved to be more selective (e.g., README-driven or guided selection), and I’m happy to explore that in a follow-up PR once this foundation is in place.

cmd/cli/app/quickstart/quickstart.go
sachin9058: Standardized error handling to use `fmt.Errorf(... %w ...)` for consistency across the flow and removed mixed error patterns.

cmd/cli/app/quickstart/quickstart.go
sachin9058: Updated the loop to use a standard `for _, v := range` iteration for clarity and consistency.

cmd/cli/app/quickstart/quickstart.go
sachin9058: Simplified the handling here by relying on proto accessors and removing redundant checks. Added a minimal fallback only where needed for safety.

cmd/cli/app/quickstart/quickstart.go
sachin9058: Made profile handling consistent by using the profile ID returned from the API for tracking created resources.

cmd/cli/app/quickstart/quickstart.go
sachin9058: Good point — switched to using a `map[string]*RuleType` for direct lookup by name and converted to a slice only where needed. This simplifies the logic and avoids maintaining parallel structures.

cmd/cli/app/quickstart/quickstart.go
sachin9058: Updated the loading flow to skip invalid rule types and profiles with warnings instead of failing the entire catalog load. This makes the behavior more robust for mixed or partially invalid inputs.

cmd/cli/app/quickstart/quickstart.go
sachin9058: Updated this to reuse existing utilities:

* `profiles.ReadProfileFromPath` for profile parsing
* `fileconvert` for rule type decoding
* `billyutil.Walk` for filesystem traversal

This removes custom parsing logic and aligns with existing patterns in the codebase.

sachin9058:
@evankanderson Thanks for the detailed feedback — I’ve made a pass to align the implementation accordingly.

* Moved catalog loading and validation into `internal/cli`, keeping `cmd/quickstart` focused on orchestration
* Reused existing utilities (`profiles.ReadProfileFromPath`, `fileconvert`, `billyutil.Walk`) instead of custom logic
* Simplified data structures and loops, and cleaned up error handling
* Updated behavior to skip invalid resources with warnings rather than failing the entire load
* Removed unnecessary checks and made resource handling consistent

I’ve kept the current loading behavior as-is for now and can iterate on making it more selective in a follow-up PR if that direction makes sense.

Let me know if there’s anything else you’d like refined.

pkg/profiles/util.go
```diff
@@ -82,6 +83,17 @@ func ReadProfileFromFile(fpath string) (*pb.Profile, error) {
 	return fileconvert.ReadResourceTyped[*pb.Profile](decoder)
 }
 
+// ReadProfileFromPath reads a pipeline profile from a billy filesystem path.
+func ReadProfileFromPath(vfs billy.Filesystem, path string) (*pb.Profile, error) {
```
evankanderson: What about changing `ReadProfileFromFile` to take a `vfs billy.Filesystem`, and writing it like so:

```golang
// ReadProfileFromFile reads a pipeline profile from a file and returns it as a protobuf
func ReadProfileFromFile(fs billy.Filesystem, fpath string) (*pb.Profile, error) {
	if fs == nil {
		fs = osfs.Default
	}
	decoder, closer := fileconvert.DecoderForFile(fs, fpath)
	if decoder == nil {
		return nil, fmt.Errorf("error opening file")
	}
	defer closer.Close()
	return fileconvert.ReadResourceTyped[*pb.Profile](decoder)
}
```

You can then simply change existing callers to pass `nil` as the first argument, and it starts to lay the groundwork for possibly being able to use Git sources for all the `apply` and `create` commands.

internal/cli/catalog.go
```diff
@@ -0,0 +1,251 @@
...
+
+// Catalog represents a loaded collection of rule types and profiles from a filesystem.
+type Catalog struct {
+	RuleTypes []*minderv1.RuleType
+	Profiles  []*minderv1.Profile
+}
+
+// LoadCatalogFromFS loads and validates all resources under the catalog directories.
+// Invalid resources are skipped and reported through warnf.
+func LoadCatalogFromFS(vfs billy.Filesystem, warnf func(string, ...any)) (*Catalog, error) {
```
evankanderson: This feels a lot like `fileconvert.ResourcesFromPaths`, except with a `vfs` first argument and the `Printer` type not explicitly declared.  (Declaring the type of the `warnf` explicitly provides some better documentation about what the arguments in the function mean.)

internal/cli/catalog.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	if len(loadedProfiles) == 0 {
+		return nil, fmt.Errorf("no valid profiles found under %s", catalogProfilesDir)
+	}
+
+	return loadedProfiles, nil
+}
+
+func validateCatalog(
```
evankanderson: Why is this not a method on `Catalog`?

internal/cli/catalog.go
```diff
@@ -0,0 +1,251 @@
...
+	tmpName := tmpFile.Name()
+	defer os.Remove(tmpName)
+
+	if _, err := io.Copy(tmpFile, reader); err != nil {
+		_ = tmpFile.Close()
+		return nil, fmt.Errorf("failed to copy rule type file %s: %w", path, err)
+	}
+	if err := tmpFile.Close(); err != nil {
+		return nil, fmt.Errorf("failed to close temp file for %s: %w", path, err)
+	}
```
evankanderson: Why do we do all these filesystem operations?  Is it because `DecoderForFile` doesn't currently take an `fs` argument?

pkg/profiles/util.go
evankanderson: Even better, I bet that `ReadResourceFromFile[T proto.Message](fs billy.Filesystem, fpath string) (T, error)` could be generic, like `ReadResourceTyped`, and live in the `fileconvert` package.

pkg/profiles/util.go
```diff
@@ -82,6 +83,17 @@ func ReadProfileFromFile(fpath string) (*pb.Profile, error) {
...
+	reader, err := vfs.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("error opening profile path %s: %w", path, err)
+	}
+	defer reader.Close()
+
+	return ParseYAML(reader)
```
evankanderson: There's a TODO above to remove usages of `ParseYAML` -- let's not add to them.

evankanderson (commented):
This is looking better, but I think with a few updates to `fileconvert`, this could be even simpler.

sachin9058:
@evankanderson  Thanks  I’ve updated the implementation to better align with the existing abstractions.

* Replaced custom file traversal and parsing with a reusable `ResourcesFromFilesystem` helper in `fileconvert`
* Reduced duplication in catalog loading by reusing that helper for both rule types and profiles
* Moved validation logic into a method on `Catalog`
* Kept CLI logic focused on orchestration

Let me know if you’d prefer further alignment with existing helpers like `ResourcesFromPaths`, or if this direction looks good.

pkg/fileconvert/encodedecode.go
```diff
@@ -32,26 +33,48 @@ type Decoder interface {
...
+func ReadResourceFromFile[T proto.Message](fs billy.Filesystem, path string) (T, error) {
+	var zero T
+
+	file, err := fs.Open(path)
+	if err != nil {
+		return zero, fmt.Errorf("error opening file %s: %w", path, err)
+	}
+	defer file.Close()
+
+	decoder, decoderErr := decoderForReader(path, file)
```
evankanderson: Why change `DecoderForFile` to move the open outside the function, rather than accepting a possibly-`nil` filesystem argument?

It's okay to break existing API compatibility in this library, or to add one method as a shell of the other:

```golang
func decoderForPath(fs billy.Filesystem, path string) (Decoder, io.Closer) {
	fs = cmp.Or(fs, osfs.Default)
	path = filepath.Clean(path)
	var builder func(io.Reader) Decoder
	// we use functions here so that we can early-exit without opening the file if the extension is unmatched
	switch filepath.Ext(path) {
	// ...
	}
	file, err := fs.Open(path)  // (aside for sachin: this path has been cleaned whereas the line 66 has not)
	if err != nil {
		return nil, nil
	}
	return decoder(file), file
}

func DecoderForFile(path string) (Decoder, io.Closer) {
	return decoderForPath(nil, string)
}
```

pkg/fileconvert/encodedecode.go
```diff
@@ -32,26 +33,48 @@ type Decoder interface {
...
+	if decoderErr != nil {
+		_ = file.Close()
+		return nil, nil
+	}
+	return decoder, file
+}
+
+// ReadResourceFromFile reads a single resource from the specified filesystem path.
+func ReadResourceFromFile[T proto.Message](fs billy.Filesystem, path string) (T, error) {
+	var zero T
```
evankanderson: Where is this variable used?

pkg/fileconvert/fileconvert_test.go
```diff
@@ -193,6 +194,27 @@ func TestReadResourceTyped(t *testing.T) {
...
+	require.NoError(t, err)
+	_, err = file.Write([]byte(`type: profile
+version: v1
+name: test-profile
+repository:
+  - type: sample-rule
+    def: {}
+`))
+	require.NoError(t, err)
+	require.NoError(t, file.Close())
```
evankanderson: Prefer `util.WriteFile` from go-billy/v5/util over adding your own.

pkg/fileconvert/fileconvert_test.go
```diff
@@ -193,6 +194,27 @@ func TestReadResourceTyped(t *testing.T) {
 	require.NoError(t, err, "Expected no error reading profile")
 }
 
+func TestReadResourceFromFile(t *testing.T) {
```
evankanderson, line 198: Add a test for attempting to read from a non-existent file?

pkg/fileconvert/collect.go
```diff
@@ -20,6 +27,47 @@ import (
 // Printer provides an interface for passing a printf-like function.
 type Printer func(string, ...any)
 
+// ResourcesFromFilesystem collects resources from a directory in a billy filesystem.
```
evankanderson: Again, let's not repeat the code from `ResourcesFromPaths`, and instead add a `fs billy.Filesystem` argument where `nil` represents the base OS filesystem.  You'll also need to introduce an `fs` argument to `util.ExpandFileArgs`, but that seems reasonable.

evankanderson: We'd also need to decide what to do with the special filename "-", but I think that can stay as "stdin", and use "./-" to reference the file otherwise.

pkg/fileconvert/collect.go
evankanderson, line 73: This is an interesting bit of logic where `minder apply -f dir-of-stuff/ -f a-specific-file.yaml` will error if `a-specific-file.yaml` is not the right type, but not if `dir-of-stuff/one-random-file-of-many.yaml` is malformed.

I'm not sure it's _needed_ for the catalog bit, but I'm not sure it hurts.

I think it would also be fine to introduce the `Catalog` struct here, and change `ResourcesFromPaths` to return a `Catalog` by type, rather than an array that each caller needs to pick apart by type (which is what we do today).

internal/cli/catalog.go
```diff
@@ -0,0 +1,180 @@
...
+
+// WarnFunc reports skipped catalog resources.
+type WarnFunc = fileconvert.Printer
+
+// LoadCatalogFromFS loads and validates all resources under the catalog directories.
+// Invalid resources are skipped and reported through warnf.
+func LoadCatalogFromFS(vfs billy.Filesystem, warnf WarnFunc) (*Catalog, error) {
+	if warnf == nil {
+		warnf = func(string, ...any) {}
+	}
```
evankanderson, line 39: I think you can just assert that `WarnFunc` shouldn't be `nil`, and you get what's coming if you don't supply one.

internal/cli/catalog.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		ruleTypesByName[ruleType.GetName()] = ruleType
+	}
+
+	if len(ruleTypesByName) == 0 {
+		return nil, fmt.Errorf("no valid rule types found under %s", catalogRuleTypesDir)
+	}
+
+	return ruleTypesByName, nil
+}
```
evankanderson, line 150: loadRuleTypesFromFS and loadProfilesFromFS seem almost identical.  Can we collapse them into one method?

evankanderson: (I suspect we'll want a `loadDataSourcesFromFS` as well)

internal/cli/catalog.go
```diff
@@ -0,0 +1,180 @@
...
+		ruleTypeNames = append(ruleTypeNames, name)
+	}
+	sort.Strings(ruleTypeNames)
+
+	ruleTypes := make([]*minderv1.RuleType, 0, len(ruleTypeNames))
+	for _, name := range ruleTypeNames {
+		ruleTypes = append(ruleTypes, ruleTypesByName[name])
+	}
+
+	catalog := &Catalog{RuleTypes: ruleTypes, Profiles: loadedProfiles}
```
evankanderson, line 57: Why not store the maps in the `Catalog`, to simplify `Validate`?

internal/cli/catalog.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// Validate validates the catalog contents and keeps only profiles whose rule types exist.
+func (c *Catalog) Validate(warnf WarnFunc) error {
+	if c == nil {
+		return fmt.Errorf("catalog is nil")
+	}
+	if warnf == nil {
+		warnf = func(string, ...any) {}
+	}
```
evankanderson, line 73: I don't think you need either of these checks (IMO).

internal/cli/catalog.go
```diff
@@ -0,0 +1,180 @@
...
+		referencedRuleTypes := make(map[string]struct{})
+		if err := profiles.TraverseRuleTypesForEntities(loadedProfile, func(_ minderv1.Entity, rule *minderv1.Profile_Rule) error {
+			if rule.GetType() != "" {
+				referencedRuleTypes[rule.GetType()] = struct{}{}
+			}
+			return nil
+		}); err != nil {
+			warnf("Skipping invalid profile %q: failed to inspect rules: %v\n", loadedProfile.GetName(), err)
+			continue
+		}
```
evankanderson, line 93: If your traversal function never returns an error, you don't need to check for an error in `TraverseRuleTypesForEntities`.  Just assign it to `_` with a comment.

internal/cli/catalog.go
```diff
@@ -0,0 +1,180 @@
...
+
+		missingRuleTypes := make([]string, 0)
+		for ruleType := range referencedRuleTypes {
+			if _, exists := ruleTypesByName[ruleType]; !exists {
+				missingRuleTypes = append(missingRuleTypes, ruleType)
+			}
+		}
```
evankanderson: This can be written more simply as:


```suggestion
		maps.DeleteFunc(referencedRuleTypes, func(k string, v struct{}) bool {
			return ruleTypesByName != nil
		}
		// referencedRuleTypes now contains only ruletypes not in ruleTypesByName
```

internal/cli/catalog.go
```diff
@@ -0,0 +1,180 @@
...
+
+		if len(missingRuleTypes) > 0 {
+			sort.Strings(missingRuleTypes)
```
evankanderson: If you have a map here, you can get a sorted list of keys with `slices.Sorted(maps.Keys(x))`

internal/cli/catalog.go
```diff
@@ -0,0 +1,180 @@
...
+
+	catalog := &Catalog{RuleTypes: ruleTypes, Profiles: loadedProfiles}
+	if err := catalog.Validate(warnf); err != nil {
+		return nil, err
+	}
+
+	return catalog, nil
+}
+
+// Validate validates the catalog contents and keeps only profiles whose rule types exist.
```
evankanderson: I think you should be a little "louder" about the fact that Validate will _mutate_ Catalog to only retain profiles whose ruletype references are contained in the catalog.

Often, users expect a Validate method to be "const"-like.

cmd/cli/app/profile/common.go
evankanderson, line 65: Not for this PR, but it would be great to remove the duplication between `util.OpenFileArg` and `fileconvert.ResourcesFromPaths` around handling the `-` argument.

pkg/profiles/util.go
```diff
@@ -82,6 +83,11 @@ func ReadProfileFromFile(fpath string) (*pb.Profile, error) {
 	return fileconvert.ReadResourceTyped[*pb.Profile](decoder)
 }
 
+// ReadProfileFromPath reads a pipeline profile from a billy filesystem path.
+func ReadProfileFromPath(vfs billy.Filesystem, path string) (*pb.Profile, error) {
+	return fileconvert.ReadResourceFromFile[*pb.Profile](vfs, path)
+}
+
```
evankanderson: Is this method still needed?

What about ParseYAML on line 52 and parseJSON on 60?

evankanderson (commented):
Some additional suggested changes -- the overall goal is to _reduce_ the amount of code that needs to be maintained / fixed in the future while increasing its capabilities -- right now this is about +250 lines of non-test library code, but I think we can get it closer to +100 while supporting adding repository load for additional sub-commands in a subsequent PR.

sachin9058:
@evankandersonThanks again for the detailed feedback this was very helpful in shaping the final direction.

I’ve made the following updates:

* Moved filesystem expansion fully into `util.ExpandFileArgs` and removed duplicate traversal logic
* Simplified `collect.go` to rely on the unified expansion flow
* Removed the unused `ReadProfileFromPath` helper and related import
* Fixed decoder path handling to preserve previous behavior for OS-backed paths
* Verified all affected packages with tests and lint (`go test` and `make lint-fix` both pass)

Let me know if this looks aligned now or if you’d like any further adjustments.

sachin9058:
@evankanderson 

I’ve folded the test coverage into this PR so implementation and tests can be reviewed together. I also refactored the tests into a table-driven format to align with existing patterns and added behavior-focused assertions for profile validation.

This should also make it easier to evaluate coverage impact as part of this PR.

Let me know if this looks good or if you’d like any further refinements.

sachin9058:
Hi @evankanderson can u take a look on this again i have made the changes that aligns with your feedback tell me if it looks good or need more refinements or improvements ??

evankanderson:
Sorry -- I somehow missed that this had more updates.  I'm taking a look now.

internal/util/helpers.go
```diff
@@ -121,25 +124,57 @@ type ExpandedFile struct {
...
-				Expanded: false,
-			})
+			appendExpandedPath(f, false)
 			continue
 		}
 
 		f = filepath.Clean(f)
-		fi, err := os.Stat(f)
+
+		if vfs == nil {
```
evankanderson, line 145: Rather than creating two implementations within this function, try one of the following:

1. Use [`osfs.New()`](https://pkg.go.dev/github.com/go-git/go-billy/v5@v5.9.0/osfs#New) to create a billy filesystem rooted in the current directory:

   ```golang
   if vfs == nil {
      vfs = osfs.New(".", osfs.WithBoundOS())
   }
   ```
2. You can convert both types to a base library `io/fs FS` interface using [billy's `iofs.New()`](https://pkg.go.dev/github.com/go-git/go-billy/v5@v5.9.0/helper/iofs#New) or [os's DirFS()](https://pkg.go.dev/os#DirFS) and then calling [io/fs's `WalkDir()`](https://pkg.go.dev/io/fs#WalkDir)
   ```golang
   var fs fs.FS
   if vfs != nil {
     fs = iofs.New(vfs)
   } else {
     fs = os.DirFS(".")
   }
   ```
   This implementation will allow reading absolute symlinks outside the current directory, while the billy implementation will perform a "soft" chroot behavior where it will block reads from outside the directory.  I don't have a strong feeling between the two approaches.

In general, when you find that you need two parallel implementations that are more than about 5 lines long, it's generally worth searching for a Go interface you can use to unify the two code paths.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -20,13 +21,12 @@ import (
 
 	"github.com/mindersec/minder/cmd/cli/app"
 	"github.com/mindersec/minder/cmd/cli/app/auth"
-	"github.com/mindersec/minder/cmd/cli/app/profile"
 	minderprov "github.com/mindersec/minder/cmd/cli/app/provider"
 	"github.com/mindersec/minder/cmd/cli/app/repo"
+	internalcli "github.com/mindersec/minder/internal/cli"
```
evankanderson, line 26: Why not put this in `internal/util/cli`, which we import 2 lines later?  Having two different directories for CLI-implementation-related modules is confusing to future authors.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -128,9 +128,6 @@ In case you have registered new repositories during this flow, the profile will
 `
 )
 
-//go:embed embed*
-var content embed.FS
-
```
evankanderson, line 133: Can we remove the `embed` directory as well if we don't need this `//go:embed` variable?

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +283,147 @@ func loadCatalog(
...
-	if err := minderv1.ParseResource(reader, rt); err != nil {
-		return cli.MessageAndError("error parsing rule type", err)
+	// Step 4 - Confirm profile creation with context
+	yes = cli.PrintYesNoPrompt(cmd,
+		fmt.Sprintf(stepPromptMsgProfile, strings.Join(registeredRepos, "\n")),
+		"Proceed?",
+		"Quickstart operation cancelled.",
+		true)
+	if !yes {
+		return nil
```
evankanderson, line 293: It's a little odd to have two "yes/no" prompts right in a row when nothing has happened between them.  Can we coalesce them into one prompt?

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +283,147 @@ func loadCatalog(
...
+	return applyCatalog(cmd, ruleClient, profileClient, catalog, project)
+}
 
-	rt.Context = &minderv1.Context{
-		Provider: &provider,
-		Project:  &project,
-	}
+const (
+	defaultQuickstartCatalogRepoURL = "https://github.com/mindersec/minder-rules-and-profiles"
+)
```
evankanderson, line 302: Slight preference to have constants at the top of the file, or at least before use.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +283,147 @@ func loadCatalog(
...
+		"Quickstart operation cancelled.",
+		true)
+	if !yes {
+		return nil
 	}
 
-	if rt.Context == nil {
-		rt.Context = &minderv1.Context{}
-	}
+	// Apply all resources (transactional flow)
```
evankanderson, line 296: Include the transactional behavior in the method godoc, not at the call site.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +283,147 @@ func loadCatalog(
...
-	}
+const (
+	defaultQuickstartCatalogRepoURL = "https://github.com/mindersec/minder-rules-and-profiles"
+)
 
-	// New context so we don't time out between steps
+// applyCatalog creates all rule types and profiles from the catalog via gRPC services.
+//
+// This function is called AFTER user confirmation and validation. It creates all
+// resources in the catalog. If any creation fails, the error is returned immediately.
```
evankanderson, line 307: This says "If any creation fails, the error is returned immediately", but this actually seems to try to clean up in some cases.

I'm not sure whether cleanup or "handle existing resources" is a better approach, but we should line up this doc with the "transactional flow" comment above.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +283,147 @@ func loadCatalog(
...
+		createdProfiles:  make([]string, 0, len(catalog.Profiles)),
 	}
 
-	// Creating the profile
-	cmd.Println("Creating profile...")
-	reader, err = content.Open("embed/profile.yaml")
-	if err != nil {
-		return cli.MessageAndError("error opening profile", err)
+	if err := createCatalogRuleTypes(ctx, cmd, ruleClient, catalog.RuleTypes, projectContext, &result); err != nil {
+		rollbackCatalogResources(ctx, ruleClient, profileClient, result)
```
evankanderson, line 329: This makes sense given two cases; for larger flows, I'd use a `defer` with a named `err` return, like so:

```golang
func DoThing() (err error) {
  created := map[string]error
  defer func() {
    if error != nil {
       for k, _ := range created {
         cleanup(k)  // not using stored error
       }
     }
  }()

  // do stuff and store in created.  Be careful not to shadow err in "if" blocks, i.e.
  err := makeStuff(created)
  if err != nil {  // using "if err :=...; err != nil" here would shadow the outer err
    return err
  }
}
```

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +283,147 @@ func loadCatalog(
...
-		p.Context = &minderv1.Context{}
+	printCatalogSummary(cmd, result)
+	return nil
+}
+
+type applyCatalogResult struct {
+	createdRuleTypes    []string
+	createdProfiles     []string
+	seenExistingProfile bool
+}
```
evankanderson, line 346: Again, prefer to declare structs _before_ usage, so someone reading from the top of file can understand what's in them before they are used.

cmd/cli/app/quickstart/quickstart.go
```diff
@@ -275,116 +283,147 @@ func loadCatalog(
...
+				cmd.Printf("Rule type %s already exists\n", ruleType.GetName())
+				continue
+			}
+			return fmt.Errorf("error creating rule type %s: %w", ruleType.GetName(), err)
+		}
+
+		name := resp.GetRuleType().GetName()
+		if name == "" {
+			name = ruleType.GetName()
+		}
```
evankanderson, line 386: Slight preference for `cmp.Or` here, to help improve density:


```suggestion
		name := cmp.Or(resp.GetRuleType().GetName(), ruleType.GetName())
```

I'm also not sure how the empty string from the server could happen.  Is this something we've seen?

internal/cli/catalog.go
evankanderson: I see this was resolved without comment.  I'd like to understand the defensive coding here vs (for example) in 'createCatalogProfiles`.  In general, defensive coding can end up adding a lot of "bulk" of code which is never actually used, but inhibits understanding and inflates or distorts test coverage.

internal/cli/catalog.go
```diff
@@ -0,0 +1,185 @@
...
+	loadedProfiles, err := loadProfilesFromFS(vfs, warnf)
+	if err != nil {
+		return nil, err
+	}
+
+	names := slices.Sorted(maps.Keys(ruleTypesByName))
+	ruleTypes := make([]*minderv1.RuleType, 0, len(names))
+	for _, name := range names {
+		ruleTypes = append(ruleTypes, ruleTypesByName[name])
+	}
```
evankanderson, line 55: Why not return a slice and then use a `SortFunc` on the slice?

internal/cli/catalog.go
```diff
@@ -0,0 +1,185 @@
...
+)
+
+// Catalog represents a loaded collection of rule types and profiles from a filesystem.
+type Catalog struct {
+	RuleTypes []*minderv1.RuleType
+	Profiles  []*minderv1.Profile
+}
+
+// WarnFunc reports skipped catalog resources.
+type WarnFunc = fileconvert.Printer
```
evankanderson, line 32: It's a little odd to copy a utility interface from another package.  What about simply adding this to the `fileconvert` package?

internal/cli/catalog.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+
+	validProfiles := make([]*minderv1.Profile, 0, len(c.Profiles))
+	for _, loadedProfile := range c.Profiles {
+		referencedRuleTypes := make(map[string]struct{})
+		if err := profiles.TraverseRuleTypesForEntities(loadedProfile, func(_ minderv1.Entity, rule *minderv1.Profile_Rule) error {
+			if rule.GetType() != "" {
+				referencedRuleTypes[rule.GetType()] = struct{}{}
+			}
```
evankanderson, line 88: If the rule's type is empty-string, what is supposed to happen?  It's not clear we need the "if" check here.

internal/cli/catalog.go
```diff
@@ -0,0 +1,185 @@
...
+		}); err != nil {
+			warnf("Skipping invalid profile %q: failed to inspect rules: %v\n", loadedProfile.GetName(), err)
+			continue
+		}
+
+		maps.DeleteFunc(referencedRuleTypes, func(ruleType string, _ struct{}) bool {
+			_, exists := ruleTypesByName[ruleType]
+			return exists
+		})
+		missingRuleTypes := slices.Sorted(maps.Keys(referencedRuleTypes))
```
evankanderson, line 99: This feels like a lot of mechanism to find the ruletypes which _aren't_ in `ruleTypesByName`.  Why not simply do:

```golang
missingRuleTypes := []string
profiles.TraverseRuleTypesForEntities(loadedProfile, func(_ minderv1.Entity, rule *minderv1.Profile_Rule) error {
  if _, ok := ruleTypesByName[rule.GetType()]; !ok {  // No rule type defined, error!
    missingRuleTypes = append(missingRuleTypes, rule.GetType())
  }
}

// could sort / dedup the slice here if necessary with with slices.Compact(slices.Sorted(missingRuleTypes))
```

internal/cli/catalog.go
```diff
@@ -0,0 +1,185 @@
...
+			)
+			continue
+		}
+
+		validProfiles = append(validProfiles, loadedProfile)
+	}
+
+	if len(validProfiles) == 0 {
+		return fmt.Errorf("no valid profiles found under %s", catalogProfilesDir)
+	}
```
evankanderson, line 115: What's the motivation for dropping profiles rather than erroring on missing references?  I'm not disagreeing, it just seems like an important behavior choice, and I'm not sure whether a loud failure for a user who's unfamiliar with the system is better or worse than warn-and-continue for a user who (again) can't tell what "right" looks like.

internal/cli/catalog.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil, err
+	}
+
+	ruleTypesByName := make(map[string]*minderv1.RuleType, len(resources))
+	for _, resource := range resources {
+		ruleType, ok := resource.(*minderv1.RuleType)
+		if !ok {
+			warnf("Skipping invalid resource type %T under %s\n", resource, catalogRuleTypesDir)
+			continue
+		}
```
evankanderson, line 133: This is going to print for every profile (and datasource) in the directory, even though the profiles _will_ be processed later.  You may want to do a single call to `ResourcesFromPaths` and then pass the results into each function, rather than re-walking the path multiple times.

You could also look at adjusting `ResourcesFromPaths` to return a `Catalog`-style resource.

internal/cli/catalog.go
evankanderson: I see this was resolved, but I don't see what changed.

internal/cli/catalog_test.go
```diff
@@ -0,0 +1,97 @@
...
+						},
+					},
+				},
+			},
+			expectErr:        false,
+			expectedProfiles: 1,
+			expectedFirst:    "valid",
+		},
+		{
+			name: "mixed valid and missing rule references",
```
evankanderson, line 53: We should have a test which contains a ruletype, but all the profiles reference at least one non-declared ruletype.

pkg/fileconvert/collect.go
```diff
@@ -20,21 +21,37 @@ import (
...
+	Expanded bool
+}
+
 // ResourcesFromPaths collects
-func ResourcesFromPaths(printer Printer, paths ...string) ([]minderv1.ResourceMeta, error) {
-	files, err := util.ExpandFileArgs(paths...)
+func ResourcesFromPaths(vfs billy.Filesystem, printer Printer, paths ...string) ([]minderv1.ResourceMeta, error) {
+	if printer == nil {
+		printer = func(string, ...any) {}
+	}
```
evankanderson, line 33: Again, I don't think we need this defensive coding.

pkg/fileconvert/collect.go
```diff
@@ -20,21 +21,37 @@ import (
...
 
 	objects := make([]minderv1.ResourceMeta, 0, len(files))
 	for _, file := range files {
 		var input Decoder
+		var closer io.Closer
 		if file.Path == "-" {
+			if vfs != nil {
+				return nil, fmt.Errorf("stdin is not supported with filesystem-backed reads")
+			}
```
evankanderson, line 51: This is odd -- why do we have this?  (Yes, we might need to be able to override `stdin` in the future, but I don't see why it needs to be a hard error right now.)

pkg/fileconvert/collect.go
evankanderson: I see this was resolved, but there were no changes or comments.  (I suggested this again on `loadRuleTypesFromVFS`)

pkg/fileconvert/encodedecode.go
```diff
@@ -32,26 +33,65 @@ type Decoder interface {
...
 	case ".yaml", ".yml":
-		builder = func(r io.Reader) Decoder { return yaml.NewDecoder(r) }
+		return yaml.NewDecoder(reader), nil
 	default:
-		return nil, nil
+		return nil, fmt.Errorf("unsupported file format: %s", path)
+	}
+}
+
+func decoderForPath(fs billy.Filesystem, path string) (Decoder, io.Closer) {
```
evankanderson, line 47: Given that this is a private function, it seems like we can assume that `fs` has been created correctly previously, no?

pkg/fileconvert/encodedecode.go
```diff
@@ -32,26 +33,65 @@ type Decoder interface {
 var _ Decoder = (*yaml.Decoder)(nil)
 var _ Decoder = (*json.Decoder)(nil)
 
-// DecoderForFile returns a Decoder for the file at the specified path,
-// or nil if the file is not of the appropriate type.
-func DecoderForFile(path string) (Decoder, io.Closer) {
-	path = filepath.Clean(path)
-	ext := filepath.Ext(path)
-	var builder func(io.Reader) Decoder
-	// we return functions here so that we can early-exit without opening the file if the extension is unmatched.
```
evankanderson, line 41: The previous code was structured to only *attempt* to open the file if it was of the appropriate type.  This code tries to open the file first, which generates more work and cleanup than the reverse approach.

We've also added an "error" to the return type from `decoderForReader`, which means we need to check and handle some more cases.

pkg/fileconvert/encodedecode.go
```diff
@@ -32,26 +33,65 @@ type Decoder interface {
...
+	}
+
+	return decoder, file
+}
+
+// DecoderForFile returns a Decoder for the file at the specified path,
+// or nil if the file is not of the appropriate type.
+func DecoderForFile(path string) (Decoder, io.Closer) {
+	return decoderForPath(nil, path)
+}
```
evankanderson, line 79: Can we simply update existing callers to the new method (and enforce they provide a vfs)?

pkg/fileconvert/encodedecode.go
```diff
@@ -32,26 +33,65 @@ type Decoder interface {
...
+
+// ReadResourceFromFile reads a single resource from the specified filesystem path.
+func ReadResourceFromFile[T proto.Message](fs billy.Filesystem, path string) (resource T, err error) {
```
evankanderson, line 82: Is this method used anywhere?

pkg/mindpak/reader/reader.go
```diff
@@ -82,7 +84,7 @@ func (b *bundleReader) GetProfile(name string) (*v1.Profile, error) {
 	defer file.Close()
 
 	// parse profile from YAML
-	profile, err := profiles.ParseYAML(file)
```
evankanderson, line 85: Did we remove all usage of `profiles.ParseYAML` in favor of `ReadResourceTyped`?  If so, can we remove the method?

evankanderson (commented):
Sorry for the delay getting back to you -- it seems like there were a number of comments which got closed as "resolved" without either action or a counter-argument ("this seems clearer" or "I don't understand the suggestion" would be okay if that's the case).

PR #6368, opened 2026-04-14
fix: prevent cache poisoning by force-resetting worktree on failed PR remediation

### LINKED ISSUE #6367: Bug: Cache poisoning vulnerability in PR remediator causes silent evaluation failures
## Describe the bug

There's a critical state-leak in [internal/engine/actions/remediate/pull_request/pull_request.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/engine/actions/remediate/pull_request/pull_request.go:0:0-0:0) where failing to fully commit and push a pull request remediation poisons the ingestion cache for all subsequent rule evaluations in that run.

When the PR remediator runs, it checks out a new branch in the `go-git` workspace (which sits directly on the shared `ingestCache.Fs` virtual filesystem). If the remediation algorithm fails *after* modifying files but *before* committing (e.g. failing during `wt.Add` or crashing during file generation), the worktree is left in a dirty state.

The code attempts to clean up in a `defer` block using `checkoutToOriginallyFetchedBranch`:
```go
func checkoutToOriginallyFetchedBranch(...) {
	err := wt.Checkout(&git.CheckoutOptions{
		Branch: originallyFetchedBranch, 
        // BUG: Missing `Force: true`
	})
}
```
Because this `wt.Checkout` call omits Force: true and `CleanOptions`{Dir: true}, go-git defaults to preserving all uncommitted/untracked modifications and carrying them into the target branch.
## Impact
Because the executor shares a single `ingestCache.Fs` across all rule evaluations for a repository during a run, this dirty state leaks across boundaries. If Rule A attempts remediation, modifies a file to be "compliant", but crashes out, Rule B will then evaluate against that dirty cache. Rule B will falsely evaluate the repository as compliant, creating a silent security bypass (false negative).
## Steps to reproduce

1. Configure two rules against the same entity (e.g., Rule A and Rule B).
2. Set Rule A to auto-remediate with PRs.
3. Simulate a failure during Rule A's remediation (after it modifies fs but before wt.Commit).
4. Observe that Rule B evaluates against the uncommitted files produced by Rule A rather than the upstream repository state.

## Expected Behavior
A failed remediation attempt should completely blow away the dirty `worktree` and restore the ingest cache entirely to its pristine upstream state.
## Suggested Fix
Update `checkoutToOriginallyFetchedBranch` to aggressively reset the workspace:
```
	err := wt.Checkout(&git.CheckoutOptions{
		Branch: originallyFetchedBranch,
		Force:  true, // Discards modified tracked files
	})
    // And also run:
    // wt.Clean(&git.CleanOptions{Dir: true}) // Removes left-behind untracked files

```

AftAb-25:
Fixes #6367

## Description
This fixes a cache poisoning bug in the PR remediator where failed remediations would leak uncommitted changes into the shared ingest cache, causing silent evaluation failures for subsequent rules.

When [pull_request.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/engine/actions/remediate/pull_request/pull_request.go:0:0-0:0) aborts a remediation and runs [checkoutToOriginallyFetchedBranch](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/engine/actions/remediate/pull_request/pull_request.go:562:0-585:1), it previously used the default `go-git` checkout which preserves modifications to the worktree/index. Because the executor uses a single shared `ingestCache.Fs` per run, those left-behind dirty files would then be incorrectly evaluated by subsequent rules.

This PR aggressively resets the worktree during the checkout cleanup phase:
- Adds `Force: true` to the checkout options to discard modified tracked files.
- Calls `wt.Clean(&git.CleanOptions{Dir: true})` to remove any untracked leftover files.

I've also added a unit test ([checkout_cleanup_test.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/engine/actions/remediate/pull_request/checkout_cleanup_test.go:0:0-0:0)) that simulates a dirty worktree and ensures both tracked modifications and untracked files are correctly wiped out.

## Checklist
- [x] Code compiles correctly
- [x] Added tests that fail without the change (if possible)
- [x] All tests passing
- [x] Extended the README / documentation, if necessary

### DIFF
--- a/internal/engine/actions/remediate/pull_request/checkout_cleanup_test.go
+++ b/internal/engine/actions/remediate/pull_request/checkout_cleanup_test.go
@@ -1,0 +1,87 @@
+// SPDX-FileCopyrightText: Copyright 2024 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package pull_request
+
+import (
+	"testing"
+	"time"
+
+	"github.com/go-git/go-billy/v5/memfs"
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/object"
+	"github.com/go-git/go-git/v5/storage/memory"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCheckoutToOriginallyFetchedBranch_CleansWorktree(t *testing.T) {
+	t.Parallel()
+
+	mfs := memfs.New()
+	storer := memory.NewStorage()
+	repo, err := git.InitWithOptions(storer, mfs, git.InitOptions{
+		DefaultBranch: "refs/heads/main",
+	})
+	require.NoError(t, err)
+
+	wt, err := repo.Worktree()
+	require.NoError(t, err)
+
+	// Create an initial commit on main so we have a valid HEAD.
+	f, err := mfs.Create("README.md")
+	require.NoError(t, err)
+	_, err = f.Write([]byte("initial"))
+	require.NoError(t, err)
+	require.NoError(t, f.Close())
+
+	_, err = wt.Add("README.md")
+	require.NoError(t, err)
+	_, err = wt.Commit("initial commit", &git.CommitOptions{
+		Author: &object.Signature{Name: "test", Email: "t@t.com", When: time.Now()},
+	})
+	require.NoError(t, err)
+
+	mainRef := plumbing.NewBranchReferenceName("main")
+
+	// Checkout a remediation branch and simulate a partial failure:
+	// modify a tracked file and create an untracked file without committing.
+	err = wt.Checkout(&git.CheckoutOptions{
+		Branch: plumbing.NewBranchReferenceName("minder_remediation"),
+		Create: true,
+	})
+	require.NoError(t, err)
+
+	// Dirty a tracked file (simulates remediation writing a compliant version).
+	f, err = mfs.Create("README.md")
+	require.NoError(t, err)
+	_, err = f.Write([]byte("dirty content from failed remediation"))
+	require.NoError(t, err)
+	require.NoError(t, f.Close())
+
+	// Create an untracked file (simulates remediation creating a new config).
+	f, err = mfs.Create("leftover.txt")
+	require.NoError(t, err)
+	_, err = f.Write([]byte("should be removed"))
+	require.NoError(t, err)
+	require.NoError(t, f.Close())
+
+	// Verify the worktree is actually dirty before we clean it.
+	status, err := wt.Status()
+	require.NoError(t, err)
+	require.False(t, status.IsClean(), "worktree should be dirty before cleanup")
+
+	// Run checkoutToOriginallyFetchedBranch — this is the code under test.
+	logger := zerolog.Nop()
+	checkoutToOriginallyFetchedBranch(&logger, wt, mainRef)
+
+	// Assert: worktree should now be clean.
+	status, err = wt.Status()
+	require.NoError(t, err)
+	require.True(t, status.IsClean(), "worktree should be clean after checkout with Force + Clean")
+
+	// Assert: untracked file should have been removed by wt.Clean.
+	_, err = mfs.Stat("leftover.txt")
+	require.Error(t, err, "untracked file should have been removed by Clean")
+}
--- a/internal/engine/actions/remediate/pull_request/pull_request.go
+++ b/internal/engine/actions/remediate/pull_request/pull_request.go
@@ -567,14 +567,22 @@ func checkoutToOriginallyFetchedBranch(
 ) {
 	err := wt.Checkout(&git.CheckoutOptions{
 		Branch: originallyFetchedBranch,
+		Force:  true,
 	})
 	if err != nil {
 		logger.Err(err).Msg(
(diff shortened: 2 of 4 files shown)


internal/engine/actions/remediate/pull_request/checkout_cleanup_test.go
evankanderson, line 1: Can you add this to `pull_request_test.go`, rather than creating a new test file (which doesn't correspond to any of the regular `.go` files in this directory)?

internal/engine/actions/remediate/pull_request/checkout_cleanup_test.go
```diff
@@ -0,0 +1,87 @@
...
+		Create: true,
+	})
+	require.NoError(t, err)
+
+	// Dirty a tracked file (simulates remediation writing a compliant version).
+	f, err = mfs.Create("README.md")
+	require.NoError(t, err)
+	_, err = f.Write([]byte("dirty content from failed remediation"))
+	require.NoError(t, err)
+	require.NoError(t, f.Close())
```
evankanderson, line 61: `util.WriteFile` from `go-billy/v5/util` might be simpler:


```suggestion
	require.NoError(util.WriteFile(mfs, "README.md", []byte("dirty content from failed remediation"), 0644))
```

internal/engine/actions/remediate/pull_request/pull_request.go
```diff
@@ -567,14 +567,22 @@ func checkoutToOriginallyFetchedBranch(
 ) {
 	err := wt.Checkout(&git.CheckoutOptions{
 		Branch: originallyFetchedBranch,
+		Force:  true,
```
evankanderson, line 570: It looks like [the default behavior is `MixedReset`](https://pkg.go.dev/github.com/go-git/go-git/v6#ResetMode), which will reset the index but not the working tree.  This _should_ prevent importing already-changed files into the branch in most cases, but the worktree corruption could affect future evaluations, since the order is currently "evaluate one rule, then evaluate its actions".

evankanderson (commented):
Nice find!  A few minor cleanups, but I'd love to merge this soon.

AftAb-25:
Hi @evankanderson, thank you for the review and the thoughtful suggestion! 

I looked into relying solely on `go-git`'s native [Checkout](cci:1://file:///C:/Users/aftab/Desktop/minder/go_git_worktree.go:159:0-198:1) logic to evaluate tracked vs untracked files, and I ran unit tests dropping the `.Clean()` call to verify. Unfortunately, it turns out that `go-git`'s [Checkout](cci:1://file:///C:/Users/aftab/Desktop/minder/go_git_worktree.go:159:0-198:1) does not natively handle explicitly untracked files. 

According to the official `go-git` [[CheckoutOptions](cci:1://file:///C:/Users/aftab/Desktop/minder/go_git_worktree.go:228:0-256:1) documentation](https://pkg.go.dev/github.com/go-git/go-git/v5/worktree#CheckoutOptions):
> *"Force, if true, any local modification to existing files will be dropped, **and any untracked files will remain.***"

My local test assertions confirmed this limitation:
1. **If we remove `Force: true`**: [Checkout](cci:1://file:///C:/Users/aftab/Desktop/minder/go_git_worktree.go:159:0-198:1) cleanly falls back to `MergeReset`. However, because the remediator strictly operates by modifying files, the checkout immediately throws an `ErrUnstagedChanges` exception. This leaves the modifications intact and leaves the cache fully poisoned.
2. **If we use `Force: true` but remove `.Clean()`**: The tracked files are successfully reverted, but any completely new, untracked files generated during the aborted remediation run (e.g. `.lock` files or build artifacts) persistently remain in the directory.

Because we are checking out over a natively shared `ingestcache` where subsequent, unrelated rule evaluations heavily depend on a pure state, both [Checkout(Force: true)](cci:1://file:///C:/Users/aftab/Desktop/minder/go_git_worktree.go:159:0-198:1) (to wipe modified tracked files) and `wt.Clean()` (to strictly purge newly created untracked files) are structurally required to guarantee a fully hermetic state after an abort.

Let me know if this context makes sense, or if you'd like me to look into implementing this via a different `go-git` mechanism (like a manual `HardReset` combination) instead!

evankanderson:
> Hi @evankanderson, thank you for the review and the thoughtful suggestion!
> 
> I looked into relying solely on `go-git`'s native Checkout logic to evaluate tracked vs untracked files, and I ran unit tests dropping the `.Clean()` call to verify. Unfortunately, it turns out that `go-git`'s Checkout does not natively handle explicitly untracked files.

Yes, I agreed with you in this comment: https://github.com/mindersec/minder/pull/6368#discussion_r3116237879

(We shouldn't have any untracked files except in very unusual circumstances, but it looks like even tracked file changes may persist in the worktree for future evaluations.)

However, I had two other requests as well:

* [Put the additional tests alongside the existing tests in `pull_request_test.go`](https://github.com/mindersec/minder/pull/6368#discussion_r3116014457)
* [Use `util.WriteFile` from Billy's util module to simplify the code](https://github.com/mindersec/minder/pull/6368#discussion_r3116173669)

AftAb-25:
Hi, @evankanderson 
Thanks for the feedback! I've updated the PR with both requests:
- Moved the tests into `pull_request_test.go`
- Updated the test file creation code to use `util.WriteFile` from Billy's util module for simplicity

evankanderson (approved):

PR #6370, opened 2026-04-14
fix: abort repository deletion if upstream webhook deregistration fails

### LINKED ISSUE #6369: Bug: Failed repository deletions leave permanent orphaned webhooks in GitHub
## Describe the bug

There's a critical resource leak in [internal/repositories/service.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/repositories/service.go:0:0-0:0) when a user attempts to delete a repository from Minder. 

When [deleteRepository](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/repositories/service.go:397:0-426:1) is called, the code first attempts to deregister the repository entity by deleting the webhook in GitHub via `client.DeregisterEntity(...)`. However, if that GitHub API call fails for any reason (e.g., temporary network issue, rate limiting, or revoked credentials), the error is only logged. The code then completely ignores the failure and proceeds to execute the database transaction that permanently wipes the repository from the Minder DB.

The result is a permanent "zombie" webhook left in the user's GitHub repository.

## The Impact
1. **Unrecoverable State:** The user is completely locked out of attempting to delete the webhook gracefully through Minder. If they try calling the delete endpoint again, Minder returns `sql.ErrNoRows` since the repo is already gone natively.
2. **Infinite Ingress Spam:** GitHub will continuously blast Minder with events for the targeted repository. Because Minder deleted the repo locally, the webhook handlers can't find it locally resulting in 404/ErrNoRows, silently dropping payloads indefinitely and wasting bandwidth, CPU, and DB queries over time.

## Steps to reproduce

1. Register a GitHub repository in Minder (verifying the webhook is created on GitHub).
2. Manually invalidate/revoke the GitHub token, or simulate a network failure or `500 Internal Server Error` using a proxy.
3. Attempt to delete the repository via Minder.
4. Observe the `error deregistering repo` in the Minder server logs.
5. Check the Minder database: The repository has been successfully deleted.
6. Check the GitHub repository settings: The webhook is still active and continuously sending data to Minder.

## Expected Behavior

Repository deletion should be a safe, atomic operation from the user's perspective. If Minder fails to deregister the webhook upstream, it should abort the local database deletion and fail fast, allowing the user to retry the deletion once credentials or upstream connectivity are restored.

## Suggested Fix

In [internal/repositories/service.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/repositories/service.go:0:0-0:0), return the error explicitly if `DeregisterEntity` fails, blocking the `db.WithTransaction` block from executing:

```go
	err = client.DeregisterEntity(ctx, pb.Entity_ENTITY_REPOSITORIES, repo.Properties)
	if err != nil {
		zerolog.Ctx(ctx).Error().
			Dict("properties", repo.Properties.ToLogDict()).
			Err(err).Msg("error deregistering repo")
        // Abort deletion to prevent an orphaned webhook state.
        return fmt.Errorf("failed to deregister webhook upstream: %w", err)
	}
    
    // Proceed with db.WithTransaction...

AftAb-25:
Fixes #6369

## Description
This fixes a critical resource leak where deleting a repository from Minder could leave a permanently orphaned "zombie" webhook in the user's GitHub repository.

Previously, when a user deleted a repo, [deleteRepository](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/repositories/service.go:397:0-427:1) in [internal/repositories/service.go](cci:7://file:///c:/Users/aftab/Desktop/minder/internal/repositories/service.go:0:0-0:0) would attempt to deregister the webhook upstream. If that GitHub API call failed (e.g., due to a temporary network issue, rate limit, or revoked token), the code simply logged the error and blithely continued to delete the repository from the Minder database.

Once deleted from Minder, the user had absolutely no way to retry the deletion, leaving the active webhook permanently installed on their GitHub repo. This webhook would continuously spray ingress traffic against Minder, which would just `404` or `sql.ErrNoRows` since the repo no longer existed in the database.

## Changes
- Added a fail-fast return to [deleteRepository](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/repositories/service.go:397:0-427:1). The DB transaction that wipes the repository is now completely skipped if `client.DeregisterEntity` fails, allowing users to safely retry the deletion once credentials or upstream connectivity are restored.
- Updated the two [DeleteRepository](cci:1://file:///c:/Users/aftab/Desktop/minder/internal/controlplane/handlers_repositories.go:263:0-286:1) unit tests that previously baked in the old leak behavior (they expected the DB deletion to succeed even when deregistration failed). These tests now correctly expect an error.

## Checklist
- [x] Code compiles correctly
- [x] Added tests that fail without the change (if possible)
- [x] All tests passing
- [x] Extended the README / documentation, if necessary

### DIFF
--- a/internal/repositories/service.go
+++ b/internal/repositories/service.go
@@ -405,6 +405,7 @@ func (r *repositoryService) deleteRepository(
 		zerolog.Ctx(ctx).Error().
 			Dict("properties", repo.Properties.ToLogDict()).
 			Err(err).Msg("error deregistering repo")
+		return fmt.Errorf("failed to deregister entity upstream, aborting deletion to prevent orphaned webhook: %w", err)
 	}
 
 	_, err = db.WithTransaction(r.store, func(t db.ExtendQuerier) (*pb.Repository, error) {
--- a/internal/repositories/service_test.go
+++ b/internal/repositories/service_test.go
@@ -191,14 +191,15 @@ func TestRepositoryService_DeleteRepository(t *testing.T) {
 			ExpectedError: "error instantiating provider",
 		},
 		{
-			Name:         "DeleteByName still works when entity cannot be deregistered",
+			Name:         "DeleteByName fails when entity cannot be deregistered",
 			DeleteType:   ByName,
-			DBSetup:      newDBMock(withSuccessfulGetByName, withSuccessfulDelete),
+			DBSetup:      newDBMock(withSuccessfulGetByName),
 			ServiceSetup: newPropSvcMock(withSuccessfulEntityWithProps),
 			ProviderManagerSetup: func(p provinfv1.Provider) pf.ProviderManagerMockBuilder {
 				return pf.NewProviderManagerMock(pf.WithSuccessfulInstantiateFromID(p))
 			},
 			ProviderSetup: newProviderMock(withFailedDeregister),
+			ExpectedError: "failed to deregister entity upstream",
 		},
 		{
 			Name:         "DeleteByName by ID fails when repo cannot be deleted from DB",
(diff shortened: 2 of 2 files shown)


evankanderson (commented):
We need to be able to delete the repository if the user uses the GitHub interface to remove the repository from the application installation.  In this case, we _have_ to leak the webhook, as we don't have permissions to update the repository (including to remove the webhook) anymore.  The best we can do is to clean up the Minder state around the repository -- the dangling webhook has become unreachable from Minder's cleanup perspective.

It's possible that all this could be avoided by registering an app-level webhook, but that would be a larger refactor of both webhook and provider logic.

evankanderson (changes requested):
The most likely cause of being unable to delete a webhook is that the app's permissions on the repository have been removed.  In that case, keeping a repository object in Minder will simply produce more failed attempts to act on a resource that we don't have permissions on.

We could potentially introduce some sort of "tombstone" process to mark these entities in the database as needing future cleanup, but preventing any policy application (e.g. filtering them from all the current RPC outputs, etc), but I'm not convinced that the payoff is worth the costs.

I'm marking this PR as "request changes" (needs to address the above issue) to help track which outstanding PRs need maintainer action vs contributor action.

evankanderson:
Closing per [this comment](https://github.com/mindersec/minder/pull/6370#pullrequestreview-4191717366) and no reply.

PR #6371, opened 2026-04-15
Fix: add panic recovery with proper cleanup to entity event handler

### LINKED ISSUE #6319: Ensure proper cleanup of cancel functions on panic in ExecutorEventHandler
While working with `ExecutorEventHandler`, I noticed that cancel functions for in-flight evaluations are tracked in a slice and cleaned up during normal execution.

However, in the case of a panic during execution, it’s not entirely clear that this cleanup is always guaranteed to run. If cleanup is skipped, it could leave stale cancel functions in the slice and lead to inconsistent internal state over time.

This isn’t causing any immediate issues in normal flows, but it could become problematic in edge cases or long-running processes.

It might be worth ensuring that cleanup logic (especially removal of cancel functions) always executes, even in panic scenarios, so that the lifecycle of these functions is fully reliable and predictable.

Happy to take this up if it makes sense.

mintRaven-05:
# Summary
Previously, the cancel function cleanup was deferred until after context setup, which meant panics occurring during executor evaluation could leave stale cancel function references in the handler's cancels slice. This change moves the cleanup into a dedicated defer block that executes first, guaranteeing cleanup even if the executor panics.

**Key changes:**
- Added panic recovery wrapper in `HandleEntityEvent` goroutine with comprehensive error logging
- Cancel function cleanup now executes reliably regardless of panic state
- Added logging of panic events with full entity context for debugging
- Prevents memory leaks and stale state in long-running handlers

Fixes #6319

# Testing

The changes were tested through:

1. **Existing tests pass** - All existing tests in `handler_test.go` continue to pass, including:
   - `TestExecutorEventHandler_handleEntityEvent`
   - `TestExecutorEventHandler_ShutdownCancelsNewEvents`

2. **New panic recovery test** - Added `TestExecutorEventHandler_CleanupOnPanic`

To test locally:
```bash
go test -json -race -v ./internal/engine | gotestfmt
```
image


sachin9058:
Nice work adding panic recovery and ensuring cleanup runs — this definitely improves robustness for long-running handlers 👍

A couple of thoughts:

* Since recover() is used here, the panic is effectively swallowed. Do we want to fully suppress it, or should there be an option to propagate or signal failure upstream? Just wondering about observability vs fault visibility trade-offs.

* In the cleanup logic, comparing function pointers (`cf == &shutdownCancel`) might be a bit fragile depending on how the cancel functions are created and referenced. It might be worth confirming this is reliable across all cases.

* In the test, using `time.Sleep` could lead to flakiness. Maybe we could use a more deterministic sync mechanism (like a channel or waitgroup) to ensure the goroutine has completed before assertions.

Overall, this looks like a solid improvement and a good step toward making the handler more resilient.

mintRaven-05:
>     * Since recover() is used here, the panic is effectively swallowed. Do we want to fully suppress it, or should there be an option to propagate or signal failure upstream? Just wondering about observability vs fault visibility trade-offs.
well... panic is logged with full context... i dont think it is required to send the failure upstream
>     * In the cleanup logic, comparing function pointers (`cf == &shutdownCancel`) might be a bit fragile depending on how the cancel functions are created and referenced. It might be worth confirming this is reliable across all cases.
Comparing function pointers is safe in Go... but maybe using a map[*context.CancelFunc]bool or ID-based tracking would be more maintainable and clearer instead of a slice (so i can remove slice and replace it with a map)... i can do this change
>     * In the test, using `time.Sleep` could lead to flakiness. Maybe we could use a more deterministic sync mechanism (like a channel or waitgroup) to ensure the goroutine has completed before assertions.

this is actually a better idea... i am fixing this by using channel

mintRaven-05:
@sachin9058 
i have updated the `handler_test.go`. It now uses a channel signal in panic recovery test instead of time.sleep(100ms)
however, i wont push the code where i use map instead of slice to track cancels without approval from @evankanderson  

**NOTE: all tests were passed**
command used: `make test` and `go test -json -race -v ./internal/engine | gotestfmt`

sachin9058:
Looks great  thanks for addressing the feedback 👍

Using a channel instead of time.Sleep makes the test much more deterministic, and the overall approach here improves robustness of the handler.

Happy with this change.

internal/engine/handler.go
```diff
@@ -119,6 +119,31 @@ func (e *ExecutorEventHandler) HandleEntityEvent(msg *message.Message) error {
 	e.wgEntityEventExecution.Add(1)
 	go func() {
 		defer e.wgEntityEventExecution.Done()
+
+		// Ensure cancel function is cleaned up even if panic occurs.
+		// This guarantees the lifecycle of cancel functions is reliable
+		// and prevents stale references from accumulating in the cancels slice.
```
evankanderson, line 125: If this panic isn't recovered, what happens?  I suspect that the cancels get cleaned up somewhat more abruptly, based on the contract for a non-recovered panic in a goroutine.

If Minder is running under a platform like Kubernetes, this will manifest as pod restarts, which tends to be a "loud" sign something is wrong, rather than a "quiet" signal like a log message.

evankanderson (commented):
I appreciate the goal of making error processing more robust in Minder, but a panic outside the rule evaluator feels like a programming error that might better be handled with the Erlang crash-and-restart approach, rather than a Java exception model.

mintRaven-05:
thanks for the review
actually your concern is right, if ensuring a crash is important... i can do few changes here to make sure that both cleanup happens and a crash occurs as well after that.

would it be better to re-panic after cleanup to make the error visible  as a pod restart in Kubernetes? That way we get both the cleanup guarantee and the visibility:
```go
defer func() {
    if panicErr := recover(); panicErr != nil {
        zerolog.Ctx(msgCtx).Error().Any("panic", panicErr)...
        panic(panicErr)  // panic to crash process
    }
}()
```
i am just trying to follow the best practices... where we zerolog errors while crashing when needed

internal/engine/handler.go
evankanderson: > If the panic isn't recovered:
> 
> 1. Goroutine dies but cleanup code never runs
> 2. &shutdownCancel stays in the map indefinitely
> 3. map grows unbounded
> 4. pod keeps running, no visibility
> 
> The defer recovery can solve this by:
> - Guaranteeing cleanup happens (map entry deleted)
> - Logging the panic with full context
> - Allowing the handler to remain responsive
> 
> This is why the recovery is necessary for reliability.

According to https://go.dev/blog/defer-panic-and-recover:

> When the function F calls panic, execution of F stops, any deferred functions in F are executed normally, and then F returns to its caller. To the caller, F then behaves like a call to panic. The process continues up the stack until all functions in the current goroutine have returned, at which point the program crashes.

My thought is that "the program crashes" will empty the cancellation map by calling `os.Exit()` and invoking the kernel's process cleanup routine.  Once the program has crashed, the map will be gone, along with any other state in the process, including open file descriptors and network sockets.

mintRaven-05: i see... so a simple log and panic is enough to resolve this i believe... the cleanup is not necessary.
should i commit the change ?

evankanderson:
> i am just trying to follow the best practices... where we zerolog errors while crashing when needed

I think `panic` without a `recover` will give us a stack trace of where the program crashed? 

I'm not trying to be difficult, but there are often advantages to leaning in to the language patterns and features over reimplementing them yourself.

mintRaven-05:
> I think `panic` without a `recover` will give us a stack trace of where the program crashed?

so we can use the stack trace output to our advantage right ?
> I'm not trying to be difficult, but there are often advantages to leaning in to the language patterns and features over reimplementing them yourself.

you are right to point that out
so what is the final suggestion... that can be implemented?
A panic without a recovery so that it provides the stack trace, crashes the process and it invokes the kernel's process cleanup routine automatically

in that case i can close this PR... since the original code already propagated the panic naturally

evankanderson:
> in that case i can close this PR... since the original code already propagated the panic naturally

Yes, that would be my take -- in general, `panic` and `recover` is best (in Go code) for specific units where the recovery action is clear -- for example, we `recover` around rule evaluation, but our recovery action in that case is to record the rule evaluation as "error" in the same way that we handle other rule evaluation errors.  We don't have a similar recovery around remediation and alert actions at the moment, because those tend to be more structured (at the moment), and we haven't had the need.

mintRaven-05:
got it... thanks for the feedback!
i am closing this PR and you can close the issue since its irrelevant

Issue #2176, opened 2024-01-23
Not all commits being added to `main` are checked during PR


gregfurman:
### Describe the issue

It is currently possible to have commits prior to HEAD of an incoming branch merged into `main` that contain vulnerabilities

`minder` seemingly does not evaluate all commits in a diff allowing for a commit with a vulnerable version to be included when merging a PR.

Although this wont cause HEAD of `main` (or the target) to have a vulnerability pulled in, there should be checks against being able to introduce any commits onto `main` that include unsafe dependencies.

### To Reproduce

- Create a branch off of main
- Change a dependency's version to include a vulnerability and commit (commit A)
- Change this dependency back to a safe version in a new commit (commit B)
- Create a PR and see how `minder` will evaluate this as safe despite commit A being present and potentially introducing a vulnerability to the version history

evankanderson:
Hey Greg,

I'm trying to understand the vulnerability here -- is the concern that there could be a merge commit on the history of `main` which includes a commit where the code is not in a safe state?  Is that a different concern than (for example) that a commit in the history of `main` contains the code in a non-working state, even though the merged commit was working and passed unit tests?

gregfurman:
Hey Evan! 

If I recall, the main concern is when rebasing/merging multiple commits into `main` -- not all commits are checked for vulnerabiltiies.

For example:
`main` contains commits `A -> B -> C`
`feature-branch` contains commits `D -> E* -> F`

Now if commit `E*` has vulnerabilities (let's assume a log4j issue) then merging in ALL commits of `feature-branch` onto `main` such that `main` looks like  `A -> B -> C -> D -> E* -> F`  then there exists a commit on `main`  (`E*`) with some vulnerability -- despite `HEAD` of `main` (`F`) being secure.

Here is a diagram of the scenario as I explained it to Jakub.

![Image](https://github.com/user-attachments/assets/181c1076-0dd1-4e46-9b69-367a6b9e4cdd)

evankanderson:
Ah, so this is primarily a concern when using rebase merges, as the other merge types will roll the changes to main up under a single commit which can't be partially stepped into.

This is pretty interesting, because I imagine that a lot of projects don't care about the intermediate states of PRs (because people as a whole aren't great at rewriting branch history to remove the vulnerable commits), but it _would_ be important if rebase merges are allowed.

gregfurman:
[...]

PR #6380, opened 2026-04-15
feat(engine): expose structured commit metadata for pull request evaluation

sachin9058:
# Summary

This PR builds on the existing commit iteration support by introducing structured commit metadata extraction for pull request evaluation.

Previously, commit data was accessed directly from provider-specific objects within the evaluation flow. This change introduces a lightweight abstraction (`CommitInfo`) to normalize commonly used fields such as SHA, message (first line), and author.

The implementation:
- Adds a new `commitinfo` package to avoid import cycles and provide a reusable abstraction
- Introduces a `CommitInfo` struct for normalized commit metadata
- Adds a helper function to extract structured data from GitHub commit objects
- Updates commit logging to use structured metadata instead of raw commit fields

This change is intentionally scoped as a foundational improvement and does not introduce any commit-level policy enforcement.

# Testing

- Verified commit metadata extraction across multiple commits
- Confirmed no changes to existing behavior
- Ran full test suite:
  go test ./...
- Ensured no import cycles or build issues after refactor

### DIFF
--- a/internal/engine/commitinfo/commitinfo.go
+++ b/internal/engine/commitinfo/commitinfo.go
@@ -1,0 +1,45 @@
+// Package commitinfo provides utilities for extracting and normalizing
+// pull request commit metadata into a provider-agnostic structure.
+//
+// It is designed to decouple commit data handling from provider-specific
+// implementations (e.g., GitHub) and avoid import cycles between engine
+// components.
+package commitinfo
+
+import (
+	"strings"
+
+	"github.com/google/go-github/v63/github"
+)
+
+// CommitInfo represents normalized pull request commit metadata.
+type CommitInfo struct {
+	SHA     string
+	Message string
+	Author  string
+}
+
+// Extract normalizes a GitHub commit into a CommitInfo struct.
+func Extract(c *github.RepositoryCommit) CommitInfo {
+	msg := ""
+	if c.GetCommit() != nil {
+		msg = c.GetCommit().GetMessage()
+	}
+
+	firstLine := msg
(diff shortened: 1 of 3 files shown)


sachin9058:
This follows up on the earlier commit iteration work and prepares the system for future commit-level policy checks.

sachin9058:
This test appears to be timing-sensitive. It passes consistently locally and may be flaky due to the short wait duration.

evankanderson:
Question: do we have a high-level design for what's being built, and how it will be used?

Given that this probably requires exposing some number of functions to e.g. the Rego evaluator, it would be handy to follow this approach:

1. Update the issue with the shape of the proposed solution.  This should look like:

   a. A problem statement (we aim to address the following types of rules...)

   b. Background that's helpful in understanding the problem

   c. An outline of the full solution (e.g. "with the following 3 primitives, we can cover all but one use case")

      1. This should include some detail on what capabilities _will not_ be available, and why.  Engineering is about tradeoffs, so it's fine to rule out functionality that might be too expensive -- for example, the functions might fail for branches with more than a certain number of commits of difference.
      1. Include at least one example of what usage will look like

   d. (optional) were any alternatives considered? If so, what made this solution preferred?

2. Once the proposed solution is accepted, break out the implementation into sub-tasks (e.g. using the "create sub-issue" button).
3. Try to line up PRs with sub issues; it doesn't need to be 1:1, but having a map of

sachin9058:
@evankanderson Thanks for the suggestion  I’ve added a proposed high-level design in #2176 outlining how the pieces fit together and example usage.

Happy to iterate on the design before proceeding further.

PR #6383, opened 2026-04-17
fix: prevent panic when Trusty API call fails mid-flight

alimx07:
## Summary
Fixes a latent panic in `getDependencyScore` that could crash the process under Trusty API errors.

The original implementation fired 4 concurrent API calls and coordinated them via channels. The errors channel was unbuffered, and all channels were closed in a defer on function return. When any call returned an error as example, the function already returned early which triggered the defer and closed all channels while the remaining 3 goroutines were still blocked trying to send to the (now-closed) errors channel. Sending to a closed channel in Go causes an **unrecovered panic**, which crashes the entire server process.

This is a realistic failure mode: the Trusty API is an external HTTP service subject to network errors, timeouts, and partial outages.

**Fix**: Replace the channel fan-out with [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) as `errgroup.WithContext` guarantees:

- When one call fails, errgroup cancels gctx, signalling the other in-flight HTTP calls to abort rather than run to completion unnecessarily
- g.Wait() blocks until all 4 goroutines finish, so the function never returns while any goroutine is still running, eliminating the send-after-return entirely

Also , I removed the per-response buffered channels (summary, metadata, alternatives, provenance) and the unbuffered errors channel, results are now written directly to pre-declared variables, read safely after g.Wait() guarantees all writes are done

## Testing
- go test ./internal/engine/eval/trusty/... -race : passes, no data races
- `TestGetDependencyScore` directly reproduces the failure scenario: Summary errors while the other 3 goroutines are blocked mid-flight; confirms error propagation and no panic.

### DIFF
--- a/internal/engine/eval/trusty/trusty.go
+++ b/internal/engine/eval/trusty/trusty.go
@@ -14,6 +14,7 @@ import (
 	"github.com/rs/zerolog"
 	trusty "github.com/stacklok/trusty-sdk-go/pkg/v2/client"
 	trustytypes "github.com/stacklok/trusty-sdk-go/pkg/v2/types"
+	"golang.org/x/sync/errgroup"
 	"google.golang.org/protobuf/reflect/protoreflect"
 
 	"github.com/mindersec/minder/internal/constants"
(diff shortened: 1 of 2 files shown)


evankanderson (approved):
The Trusty API has been closed for approximately the last year; we should retire this evaluator.

I'm willing to accept this PR (this could also have swapped the push to `errors` and the data channels for the same effect), but I don't expect it to have much current utility.

PR #6386, opened 2026-04-20
fix(cli): address review feedback and stabilize ruletype test suite

DharunMR:
# Summary

This PR addresses the review comments on [#6286 ](https://github.com/mindersec/minder/pull/6286) and refactors the `ruletype` CLI commands to properly support the isolated, deterministic test suite.

### Changes Made

- **Removed Test Execution Wrapper (execfunc):** Refactored the CLI test harness (internal/util/cli/testing.go) to remove the custom execfunc wrapper, transitioning instead to Cobra's native cmd.ExecuteContextC(ctx).

- **Impact:** Tests now accurately simulate real-world CLI execution by triggering the complete Cobra command lifecycle (including Args validation and PreRunE configuration binding). This eliminates the "bypassed mock" issue, ensures Viper contexts are properly isolated, and allows the command definitions to follow standard CLI design patterns without breaking the test suite.

- **Improved Table:** Fixed the unexpected spaces and improved table output

### DIFF
--- a/cmd/cli/app/ruletype/ruletype.go
+++ b/cmd/cli/app/ruletype/ruletype.go
@@ -25,23 +25,28 @@ var ruleTypeCmd = &cobra.Command{
 
 // getRuleTypeClient returns the RuleTypeServiceClient, a cleanup function to close the connection and an error
 func getRuleTypeClient(cmd *cobra.Command) (minderv1.RuleTypeServiceClient, func(), error) {
-	ctx := cmd.Context()
+	ctx, cancel := cli.GetAppContext(cmd.Context(), viper.GetViper())
+	cmd.SetContext(ctx)
 
 	// Check the backpack. Are we running inside a test?
 	if mockClient, ok := cli.GetRPCClient[minderv1.RuleTypeServiceClient](ctx); ok {
-		return mockClient, func() {}, nil
+		return mockClient, func() { cancel() }, nil
 	}
 
 	// The backpack is empty. We are in production, make a real connection.
 	conn, err := cli.GrpcForCommand(cmd, viper.GetViper())
 	if err != nil {
+		cancel()
 		return nil, nil, err
 	}
 
 	client := minderv1.NewRuleTypeServiceClient(conn)
 
 	// Return the client and the closer so the subcommand can manage the lifecycle
-	return client, func() { _ = conn.Close() }, nil
+	return client, func() {
+		cancel()
+		_ = conn.Close()
+	}, nil
 }
 
 func init() {
--- a/cmd/cli/app/ruletype/ruletype.go
+++ b/cmd/cli/app/ruletype/ruletype.go
@@ -5,10 +5,8 @@
 package ruletype
 
 import (
-	"context"
-
 	"github.com/spf13/cobra"
-	"google.golang.org/grpc"
+	"github.com/spf13/viper"
 
 	"github.com/mindersec/minder/cmd/cli/app"
 	"github.com/mindersec/minder/internal/util/cli"
@@ -25,14 +23,25 @@ var ruleTypeCmd = &cobra.Command{
 	},
 }
 
-func getRuleTypeClient(ctx context.Context, conn grpc.ClientConnInterface) minderv1.RuleTypeServiceClient {
-	// 1. Check the backpack. Are we running inside a test?
+// getRuleTypeClient returns the RuleTypeServiceClient, a cleanup function to close the connection and an error
+func getRuleTypeClient(cmd *cobra.Command) (minderv1.RuleTypeServiceClient, func(), error) {
+	ctx := cmd.Context()
+
+	// Check the backpack. Are we running inside a test?
 	if mockClient, ok := cli.GetRPCClient[minderv1.RuleTypeServiceClient](ctx); ok {
-		return mockClient
+		return mockClient, func() {}, nil
 	}
 
-	// 2. The backpack is empty. We are in production, make a real connection.
-	return minderv1.NewRuleTypeServiceClient(conn)
+	// The backpack is empty. We are in production, make a real connection.
+	conn, err := cli.GrpcForCommand(cmd, viper.GetViper())
+	if err != nil {
+		return nil, nil, err
+	}
+
+	client := minderv1.NewRuleTypeServiceClient(conn)
+
+	// Return the client and the closer so the subcommand can manage the lifecycle
+	return client, func() { _ = conn.Close() }, nil
 }
 
 func init() {
--- a/cmd/cli/app/ruletype/ruletype_apply.go
+++ b/cmd/cli/app/ruletype/ruletype_apply.go
@@ -107,8 +107,7 @@ func applyCommand(cmd *cobra.Command, args []string) error {
 		if f.Path != "-" && shouldSkipFile(f.Path) {
 			continue
 		}
-		ctx, cancel := cli.GetAppContext(cmd.Context(), viper.GetViper())
-		if err = execOnOneRuleType(ctx, table, f.Path, os.Stdin, project, applyFunc); err != nil {
+		if err = execOnOneRuleType(cmd.Context(), table, f.Path, os.Stdin, project, applyFunc); err != nil {
 			if f.Expanded && minderv1.YouMayHaveTheWrongResource(err) {
 				cmd.PrintErrf("Skipping file %s: not a rule type\n", f.Path)
 				// We'll skip the file if it's not a rule type
@@ -116,7 +115,6 @@ func applyCommand(cmd *cobra.Command, args []string) error {
 			}
 			return cli.MessageAndError(fmt.Sprintf("error applying rule type from %s", f.Path), err)
 		}
-		cancel()
 	}
 	// Render the table
 	table.Render()
--- a/cmd/cli/app/ruletype/ruletype_apply.go
+++ b/cmd/cli/app/ruletype/ruletype_apply.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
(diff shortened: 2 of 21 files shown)


cmd/cli/app/ruletype/testdata/apply_create.table.golden
evankanderson, line 1: This output looks much nicer, thank you!

cmd/cli/app/ruletype/testdata/get_by_id.table.golden
```diff
@@ -22,10 +22,9 @@
                    │ can use secret scanning to prevent supported secrets from being pushed into    
                    │ your repository by enabling secret scanning push protection. For more          
                    │ information, see GitHub's documentation                                        
-                   │ https://docs.github.com/en/code-security/secret-                               
-                   │ scanning/push-protection-for-repositories-and-                                 
-                   │ organizations#enabling-secret-scanning-as-a-push- protection-for-              
-                   │ a-repository.                                                                  
+                   │ https://docs.github.com/en/code-security/secret-scanning/push-protection-for-r 
+                   │ epositories-and-organizations#enabling-secret-scanning-as-a-push-protection-fo 
+                   │ r-a-repository.
```
evankanderson, line 27: I don't have a good solution, but it's a shame that this link won't be clickable in the terminal.

cmd/cli/app/ruletype/ruletype.go
```diff
@@ -25,14 +23,26 @@ var ruleTypeCmd = &cobra.Command{
 	},
 }
 
-func getRuleTypeClient(ctx context.Context, conn grpc.ClientConnInterface) minderv1.RuleTypeServiceClient {
+func getRuleTypeClient(cmd *cobra.Command) (minderv1.RuleTypeServiceClient, func(), error) {
```
evankanderson, line 27: Can you add a comment about the type of the second argument?  You can either use a typedef to create a "Closer" or "Cleanup" interface, or simply put a comment above the function.

cmd/cli/app/ruletype/ruletype_apply.go
```diff
@@ -23,89 +22,103 @@ var applyCmd = &cobra.Command{
...
-		return cli.MessageAndError("Error validating files", err)
-	}
+		return nil
+	},
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// No longer print usage on returned error, since we've parsed our inputs
+		// See https://github.com/spf13/cobra/issues/340#issuecomment-374617413
+		cmd.SilenceUsage = true
 
-	files, err := util.ExpandFileArgs(allFiles...)
```
evankanderson, line 51: We used to expand all the files before turning off the usage message; now we do that later.

(We also don't print usage if we can't create the connection to the server.)

cmd/cli/app/ruletype/ruletype_apply.go
```diff
@@ -23,89 +22,103 @@ var applyCmd = &cobra.Command{
 	Use:   "apply [files...]",
 	Short: "Apply a rule type",
 	Long:  `The ruletype apply subcommand lets you create or update rule types for a project within Minder.`,
-	RunE:  cli.GRPCClientWrapRunE(applyCommand),
-	Args:  cobra.ArbitraryArgs,
-}
```
evankanderson, line 28: This switch from assigning to `RunE` and `Args` to using in-line functions is a bit much for GitHub diff to handle.  Do you mind keeping the `applyCommand`, and simply writing e.g.:

```golang
	RunE:  applyCommand,
```

evankanderson (commented):
While the switch to anonymous functions in `RunE` might be a nice change in the long run, I think the diff will be easier to read if we keep the per-command functions and simply strip the `conn *grpc.ClientConn` argument for now.

cmd/cli/app/ruletype/testdata/get_by_id.table.golden
DharunMR: yea i thought same, i tried fixing it but because of the tablelines i coundnt fix it .Maybe using hyperlink a solution?

evankanderson: Not sure how to do a hyperlink in terminal output.  😁 

I think this is fine for now; having the golden file makes it easier to see these problems, even if it's not obvious how to fix them.

cmd/cli/app/ruletype/ruletype_apply.go
```diff
@@ -92,20 +107,21 @@ func applyCommand(ctx context.Context, cmd *cobra.Command, args []string, conn *
...
 		}
-		// cmd.Context() is the root context. We need to create a new context for each file
-		// so we can avoid the timeout.
-		if err = execOnOneRuleType(cmd.Context(), table, f.Path, os.Stdin, project, applyFunc); err != nil {
+		ctx, cancel := cli.GetAppContext(cmd.Context(), viper.GetViper())
+		if err = execOnOneRuleType(ctx, table, f.Path, os.Stdin, project, applyFunc); err != nil {
 			if f.Expanded && minderv1.YouMayHaveTheWrongResource(err) {
 				cmd.PrintErrf("Skipping file %s: not a rule type\n", f.Path)
 				// We'll skip the file if it's not a rule type
 				continue
```
evankanderson, line 114: We miss calling `cancel()` here when we `continue` the for loop.  You may want to adjust `execOnOneRuleType` to call `cli.GetAppContext()`, so that you can use `defer` on the returned cancel.

cmd/cli/app/ruletype/ruletype_create.go
```diff
@@ -22,23 +21,31 @@ var createCmd = &cobra.Command{
...
-	client := getRuleTypeClient(ctx, conn)
+		fileFlag, err := cmd.Flags().GetStringArray("file")
+		if err != nil {
+			return cli.MessageAndError("Error parsing file flag", err)
+		}
 
-	project := viper.GetString("project")
+		if len(fileFlag) == 0 {
+			return fmt.Errorf("required flag(s) \"file\" not set")
+		}
```
evankanderson: We already marked `file` required, right?

cmd/cli/app/ruletype/ruletype_create.go
```diff
@@ -49,6 +56,17 @@ func createCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *gr
...
 	cmd.SilenceUsage = true
 
+	client, closeConn, err := getRuleTypeClient(cmd)
+	if err != nil {
+		return cli.MessageAndError("Error connecting to server", err)
+	}
+	defer closeConn()
+
+	ctx, cancel := cli.GetAppContext(cmd.Context(), viper.GetViper())
+	defer cancel()
```
evankanderson: We're calling this here because `GRPCClientWrapRunE` used to do it, correct?

evankanderson: If we're feeling a bit naughty, `getRuleTypeClient` could mutate its argument by calling `cli.GetAppContext` and then passing the context into `cmd.SetContext` and rolling the cancel into its cancellation value.

cmd/cli/app/ruletype/ruletype_get.go
```diff
@@ -19,38 +17,61 @@ import (
 	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
 )
 
+type ruleTypeGetter interface {
+	protoreflect.ProtoMessage
+	GetRuleType() *minderv1.RuleType
+}
```
evankanderson, line 23: ❤️  defining these micro-interfaces to simplify the code.

cmd/cli/app/ruletype/ruletype_get.go
```diff
@@ -19,38 +17,61 @@ import (
...
 
-// getCommand is the ruletype get subcommand
-func getCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc.ClientConn) error {
-	client := getRuleTypeClient(ctx, conn)
+		if id == "" && name == "" {
+			return fmt.Errorf("at least one of the flags in the group [id name] is required")
+		}
+		if id != "" && name != "" {
+			return fmt.Errorf("if any flags in the group [id name] are set they must all be set; but mutually exclusive flags were passed")
+		}
```
evankanderson, line 40: Where did this come from?  In particular, I'm confused by "if any flags are set they must all be set" along with "mutually exclusive".

Maybe the wording of the error is off?

internal/util/cli/testing.go
```diff
@@ -19,18 +19,32 @@ import (
...
 	"google.golang.org/protobuf/proto"
-
-	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
-	mockv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1/mock"
 )
 
 var update = flag.Bool("update", false, "update golden files")
 
+// ResetEntireTree wipes flags and contexts so tests don't "bleed" into each other
+func ResetEntireTree(c *cobra.Command) {
```
evankanderson, line 27: Ah, interesting!  I hadn't thought about how calling `viper.BindPFlags(cmd.Flags())` over and over might lead to problems.  I wonder if we should be using / supporting a `New()` viper instance for each cmd.

(This is me thinking -- don't do that in this PR!)

evankanderson (dismissed):
I'm happy to either merge this work as-is (i.e. it's good enough to go in), or to give you a chance to address the cancellations and one confusing error message.

It also seems worth an explanation of redoing the flag-validation in `PreRunE` given that we register various flag constraints in `init` for the specific commands.

Anyway, feel free to either say "commit this" in a comment, or to push another commit.  If you don't comment in a day or two, I'll default to merging this and having you submit a follow-on if you want.

cmd/cli/app/ruletype/ruletype_apply.go
DharunMR: actually when removing wrapper i copy pasted whole thing, i just stripped GetAppContext completely out of the loops and execOnOneRuleType now getRuleTypeClient handles every ctx, its clean now

cmd/cli/app/ruletype/ruletype_create.go
DharunMR: did as you said

cmd/cli/app/ruletype/ruletype_get.go
DharunMR: yes! because i couldnt get the the sentence, but ops i copy pasted same error msg (changed it)

evankanderson (commented):
I have a slight concern that we might hit the 20s default timeout from `cli.GetAppContext` during a large batch of files in one ruletype apply, but I'm not sure we wouldn't hit that error anyway with the current code.

evankanderson (approved):

PR #6406, opened 2026-04-23
Fix SARIF uploads by granting security-events: write

evankanderson:
# Summary

Currently, the upload-codeql action fail to upload sarif results after Trivy produces them, because the GITHUB_TOKEN doesn't have permissions to report the security events.

# Testing

Will verify with this run (and fix any currently-detected issues).

PR #6412, opened 2026-04-23
Add security-events: write to main workflows, granular permissions

evankanderson:
# Summary

#6406 adjusted `security.yml` to grant `security-events: write`, but missed adding these to `main.yml` for post-merge CI runs.  It also granted the access over-broadly, so reduce the permissions scopes in the CI as well.

# Testing

Adjusting the PR workflow to follow the same pattern allows checking that the settings in `main.yml` are likely to work.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,6 @@ on:
       - 'docs/**'
 permissions:
   contents: read
-  packages: write
 
 jobs:
   license-check:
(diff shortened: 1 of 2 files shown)


evankanderson:
(I think Ozz may be out, and this is actually breaking the post-submit workflow)

rdimitrov (approved):

rdimitrov:
> (I think Ozz may be out, and this is actually breaking the post-submit workflow)

Merged 👍

PR #6417, opened 2026-04-24
feat(cli): improve evaluation reasoning output using existing fields

### LINKED ISSUE #6416: Improve visibility and debugging of evaluation results in CLI and documentation
### Problem

Minder already captures a rich set of evaluation data, including:
- allow/violations (via rule logic)
- details and message fields
- structured output (EvalResultOutput)
- remediation and alert metadata

However, this information is not always easy to discover or interpret from the CLI or documentation.

As a result, users may find it difficult to understand:
- why a rule passed or failed
- how violations and messages relate to evaluation logic
- how to access and interpret structured output

This creates a usability gap, especially as policies become more complex (e.g., commit-level evaluation, selectors).

---

### Solution Proposal

Improve how existing evaluation data is surfaced and documented, instead of introducing new engine-level concepts.

This could include:

1. CLI improvements:
   - Present evaluation data in a more readable, reasoning-first format
   - Aggregate fields like:
     - details
     - guidance
     - remediation_details
     - alert metadata
     - structured output
   - Make it easier to inspect rule-level results without relying on raw JSON/YAML output

2. Documentation improvements:
   - Add a dedicated section on “Writing and Debugging Rules”
   - Explain how allow, violations, message, and output map to user-visible results
   - Provide examples using real rules from minder-rules-and-profiles

3. (Optional future work)
   - Introduce a dedicated CLI command (e.g., `minder results` or `minder eval`)
   - Support exporting results (CSV/HTML)
   - Explore capturing additional debug signals (e.g., Rego `print`)

---

- Adding a new “explain mode” at the engine level  
  → Not aligned with current architecture, since evaluation already captures relevant data

- Logging-based debugging  
  → Not user-friendly and not easily accessible from CLI

- External tooling  
  → Not practical since evaluation context is internal to Minder

---

### Additional context

Maintainers noted that:
- evaluation data already exists but is not well documented
- CLI currently does not surface all available information clearly

This work focuses on improving usability and discoverability of existing capabilities.

---

### Acceptance Criteria

- CLI output presents evaluation data in a clearer, reasoning-first format
- Users can understand why a rule passed/failed without inspecting raw JSON
- No changes to evaluation engine behavior
- Documentation explains how evaluation fields map to user-visible results
- At least one concrete example (rule) is used to demonstrate behavior

sachin9058:
# Summary

This PR improves how evaluation results are surfaced in the CLI by making existing evaluation data easier to understand.

Previously, CLI output primarily focused on pass/fail status, while additional context (details, guidance, remediation details, alerts, and structured output) was harder to interpret. This change aggregates and presents that information in a clearer, reasoning-first format.

# Key Changes

- Added a "Reasoning" column to evaluation-related CLI outputs
- Aggregated existing fields into a readable format:
  - details
  - guidance
  - remediation_details
  - alert.details and alert.url
  - output (structured evaluation output)
- Introduced shared helpers for consistent rendering:
  - `FormatEvaluationReasoning`
  - `RuleDisplayName`
- Removed duplication by centralizing reasoning logic across CLI commands
- Improved readability with bullet-formatted output

# Design Rationale

This change builds on the existing evaluation model instead of introducing new engine-level concepts.

- No changes to evaluation logic or engine behavior
- Reuses already available evaluation data
- Focuses on improving visibility and usability in CLI
- Keeps changes incremental and aligned with current architecture

# CLI Changes

Example improvement:

Before:
    Rule: branch_protection_enabled → FAIL

After:
    Rule: branch_protection_enabled
    Result: FAIL
    Reasoning:
    - branch protection is disabled
    - Guidance: enable branch protection
    - Output:
      ...

This makes it easier to understand *why* a rule passed or failed without needing to inspect raw JSON/YAML output.

# Testing

- Ran:
    go test ./cmd/cli/app/profile/... ./cmd/cli/app/artifact ./cmd/cli/app/history
- Verified CLI output formatting for:
  - profile status
  - artifact evaluation status
- Confirmed no changes to evaluation engine behavior

# Notes

This PR focuses on CLI improvements only. It does not introduce any new evaluation fields or modify rule execution.

Future improvements could include:
- a dedicated `results` / `eval` CLI command for deeper inspection
- documentation improvements for rule debugging
- additional formatting options (e.g., CSV/HTML export)

Fixes #6416


evankanderson (commented):
I don't think you need any of the changes in `internal/engine/eval/` and `pkg/engine/v1` for this change -- we already wire up `output`, `status`, `details`, `remediation_details` and `alert.details`.  We don't necessarily surface all of these in the CLI yet, but I'd like to see some more design detail about how explain would differ from the existing information collected (and how it would be used) before making rule engine changes.

sachin9058:
> I don't think you need any of the changes in `internal/engine/eval/` and `pkg/engine/v1` for this change -- we already wire up `output`, `status`, `details`, `remediation_details` and `alert.details`. We don't necessarily surface all of these in the CLI yet, but I'd like to see some more design detail about how explain would differ from the existing information collected (and how it would be used) before making rule engine changes.

@evankanderson Thanks, that’s very helpful.

I see your point about the existing fields (output, status, details, remediation_details, alert.details) already carrying structured information from evaluation.

My intention with “explain” was to surface evaluation reasoning more explicitly (e.g., why a rule passed/failed, intermediate checks like allow/violations), rather than just the final outcome or message.

That said, I agree the current PR doesn’t clearly differentiate this from existing fields, and it likely needs a clearer design before making engine-level changes.

I’ll put together a short design proposal outlining:
- what “explain” represents conceptually
- how it differs from existing fields like output and details
- whether it can reuse existing structures instead of introducing new ones
- how it would be consumed (CLI / API)

I can share that here for feedback before continuing implementation.

sachin9058:
@evankanderson Following up with a quick design proposal:

## Goal

“Explain” is intended to expose evaluation reasoning, not just final results.

This includes:
- allow/deny decisions
- constraint violations
- intermediate evaluation checks

## Current State

We already have:
- `output`: user-facing result (message / structured output)
- `details`, `remediation_details`, `alert.details`: structured metadata

These represent results, but not necessarily the reasoning behind them.

## Proposed Direction

Instead of introducing a new engine-level field, “explain” can be implemented as an extension of the existing evaluation model.

1. Reuse existing structures  
   - Leverage `output` and/or `details` to carry additional reasoning data  
   - Avoid adding a new top-level field unless a clear gap is identified  
   - Keep compatibility with existing APIs and storage

2. Evaluator-level enrichment  
   - When explain mode is enabled, evaluators (e.g., deny-by-default, constraints) can include additional context such as:
     - evaluation decisions (e.g., allowed: true/false)
     - matched conditions or violations
     - key inputs used in the decision
   - When explain is disabled, behavior remains unchanged (no extra data populated)

3. Minimal engine involvement  
   - The engine only needs to propagate an “explain enabled” signal (e.g., via context)
   - No structural changes to evaluation result types are required initially

4. CLI and API consumption  
   - CLI can surface this additional data when `--explain` is used
   - Initially, this can reuse existing output rendering paths
   - Future iterations can improve formatting or explicitly surface structured reasoning

This approach keeps the change incremental:
- avoids introducing new concepts prematurely  
- builds on existing fields  
- allows gradual refinement based on usage and feedback
## Usage

Explain would be a debugging/visibility feature:

    minder <command> --explain

Used to:
- understand why a rule passed/failed
- debug policies
- inspect evaluator behavior


Happy to refine this based on what aligns best with the current model.

evankanderson:
> 2. Evaluator-level enrichment
>    
>    * When explain mode is enabled, evaluators (e.g., deny-by-default, constraints) can include additional context such as:
>      
>      * evaluation decisions (e.g., allowed: true/false)
>      * matched conditions or violations
>      * key inputs used in the decision
>    * When explain is disabled, behavior remains unchanged (no extra data populated)

I'm trying to understand how this part works.  The rest mostly makes sense, but how does the `explain` context setting change the rule execution (if at all)?  **What does it record that isn't previously recorded?**  It might be easiest to pick an example rule from the existing rules to use as an example.

The rest of what you're proposing mostly makes sense, except for one detail:

> 1. When explain mode is used ...
> 2. CLI can surface this additional data when `--explain` is used

Rule evaluation takes place independently of the CLI ever asking for a report on what happened.  In many cases (success and failure) the main user interaction that takes place happens outside of Minder -- for example, https://github.com/mindersec/minder/pull/6184 happened entirely in the GitHub UI, and I never looked at the rule status after loading it.  (Now, I'd tested it elsewhere before enabling it, but we can't know who's debugging rules and who is just consuming them in the current architecture.)

sachin9058:
@evankanderson thanks for calling this out, this helps clarify the gap.

You're right that evaluators already compute and expose the core reasoning signals (allow/violations, message, output), and there isn't really additional hidden state to capture at execution time.

Given that, I think my earlier framing of “explain mode” as something that changes rule execution isn’t the right direction.

Instead, it seems the real gap is:
- how this existing reasoning data is surfaced and understood
- especially across CLI, RPC responses, and documentation

So rather than adding evaluator-level changes, a better approach would be:
- improving how existing fields (output, details, EvalResultOutput) are exposed
- making it clearer how to interpret allow/violations/message/output in practice
- potentially improving CLI output to present this in a more “explain-like” way

I’ll pivot this PR in that direction and remove the evaluator changes.

For example, I can:
- take an existing rule from minder-rules-and-profiles
- show how its allow/violations/message/output map to what a user sees
- and improve CLI output / documentation around that

Let me know if that direction aligns better.

evankanderson:
> @evankanderson thanks for calling this out, this helps clarify the gap.
> 
> You're right that evaluators already compute and expose the core reasoning signals (allow/violations, message, output), and there isn't really additional hidden state to capture at execution time.

There is one tool that we don't currently support (but I've considered adding) -- Rego has a built-in `print` function that we don't wire up or collect today, except in the `mindev` command.  We could store that alongside the existing structured output to assist with debugging.  Note that it would be always-on, rather than a "mode" that you could use for one invocation.

> Given that, I think my earlier framing of “explain mode” as something that changes rule execution isn’t the right direction.

I think it's a really interesting idea -- unfortunately, it doesn't quite match the model (background execution on event, rather than triggered by users) and primitives (Rego, JQ evaluators) we currently have.

> Instead, it seems the real gap is:
> 
> * how this existing reasoning data is surfaced and understood

FWIW, I wouldn't describe this as "reasoning data" -- this is basically intermediate state.  If you're looking for equivalent / conceptually-similar notions, tools like Rookout and Lightrun offer stopless debugging -- there are also approaches that save enough state to allow "time-travel" or backwards jumps in execution.  I'm not sure to what extent our current tools support that, though.

> * especially across CLI, RPC responses, and documentation

Yes, this is a big gap!  And writing documentation seems to be harder than writing code.  😁 

Our current docs are somewhat organized, but I think a better section focused on writing and debugging rules would probably be helpful.  I generally like the https://diataxis.fr/ model for writing documentation; if you haven't written much documentation before, a brief read through that site may be helpful.

> So rather than adding evaluator-level changes, a better approach would be:
> 
> * improving how existing fields (output, details, EvalResultOutput) are exposed

In particular, the CLI currently doesn't make much use of EvalResultService and prefers to use ProfileService with some extra flags.  I'd be open to a new top-level `results` (or some other useful word, like `eval`) command for viewing and analyzing results.  CSV export might also be a useful feature there -- or HTML with an embedded webserver, though I'm also working on a full HTML UI, so that may not be as helpful.

> * making it clearer how to interpret allow/violations/message/output in practice

Yes, and documenting patterns for "what goes where".  We have a bunch of rules that stuff things into `violations` in order to have nice messages where there might be better ways to do it, or ways to use `message` in a more consistent and conventional way.

> * potentially improving CLI output to present this in a more “explain-like” way

Yes, I think being able to zoom in on a particular rule + entity and get a full readout of what happened without resorting to `-o yaml` and scrolling around could be handy.

> I’ll pivot this PR in that direction and remove the evaluator changes.

👍 

> For example, I can:
> 
> * take an existing rule from minder-rules-and-profiles
> * show how its allow/violations/message/output map to what a user sees
> * and improve CLI output / documentation around that

That sounds like a great plan -- and will give you something concrete to poke at.  If you find rules that don't seem to behave correctly, feel free to bring them to the `#minder` OpenSSF Slack channel with questions and I'll be happy to help out.

sachin9058:
@evankanderson Thanks, this is super helpful  I really appreciate the detailed guidance.

I’ll update this PR to remove the evaluator/engine changes and pivot it toward:

- improving how existing evaluation data is surfaced
- exploring a more focused CLI experience (e.g., a results/eval command or improved output)
- and documenting how allow/violations/message/output map to user-visible results

I’ll start by picking an example rule from minder-rules-and-profiles and building from there.

Will push an update shortly.

sachin9058:
@evankanderson Thanks for the guidance — I’ve updated the PR to align with your feedback.

- removed the earlier engine-level changes
- focused on improving how existing evaluation data is surfaced in the CLI
- centralized reasoning formatting and applied it consistently

This now only builds on existing fields (details, guidance, remediation, alerts, output) without changing evaluation behavior.

Let me know if this direction looks good or if you'd like any further adjustments.

evankanderson (commented):
Can you include some screenshots / extracts of how this renders?  I'm concerned that information density and formatting may be difficult to pull off in the table view.

sachin9058:
@evankanderson Good point .. I extracted the actual rendering from the reasoning helper used by the CLI.

Here’s the real output produced:



Screenshot from 2026-04-25 11-13-59





This is exactly what gets rendered inside the “Reasoning” column, so multi-line values appear stacked within that cell.

To keep it readable, I:
- use bullet prefixes
- group related fields (details, guidance, remediation, alerts, output)
- indent structured output

One small thing I noticed is that the output value is currently wrapped in quotes (coming from the structpb value formatting). I can clean that up in a follow-up if needed.
This is exactly what gets rendered inside the “Reasoning” column, so multi-line values appear stacked within that cell.

I don’t currently have a local backend running to capture a full end-to-end CLI screenshot, but this output is directly produced by the same helper used in table rendering.

Happy to adjust formatting further (e.g., truncation or alternative layouts) if this feels too dense in table view.

cmd/cli/app/artifact/artifact_get.go
```diff
@@ -193,13 +194,16 @@ func printEvalStatus(
...
-			[]string{"Profile", "Rule", "Status", "Message"})
+			[]string{"Profile", "Rule", "Result", "Reasoning"})
 		for _, status := range evalStatus {
+			ruleName := profile.RuleDisplayName(status)
+			reasoning := profile.FormatEvaluationReasoning(status)
+
 			ta.AddRow(
 				status.ProfileId,
-				status.RuleTypeName,
+				fmt.Sprintf("%s\n[%s]", ruleName, status.GetRuleTypeName()),
```
DharunMR: use cmd.Printf or wrap it inside cmd.Println so that it will write in cobra's output buffer

cmd/cli/app/profile/table_render.go
```diff
@@ -32,6 +32,19 @@ func marshalStructOrEmpty(v *structpb.Struct) string {
...
 
+func marshalValueOrEmpty(v *structpb.Value) string {
+	if v == nil {
+		return ""
+	}
+
+	out, err := yaml.Marshal(v.AsInterface())
+	if err != nil {
+		return ""
+	}
```
evankanderson: I think `util.GetYamlFromProto` should work here without needing nil checks.  It does return `{}` (empty object) for a nil proto message, but that seems okay.

cmd/cli/app/profile/table_render.go
```diff
@@ -94,11 +107,64 @@ func RenderProfileStatusTable(ps *minderv1.ProfileStatus, t table.Table, emoji b
...
+	if output := marshalValueOrEmpty(eval.GetOutput()); output != "" {
+		parts = append(parts, fmt.Sprintf("- Output:\n%s", output))
+	}
+
+	return strings.Join(parts, "\n")
+}
+
+// RuleDisplayName returns the most user-friendly rule name available.
+func RuleDisplayName(eval *minderv1.RuleEvaluationStatus) string {
+	return firstNonEmpty(eval.GetRuleDisplayName(), eval.GetRuleDescriptionName(), eval.GetRuleTypeName())
```
evankanderson: You probably want to prefer `rule_description_name` (the `name` from the profile) if set over `rule_display_name` (the human-readable field in the rule type); it's possible to have one profile which includes the same ruletype with different parameters, and they'll only be distinguished by the `rule_description_name`.

evankanderson: Also, both `rule.name` and `display_name` have validation which ensures that the first character is non-whitespace, so you can probably just use `cmp.Or` here:


```suggestion
	return strings.TrimSpace(cmp.Or(eval.GetRuleDescriptionName(), eval.GetRuleDisplayName(), eval.GetRuleTypeName()))
```

cmd/cli/app/profile/table_render.go
```diff
@@ -94,11 +107,64 @@ func RenderProfileStatusTable(ps *minderv1.ProfileStatus, t table.Table, emoji b
...
+}
+
+// FormatEvaluationReasoning returns a rendered reasoning block for CLI output.
+func FormatEvaluationReasoning(eval *minderv1.RuleEvaluationStatus) string {
+	reasoning := formatReasoning(eval)
+	if reasoning == "" {
+		return "-"
+	}
+
+	return reasoning
```
evankanderson: Again, you could use `cmp.Or` here:


```suggestion
	return cmp.Or(formatReasoning(eval), "-")
```

cmd/cli/app/profile/table_render.go
```diff
@@ -114,11 +180,13 @@ func RenderRuleEvaluationStatusTable(
...
 	for _, eval := range statuses {
 		evalInfo := types.RuleEvalStatus(eval)
+		ruleName := RuleDisplayName(eval)
+		reasoning := FormatEvaluationReasoning(eval)
 		t.AddRowWithColor(
 			layouts.NoColor(fmt.Sprintf("%s\n[%s]", eval.EntityInfo["name"], eval.Entity)),
-			layouts.NoColor(fmt.Sprintf("%s\n[%s]", eval.RuleDescriptionName, eval.RuleTypeName)),
+			layouts.NoColor(fmt.Sprintf("%s\n[%s]", ruleName, eval.GetRuleTypeName())),
 			table.GetStatusIcon(evalInfo, emoji),
-			table.BestDetail(evalInfo),
```
evankanderson, line 121: BestDetail was previously selecting (preferentially):

1. `remediation_detail` if it was set
1. `alert_detail` if not
1. `status_detail` as a default if neither was set.

I'm presuming this was for more-compact output; for example, the `security-baseline-level-1` profile has about 20 ruletypes, and registering it for 4 repos can produce 80 rows of evaluation output.

cmd/cli/app/profile/table_render.go
```diff
@@ -94,11 +107,64 @@ func RenderProfileStatusTable(ps *minderv1.ProfileStatus, t table.Table, emoji b
 // NewRuleEvaluationsTable creates a new table for rendering rule evaluations
 func NewRuleEvaluationsTable(out io.Writer) table.Table {
 	return table.New(table.Simple, layouts.Default, out,
-		[]string{"Entity", "Rule Name", "Status", "Details"}).
+		[]string{"Entity", "Rule", "Result", "Reasoning"}).
```
evankanderson: I'm trying to decide whether my reaction to renaming the status and details is knee-jerk resistance to change.  😁 

"Status" represents an ongoing state, while "result" focuses on the result of the most recent evaluation.  I can see the argument that users are interested in what happened most recently, though I'd also like to push them towards thinking about Minder as a continuous process, rather than something explicitly triggered.  I think I can go either way on this rename.

"Reasoning" gives me a bit more pause when compared with "details".  "Reasoning" suggests that the process may be stochastic (e.g. like an LLM), rather than a deterministic process.  Additionally, since we're presenting outputs defined by the rule author, they may or may not actually look like "reasoning" to the viewer.

Looking at the [dependabot rule](https://github.com/custcodian/minder-rules-and-profiles/blob/main/rule-types/github/sca_dependabot.yaml) given a particular repo, it looks like the "reasoning" would look like:

```
- [{"dirs":["/rule-types/common/python_lock_file_exists.testdata/no_lock/","/rule-types/common/python_lock_file_exists.testdata/pdm_lock/","/rule-types/common/python_lock_file_exists.testdata/pinned_requirements_txt/","/rule-types/common/python_lock_file_exists.testdata/poetry_lock/","/rule-types/common/python_lock_file_exists.testdata/some_unpinned_requirements_txt/"],"ecosystem":"pip"},{"dirs":["/rule-types/github/sca_dependabot.testdata/all_ecosystems/","/rule-types/github/sca_dependabot.testdata/no_dependabot_yml/","/rule-types/github/sca_dependabot.testdata/some_ecosystems/frontend/","/security-baseline/rule-types/github/osps-qa-02-01.testdata/package_json_no_lock/","/security-baseline/rule-types/github/osps-qa-02-01.testdata/package_json_with_package_lock/","/security-baseline/rule-types/github/osps-qa-02-01.testdata/package_json_with_yarn_lock/"],"ecosystem":"npm"},{"dirs":["/rule-types/github/sca_dependabot.testdata/some_ecosystems/"],"ecosystem":"maven"},{"dirs":["/security-baseline/rule-types/github/osps-qa-02-01.testdata/gemfile_no_lock/","/security-baseline/rule-types/github/osps-qa-02-01.testdata/gemfile_with_lock/"],"ecosystem":"bundler"}]
- Guidance: Ensure that Dependabot is configured and enabled for all language
ecosystems used in the repository.

Dependabot enables Automated dependency updates for repositories.
It is recommended that repositories have some form of automated
dependency updates enabled to ensure that vulnerabilities are not
introduced into the codebase.

For more information, see [GitHub's
documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file).
- Output:
- dirs:
    - /rule-types/common/python_lock_file_exists.testdata/no_lock/
    - /rule-types/common/python_lock_file_exists.testdata/pdm_lock/
    - /rule-types/common/python_lock_file_exists.testdata/pinned_requirements_txt/
    - /rule-types/common/python_lock_file_exists.testdata/poetry_lock/
    - /rule-types/common/python_lock_file_exists.testdata/some_unpinned_requirements_txt/
  ecosystem: pip
- dirs:
    - /rule-types/github/sca_dependabot.testdata/all_ecosystems/
    - /rule-types/github/sca_dependabot.testdata/no_dependabot_yml/
    - /rule-types/github/sca_dependabot.testdata/some_ecosystems/frontend/
    - /security-baseline/rule-types/github/osps-qa-02-01.testdata/package_json_no_lock/
    - /security-baseline/rule-types/github/osps-qa-02-01.testdata/package_json_with_package_lock/
    - /security-baseline/rule-types/github/osps-qa-02-01.testdata/package_json_with_yarn_lock/
  ecosystem: npm
- dirs:
    - /rule-types/github/sca_dependabot.testdata/some_ecosystems/
  ecosystem: maven
- dirs:
    - /security-baseline/rule-types/github/osps-qa-02-01.testdata/gemfile_no_lock/
    - /security-baseline/rule-types/github/osps-qa-02-01.testdata/gemfile_with_lock/
  ecosystem: bundler
```

Part of this ugliness is that `output` is derived from `violations` if `violation_format: json` is set in the rule type (from before we had an explicit `output` field in the Rego evaluation), but even without, it seems like we might need a different (detailed) command to be able to make this digestable.

Running `profile status list --detailed` on this might produce interesting results; the 0.1.2 release produces:

```
╭──────────────────────────────────────┬───────────────────────────────────────────────────────┬────────┬──────────────────────────────────────╮
│ Entity                               │ Rule Name                                             │ Status │ Details                              │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/.github                   │ Enable Dependabot SCA                                 │ ✅     │                                      │
│ [repository]                         │ [sca_dependabot]                                      │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/.github                   │ Require pull requests for default branch changes      │ 🔧     │ Changes can be made to the default   │
│ [repository]                         │ [default_branch_pr_required]                          │        │ branch without a pull request for    │
│                                      │                                                       │        │ custcodian/.github                   │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/.github                   │ Ensure immutable version of GitHub action             │ ✅     │                                      │
│ [repository]                         │ [actions_check_pinned_tags]                           │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/.github                   │ Enable Semgrep SAST                                   │ ✅     │                                      │
│ [repository]                         │ [sast_semgrep]                                        │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/.github                   │ Release tags should not be updated after creation     │ ✅     │                                      │
│ [repository]                         │ [release_tags_immutable]                              │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/.github                   │ Ensure GitHub Actions workflows set their permissions │ ✅     │                                      │
│ [repository]                         │ [actions_check_default_permissions]                   │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/minder-rules-and-profiles │ Ensure immutable version of GitHub action             │ ✅     │                                      │
│ [repository]                         │ [actions_check_pinned_tags]                           │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/minder-rules-and-profiles │ Ensure GitHub Actions workflows set their permissions │ ✅     │                                      │
│ [repository]                         │ [actions_check_default_permissions]                   │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/minder-rules-and-profiles │ Enable Dependabot SCA                                 │ ⛔     │ [{"dirs":["/rule-                    │
│ [repository]                         │ [sca_dependabot]                                      │        │ types/common/python_lock_file_exists │
│                                      │                                                       │        │ .testdata/no_lock/","/rule-          │
│                                      │                                                       │        │ types/common/python_lock_file_exists │
│                                      │                                                       │        │ .testdata/pdm_lock/","/rule-         │
│                                      │                                                       │        │ types/common/python_lock_file_exists │
│                                      │                                                       │        │ .testdata/pinned_requirements_txt/", │
│                                      │                                                       │        │ "/rule-                              │
│                                      │                                                       │        │ types/common/python_lock_file_exists │
│                                      │                                                       │        │ .testdata/poetry_lock/","/rule-      │
│                                      │                                                       │        │ types/common/python_lock_file_exists │
│                                      │                                                       │        │ .testdata/some_unpinned_requirements │
│                                      │                                                       │        │ _txt/"],"ecosystem":"pip"},{"dirs":[ │
│                                      │                                                       │        │ "/rule-                              │
│                                      │                                                       │        │ types/github/sca_dependabot.testdata │
│                                      │                                                       │        │ /all_ecosystems/","/rule-            │
│                                      │                                                       │        │ types/github/sca_dependabot.testdata │
│                                      │                                                       │        │ /no_dependabot_yml/","/rule-         │
│                                      │                                                       │        │ types/github/sca_dependabot.testdata │
│                                      │                                                       │        │ /some_ecosystems/frontend/","/securi │
│                                      │                                                       │        │ ty-baseline/rule-types/github/osps-  │
│                                      │                                                       │        │ qa-02-                               │
│                                      │                                                       │        │ 01.testdata/package_json_no_lock/"," │
│                                      │                                                       │        │ /security-baseline/rule-             │
│                                      │                                                       │        │ types/github/osps-qa-02-             │
│                                      │                                                       │        │ 01.testdata/package_json_with_packag │
│                                      │                                                       │        │ e_lock/","/security-baseline/rule-   │
│                                      │                                                       │        │ types/github/osps-qa-02-             │
│                                      │                                                       │        │ 01.testdata/package_json_with_yarn_l │
│                                      │                                                       │        │ ock/"],"ecosystem":"npm"},{"dirs":[" │
│                                      │                                                       │        │ /rule-                               │
│                                      │                                                       │        │ types/github/sca_dependabot.testdata │
│                                      │                                                       │        │ /some_ecosystems/"],"ecosystem":"mav │
│                                      │                                                       │        │ en"},{"dirs":["/security-            │
│                                      │                                                       │        │ baseline/rule-types/github/osps-qa-  │
│                                      │                                                       │        │ 02-                                  │
│                                      │                                                       │        │ 01.testdata/gemfile_no_lock/","/secu │
│                                      │                                                       │        │ rity-baseline/rule-                  │
│                                      │                                                       │        │ types/github/osps-qa-02-             │
│                                      │                                                       │        │ 01.testdata/gemfile_with_lock/"],"ec │
│                                      │                                                       │        │ osystem":"bundler"}]                 │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/minder-rules-and-profiles │ Require pull requests for default branch changes      │ 🔧     │ Changes can be made to the default   │
│ [repository]                         │ [default_branch_pr_required]                          │        │ branch without a pull request for    │
│                                      │                                                       │        │ custcodian/minder-rules-and-profiles │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/minder-rules-and-profiles │ Release tags should not be updated after creation     │ ✅     │                                      │
│ [repository]                         │ [release_tags_immutable]                              │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/minder-rules-and-profiles │ Enable Semgrep SAST                                   │ ✅     │                                      │
│ [repository]                         │ [sast_semgrep]                                        │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/website                   │ Ensure immutable version of GitHub action             │ ✅     │                                      │
│ [repository]                         │ [actions_check_pinned_tags]                           │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/website                   │ Enable Dependabot SCA                                 │ ⛔     │ [{"dirs":["/"],"ecosystem":"npm"}]   │
│ [repository]                         │ [sca_dependabot]                                      │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/website                   │ Enable Semgrep SAST                                   │ ✅     │                                      │
│ [repository]                         │ [sast_semgrep]                                        │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/website                   │ Ensure GitHub Actions workflows set their permissions │ ✅     │                                      │
│ [repository]                         │ [actions_check_default_permissions]                   │        │                                      │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/website                   │ Require pull requests for default branch changes      │ 🔧     │ Changes can be made to the default   │
│ [repository]                         │ [default_branch_pr_required]                          │        │ branch without a pull request for    │
│                                      │                                                       │        │ custcodian/website                   │
├──────────────────────────────────────┼───────────────────────────────────────────────────────┼────────┼──────────────────────────────────────┤
│ custcodian/website                   │ Release tags should not be updated after creation     │ ✅     │                                      │
│ [repository]                         │ [release_tags_immutable]                              │        │                                      │
╰──────────────────────────────────────┴───────────────────────────────────────────────────────┴────────┴──────────────────────────────────────╯
```

(The loaded branch protection rule doesn't take into account repository rulesets, which is why there are a bunch of fix icons)

There's also a `remediation_url` which may be interesting and isn't currently included.

cmd/cli/app/history/history_list.go
```diff
@@ -229,6 +229,12 @@ func renderRuleEvaluationStatusTable(
 			layouts.NoColor(eval.Rule.Name),
 			table.GetStatusIcon(types.HistoryStatus(eval), emoji),
 		)
+		timeValue := layouts.NoColor(eval.EvaluatedAt.AsTime().Format(time.DateTime))
+		entityValue := layouts.NoColor(eval.Entity.Name)
+		ruleValue := layouts.NoColor(eval.Rule.Name)
+		statusValue := table.GetStatusIcon(types.HistoryStatus(eval), emoji)
+
+		t.AddRowWithColor(timeValue, entityValue, ruleValue, statusValue)
```
evankanderson: Why duplicate these outputs?

evankanderson (commented):
I suspect we want to introduce a _new_ command to get all the detail information about a specific rule evaluation on a selected entity -- getting all the details for _every_ rule evaluation in a great big list can be overwhelming when you either have e.g. 20 ruletypes or 20 repositories you're managing.

sachin9058:
@evankanderson Thanks  I’ve updated the PR based on your feedback.

- kept the table output compact by reducing details to a single-line summary
- reverted naming back to “Details”
- applied the suggested improvements (util usage, cmp.Or, rule naming order)
- removed duplication in history output

This keeps the table readable while still surfacing useful context.

I agree that a separate command for detailed inspection would be a better place for full output, and I’d be happy to explore that in a follow-up.

Let me know if this looks better.

cmd/cli/app/artifact/artifact_get.go
```diff
@@ -193,13 +194,16 @@ func printEvalStatus(
 	switch format {
 	case app.Table:
 		ta := table.New(table.Simple, layouts.Default, cmd.OutOrStdout(),
-			[]string{"Profile", "Rule", "Status", "Message"})
+			[]string{"Profile", "Rule", "Result", "Reasoning"})
```
evankanderson: Adjust this back to details?

cmd/cli/app/profile/table_render.go
```diff
@@ -114,11 +177,13 @@ func RenderRuleEvaluationStatusTable(
 
 	for _, eval := range statuses {
 		evalInfo := types.RuleEvalStatus(eval)
+		ruleName := RuleDisplayName(eval)
+		reasoning := FormatEvaluationReasoning(eval)
 		t.AddRowWithColor(
 			layouts.NoColor(fmt.Sprintf("%s\n[%s]", eval.EntityInfo["name"], eval.Entity)),
-			layouts.NoColor(fmt.Sprintf("%s\n[%s]", eval.RuleDescriptionName, eval.RuleTypeName)),
+			layouts.NoColor(fmt.Sprintf("%s\n[%s]", ruleName, eval.GetRuleTypeName())),
```
evankanderson: If `rule.name` in the profile and `display_name` in the rule type aren't set, it seems like this will double-display the rule type name. 

Maybe move this formatting into the `RuleDisplayName` function?

cmd/cli/app/profile/table_render.go
```diff
@@ -94,11 +107,61 @@ func RenderProfileStatusTable(ps *minderv1.ProfileStatus, t table.Table, emoji b
...
+		}
+	}
+	if output := strings.TrimSpace(marshalValueOrEmpty(eval.GetOutput())); output != "" {
+		return fmt.Sprintf("- Output: %s", firstLine(output))
+	}
+
+	return ""
+}
+
+func formatReasoning(eval *minderv1.RuleEvaluationStatus) string {
```
evankanderson: Can we just omit this function since it is a pass through?

cmd/cli/app/profile/table_render.go
```diff
@@ -94,11 +107,61 @@ func RenderProfileStatusTable(ps *minderv1.ProfileStatus, t table.Table, emoji b
...
+
+func buildDetailSummary(eval *minderv1.RuleEvaluationStatus) string {
+	if detail := strings.TrimSpace(eval.GetDetails()); detail != "" {
+		return fmt.Sprintf("- %s", detail)
+	}
+	if guidance := strings.TrimSpace(eval.GetGuidance()); guidance != "" {
+		return fmt.Sprintf("- Guidance: %s", guidance)
+	}
+	if remDetails := strings.TrimSpace(eval.GetRemediationDetails()); remDetails != "" {
+		return fmt.Sprintf("- Remediation: %s", remDetails)
```
evankanderson: If we're going to list the alert URL, we should list the remediation URL as well.

cmd/cli/app/profile/table_render.go
```diff
@@ -94,11 +107,61 @@ func RenderProfileStatusTable(ps *minderv1.ProfileStatus, t table.Table, emoji b
...
+	if alert := eval.GetAlert(); alert != nil {
+		if alertDetails := strings.TrimSpace(alert.GetDetails()); alertDetails != "" {
+			return fmt.Sprintf("- Alert: %s", alertDetails)
+		}
+		if alertURL := strings.TrimSpace(alert.GetUrl()); alertURL != "" {
+			return fmt.Sprintf("- Alert URL: %s", alertURL)
+		}
+	}
+	if output := strings.TrimSpace(marshalValueOrEmpty(eval.GetOutput())); output != "" {
+		return fmt.Sprintf("- Output: %s", firstLine(output))
```
evankanderson: If the output is pretty-printed yaml, the first line seems unlikely to be useful on its own.  (It should also always be non-empty, so `strings.SplitN(output, "\n", 1)[1]` should be sufficient.)

sachin9058:
@evankanderson Thanks I’ve incorporated the feedback:

- switched column naming back to “Details”
- fixed rule name duplication via RuleDisplayName
- removed the pass-through function
- added remediation URL alongside alert URL
- simplified the table to a compact summary (no YAML/output in table)

Let me know if this looks good or if you'd like any further tweaks.

cmd/cli/app/profile/table_render.go
sachin9058: I’ve added remediation URL handling alongside alert URL.

Since the table is compact and returns a single summary line, only one of these will be shown depending on priority, but both are now supported in the rendering logic.

Happy to adjust if you'd prefer showing multiple fields instead.

sachin9058:
@evankanderson hey can u take a look on this again ?? I have  improved the pr and also made requested changes and tell me wheather it looks good or need further improvements ??

cmd/cli/app/profile/table_render.go
```diff
@@ -114,11 +149,13 @@ func RenderRuleEvaluationStatusTable(
 
 	for _, eval := range statuses {
 		evalInfo := types.RuleEvalStatus(eval)
+		ruleName := RuleDisplayName(eval)
+		reasoning := FormatEvaluationReasoning(eval)
 		t.AddRowWithColor(
 			layouts.NoColor(fmt.Sprintf("%s\n[%s]", eval.EntityInfo["name"], eval.Entity)),
-			layouts.NoColor(fmt.Sprintf("%s\n[%s]", eval.RuleDescriptionName, eval.RuleTypeName)),
+			layouts.NoColor(ruleName),
```
evankanderson, line 175: We don't necessarily have the ruletype name visible here anymore, do we?

evankanderson (commented):
I think it would be helpful to add the golden-output type tests like #6418 or #6419 for `artifact get` or `profile status list` before merging this change -- the earlier comments have a screenshot or two, but being able to evaluate the output changes in a diff feels useful.

sachin9058:
@evankanderson Thanks !!!

I’ve added golden-output tests for `artifact get` in this PR following the same pattern as #6418/#6419, so the CLI output changes can now be evaluated directly via diffs.

To support this, I introduced test-friendly RPC injection using `cli.WithRPCClient`. Previously, the CLI wrapper would always initialize a real gRPC connection, which caused tests to trigger login/network flows. I added a small context-based check so that when an injected client is present, the command uses it instead of dialing.

This behavior only applies when a client is explicitly injected in tests, so normal CLI execution remains unchanged.

I used lightweight stubs for the artifact and profile clients to keep the tests self-contained, but I’m happy to switch to generated mocks if that’s preferred.

Let me know if you’d like additional commands covered or adjustments to the injection approach.

cmd/cli/app/artifact/artifact_get.go
```diff
@@ -73,6 +75,22 @@ func getCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc.
...
+	return minderv1.NewArtifactServiceClient(conn)
+}
+
+func getProfileClient(ctx context.Context, conn *grpc.ClientConn) minderv1.ProfileServiceClient {
+	if mockClient, ok := cli.GetRPCClient[minderv1.ProfileServiceClient](ctx); ok {
+		return mockClient
+	}
+
+	return minderv1.NewProfileServiceClient(conn)
+}
```
evankanderson: Put this in `cmd/cli/app/profile/profile.go` and export it, like in #6420.  (Actually, that has an even-better approach now, but you're addressing this feedback to some extent: https://github.com/mindersec/minder/pull/6420/changes#r3150092088)

evankanderson: (Use the new `cli.GetCLIClient(cmd, minderv1.NewProfileServiceClient)` format, and then you don't need to write these at all.  You do still need to handle the closer returned from the `GetCLIClient` call.)

cmd/cli/app/artifact/artifact_get_test.go
```diff
@@ -0,0 +1,154 @@
...
+		mockResp := &minderv1.GetArtifactByIdResponse{}
+		cli.LoadFixture(t, "mock_artifact_get.json", mockResp)
+
+		client.getArtifactByIDResp = mockResp
+		client.getArtifactByIDErr = nil
+
+		profileClient := &profileClientStub{
+			listProfilesResp: &minderv1.ListProfilesResponse{},
+			statusResp:       &minderv1.GetProfileStatusByNameResponse{},
+		}
```
evankanderson: While I'm not normally a huge fan of mocks, #6386 and follow-ons like #6418 and #6420 have used them pretty effectively to avoid needing to write an entire stub service in the test.

cmd/cli/app/artifact/artifact_get_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+		mockResp := &minderv1.GetArtifactByIdResponse{}
+		cli.LoadFixture(t, "mock_artifact_get.json", mockResp)
+
+		client.getArtifactByIDResp = mockResp
+		client.getArtifactByIDErr = nil
+
+		profileClient := &profileClientStub{
+			listProfilesResp: &minderv1.ListProfilesResponse{},
+			statusResp:       &minderv1.GetProfileStatusByNameResponse{},
```
evankanderson: I think you need to fill out these responses so that we can see how the `artifact get` output differs with your formatting.

The good news is that if you sync past #6420, you'll get the current "profiles status list --detailed" output for free, and we can look at those diffs.

evankanderson: I think you should be able to run the tests with the `--update=true` flag to have the files updated automatically.

internal/util/cli/cli.go
```diff
@@ -83,14 +83,17 @@ func GRPCClientWrapRunE(
...
+		}
+
 		c, err := GrpcForCommand(cmd, viper.GetViper())
 		if err != nil {
 			return err
 		}
-
-		ctx, cancel := GetAppContext(cmd.Context(), viper.GetViper())
-		defer cancel()
-
```
evankanderson: I think I'd prefer that we move towards a standard `RunE` function and discard the `GRPCClientWrapRunE` method if possible.

internal/util/cli/context.go
```diff
@@ -12,10 +12,13 @@ type rpcKey struct {
...
 }
 
+type rpcInjectedKey struct{}
+
 // WithRPCClient injects the provided RPC client into the context,
 func WithRPCClient[T any](ctx context.Context, client T) context.Context {
 	key := rpcKey{clientType: reflect.TypeOf((*T)(nil)).Elem()}
-	return context.WithValue(ctx, key, client)
+	ctx = context.WithValue(ctx, key, client)
+	return context.WithValue(ctx, rpcInjectedKey{}, struct{}{})
```
evankanderson: Why add a second injection (which overrides the typed injection)?

sachin9058:
@evankanderson 
Thanks I'’ve rebased this PR on the latest main and aligned it with the pattern introduced in #6420.

* removed the custom injection logic and restored standard CLI behavior
* updated `artifact get` to use `cli.GetCLIClient` for both artifact and profile clients
* added golden-output tests for `artifact get` using gomock with realistic evaluation data
* regenerated profile/status goldens to match the current CLI output after the rebase


Let me know if anything else needs adjustment.

sachin9058:
@evankanderson 

I also updated the artifact get goldens (the failure was due to JSON formatting differences only — spacing changes, not data changes).

Locally everything is passing:

* go test ./cmd/cli/app/artifact
* go test ./cmd/cli/...

However, CI is still failing on the artifact tests, and I’m not able to reproduce the failure locally.

Could you help point out what’s different in CI or if there’s something I might be missing (e.g., environment-specific formatting or test expectations)?

Happy to adjust once I understand the discrepancy.

evankanderson:
The test is claiming an extra space after the ":" in the JSON:
```
          	            	Diff:
          	            	--- Expected
          	            	+++ Actual
          	            	@@ -1,10 +1,10 @@
          	            	 {
          	            	-  "artifact": {
          	            	-    "artifactPk": "111",
          	            	-    "owner": "owner-1",
          	            	-    "name": "artifact-1",
          	            	-    "type": "image",
          	            	-    "visibility": "public",
          	            	-    "repository": "org/repo",
          	            	-    "createdAt": "2024-01-02T15:04:05Z"
          	            	+  "artifact":  {
          	            	+    "artifactPk":  "111",
          	            	+    "owner":  "owner-1",
          	            	+    "name":  "artifact-1",
          	            	+    "type":  "image",
          	            	+    "visibility":  "public",
          	            	+    "repository":  "org/repo",
          	            	+    "createdAt":  "2024-01-02T15:04:05Z"
          	            	   }
```

I'll see if I can figure out why that might be.

evankanderson:
[From protojson's `MarshalOptions.Marshal`](https://pkg.go.dev/google.golang.org/protobuf@v1.36.11/encoding/protojson#MarshalOptions.Marshal):

> Do not depend on the output being stable. Its output will change across different builds of your program, even when using the same version of the protobuf module.

And they mean it: https://github.com/protocolbuffers/protobuf-go/blob/master/internal/encoding/json/encode.go#L272

Unfortunately, `detrand.Disable()` is in the `internal` package, so we may need to do something to strip out their extra whitespace (possibly replace all runs of spaces with a single space for JSON).

evankanderson:
Alternatively, we can write a nasty note (meaning replacing each JSON case with a comment referencing a bug in the Minder repo that the JSON output is not stable) and not test the JSON output, which seems fairly safe.

sachin9058:
@evankanderson

I’ve removed the JSON golden tests for `artifact get` since the protojson output isn’t stable across environments, and kept the table and YAML output tests. I also added a note in the test explaining this decision.

All CLI tests are passing locally:

Let me know if anything else needs adjustment.

cmd/cli/app/profile/status/testdata/status_get.json.golden
evankanderson, line 1: I think you want to remove these .json.golden files now that you've filed #6430

cmd/cli/app/artifact/fixture/mock_artifact_get.json
evankanderson, line 1: You can use additional files to mock `GetProfileStatusByNameResponse` and `ListProfilesResponse` in the same way you mock `GetArtifactsByIdResponse`.

cmd/cli/app/artifact/testdata/artifact_get.yaml.golden
```diff
@@ -0,0 +1,20 @@
...
+- details: artifact attestation is disabled for this image
+  entity: artifact
+  entityInfo:
+    name: owner-1/artifact-1
+  guidance: enable artifact attestations before release
+  profileId: artifact-security-baseline
+  remediationUrl: https://example.com/remediate/artifact-111
+  ruleDescriptionName: Require artifact attestation
+  ruleTypeName: artifact_attestation_slsa
+  status: failure
```
evankanderson: This is really interesting!  We're not producing valid YAML (or JSON) from `artifact get`.  I don't know if we should change it (we should probably have a consistent design), but I wouldn't have noticed without this test file.

cmd/cli/app/profile/status/testdata/status_list_table_detailed.txt.golden
```diff
@@ -2,10 +2,10 @@
...
-                         ├──────────────────────┼─────────┼─────────────────────────────────────────
-                         │                      │ ⛔      │ Mock rule evaluation failed.            
-                         │ [codeql_enabled]     │         │                                         
+ ENTITY              │ RULE                                      │ RESULT │ DETAILS                 
+─────────────────────┼───────────────────────────────────────────┼────────┼─────────────────────────
+ acme-corp/mock-repo │ Enable secret scanning to detect          │ ✅     │ - Mock rule evaluation  
+ [repository]        │ hardcoded secrets                         │        │ succeeded.              
+                     ├───────────────────────────────────────────┼────────┼─────────────────────────
+                     │ Enable CodeQL for vulnerability scanning  │ ⛔     │ - Mock rule evaluation  
+                     │                                           │        │ failed.
```
evankanderson: I like that we now have clear rule names rather than a blank line.  I'm not sure how I feel about the size that "rule" takes up vs "details".  The leading "-" on "Details" also feels off to me.

cmd/cli/app/profile/table_render.go
```diff
@@ -114,11 +149,13 @@ func RenderRuleEvaluationStatusTable(
...
 			table.GetStatusIcon(evalInfo, emoji),
-			table.BestDetail(evalInfo),
```
evankanderson, line 121: is `BestDetail` used anywhere anymore?

cmd/cli/app/profile/table_render.go
```diff
@@ -94,11 +94,46 @@ func RenderProfileStatusTable(ps *minderv1.ProfileStatus, t table.Table, emoji b
...
+	}
+	if guidance := strings.TrimSpace(eval.GetGuidance()); guidance != "" {
+		return fmt.Sprintf("- Guidance: %s", guidance)
+	}
+	if remURL := strings.TrimSpace(eval.GetRemediationUrl()); remURL != "" {
+		return fmt.Sprintf("- Remediation URL: %s", remURL)
+	}
+	if remDetails := strings.TrimSpace(eval.GetRemediationDetails()); remDetails != "" {
+		return fmt.Sprintf("- Remediation: %s", remDetails)
+	}
```
evankanderson: What do you think about making this 2 lines to include both the comment and the URL (if present -- and the same for Alert)?

```suggestion

	if remDetails := strings.TrimSpace(eval.GetRemediationDetails()); remDetails != "" {
		remediated := fmt.Sprintf("Remediation: %s", remDetails)
		if remURL := eval.GetRemediationUrl(); remURL != "" {
			remediated += fmt.Sprintf("\nURL: %s", remURL)
		}
		return remediated
	}
```

I'm also wondering if we want to return in priority order:

* Alert, if it happened
* Remediation, if it happened
* Details (message)
* Guidance

evankanderson (commented):
Just a couple comments -- mostly on cleaning up the json golden files, but I'm also wondering about our current priority of messages, and a few other cleanups.

cmd/cli/app/profile/status/testdata/status_get.json.golden
sachin9058: Yes i have removed them

cmd/cli/app/profile/table_render.go
sachin9058: I have fixed this now they are returning in given priority order

sachin9058:
@evankanderson Thanks  I’ve addressed the feedback.

* removed remaining JSON golden tests and test cases
* switched to fixture-backed mocks for consistency
* updated reasoning formatting (Alert → Remediation → Details → Guidance) with multiline remediation support
* removed unused helpers and added test coverage for the updated formatting

All CLI tests are passing:


Let me know if anything else needs adjustment.

sachin9058:
I tried reproducing the race locally with `go test -race ./...`, and everything passes (including the `internal/eea` tests).
This might be a timing-sensitive or flaky race condition that’s showing up in CI but not locally. Since these failures are in the eventing layer and not directly related to the CLI changes in this PR.
Let me know if you’d like me to dig deeper or if this should be tracked separately.

evankanderson (approved):

evankanderson:
The EEA test looks like a flake to me.  Re-running

PR #6420, opened 2026-04-27
added test files for profile and profile/status command

DharunMR:
## Description

- This PR introduces a comprehensive refactor of the `cmd/cli/app/profile` and `cmd/cli/app/profile/status`  package to modernize its testing architecture, improve separation of concerns. Stripped out manual `ctx` and `grpc.ClientConn` arguments from command execution functions to align with standard Cobra `RunE` signatures. Implemented the "backpack" pattern using `cmd.SetContext(ctx)` and `cli.WithRPCClient` to inject mock clients cleanly during testing.
- Modified the `apply` command to accept and process multiple YAML file paths

## 🧪 How to Test
1. Run standard tests: `go test ./...` (Should PASS)
2. Run test if any changes made in cmd code then add `-update` flag: `go test ./cmd/cli/app/profile/ -update` and `go test ./cmd/cli/app/profile/status -update`


cmd/cli/app/datasource/datasource.go
```diff
@@ -24,3 +27,24 @@ func init() {
...
+	if err != nil {
+		cancel()
+		return nil, nil, err
+	}
+
+	return minderv1.NewDataSourceServiceClient(conn), func() {
+		cancel()
+		_ = conn.Close()
+	}, nil
+}
```
evankanderson: I'm noticing that these seem very repetitive; can we use a generic version?


```suggestion
type Cleanup = func()

// GetCLIClient takes a factory for a GRPC client service and returns
// a client, a cleanup function to close the connection and an error
func GetCLIClient[T interface](cmd *cobra.Command, client func(grpc.ClientConnInterface) T) (T, Cleanup, error) {
	ctx, cancel := cli.GetAppContext(cmd.Context(), viper.GetViper())
	cmd.SetContext(ctx)

	if mockClient, ok := cli.GetRPCClient[T](ctx); ok {
		return mockClient, func() { cancel() }, nil
	}

	conn, err := cli.GrpcForCommand(cmd, viper.GetViper())
	if err != nil {
		cancel()
		return nil, nil, err
	}

	return client(conn), func() {
		cancel()
		_ = conn.Close()
	}, nil
}
```

cmd/cli/app/profile/status/status_get.go
```diff
@@ -123,7 +134,7 @@ func getProfileStatusByName(
 }
 
 type protoWithProfileStatus interface {
-	proto.Message
+	protoreflect.ProtoMessage
```
evankanderson: we should prefer `proto.Message`, but a bunch of early usage (before my arrival) used the `protoreflect` package, and it got copied lots of places.


```suggestion
	proto.Message
```

cmd/cli/app/profile/status/testdata/status_list_table_detailed.txt.golden
evankanderson, line 1: I think we need a detailed with no emoji to check that output.

I'm also wondering about empty entity and rule name fields -- can we fill those in for the fixture?

cmd/cli/app/profile/status/status_list.go
```diff
@@ -23,26 +21,37 @@ var listCmd = &cobra.Command{
...
-
+func listCommand(cmd *cobra.Command, _ []string) error {
 	project := viper.GetString("project")
 	profileName := viper.GetString("name")
 	detailed := viper.GetBool("detailed")
 	ruleType := viper.GetString("ruleType")
 	ruleName := viper.GetString("ruleName")
+
+	cmd.SilenceUsage = true
+
```
evankanderson: `SilenceUsage` should go _after_ the output support check (line 44), as we should probably print usage alongside the "not supported" message.

evankanderson: I could make an argument that it should go after the `GetProfileClient`, because the help will point you to things like `--grpc-host` and `--grpc-insecure` that could help you fix a broken connection.

cmd/cli/app/profile/testdata/apply_create.table.golden
evankanderson, line 1: This looks terrible (yaml wrapped, etc), but it is what it is.  Maybe file an issue that "Rule definition" is not very helpful in the current apply output?

evankanderson: (Thanks for doing this, btw!  It makes it much easier to see where the formatting sucks)

cmd/cli/app/profile/testdata/apply_multiple.table.golden
```diff
@@ -0,0 +1,18 @@
+Successfully created new profile named: mock-attestation-profile
+Successfully updated existing profile named: mock-dependabot-profile
+ ENTITY     │ RULE                      │ RULE PARAMS              │ RULE DEFINITION
```
evankanderson, line 2: Interesting... if one of these profiles has two rules, what does the output look like?  It feels like we might need a delimiter (extra column wouldn't fit) for the profile name.

cmd/cli/app/profile/testdata/export_by_id.yaml.golden
```diff
@@ -0,0 +1,17 @@
...
+    eval:
+        data_sources:
+            - name: mock-data-source
+name: mock-rule-type
+type: rule-type
+version: v1
+---
+name: mock-data-source
+type: data-source
+version: v1
```
evankanderson, line 54: This output is weird -- can we fill in enough fields that it would make sense?  (It's also annoying that the field sort "feels" random when it's alphabetical rather than by some other canonical ordering ala Kubernetes.)

cmd/cli/app/profile/export.go
```diff
@@ -50,18 +49,60 @@ func exportCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *gr
...
-	// Fetch associated resources
-	rulesClient := minderv1.NewRuleTypeServiceClient(conn)
+	// Fetch associated Rule Types
+	rulesClient, closeRules, err := ruletype.GetRuleTypeClient(cmd)
+	if err != nil {
+		return cli.MessageAndError("Error connecting to rule service", err)
+	}
+	defer closeRules()
 
 	// TODO: it would be nice if this were just a list of rules...
```
evankanderson: I think this TODO may be done?

cmd/cli/app/profile/get.go
```diff
@@ -39,15 +41,25 @@ func getCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc.
...
 	}
+	if id == "" && name == "" {
+		return cli.MessageAndError("Error getting profile", fmt.Errorf("id or name required"))
+	}
 
 	// No longer print usage on returned error, since we've parsed our inputs
 	// See https://github.com/spf13/cobra/issues/340#issuecomment-374617413
 	cmd.SilenceUsage = true
 
+	// 3. NOW SETUP GRPC
```
evankanderson: Comment here is a bit strange.

cmd/cli/app/profile/export.go
DharunMR: TODO actually got misplaced during a refactor. i think its still unresolved because the protobuf schema still partitions rules by entity type?

cmd/cli/app/datasource/datasource.go
DharunMR: done 👍

cmd/cli/app/profile/status/testdata/status_list_table_detailed.txt.golden
DharunMR: added that 👍

cmd/cli/app/profile/testdata/apply_create.table.golden
DharunMR: Yea i'll file an issue

cmd/cli/app/profile/testdata/apply_multiple.table.golden
DharunMR: i made table to render after every success profile creation , does it look better now?

cmd/cli/app/profile/testdata/export_by_id.yaml.golden
DharunMR: yes the alphabetical sort makes these hard to parse visually, i think i can open a follow-up pr for canonical ordering of fields

cmd/cli/app/profile/status/testdata/status_list_table_detailed_no_emoji.txt.golden
```diff
@@ -0,0 +1,11 @@
+ NAME                          │ STATUS         │ EVALUATED AT                                      
+───────────────────────────────┼────────────────┼───────────────────────────────────────────────────
+ mock-profile                  │ Ok             │ 2024-01-01T00:00:00Z                              
+
+ ENTITY                  │ RULE NAME            │ STATUS  │ DETAILS                                 
+─────────────────────────┼──────────────────────┼─────────┼─────────────────────────────────────────
+ acme-corp/mock-repo     │                      │ Ok      │ Mock rule evaluation succeeded.
```
evankanderson, line 7: Empty rule name here is weird, but it seems like we're passing the tests, so this should be a follow up fix in coordination with #6417.

cmd/cli/app/profile/testdata/apply_multiple.table.golden
evankanderson: Modulo https://github.com/mindersec/minder/pull/6420/changes#r3150092088, yes.

Thanks for doing this -- I'm inclined to defer _output_ cleanup on the CLI until after we have the existing output captured by these tests (but maybe we file issues for the cleanup).

cmd/cli/app/profile/testdata/export_by_id.yaml.golden
evankanderson: #6428

cmd/cli/app/profile/apply.go
```diff
@@ -69,38 +91,54 @@ func applyCommand(_ context.Context, cmd *cobra.Command, args []string, conn *gr
...
-	table.Render()
 	return nil
 }
 
 func init() {
 	ProfileCmd.AddCommand(applyCmd)
 	// Flags
-	applyCmd.Flags().StringP("file", "f", "", "Path to the YAML defining the profile (or - for stdin)")
+	applyCmd.Flags().StringArrayP("file", "f", []string{},
+		"Path to the YAML defining the profile (or - for stdin). Can be specified multiple files")
```
evankanderson, line 143: It's okay this time, but let's try to separate behavior changes from e.g. test and refactoring changes that don't change behavior.

cmd/cli/app/profile/export.go
```diff
@@ -80,38 +121,49 @@ func exportCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *gr
 	ruletypes = slices.Compact(ruletypes)
 
 	// Collect the referenced datasources from the rule types as we process them.
-	datasources := make([]string, 0)
-	for _, ruletype := range ruletypes {
-		resp, err := rulesClient.GetRuleTypeByName(ctx, &minderv1.GetRuleTypeByNameRequest{
+	var datasources []string
+	for _, rtName := range ruletypes {
```
evankanderson, line 125: This refactor makes it harder to review the PR because I now have to start a second context (what's going on with these functions?) in addition to the rule output changes.  While it's super-tempting to fix things as you see them, it's easier to review 2 focused PRs than one PR that mixes behavior.  (Renaming all the `ruletype` variables also introduces a bunch of extra diffs.)

internal/util/cli/table/table.go
```diff
@@ -101,7 +102,7 @@ func (g *goPrettyTable) updateWidths(row []string) {
 		}
 		w := 0
 		for _, line := range strings.Split(val, "\n") {
-			lw := utf8.RuneCountInString(line)
+			lw := text.RuneWidthWithoutEscSequences(line)
```
evankanderson, line 105: Ah, was this a bug?

evankanderson (approved):
I'm going to merge this to un-block #6417 and others.

internal/util/cli/table/table.go
DharunMR: yea, this printing color and emoji as code instead as it is

cmd/cli/app/profile/export.go
DharunMR: sure i will follw it

PR #6432, opened 2026-05-01
feat: order YAML output fields for Profile, RuleType, DataSource

### LINKED ISSUE #6428: YAML output (and export) field ordering is alphabetical, feels random
Inspired by https://github.com/mindersec/minder/pull/6420/changes#r3150107180

We have a standard field ordering in https://github.com/mindersec/minder-rules-and-profiles for Profiles, RuleTypes, and DataSources.  It's generally:

* version
* type
* name
* context (if needed/present)
* (other fields, generally descriptive / metadata before deep structure)

It would be great if places where we print YAML could order the fields -- it's not clear how hard this is, but it might be a good first issue.

maryam22001:
## Summary

When printing YAML output for profiles, rule types, and data sources, fields were ordered by protobuf field number rather than the canonical order defined in the project conventions. This PR adds a `GetOrderedYamlFromProto` helper in `internal/util/` that marshals fields in the standard order (`version → type → name → context → ...`) for Profile, RuleType, and DataSource resources. Non-resource protos fall back to alphabetical order to avoid regressions.

Fixes #6428

## Changes

- Added `GetOrderedYamlFromProto` helper in `internal/util/yaml_order.go`
- Updated `cmd/cli/app/profile/get.go`, `ruletype/get.go`, and `datasource/get.go` to use the new helper
- Added unit tests in `internal/util/yaml_order_test.go`
- Updated golden files to reflect correct field order

### Unit tests

New tests added in `internal/util/yaml_order_test.go`:

```bash
export PATH=$HOME/go-install/go/bin:$PATH
go test ./internal/util/... -run "TestGetOrdered" -v
```

| Test | Verifies |
|------|----------|
| `TestGetOrderedYamlFromProto_Profile` | `version → type → name` are fields 1–3 |
| `TestGetOrderedYamlFromProto_RuleType` | `version → type → name` are fields 1–3 |
| `TestGetOrderedYamlFromProto_DataSource` | `version → type → name` are fields 1–3 |
| `TestGetOrderedYamlFromProto_ContextComesAfterName` | `context` is field 4 when present |
| `TestGetOrderedYamlFromProto_NonResourceFallsBackToAlpha` | Non-resource protos still marshal without error |
| `TestGetOrderedYamlFromProto_NilReturnsEmptyObject` | `nil` input returns `{}\n` with no error |

All 6 pass 

### Existing CLI golden file tests

Golden files updated to reflect the new field order:

```bash
go test ./cmd/cli/app/profile/... ./cmd/cli/app/ruletype/... ./cmd/cli/app/datasource/...
```

Updated files: `list_profiles.yaml.golden`, `status_get.yaml.golden`, `status_list.yaml.golden`, `get_by_name.yaml.golden`, `list_populated.yaml.golden`

### Smoke test

```bash
make build
./bin/minder profile get    -n <name> -o yaml
./bin/minder ruletype get   -n <name> -o yaml
./bin/minder datasource get -n <name> -o yaml
```

Expected — every resource starts with `version → type → name → context`:

```yaml
version: v1
type: profile
name: secret-scanning
context:
    project: my-project
```

### Configuration
- Go `1.26.2`
- No new dependencies introduced

### DIFF
--- a/cmd/cli/app/profile/status/testdata/status_get.yaml.golden
+++ b/cmd/cli/app/profile/status/testdata/status_get.yaml.golden
@@ -1,39 +1,39 @@
 profileStatus:
-  lastUpdated: "2024-01-01T00:00:00Z"
-  profileId: 11111111-1111-1111-1111-111111111111
-  profileName: mock-profile
-  profileStatus: success
+    lastUpdated: "2024-01-01T00:00:00Z"
+    profileId: 11111111-1111-1111-1111-111111111111
+    profileName: mock-profile
+    profileStatus: success
 ruleEvaluationStatus:
-  - details: Mock rule evaluation succeeded.
-    entity: repository
-    entityInfo:
-      entity_type: repository
-      name: acme-corp/mock-repo
-      provider: github-app-mock-provider
-      repo_name: mock-repo
-      repo_owner: acme-corp
-      repository_id: 22222222-2222-2222-2222-222222222222
-    lastUpdated: "2024-01-01T00:00:00Z"
-    profileId: 11111111-1111-1111-1111-111111111111
-    ruleDisplayName: Enable secret scanning to detect hardcoded secrets
-    ruleId: mock-rule-123
-    ruleName: secret_scanning
-    ruleTypeName: secret_scanning
-    status: success
-  - details: Mock rule evaluation failed.
-    entity: repository
-    entityInfo:
-      entity_type: repository
-      name: acme-corp/mock-repo
-      provider: github-app-mock-provider
-      repo_name: mock-repo
-      repo_owner: acme-corp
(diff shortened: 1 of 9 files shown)


evankanderson (commented):
How will this scale as we add new fields to protocol buffers?  I'm concerned that we'll end up forgetting to adjust the helpers, and end up omitting new fields in input and output.

Thinking out loud, it almost feels like we want to have a known set of document nodes which we serialize first, then serialize everything else alphabetically.

Looking at e.g. https://github.com/mindersec/minder-rules-and-profiles/blob/main/rule-types/github/branch_protection_lock_branch.yaml, it also seems like we serialize consistently within the `def` of the RuleType; `ingest` comes before `eval` comes before `remediate` comes before `alert` (which is the order that evaluation happens in).

PR #6433, opened 2026-05-01
test(cli): support JSON stream decoding in golden tests to fix protojson instability

### LINKED ISSUE #6430: Unstable JSON output in CLI tests due to protojson formatting
### Describe the issue

CLI JSON output generated via protojson is not stable across environments. 
Specifically, spacing differences (e.g., extra spaces after `:`) can vary between local runs and CI, even when the underlying data is identical.

This causes golden tests that rely on exact JSON string matching to fail intermittently.

### To Reproduce

1. Run CLI tests that include JSON golden output (e.g., artifact get)
2. Compare outputs across different environments (local vs CI)
3. Observe differences in spacing/formatting (extra spaces after `:`)
4. Tests fail due to string mismatch despite identical data

Latest main (after #6420 / #6417 changes)

sachin9058:
# Summary

This PR fixes unstable CLI golden tests caused by protojson formatting differences and newline-delimited JSON output.

Currently, CLI outputs can be a JSON stream (multiple JSON objects), which makes string-based comparisons and JSONEq unsuitable. This results in flaky tests across environments (local vs CI).

This change introduces a decoder-driven approach using json.Decoder to parse JSON streams and compare structured data instead of raw strings.

Key changes:

* Added support for newline-delimited JSON (JSON streams) in golden file comparison
* Retained JSONEq for single JSON objects (backward compatibility)
* Added graceful fallback to string comparison for non-JSON outputs (YAML, tables, etc.)
* Removed heuristic-based detection in favor of decoder-driven parsing
* Added unit tests for:

  * JSON stream parsing
  * Single JSON object parsing
  * Invalid JSON handling
  * Non-JSON input (YAML/text)
  * Empty JSON array edge case

This ensures deterministic and environment-independent test behavior.

Fixes #6430

---

# Testing

The changes were tested using the existing CLI test suite and additional unit tests.

### Steps:

1. Run all tests:

   ```
   go test ./...
   ```

2. Verified:

   * JSON stream outputs are correctly parsed and compared structurally
   * Single JSON outputs continue to work with JSONEq
   * Non-JSON outputs (YAML, tables) fall back to string comparison
   * No flaky failures observed across runs

All existing tests pass, and new tests validate the updated behavior.

### DIFF
--- a/cmd/cli/app/artifact/artifact_list.go
+++ b/cmd/cli/app/artifact/artifact_list.go
@@ -30,7 +30,11 @@ var listCmd = &cobra.Command{
 
 // listCommand is the artifact list subcommand
 func listCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc.ClientConn) error {
-	client := minderv1.NewArtifactServiceClient(conn)
+	client, cleanup, err := cli.GetCLIClient(cmd, minderv1.NewArtifactServiceClient)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
 
 	provider := viper.GetString("provider")
 	project := viper.GetString("project")
@@ -74,11 +78,14 @@ func listCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc
 		}
 		t.Render()
 	case app.JSON:
-		out, err := util.GetJsonFromProto(artifactList)
-		if err != nil {
-			return cli.MessageAndError("Error getting json from proto", err)
+		// Emit each artifact as a separate JSON object (JSON stream)
+		for _, art := range artifactList.Results {
+			out, err := util.GetJsonFromProto(art)
+			if err != nil {
+				return cli.MessageAndError("Error getting json from proto", err)
+			}
+			cmd.Println(out)
 		}
-		cmd.Println(out)
 	case app.YAML:
 		out, err := util.GetYamlFromProto(artifactList)
 		if err != nil {
--- a/cmd/cli/app/artifact/artifact_list.go
+++ b/cmd/cli/app/artifact/artifact_list.go
@@ -4,14 +4,12 @@
 package artifact
 
 import (
-	"context"
 	"fmt"
 	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"google.golang.org/grpc"
 
 	"github.com/mindersec/minder/cmd/cli/app"
 	"github.com/mindersec/minder/internal/util"
@@ -25,16 +23,22 @@ var listCmd = &cobra.Command{
 	Use:   "list",
 	Short: "List artifacts from a provider",
 	Long:  `The artifact list subcommand will list artifacts from a provider.`,
-	RunE:  cli.GRPCClientWrapRunE(listCommand),
+	RunE:  listCommand,
 }
 
 // listCommand is the artifact list subcommand
-func listCommand(ctx context.Context, cmd *cobra.Command, _ []string, _ *grpc.ClientConn) error {
+func listCommand(cmd *cobra.Command, _ []string) error {
+	if err := viper.BindPFlags(cmd.Flags()); err != nil {
+		return fmt.Errorf("error binding flags: %w", err)
+	}
+
 	client, cleanup, err := cli.GetCLIClient(cmd, minderv1.NewArtifactServiceClient)
 	if err != nil {
 		return err
 	}
 	defer cleanup()
+
+	ctx := cmd.Context()
 
 	provider := viper.GetString("provider")
 	project := viper.GetString("project")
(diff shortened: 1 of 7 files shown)


DharunMR:
IMO you need to add those json testdata again to verify it passes

DharunMR:
actually you should probably need to add the json stream testdata and those profile, repo testcase that was removed in this pr #6417 since they fall directly under the scope of what we're fixing here

sachin9058:
@DharunMR Yes, thank you. I actually forgot to add the golden earlier I had only included parsing tests. I’ve now added them, and CI is passing. However, the NATS test is still failing. It looks flaky because when I run it locally multiple times, it consistently passes. For now, I’ve retriggered the CI Let’s see what happens.

sachin9058:
Yes, Once the CI passes and it gets merged, I will rebase and revert the golden files that I removed there.

sachin9058:
Well CI is green now ... That JSON stream error is passing now let see what @evankanderson thinks about this fix...

sachin9058:
@DharunMR 
The artifact list command now emits newline-delimited JSON, so the stream parsing path is exercised in tests. This PR mainly ensures the comparison logic is robust for such outputs.
Happy to extend coverage further if needed, but wanted to keep this PR focused.

DharunMR:
yep, i thought adding those testcase back is good but we can wait for @evankanderson

sachin9058:
@evankanderson Hey evan can u take a look on this ???

cmd/cli/app/artifact/artifact_list.go
```diff
@@ -74,17 +82,24 @@ func listCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc
...
-		cmd.Println(out)
 	case app.YAML:
 		out, err := util.GetYamlFromProto(artifactList)
 		if err != nil {
 			return cli.MessageAndError("Error getting yaml from proto", err)
 		}
-		cmd.Println(out)
+		if _, err := cmd.OutOrStdout().Write([]byte(out + "\n")); err != nil {
+			return cli.MessageAndError("Error writing yaml to stdout", err)
+		}
```
evankanderson, line 102: This is basically the same thing as `cmd.Println()` except that it logs an error message if we try to redirect to something which doesn't accept writes.  I'm not sure the error message in that case is a big improvement, but I don't want to block the rest of the value of this PR on that.

cmd/cli/app/artifact/artifact_list.go
```diff
@@ -74,17 +82,24 @@ func listCommand(ctx context.Context, cmd *cobra.Command, _ []string, conn *grpc
 		}
 		t.Render()
 	case app.JSON:
-		out, err := util.GetJsonFromProto(artifactList)
-		if err != nil {
-			return cli.MessageAndError("Error getting json from proto", err)
+		// Emit each artifact as a separate JSON object (JSON stream)
+		for _, art := range artifactList.Results {
```
evankanderson, line 86: I think it's fine for the output to be JSON-LD; it also seems fine to define that we'll only emit a _single_ JSON element.

In either case, it would be great to record CLI patterns somewhere -- maybe in `DEVELOPMENT.md`?  Right now, we're not totally consistent about any of the following:

* Output columns and names
* JSON output contents
* Positional parameters and flags like `--file`
* Tabular output and multi-row output

It would be great to have a "design philosophy" around these that people could reference when making changes, such that the CLI starts to feel more consistently and thoughtfully designed.

I don't think it needs to happen in this PR, but it would be useful to start making some of these decisions explicit.

internal/util/cli/context.go
```diff
@@ -22,6 +22,12 @@ func WithRPCClient[T any](ctx context.Context, client T) context.Context {
 	return context.WithValue(ctx, key, client)
 }
 
+// WithCLIClient is an alias for WithRPCClient kept for backwards/alternate usage
+// by tests or callers expecting a "CLI"-named helper.
+func WithCLIClient[T any](ctx context.Context, client T) context.Context {
+	return WithRPCClient[T](ctx, client)
+}
+
```
evankanderson, line 30: I don't understand why we wouldn't just switch over to one consistent name here -- this is in `internal`, so nothing outside the codebase should be depending on it.

internal/util/cli/testing.go
```diff
@@ -122,13 +124,54 @@ func checkGoldenFile(t *testing.T, filename string, actual string) {
...
 
+	// Try standard JSON comparison first
 	if json.Valid(expected) && json.Valid([]byte(actual)) {
-		// if it's valid json compare the objects (ignores spaces/newlines)
 		require.JSONEq(t, string(expected), actual, "JSON Output does not match golden file")
-	} else {
-		// if it's a table, txt, or yaml fallback to exact string matching
-		require.Equal(t, string(expected), actual, "Output does not match golden file")
+		return
 	}
```
evankanderson, line 131: I'm concerned that using `require.JSONeq` here or some other non-byte comparison will cause the golden files to "churn" when we run `--update`.

Taking the output from protojson and running it through the standard Go `encoding/json` module feels like it might be the better approach (and would avoid needing to add a bunch of complex test evaluation logic)

evankanderson: It's less efficient, but we're talking about some in-memory operations on strings <10KB vs filesystem output, so the extra parsing is probably in the noise.

evankanderson (commented):
I think this will probably work, but I'm concerned that we're only partially hiding the underlying non-determinism, because the golden files on-disk won't necessarily have the same formatting each time they are generated.

evankanderson:
(Sorry, was out Friday and had a busy weekend.)  Thanks to @DharunMR for taking a first look at this PR and helping get some of the details correct.

PR #6451, opened 2026-05-11
Add generic entity commands

evankanderson:
# Summary

Implements CRUD CLI operations for interacting with the generic EntityInstanceService

Related to #5377 and possibly #6105

Assisted by Claude Sonnet 4.6 with the following prompts:

> Create CRUD commands for generic Minder entities in the `cmd/cli/app/entity` directory, modelled on the commands in `cmd/cli/app/repo`.

> Switch the commands in `cmd/cli/app/entity` to use `cli.GetCLIClient` rather than `cli.GRPCClientWrapRunE`, and generate tests similar to the tests in `cmd/cli/app/ruletype`.  To generate the golden testdata files, write the tests, and then run the tests with the `--update` flag.  After running the `--update` command once, run the test without the flag to verify that the output is correct.

> Add emoji output to get and list entity commands in table-output mode, where each entity type will be represented by a different emoji.  Fall back to text using the same pattern as `GetStatusIcon` from `internal/util/cli/table`.

And the following Gemini 3 Flash (accidentally, in a different session):
> Update the `entity_list` tests to include test cases for the "--property" flag, using the `-update` flag to the tests to re-generate the golden files.
>
> Also, update the table output for `entity_list` to only show the ID if no properties were requested.

# Testing

Tested manually and with the testing framework from #6386.  Particularly, review and suggest output or usage improvements based on the current files.


evankanderson:
I'd prompt @DharunMR for a review, but his org invite is still outstanding.

Once @sachin9058 and/or @dakshhhhh16 have done their reviews, I'll assign it to someone with CODEOWNERS to do a final pass.

evankanderson:
This also automates maintenance of the GRPC service mocks using `make mock`, the same way we run `mockgen` automatically elsewhere.

cmd/cli/app/entity/entity_register.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return cli.MessageAndError("Error creating gRPC client", err)
+	}
+	defer closeConn()
+
+	project := viper.GetString("project")
+	provider := viper.GetString("provider")
+	format := viper.GetString("output")
+	entityTypeStr := viper.GetString("type")
+	properties := viper.GetStringSlice("property")
```
dakshhhhh16, line 55: Small thing here, the flag is registered with `StringArrayP`, but it's being read back using `viper.GetStringSlice`. Those two behave a bit differently since `StringSlice` will split on commas, while `StringArray` preserves the raw value.

So something like:

```
--property "a=b,c=d"
```
would end up getting split into two entries when going through Viper.

Since property values can reasonably contain commas, I think StringArray is the correct choice for the flag here. But in that case it probably makes more sense to read it directly via:
```
cmd.Flags().GetStringArray("property")
```
instead of viper.GetStringSlice, otherwise values containing commas can get silently corrupted.

cmd/cli/app/entity/entity_list.go
```diff
@@ -0,0 +1,131 @@
...
+	Use:   "list",
+	Short: "List entities",
+	Long:  `The entity list subcommand is used to list entity instances within Minder.`,
+	PreRunE: func(cmd *cobra.Command, _ []string) error {
+		if err := viper.BindPFlags(cmd.Flags()); err != nil {
+			return fmt.Errorf("error binding flags: %w", err)
+		}
+		return nil
+	},
+	RunE: listCommand,
```
dakshhhhh16, line 39: One thing I noticed is that the existing `repo list`, `ruletype list`, `profile list`, etc. commands all validate the output format in `PreRunE` using `app.IsOutputFormatSupported(format)`.

The entity commands seem to skip that check right now, so something like:

```bash
minder entity list --type repository -o foobar
```
just exits without output instead of returning an error.

Might be worth adding for consistency with the other list commands. Looks like it would just be a small one-line check in each command.

cmd/cli/app/entity/entity_get.go
```diff
@@ -0,0 +1,129 @@
...
+		})
+		if err != nil {
+			return cli.MessageAndError("Error getting entity by ID", err)
+		}
+		entity = resp.GetEntity()
+	} else {
+		entityType := minderv1.EntityFromString(entityTypeStr)
+		if entityType == minderv1.Entity_ENTITY_UNSPECIFIED {
+			return fmt.Errorf("invalid or unspecified entity type %q; required when using --name", entityTypeStr)
+		}
```
dakshhhhh16: Just a suggestion as `--type` is only required when `--name` is provided, would it make sense to wire this up using Cobra's:

```go
MarkFlagsRequiredTogether("name", "type")
```

instead of the manual runtime check?

The current approach works, but using Cobra here would give the standard flag validation error and also surface the relationship in `--help` output.

Right now the check runs after `cmd.SilenceUsage = true`, so the user gets the error message but misses the usage hint that would probably help them fix it faster.

cmd/cli/app/entity/entity_register.go
evankanderson: Thanks for flagging this.  It looks like `viper.GetStringSlice` reads the value from `StringArrayP`, rather than re-parsing the slice itself.  I've added a test to verify.

cmd/cli/app/entity/entity_get.go
evankanderson: Good suggestion, thanks!

cmd/cli/app/entity/entity_list.go
evankanderson: Thanks, added for `get`, `list`, and `register` commands that all have `output` fields.

evankanderson (commented):
Thanks for the review; I've addressed all three comments -- the two command-line validation ones by implementing the validation, and the StringSlice/StringArray discrepancy by adding a test that shows that the usage is actually correct, though it may seem strange.

cmd/cli/app/entity/entity_register.go
dakshhhhh16: Rest it looks good from my side!

evankanderson:
Two of the tests were failing (trying to reach the actual Minder server) due to eagerly constructing the gRPC client before checking arguments.  I made the client construction lazier (`list` and `register` had this problem).

JAORMX (approved):

JAORMX (approved):

PR #6462, opened 2026-05-19
Migrate test policies to Rego v1

dakshhhhh16:
## Summary

This PR migrates Minder's embedded Rego test fixtures, sample rule YAML, and Rego docs examples to Rego v1 syntax as part of Phase 2 of #5262. (Parent issue)
Child issue #6463 

The production evaluator already supports Rego version detection. This change moves the policy corpus we own toward the Rego v1 baseline by replacing v0 syntax and `future.keywords` examples with `import rego.v1`.

## What changed

- Migrated embedded Rego snippets in evaluator and rtengine tests to Rego v1 syntax.
- Updated Minder sample/fixture rule YAML and quickstart policy YAML to `import rego.v1`.
- Updated Rego documentation snippets to stop recommending `future.keywords` imports.
- Refreshed the affected mindpak fixture hash after the YAML policy migration.

## Validation

- `git diff --check`
- `go test ./pkg/mindpak ./internal/engine/eval/... ./pkg/engine/v1/rtengine ./pkg/api/protobuf/go/minder/v1 ./cmd/cli/app/ruletype ./cmd/cli/app/quickstart`
- `go test ./...` was also attempted. It reaches unrelated failures in `internal/engine/actions/remediate/pull_request` where tests cannot create a temporary git repository in this local environment.

## Notes for reviewers

This is intended to be a syntax-only migration. The remaining v0 snippets in `internal/engine/eval/rego/version_test.go` are deliberate detector coverage and do not use `future.keywords`.

### DIFF
--- a/cmd/cli/app/quickstart/embed/secret_scanning.yaml
+++ b/cmd/cli/app/quickstart/embed/secret_scanning.yaml
@@ -59,10 +59,12 @@ def:
       def: |
         package minder
 
-        import future.keywords.if
+        import rego.v1
 
         default allow := false
+
         default skip := false
+
         default message := "Secret scanning is disabled"
 
         allow if {
--- a/internal/engine/eval/rego/version_test.go
+++ b/internal/engine/eval/rego/version_test.go
@@ -50,20 +50,17 @@ allow if {
 			wantStr: "v1",
 		},
 		{
-			name: "v0 policy with future.keywords parses as v1",
+			name: "v1 policy with keyword syntax",
 			def: `
 package minder
 
-import future.keywords.if
-import future.keywords.in
+import rego.v1
 
-default allow = false
+default allow := false
 
 allow if {
 	"admin" in input.profile.roles
 }`,
-			// future.keywords policies are accepted by OPA's V1 parser
-			// for backward compatibility during migration.
 			want:    ast.RegoV1,
 			wantStr: "v1",
 		},
--- a/internal/engine/eval/rego/version_test.go
+++ b/internal/engine/eval/rego/version_test.go
@@ -50,11 +50,27 @@ allow if {
 			wantStr: "v1",
 		},
 		{
-			name: "v1 policy with keyword syntax",
+			name: "v0 policy with future.keywords parses as v1",
 			def: `
 package minder
 
-import rego.v1
+import future.keywords.if
+import future.keywords.in
+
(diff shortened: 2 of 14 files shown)


dakshhhhh16:
@JAORMX , @evankanderson 
You can have a look into this now!

cmd/cli/app/quickstart/embed/secret_scanning.yaml
```diff
@@ -59,10 +59,12 @@ def:
       def: |
         package minder
 
-        import future.keywords.if
+        import rego.v1
```
evankanderson, line 62: Do we _need_ `import rego.v1` with the detection we've implemented?  (I think it's fine to leave it in, just curious.)

internal/engine/eval/rego/version_test.go
```diff
@@ -50,20 +50,17 @@ allow if {
...
-import future.keywords.if
-import future.keywords.in
+import rego.v1
 
-default allow = false
+default allow := false
 
 allow if {
 	"admin" in input.profile.roles
 }`,
```
evankanderson: Doesn't this change this test into the next case?
(I think you want to revert this specific test change)

It would also be good to have a test which uses the v1 syntax but does not `import rego.v1` explicitly, to show whether that is accepted or an error.

evankanderson (dismissed):
Sorry for the delayed review, and thanks for cleaning these up!  A couple questions before I merge.

cmd/cli/app/quickstart/embed/secret_scanning.yaml
dakshhhhh16: Yepp!

dakshhhhh16:
@evankanderson 
Minor change done!

evankanderson (approved):
Thanks for adding the test that confirmed my behavior expectation.  It's possible in the future I'll come in and remove some of these `import rego.v1` lines since they aren't necessary (and may distract from the test), but this change seems fine for now.

dakshhhhh16:
> Thanks for adding the test that confirmed my behavior expectation. It's possible in the future I'll come in and remove some of these `import rego.v1` lines since they aren't necessary (and may distract from the test), but this change seems fine for now.

Same here! I'll myself make these changes after all phases are complete. Will soon come up with the third phase pr.

PR #6479, opened 2026-06-01
Fix GitLab Service Account enrollment

evankanderson:
# Summary

#6478 fixed the API validation for GitLab tokens, but the underlying code still expected tokens to be of the "OAuth" format, even when a PAT was explicitly provided.  Relax this constraint (which probably fixes GitHub as well, but the app flow for GitHub generally automates the service account creation and token rotation, so we still prefer that for GitHub).

Additionally, simplify ListAllRepositories (used to populate the list in `repo register`) for GitLab, using the `min_access_level` filter.  This should avoid recommending repos that the token doesn't have sufficient access to.

Includes user documentation on setting up GitLab service account PATs.

# Testing

Manual testing alongside API double-checks with the `glab api` tool.  I was able to create a service account ("Custcodian-minder"), assign it a role in the "custcodian" group, and then use it to list and enroll a repository.

### DIFF
--- a/internal/providers/gitlab/manager/auth.go
+++ b/internal/providers/gitlab/manager/auth.go
@@ -6,6 +6,7 @@ package manager
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/gitlab"
@@ -46,25 +47,19 @@ func (g *providerClassManager) NewOAuthConfig(_ db.ProviderClass, cli bool) (*oa
 func (*providerClassManager) ValidateCredentials(
 	_ context.Context, cred provv1.Credential, _ *m.CredentialVerifyParams,
 ) error {
-	tokenCred, ok := cred.(provv1.OAuth2TokenCredential)
-	if !ok {
+	switch c := cred.(type) {
+	case provv1.OAuth2TokenCredential:
+		_, err := c.GetAsOAuth2TokenSource().Token()
+		if err != nil {
+			return fmt.Errorf("cannot get token from credential: %w", err)
+		}
+	case string:
+		if !strings.HasPrefix(c, "glpat-") {
+			return fmt.Errorf("expected token to start with 'glpat-'")
+		}
+	default:
 		return fmt.Errorf("invalid credential type: %T", cred)
 	}
-
-	_, err := tokenCred.GetAsOAuth2TokenSource().Token()
-	if err != nil {
-		return fmt.Errorf("cannot get token from credential: %w", err)
-	}
-
-	// TODO: verify token identity
-	// if params.RemoteUser != "" {
-	// 	err := g.ghService.VerifyProviderTokenIdentity(ctx, params.RemoteUser, token.AccessToken)
-	// 	if err != nil {
-	// 		return fmt.Errorf("error verifying token identity: %w", err)
-	// 	}
-	// } else {
-	// 	zerolog.Ctx(ctx).Warn().Msg("RemoteUser not found in session state")
-	// }
 
 	return nil
 }
--- a/internal/providers/gitlab/manager/manager.go
+++ b/internal/providers/gitlab/manager/manager.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"net/url"
 	"slices"
+	"strings"
 	"time"
 
 	"github.com/rs/zerolog"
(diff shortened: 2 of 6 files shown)


internal/providers/gitlab/manager/auth.go
```diff
@@ -46,26 +47,20 @@ func (g *providerClassManager) NewOAuthConfig(_ db.ProviderClass, cli bool) (*oa
 func (*providerClassManager) ValidateCredentials(
```
evankanderson, line 47: [The `StoreProviderToken` RPC in handlers_oauth.go](https://github.com/mindersec/minder/blob/main/internal/controlplane/handlers_oauth.go#L554) calls ValidateCredentials with the "AccessToken" RPC field (string type) as an argument.  I didn't refactor the naming in the caller, but this is really just "a bearer token that works for auth" and is not necessarily an OAuth access token.

internal/providers/gitlab/manager/manager.go
```diff
@@ -255,5 +256,7 @@ func (m *providerClassManager) persistToken(
 }
 
 func tokenNeedsRefresh(token oauth2.Token) bool {
-	return !token.Valid() || token.Expiry.UTC().Add(tokenExpirationThreshold).Before(time.Now().UTC())
+	bufferedExpiration := time.Now().UTC().Add(-1 * tokenExpirationThreshold)
+	return !strings.HasPrefix(token.AccessToken, "glpat-") && // is a PAT, not an OAuth token
```
evankanderson, line 260: Again, we store an (encrypted) JSON object that has the same form as `oauth2.Token`, but we only set the `AccessToken` field if it comes from `StoreAccessToken`.  I could also key off the presence of a refresh token or `expiry==0` if that feels cleaner than the string-prefix approach.

internal/providers/gitlab/manager/auth.go
intelligent-ears: Makes sense — the naming is a bit misleading but the type switch makes the actual behavior clear. Could be worth a follow-up to rename the field or add a comment in handlers_oauth.go to clarify it accepts any bearer token, not just OAuth tokens

internal/providers/gitlab/manager/manager.go
intelligent-ears: The glpat- prefix approach works and is explicit. Keying off expiry==0 or absent refresh token might be more robust long-term if GitLab ever changes their token format, but that seems unlikely. I'd leave it as-is and add a comment explaining why we check the prefix — something like // GitLab PATs always start with "glpat-"; OAuth tokens do not.

intelligent-ears:
Reviewed the changes. The type switch in `auth.go` is clean and the `glpat-` prefix validation is consistent with GitLab's token format guarantee. The `repo_lister.go` simplification is a nice improvement — single API call vs iterating groups.

One minor question: `minAccessLevelControl` is defined but I don't see it used in the final `glRESTGet` call — it's hardcoded as `40` in the URL string. Was that intentional, or should `minAccessLevelControl` be used there?

Otherwise LGTM.

internal/providers/gitlab/repo_lister.go
```diff
@@ -6,57 +6,40 @@ package gitlab
...
-	if err := glRESTGet(ctx, c, "groups", &groups); err != nil {
-		return nil, fmt.Errorf("failed to get groups: %w", err)
-	}
+// Access levels, from https://docs.gitlab.com/api/projects/#list-all-projects:
+// 20 => "Reporter"
+// 25 => "Security Manager"
+// 30 => "Developer"
+// 40 => "Maintainer"
+// 50 => "Owner"
+var minAccessLevelControl = 25 // "Security Manager"
```
intelligent-ears, line 22: This variable is defined but the `glRESTGet` call hardcodes `40` in the URL string (projects?min_access_level=40). Should `minAccessLevelControl` be used here instead, or is the hardcoded 40 intentional?

rdimitrov (dismissed):
LGTM! Note the lint job fails though 👍

evankanderson:
> One minor question: `minAccessLevelControl` is defined but I don't see it used in the final `glRESTGet` call — it's hardcoded as `40` in the URL string. Was that intentional, or should `minAccessLevelControl` be used there?

I apparently failed to save the last edit to the file, fixed.

intelligent-ears (approved):

evankanderson:
@rdimitrov - could you approve?

rdimitrov (approved):

PR #6482, opened 2026-06-02
Implement capabilities list via reflection

evankanderson:
# Summary

Implement reporting provider types (facets) via reflection on the provider interface.  This should eliminate possible errors in double-bookkeeping.  Added tests to illustrate the current values reported by the providers.

# Testing

Unit testing, as this is a refactoring change only.

### DIFF
--- a/internal/providers/github/metadata.go
+++ b/internal/providers/github/metadata.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/mindersec/minder/internal/db"
 	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+	provifv1 "github.com/mindersec/minder/pkg/providers/v1"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


internal/providers/github/provider_test.go
```diff
@@ -21,3 +26,75 @@ func TestRegistration(t *testing.T) {
...
+		minderv1.Entity_ENTITY_PULL_REQUESTS,
+		minderv1.Entity_ENTITY_ARTIFACTS,
+		minderv1.Entity_ENTITY_RELEASE,
+	}
+	wantTypes := []minderv1.ProviderType{
+		minderv1.ProviderType_PROVIDER_TYPE_GITHUB,
+		minderv1.ProviderType_PROVIDER_TYPE_GIT,
+		minderv1.ProviderType_PROVIDER_TYPE_REST,
+		minderv1.ProviderType_PROVIDER_TYPE_REPO_LISTER,
+		minderv1.ProviderType_PROVIDER_TYPE_IMAGE_LISTER,
```
evankanderson, line 44: Note that we somehow dropped "image_lister" from the GitHub types during a refactor (which I think was what caused me to notice that this could be done via reflection).

internal/providers/github/metadata.go
```diff
@@ -8,114 +8,68 @@ import (
...
-func ProviderClassInfo(class db.ProviderClass) (*minderv1.ProviderClassInfo, error) {
-	switch class {
+// ProviderClassInfo implements the Provider interface.
+func (c *GitHub) ProviderClassInfo() *minderv1.ProviderClassInfo {
+	supportedTypes := provifv1.ProviderTypesFromImpl(c)
+	supportedEntities := []minderv1.Entity{
+		minderv1.Entity_ENTITY_REPOSITORIES,
+		minderv1.Entity_ENTITY_PULL_REQUESTS,
+		minderv1.Entity_ENTITY_ARTIFACTS,
+		minderv1.Entity_ENTITY_RELEASE,
```
evankanderson, line 37: I'm wondering if we can derive these from the supported facets.  We'd need to add a facet for pull requests, which seems useful anyway.

evankanderson:
(assigning to maintainer-interested folks for review first)

JAORMX (dismissed):

intelligent-ears:
The reflection approach is clean, removes the risk of the manual list getting out of sync with actual implementation, which is exactly what happened with `image_lister` on GitHub. The `ProviderTypesFromImpl` function is easy to understand and the nil-pointer receiver pattern for `ClassInfo()` is a nice touch.

intelligent-ears (dismissed):

evankanderson:
@JAORMX - this needs another review after fixing the lint errors (exhaustive switch, dangling ref to AppImplements)

JAORMX (approved):

Issue #6434, opened 2026-05-04
Rule Testing Infrastructure


evankanderson:
### Please describe the enhancement

Significantly enhance the rule testing and Continuous Integration (CI) tooling for Minder's security policies.

## [Taken from the LFX Mentorship proposal](https://docs.google.com/document/d/1laRud0GSPqVg_rZ3ahD8GfRJYY2Bf-JL66nEXGZJ4kk/edit)

The successful completion of this project will provide Minder users with a robust, automated, and maintainable framework for ensuring the quality and correctness of custom security rules.

* **Improve Rule Testing Tools**: Develop a standalone testing command for Minder rules.
* **Robust Testing**: Allow for the definition and execution of multiple distinct test cases against a single rule.
* **Focus on Local Execution**: Tests should be able to execute locally without requiring network access to external resources.
* **Support Diverse Data Ingestion**: Design the framework to accommodate tests for rules covering all current data ingestion methods: REST API, Git repositories, and declared [data sources](https://docs.mindersec.dev/understand/data_sources).
* **Provide Automated CI Tooling**: Implement CI workflows to automatically detect and run tests.
* **Update Existing Rules**: Migrate and update existing tests within the github.com/mindersec/minder-rules-and-profiles repository to utilize the new framework.
* **Documentation and UX**: Create documentation for the new tooling and address potential client-side UX issues related to testing.

### Solution Proposal

TODO (part of the LFX mentorship to design the solution)

### Acceptance Criteria

Following documentation on the https://docs.mindersec.dev/ website, it should be clear how to set up automated testing for any Minder rules using a GitHub Action.

* The Action should not need to check out or build the Minder source code, and should use (and verify) a released artifact
* The test command should exit `0` on all tests passing (e.g. standard Unix semantics)
* Command output from failed tests should be clear and enable easy debugging and fixing of the rule
* It should be easy to create and store realistic or simplified rule inputs e.g. from `gh api` output
* Testing should support all our existing input types, including data sources as well as REST and Git ingestion
   * It should be easy and clear how to extend the test command for new ingestion types as they are written

evankanderson:
Will assign this to @krrish175-byte when he's added to the mindersec org.  Expected start is June 1.

evankanderson:
> Is the sanitization layer primarily deterministic/rule-based or model-assisted?

I'm not sure what sanitization layer you're referring to (I don't see it in the issue).  Can you explain further what you're asking about?

evankanderson:
[...]

PR #6492, opened 2026-06-09
docs: add rule testing framework design document

krrish175-byte:
## Summary

This PR adds the initial design specification for the Minder rule testing framework, proposed in #6434.

The document follows the structure Evan suggested: Abstract, Background, Goals/Non-Goals, Overview, Detailed Design, Alternatives Considered.

## Key Design Decisions Covered

| Section | Decision |
|---|---|
| DD1 | **Starlark** as the test file format (over Hybrid DSL and txtar+TOML) |
| DD2 | **HTTP-level datasource mocking** (recommended) vs abstract box mocking |
| DD3 | Test discovery and file layout options (flat vs `tests/` subdir) |
| DD4 | Test case declaration -- four approaches under consideration |
| DD5 | **Expect semantics** for assertions (report all failures, don't stop on first) |
| DD6 | Go builtins implementation sketches (`eval`, `read_file`, `txtar`, `check`) |
| DD7 | Error propagation model |
| DD8 | Standalone `ruletest` binary (not a `minder` subcommand for v1) |

## Open Questions

Several design questions are flagged for discussion in the doc:

- `test_*` vs `test()` vs tuples vs `cases` -- which approach ships first?
- How are datasource definitions discovered and loaded?
- File layout: flat (Option A) vs `tests/` subdirectory (Option B)?
- Binary name: `ruletest` vs `mindev`?

## Format Prototypes (Gists)

- [Pure TOML](https://gist.github.com/krrish175-byte/4151ef69e8e0093724faf6d046db20f4)
- [txtar + TOML hybrid](https://gist.github.com/krrish175-byte/73882e069fafe5ba50212a6f93a074ed)
- [Starlark](https://gist.github.com/krrish175-byte/859deffc008d1a8395c7f5d16dd9559b)

Related: #6434

### DIFF
--- a/docs/docs/developer_guide/designs/rule-testing.md
+++ b/docs/docs/developer_guide/designs/rule-testing.md
@@ -1,0 +1,793 @@
+# Minder Rule Testing Framework -- Design
+
+**Author:** Krrish Biswas
+**Mentor:** Evan Anderson
+**Status:** Draft -- Week 1 Design
+
+---
+
+## Abstract
+
+Minder's security policy rules are currently untestable without a live Minder
+server, a real GitHub token, and network access to external APIs. This document
+proposes a standalone rule testing framework that allows rule authors to verify
+rule correctness offline, using mocked HTTP responses and inline file fixtures.
+The framework is built on [Starlark](https://github.com/google/starlark-go)
+embedded in Go, and integrates with Go's native `testing` package for reporting.
+A dedicated binary (`ruletest`) discovers and executes `*.star` test files
+recursively from a directory.
+
+---
+
+## Background
+
+### How Minder rules work
+
+A Minder rule type defines a security policy check for an entity (repository,
+artifact, pull request). Each rule has four phases:
+
+<!-- markdownlint-disable MD013 -->
+
+```text
+  ┌─────────┐     ┌──────────┐     ┌───────────┐     ┌─────────┐
+  │ Ingest  │────▶│ Evaluate │────▶│ Remediate │────▶│  Alert  │
+  └─────────┘     └──────────┘     └───────────┘     └─────────┘
+```
+
+<!-- markdownlint-enable MD013 -->
+
+**Ingest** fetches data from an external source. Three ingest types exist:
+
+| Type | Mechanism | Example |
+|---|---|---|
+| `rest` | HTTP GET to GitHub API | Branch protection settings |
+| `git` | Clone repo, expose `file.ls()` / `file.read()` | GitHub Actions workflows |
+| `datasource` | Named HTTP client callable from Rego | Repository rulesets API |
+
+**Evaluate** runs either a `jq` expression (simple field comparison) or a
+`Rego/OPA` policy against the ingested data.
+
+**Datasources** are named HTTP clients declared in the rule definition. When
+Rego calls
+`minder.datasource.baselineghapi.branch_protection_status({...})`,
+the datasource fetches from a real API endpoint defined in a separate datasource
+YAML. Responses are returned to Rego wrapped in
+`{"body": <payload>, "status": 200}`.
+
+### The current testing gap
+
+There is no way to test a rule without:
+
+- A running Minder server
+- A real GitHub API token
+- Live network access
+
+This means a contributor who writes a new rule type has no way to verify it
+works before opening a PR. Debugging requires a full round-trip: deploy, wait
+for evaluation, observe result, repeat. The datasource `{"body"/"status"}`
+wrapper has caused multi-hour debugging sessions because rule authors assume
+they receive the raw payload.
+
+### Related repositories
+
+- `mindersec/minder` -- Core server, `pkg/engine`, `internal/datasources`
+- `mindersec/minder-rules-and-profiles` -- Rule type YAML files and profiles
+
+---
+
+## Goals and Non-Goals
+
+### Goals
+
+- Allow rule authors to run rule tests **locally**, with no network access and
+  no credentials
+- Support all three ingest types: REST, git, and datasources
+- Integrate with Go's `testing` package for standard test output and CI
+  reporting
+- Make simple tests trivially easy to write (single-function test cases under
+  20 lines)
+- Make hard tests possible (multi-source mocking, parameterized fixtures)
+- **Auto-discover** test files in a directory tree -- no manual test registration
+
+### Non-Goals (v1)
+
+- Remediation output testing (verifying auto-fix PR content)
+- Alert output testing
+- Testing Rego `error()` abort behavior
+- Timeout simulation
+- Becoming a `minder` subcommand in v1 -- ships as a standalone binary first
+
+---
+
+## Overview
+
+```text
+  ┌──────────────────────────────────────┐
+  │    ruletest --dir ./rule-types       │
+  └──────────────┬───────────────────────┘
+                 │
+                 ▼
+  ┌──────────────────────────────────────┐
+  │    Test Discovery                    │
+  │    *.star file walker                │
+  └──────────────┬───────────────────────┘
+                 │
+                 ▼
+  ┌──────────────────────────────────────┐
+  │    Module Loader                     │
+  │    go.starlark.net ExecFile          │
+  └──────────────┬───────────────────────┘
+                 │
+                 ▼
+  ┌──────────────────────────────────────┐
+  │    Predeclared Builtins              │
+  │    eval / read_file / txtar /        │
+  │    check / test / body / code        │
+  └───┬──────────┬───────────┬───────────┘
+      │          │           │
+      ▼          ▼           ▼
+  ┌────────┐ ┌────────┐ ┌────────────┐
+  │  eval  │ │read_file│ │  check.*   │
+  │        │ │+ txtar  │ │  test()    │
+  └───┬────┘ └───┬────┘ └─────┬──────┘
+      │          │             │
+      ▼          ▼             ▼
+  ┌────────┐ ┌────────┐ ┌────────────┐
+  │pkg/    │ │Mock FS │ │Go testing.T│
+  │engine  │ │io/fs.FS│ │Pass / Fail │
+  │Evaluate│ │from    │ │Error       │
+  │Offline │ │txtar   │ └────────────┘
+  └───┬────┘ └────────┘
+      │
+      ▼
+  ┌──────────────────────────────────────┐
+  │    Mock HTTP Layer                   │
+  │    http.RoundTripper intercept       │
+  │                                      │
+  │    - REST ingest calls               │
+  │    - Datasource calls                │
+  └──────────────────────────────────────┘
+```
+
+The test runner (`ruletest`) walks a directory tree, finds all `*.star` files,
+and executes each using an embedded Starlark interpreter. A set of predeclared
+Go builtins (`eval`, `read_file`, `txtar`, `check`, `test`) are injected into
+each module's environment. The builtins delegate to the existing `pkg/engine`
+evaluation pipeline -- the same jq and Rego engines used in production -- with
+HTTP calls intercepted by a mock layer.
+
+---
+
+## Detailed Design
+
+### DD1: Test File Format -- Starlark
+
+After prototyping three formats (Hybrid DSL, txtar+TOML, Starlark), Starlark
+is selected as the test file format. See
+[Alternatives Considered](#alternatives-considered) for the comparison.
+
+The critical advantage: a real programming language gives parameterized
+fixtures, shared helpers, and composable defaults without inventing a custom
+templating syntax.
+
+**Wrapper functions replace `[defaults]` blocks:**
+
+```python
+# Default args carry shared setup -- tests only specify what varies
+def workflow_rule(ref, files="check_pinned.txtar", entity=DEFAULT_ENTITY):
+    fs = {}
+    if files != None:
+        fs = {k: v.format(ref=ref) for k, v in txtar(read_file(files)).items()}
+    return eval(
+        rule   = "actions_check_pinned_tags",
+        entity = entity,
+        files  = fs,
+    )
+
+# One template txtar, two test variants -- no duplication
+def test_pinned():   check.eq(workflow_rule(PINNED_SHA).status, "pass")
+def test_floating(): check.eq(workflow_rule("v4").status, "fail")
+```
+
+**txtar as file-bundling layer:**
+
+```text
+-- .github/workflows/ci.yml --
+name: CI
+on: [push]
+jobs:
+  build:
+    steps:
+      - uses: actions/checkout@{ref}
+```
+
+The `{ref}` placeholder is substituted via Starlark's `.format()` -- enabling
+parameterized git fixture files without duplicating workflow content per test
+case.
+
+---
+
+### DD2: Datasource Mocking Strategy
+
+This is the most significant design decision in the framework.
+
+**Two approaches:**
+
+```text
+  Approach A -- Abstract Box          Approach B -- HTTP-Level (recommended)
+  ┌───────────────────────┐          ┌───────────────────────┐
+  │ test mocks             │          │ test mocks             │
+  │ datasource:name/method │          │ GET api.github.com/... │
+  └──────────┬────────────┘          └──────────┬────────────┘
+             │                                  │
+             ▼                                  ▼
+  ┌───────────────────────┐          ┌───────────────────────┐
+  │ Mock Layer             │          │ Mock HTTP interceptor  │
+  │ skips datasource       │          └──────────┬────────────┘
+  └──────────┬────────────┘                      │
+             │                                   ▼
+             │                       ┌───────────────────────┐
+             │                       │ Datasource definition  │
+             │                       │ processes real response │
+             │                       └──────────┬────────────┘
+             ▼                                  ▼
+  ┌───────────────────────┐          ┌───────────────────────┐
+  │ Rego                   │          │ Rego                   │
+  │ receives mocked payload│          │ receives datasource    │
+  └───────────────────────┘          │ output                 │
+                                     └───────────────────────┘
+```
+
+#### Approach A -- Abstract box mocking
+
+The test mocks the *output* of the datasource call directly, bypassing the
+datasource definition:
+
+```python
+mocks = {
+    "datasource:baselineghapi/branch_protection_status": {
+        "body": {"applied_rulesets": [{"type": "non_fast_forward"}]},
+    },
+}
+```
+
+- **Advantage:** Simple to write. No knowledge of the underlying API shape
+  needed.
+- **Disadvantage:** The datasource definition is never exercised. If the
+  datasource has a bug (wrong endpoint, wrong field mapping), tests still pass.
+  Authors also need to know the datasource's output shape, which is the
+  `{"body"/"status"}` wrapper.
+
+#### Approach B -- HTTP-level mocking (recommended)
+
+The test mocks at the *HTTP boundary*. The datasource definition is loaded, and
+its HTTP calls are intercepted. Rule authors write mocks against real GitHub API
+shapes:
+
+```python
+# Test: classic branch protection blocks force pushes
+mocks = {
+    "GET https://api.github.com/repos/acme-corp/widgets/branches/main/protection":
+        body({"allow_force_pushes": {"enabled": False}}),
+    "GET https://api.github.com/repos/acme-corp/widgets/rules/branches/main":
+        body([{"ruleset_source_type": "Repository", "type": "non_fast_forward"}]),
+}
+```
+
+```python
+# Test: branch not protected (separate test case -- different mock for same endpoint)
+mocks = {
+    "GET https://api.github.com/repos/acme-corp/widgets/branches/main/protection":
+        code(404),
+}
+```
+
+`body(payload)` and `code(status)` are Go-supplied Starlark builtins:
+
+- `body(x)` -- returns a 200 response with payload `x`
+- `code(n)` -- returns an empty response with HTTP status `n`
+
+**Advantages:**
+
+- The datasource definition is loaded and exercised.
+- Authors can go directly from GitHub API docs to test mocks -- no knowledge of
+  datasource internals needed.
+- The `{"body"/"status"}` wrapper is applied by the datasource as in
+  production, so authors don't need to think about it.
+
+**Disadvantages:**
+
+- Requires loading datasource definitions.
+- Needs a discovery mechanism for where datasource YAMLs live.
+- URL patterns need glob support for path params (e.g.
+  `repos/*/branches/*`).
+
+**Recommendation:** Approach B. The test exercises more of the real pipeline
+and aligns test mocks with GitHub API documentation rather than internal Minder
+abstractions.
+
+**Open question:** How are datasource definitions discovered? Options:
+
+1. Inferred from the rule YAML (which declares
+   `data_sources: [{name: baselineghapi}]`)
+2. Co-located YAML in the same directory
+3. Declared explicitly in the test file
+
+---
+
+### DD3: Test Discovery and File Layout
+
+#### File layout options
+
+The current `minder-rules-and-profiles` structure places rule YAML files in
+flat directories:
+
+```text
+rule-types/github/
+├── branch_protection_allow_force_pushes.yaml
+├── osps-ac-03-01.yaml
+└── actions_check_pinned_tags.yaml
+```
+
+Two co-location options for test files:
+
+**Option A -- Flat directory (alongside rule YAML):**
+
+```text
+rule-types/github/
+├── branch_protection_allow_force_pushes.yaml
+├── branch_protection_allow_force_pushes.star   # co-located
+├── osps-ac-03-01.yaml
+└── osps-ac-03-01.star
+```
+
+- **Advantage:** Easy to see if a rule has a test -- it's right next to the
+  YAML.
+- **Disadvantage:** Directory grows with two files per rule; no visual
+  separation between "config" and "test".
+
+**Option B -- Subdirectory for tests:**
+
+```text
+rule-types/github/
+├── branch_protection_allow_force_pushes.yaml
+├── osps-ac-03-01.yaml
+└── tests/
+    ├── branch_protection_allow_force_pushes.star
+    └── osps-ac-03-01.star
+```
+
+- **Advantage:** Clear separation between rule definitions and test code.
+  Easier to see test coverage at a glance.
+- **Disadvantage:** Adding a test requires navigating to a different directory;
+  rule and test are farther apart.
+
+Avoid **one directory per rule** -- creates many tiny directories, especially
+when rules have no test yet.
+
+#### Discovery
+
+```bash
+ruletest --dir ./rule-types
+```
+
+Recursive walk finds all `*.star` files. The directory path acts as the
+implicit test suite -- no explicit suite configuration in v1.
+
+**Rule loading:** Test runner loads all `*.yaml` files in the same directory
+(or parent directory if using `tests/` subdir) as the `*.star` file, then
+filters by the name passed to `eval(rule="name")`. Duplicate rule names in the
+same directory: "don't do that" is sufficient for v1.
+
+---
+
+### DD4: Test Case Declaration
+
+Four approaches are under active consideration.
+
+#### Approach A -- `test_*` function discovery
+
+Runner scans module globals after execution for no-arg callables whose names
+start with `test_`. Function name becomes the test identifier.
+
+```python
+def test_force_pushes_disabled():
+    result = eval(rule="branch_protection_allow_force_pushes",
+                  entity=DEFAULT_ENTITY,
+                  mocks={ENDPOINT: body({"allow_force_pushes": {"enabled": False}})})
+    check.eq(result.status, "pass")
+```
+
+- **Advantage:** Familiar (pytest, Go `TestXxx`). Function name is naturally
+  the test name. Custom assertions are straightforward.
+- **Disadvantage:** Implementation requires invoking Starlark functions after
+  module definition with correct thread context. Table-driven tests require
+  all cases inside one `test_*` function.
+
+**Implementation note:** This requires using `go.starlark.net`'s `Value`
+interface to scan module globals and call functions. Worth prototyping with a
+small Go program before committing to this approach.
+
+#### Approach B -- `test()` builtin with `expect`
+
+Test evaluation is a side effect of module loading. `test()` wraps both rule
+evaluation and result validation:
+
+```python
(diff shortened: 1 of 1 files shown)


docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+**Author:** Krrish Biswas
+**Mentor:** Evan Anderson
+**Status:** Draft -- Week 1 Design
+
+---
+
+## Abstract
+
+Minder's security policy rules are currently untestable without a live Minder
+server, a real GitHub token, and network access to external APIs. This document
```
evankanderson: This isn't true.  Let's not overstate the case in order to justify it.  Currently, there are two ways to test Minder rules without a live server:

* `mindev rtst` - this runs a single evaluation against live providers with a manually-specified entity.  It requires a full profile definition and rule alongside the entity, and live credentials for the provider.  The live credentials and backend services makes this a problem for automated testing.
* `rules_test.go` from https://github.com/mindersec/minder-rules-and-profiles - this runs against test data and supports batch operation, but has several suboptimal features.  The framework does not support datasources, and is verbose when testing multiple cases.  Additionally, file test data representing vulnerabilities appears "live" to tools like dependabot or trivy, and it is hard to re-use the tool outside that particular repo. Finally, having a stand-alone binary in a different repository introduces a bunch of dependency churn management that wouldn't be required if it were an official binary in the main repo.

Given that `rules_test.go` is about 300 lines, it's not unreasonable to simply rewrite the functionality rather than attempting to extend it.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+A Minder rule type defines a security policy check for an entity (repository,
+artifact, pull request). Each rule has four phases:
+
+<!-- markdownlint-disable MD013 -->
+
+```text
+  ┌─────────┐     ┌──────────┐     ┌───────────┐     ┌─────────┐
+  │ Ingest  │────▶│ Evaluate │────▶│ Remediate │────▶│  Alert  │
+  └─────────┘     └──────────┘     └───────────┘     └─────────┘
+```
```
evankanderson: Consider using [mermaid](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams) diagrams here rather than text diagrams.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+
+<!-- markdownlint-enable MD013 -->
+
+**Ingest** fetches data from an external source. Three ingest types exist:
+
+| Type | Mechanism | Example |
+|---|---|---|
+| `rest` | HTTP GET to GitHub API | Branch protection settings |
+| `git` | Clone repo, expose `file.ls()` / `file.read()` | GitHub Actions workflows |
+| `datasource` | Named HTTP client callable from Rego | Repository rulesets API |
```
evankanderson: Technically, datasources happen during the `Evaluate` phase, rather than the `Ingest` phase.  This allows them to use data determined during evaluation to fetch more data.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+  └─────────┘     └──────────┘     └───────────┘     └─────────┘
+```
+
+<!-- markdownlint-enable MD013 -->
+
+**Ingest** fetches data from an external source. Three ingest types exist:
+
+| Type | Mechanism | Example |
+|---|---|---|
+| `rest` | HTTP GET to GitHub API | Branch protection settings |
```
evankanderson: This is the provider's API surface; the existing examples are GitHub, but for a GitLab provider, this would go to the GitLab API, and the DockerHub provider would go to that API (which is a superset of the OCI API and includes managing registry permissions and the like).

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+| `git` | Clone repo, expose `file.ls()` / `file.read()` | GitHub Actions workflows |
+| `datasource` | Named HTTP client callable from Rego | Repository rulesets API |
+
+**Evaluate** runs either a `jq` expression (simple field comparison) or a
+`Rego/OPA` policy against the ingested data.
+
+**Datasources** are named HTTP clients declared in the rule definition. When
```
evankanderson: You don't define "remediate" or "alert" steps but include them in the diagram above.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+Rego calls
+`minder.datasource.baselineghapi.branch_protection_status({...})`,
+the datasource fetches from a real API endpoint defined in a separate datasource
+YAML. Responses are returned to Rego wrapped in
+`{"body": <payload>, "status": 200}`.
```
evankanderson: Not sure if you want to reference https://docs.mindersec.dev/understand/data_sources here.

(And yes, that documentation might benefit from updates)

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+
+- A running Minder server
+- A real GitHub API token
+- Live network access
+
+This means a contributor who writes a new rule type has no way to verify it
+works before opening a PR. Debugging requires a full round-trip: deploy, wait
+for evaluation, observe result, repeat. The datasource `{"body"/"status"}`
+wrapper has caused multi-hour debugging sessions because rule authors assume
+they receive the raw payload.
```
evankanderson: Incorporate the information from https://github.com/mindersec/minder/pull/6492/changes#r3390004310 rather than repeating the abstract.

docs/docs/developer_guide/designs/rule-testing.md
evankanderson: For the abstract, I'd write "there isn't a standardized way to test Minder rules outside of a live environment" (which is true, because `rules_test.go` is not shipped as a Minder artifact).

The rest of this info probably belongs in the "background" section.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+- Alert output testing
+- Testing Rego `error()` abort behavior
+- Timeout simulation
+- Becoming a `minder` subcommand in v1 -- ships as a standalone binary first
+
+---
+
+## Overview
+
+```text
```
evankanderson: Again, I think mermaid will be easier to write, read, and update.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+  │    Mock HTTP Layer                   │
+  │    http.RoundTripper intercept       │
+  │                                      │
+  │    - REST ingest calls               │
+  │    - Datasource calls                │
+  └──────────────────────────────────────┘
+```
+
+The test runner (`ruletest`) walks a directory tree, finds all `*.star` files,
+and executes each using an embedded Starlark interpreter. A set of predeclared
```
evankanderson: Question: can one `*.star` file import another for e.g. a shared library of additional functions?

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+Go builtins (`eval`, `read_file`, `txtar`, `check`, `test`) are injected into
+each module's environment. The builtins delegate to the existing `pkg/engine`
+evaluation pipeline -- the same jq and Rego engines used in production -- with
+HTTP calls intercepted by a mock layer.
```
evankanderson:
```suggestion
HTTP and Git filesystem calls intercepted by a mock layer.
```

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+
+### DD1: Test File Format -- Starlark
+
+After prototyping three formats (Hybrid DSL, txtar+TOML, Starlark), Starlark
+is selected as the test file format. See
+[Alternatives Considered](#alternatives-considered) for the comparison.
+
+The critical advantage: a real programming language gives parameterized
+fixtures, shared helpers, and composable defaults without inventing a custom
+templating syntax.
```
evankanderson: I'd be less "punchy" in a design document.  Something more like "Programming language execution provides benefits such as ..." rather than "The critical advantage: "

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+        entity = entity,
+        files  = fs,
+    )
+
+# One template txtar, two test variants -- no duplication
+def test_pinned():   check.eq(workflow_rule(PINNED_SHA).status, "pass")
+def test_floating(): check.eq(workflow_rule("v4").status, "fail")
+```
+
+**txtar as file-bundling layer:**
```
evankanderson: This seems like a separate heading?

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+parameterized git fixture files without duplicating workflow content per test
+case.
+
+---
+
+### DD2: Datasource Mocking Strategy
+
+This is the most significant design decision in the framework.
+
+**Two approaches:**
```
evankanderson: If you have two approaches, highlight the one that you've chosen, and put the other in "alternatives considered".  It makes it easier for readers to understand what's being built, and still gives a record of what was considered and rejected when someone wants to revisit the decision.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+        code(404),
+}
+```
+
+`body(payload)` and `code(status)` are Go-supplied Starlark builtins:
+
+- `body(x)` -- returns a 200 response with payload `x`
+- `code(n)` -- returns an empty response with HTTP status `n`
+
+**Advantages:**
```
evankanderson, line 255: One other advantage is that datasources and ingest fetches should be consistent.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+**Recommendation:** Approach B. The test exercises more of the real pipeline
+and aligns test mocks with GitHub API documentation rather than internal Minder
+abstractions.
+
+**Open question:** How are datasource definitions discovered? Options:
+
+1. Inferred from the rule YAML (which declares
+   `data_sources: [{name: baselineghapi}]`)
+2. Co-located YAML in the same directory
+3. Declared explicitly in the test file
```
evankanderson: Look at the existing usage and see what options make sense.

I think 1 and 2 don't really work, so you'll need some form of 3.  I don't know if this should be a module-wide (global) state change, or if this should be part of the rule evaluation.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+flat directories:
+
+```text
+rule-types/github/
+├── branch_protection_allow_force_pushes.yaml
+├── osps-ac-03-01.yaml
+└── actions_check_pinned_tags.yaml
+```
+
+Two co-location options for test files:
```
evankanderson: Again, pick an option.  You can mention the alternative if it makes sense, but this should spell out what we're going to do.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+├── branch_protection_allow_force_pushes.yaml
+├── branch_protection_allow_force_pushes.star   # co-located
+├── osps-ac-03-01.yaml
+└── osps-ac-03-01.star
+```
+
+- **Advantage:** Easy to see if a rule has a test -- it's right next to the
+  YAML.
+- **Disadvantage:** Directory grows with two files per rule; no visual
+  separation between "config" and "test".
```
evankanderson: With `*.star` as a suffix, you can at least use globs to distinguish.  (`*.test.yaml` unfortunately is matched by `*.yaml` today).

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+ruletest --dir ./rule-types
+```
+
+Recursive walk finds all `*.star` files. The directory path acts as the
+implicit test suite -- no explicit suite configuration in v1.
+
+**Rule loading:** Test runner loads all `*.yaml` files in the same directory
+(or parent directory if using `tests/` subdir) as the `*.star` file, then
+filters by the name passed to `eval(rule="name")`. Duplicate rule names in the
+same directory: "don't do that" is sufficient for v1.
```
evankanderson: I'd translate "don't do that" to "is detected and generates an error" when specifying what to do.  It's a difference between informal and formal writing.  😁

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+
+---
+
+### DD4: Test Case Declaration
+
+Four approaches are under active consideration.
```
evankanderson: Pick a path.  We can revisit this and update the design doc later with learnings, but we need to pick _some_ solution to get moving.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+                  entity=DEFAULT_ENTITY,
+                  mocks={ENDPOINT: body({"allow_force_pushes": {"enabled": False}})})
+    check.eq(result.status, "pass")
+```
+
+- **Advantage:** Familiar (pytest, Go `TestXxx`). Function name is naturally
+  the test name. Custom assertions are straightforward.
+- **Disadvantage:** Implementation requires invoking Starlark functions after
+  module definition with correct thread context. Table-driven tests require
+  all cases inside one `test_*` function.
```
evankanderson: Have you looked at how hard this is to do?  If it's easy, I'm fine with doing this for now.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+
+cases = {
+    "pinned":   {"ref": PINNED_SHA,   "expect": "pass"},
+    "floating": {"ref": FLOATING_TAG, "expect": "fail"},
+}
+[run_test(k, v) for k, v in cases.items()]
+```
+
+- **Advantage:** Most compact for parametrized tests.
+- **Disadvantage:** Dict key is the test name (data, not structure) -- harder
```
evankanderson: If we do this, you could do a "list of objects" instead of a dictionary, e.g.

```
cases = [
  { "name": "pinned", "ref": PINNED_SHA, "expect": "pass" },
  { "name": "floating", "ref": FLOATING_TAG, "expect": "fail" },
]
```

This would be closer to a typical table-driven test in Golang.  I think that makes this work for either approach B or approach C.  I'm not sure how it would work for approach A (maybe the check.eq would be in a list comprehension?).

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+check.eq(result.status, "pass")           # continue even if fails
+check.eq(len(result.violations), 1)       # continue even if fails
+check.contains(result.violations[0].msg, "unpinned")
+```
+
+All failures are collected and reported together when the test case completes.
+
+**Implementation:** The `check` builtin maintains a thread-local error list
+via `starlark.Thread`'s local storage. At test completion, the runner reads
+this list and reports all collected failures.
```
evankanderson: 👍

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+
+#### Extended check helpers as a Starlark module
+
+Rather than implementing every check variant as a Go builtin, a Starlark module
+can define higher-level helpers that call the primitive `check.eq` and
+`check.contains` builtins:
+
+```python
+# checks.star (shipped with ruletest, auto-loaded)
```
evankanderson: Is this something that could simply live as another `*.star` file which is [`load`ed](https://starlark-lang.org/spec.html#load-statements)?

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+    "check":     makeCheckModule(),       // expect-style assertions
+    "body":      makeBody(),              // mock response helper
+    "code":      makeCode(),              // mock error response helper
+}
+
+thread := &starlark.Thread{Name: filepath.Base(filename)}
+globals, err := starlark.ExecFile(thread, filename, nil, predeclared)
+```
+
+#### `eval` -- rule evaluator
```
evankanderson: I don't think you need these proposed implementations in the design document.  They may be handy for your initial draft, but the design document should be focused on:

1. The external interfaces (what users see)
2. Feasibility of implementation (are there blocks on implementing 1)

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+
+# Single file
+ruletest --file rule-types/github/branch_protection_allow_force_pushes.star
+
+# Output: standard Go test output
+--- PASS: branch_protection_allow_force_pushes/test_force_pushes_disabled (0.03s)
+--- FAIL: branch_protection_allow_force_pushes/test_force_pushes_enabled (0.02s)
+    check.eq: got "pass", want "fail"
+FAIL
+```
```
evankanderson: This doesn't seem like an important design element, so I'd probably omit it.

evankanderson: (For example, we might later decide to include it in some other binary, and it wouldn't have much effect on the rest of the design.)

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+  `[[array.of.tables]]` gives clear test boundaries. TOML dotted-key notation
+  (`body.allow_force_pushes.enabled = false`) is clean for shallow data.
+- **Cons:** `[[array.of.tables]]` nesting becomes hard to follow when a test
+  case has multiple mock sources. No parameterization. Mock data and test
+  definition are in the same file, coupling "dumb data" with "active logic".
+
+txtar remains useful as a **file-bundling layer** alongside Starlark for
+git-ingest test fixtures.
+
+### A3: Embedded real language (Python, Lua, JavaScript)
```
evankanderson: Since this is "alternatives", I'd Skylark out of the list and say:


```suggestion
### A3: Alternative embedded language
```

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+- **Starlark (chosen):** Designed for embedding in Go tools. Deterministic
+  (no I/O, no randomness by default). `go.starlark.net` is mature (used by
+  Bazel). Python-like syntax. Sandboxable.
+
+### A4: Datasource abstract box mocking
+
+Mock the output of the datasource call directly (see DD2 -- Approach A).
+Simpler to write but doesn't exercise the datasource definition. The
+`{"body"/"status"}` wrapper becomes a concern for test authors. Rejected in
+favor of HTTP-level mocking.
```
evankanderson: I'd put the "Approach A" here if you've chosen approach B.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,793 @@
...
+## Open Questions
+
+| # | Question | Status |
+|---|---|---|
+| Q1 | `test_*` vs `test()` vs tuples vs `cases` -- which approach ships first? | Open -- prototype all four |
+| Q2 | How are datasource definitions discovered and loaded? | Open |
+| Q3 | File layout: Option A (flat) vs Option B (`tests/` subdir)? | Open |
+| Q4 | `check` module: ship as Go builtins or auto-loaded Starlark module? | Leaning Starlark module |
+| Q5 | Binary name: `ruletest` vs `mindev`? | Open |
+| Q6 | Glob pattern matching for URL mocks (`repos/*/branches/*`)? | Needed for Approach B |
```
evankanderson: Pick an answer for these.  We can always adjust later, but it's helpful if the design doc lays out clearly what will be built (to the best of our knowledge), rather than what **can** be built.

(Of course, it's fine to have these in while drafting a design document, but I'd like to get this completed so we can get on to the "build it and live with our choices" phase.  😁 )

evankanderson (commented):
A bunch of comments, but this is looking pretty good.  Once you've made the adjustments from these comments, we should publicize this on the `#minder` OpenSSF channel for additional comments.

krrish175-byte:
Thanks for the thorough review! I have pushed an updated version that addresses all the comments:

- **Picked concrete paths for the open questions** (Starlark `test_*` function discovery with list-of-objects for table-driven tests, HTTP-level mocking, explicitly defining datasources in the test file, flat directory layout).
- **Restructured the document** so the chosen paths are in the main design, and rejected options are moved to 'Alternatives Considered'.
- **Replaced the ASCII diagrams** with Mermaid.
- **Removed the Go implementation code** as suggested to focus on the external interfaces.
- **Addressed the factual corrections** regarding `rules_test.go`, ingest vs evaluate phases, and REST providers.
- **Adjusted tone** and removed the binary name discussion.

Let me know if there's anything else needed before sharing this with the `#minder` channel!

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+| `rest` | HTTP GET to provider API | Branch protection settings |
+| `git` | Clone repo, expose `file.ls()` / `file.read()` | GitHub Actions workflows |
+
+The provider determines which API surface is used. For a GitHub
+provider, `rest` calls the GitHub API; for a GitLab provider, it
+calls the GitLab API; and so on.
+
+**Evaluate** runs either a `jq` expression (simple field comparison)
+or a `Rego/OPA` policy against the ingested data. During evaluation,
+rules may also call **datasources** -- named HTTP clients declared
```
evankanderson, line 50: Does it make sense to illustrate data sources in the diagram (to/from arrow with `Evaluate`)?

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+The `{ref}` placeholder is substituted via Starlark's `.format()`,
+enabling parameterized git fixture files without duplicating
+workflow content per test case.
+
+---
+
+### DD2: Datasource Mocking -- HTTP-Level Interception
+
+Datasource mocking is the most significant design decision in the
+framework. The chosen approach mocks at the **HTTP boundary**. The
```
evankanderson: Again, I'd strike the "most significant design decision" language.  You shouldn't need to "sell" in a design document -- lay out the facts and let the reader reach their own conclusions.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+```python
+# Test: classic branch protection blocks force pushes
+mocks = {
+    "GET https://api.github.com/repos/acme-corp/widgets"
+    "/branches/main/protection":
+        body({"allow_force_pushes": {"enabled": False}}),
+    "GET https://api.github.com/repos/acme-corp/widgets"
+    "/rules/branches/main":
+        body([{"ruleset_source_type": "Repository",
+               "type": "non_fast_forward"}]),
```
evankanderson, line 227: Not sure if you want to put in an example here with `read_file` and `txtar`:


```suggestion
        body(txtar(read_file("api_responses.txtar"))["main-protected.json"]),
```

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+    datasources = ["path/to/baselineghapi.yaml"],
+    mocks = {
+        "GET https://api.github.com/repos/*/branches/*/protection":
+            body({"allow_force_pushes": {"enabled": False}}),
+    },
+)
+```
+
+See [Alternatives Considered](#a4-datasource-abstract-box-mocking)
+for the rejected approach of mocking datasource outputs directly.
```
evankanderson: I don't think you need to specify that the approach was rejected (since that's implicit in the document).  I think it's also fine to simply avoid the forward-references to "alternatives considered" if we have a table of contents at the beginning.

```suggestion
for other approaches to mocking datasource outputs.
```

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+            "allow_force_pushes": {"enabled": True},
+        })},
+    )
+    check.eq(result.status, "fail")
+```
+
+This approach requires scanning `starlark.StringDict` globals after
+`ExecFile` and invoking matching functions. The `go.starlark.net`
+library provides `starlark.Call()` for this, making implementation
+straightforward.
```
evankanderson, line 357: 👍

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+                 label=case["name"])
+```
+
+---
+
+### DD5: Assertions -- Expect Semantics
+
+The `check` module uses **expect semantics** rather than assert
+semantics. Per the
+[Google Testing Blog](https://testing.googleblog.com/2008/07/tott-expect-vs-assert.html):
```
evankanderson: I don't think you need to name the link, just link the contents:

```suggestion
The `check` module uses [**expect semantics** rather than assert
semantics](https://testing.googleblog.com/2008/07/tott-expect-vs-assert.html):
```

(I got the google blog link from a simple Kagi search on "expect vs assert".  There are probably a bunch of other reasonable links.)

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+    found = False
+    for v in result.violations:
+        if msg in v.msg:
+            found = True
+    check.eq(found, True)
+```
+
+```python
+# In a test file:
+load("checks.star", "violations_count", "violation_contains")
```
evankanderson, line 433: ❤️  for "no magic" and just using existing features.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+
+### DD7: Error Propagation
+
+```mermaid
+sequenceDiagram
+    participant Runner
+    participant Starlark
+    participant Check as Check Module
+
+    Runner->>Starlark: ExecFile(test.star)
```
evankanderson, line 476: Won't this return a set of function definitions, and then the Runner will call Starlark with `starlark.Call(function)` for each function, and catch the error from each function independently?

This also allows `check.*` methods to be associated with the corresponding test function.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+    Starlark-->>Runner: module complete
+    Runner->>Check: read thread-local error list
+    Note over Runner: Report all 2 failures
+```
+
+`eval()` errors (e.g. unmocked HTTP call, unknown rule name) return
+Go `error`, which surfaces as a Starlark exception and stops the
+current function. This is appropriate because a broken test setup
+should fail immediately rather than silently continue with invalid
+state.
```
evankanderson: I think _many_ times, the `eval()` will be inside a `def test_...()` function, so it won't actually be called / return error on the module load, but instead when the function is `starlark.Call()`ed.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+          enabled: false
+  then:
+    result: pass
+```
+
+- **Pros:** Readable for non-programmers. Clear visual separation
+  of inputs and assertions.
+- **Cons:** Requires custom tokenizer and BNF grammar. No
+  parameterization or loops. No sharing of test setup across cases.
+  Every check variant requires a grammar change.
```
evankanderson, line 527: Oof, this is particularly painful.  Glad we figured this out.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,630 @@
...
+Rejected because co-location makes it immediately visible whether
+a rule has a test, and the `*.star` suffix provides sufficient
+distinction from `*.yaml` files.
+
+### A6: Alternative test declaration approaches
+
+Three other test declaration approaches were considered alongside
+`test_*` function discovery (DD4):
+
+**`test()` builtin with `expect`:**
```
evankanderson:
```suggestion
**`test()` builtin with `expect` argument:**
```

evankanderson (dismissed):
Looking good!  I'm okay merging this as-is, or doing one more round of document updates.  Most of the comments are just language tweaks that you can ignore, but the flow diagram in DD07 has some practical changes that may be worth updating the doc with.

I'm also wondering if you'd want to hoist the DD07 diagram to earlier in the document, but I think it's okay either way.

krrish175-byte:
Pushed another revision addressing the second round of comments:

- **Pipeline diagram**: Added `Datasource` node with bidirectional arrow to `Evaluate`.
- **DD7 sequence diagram**: Fixed to show the actual execution flow -- `ExecFile` loads the module and returns globals, then the runner calls each `test_*` function individually via `starlark.Call()`. Errors and `check.*` failures are now scoped per function.
- **Tone**: Removed "most significant design decision" language.
- **txtar for mock bodies**: Added example showing `body(txtar(read_file("api_responses.txtar"))["main-protected.json"])`.
- **Minor wording fixes**: Simplified the alternatives forward-reference, inlined the Google Blog link, fixed A6 heading.

Thanks for the approval! Let me know if you'd like any more adjustments before sharing on `#minder`.

docs/docs/developer_guide/designs/rule-testing.md
```diff
@@ -0,0 +1,649 @@
...
+
+def test_workflow_refs():
+    for case in cases:
+        result = eval(
+            rule = "actions_check_pinned_tags",
+            entity = DEFAULT_ENTITY,
+            files = workflow_files(case["ref"]),
+        )
+        check.eq(result.status, case["expect"],
+                 label=case["name"])
```
evankanderson, line 379: The "label" kwarg is a nice usage of a Starlark capability that Go doesn't have to allow us to name _some_ check results if necessary, without needing a name for all of them.  (You don't need to update the doc, I just wanted to call out this nice pattern.)

evankanderson (approved):

PR #6499, opened 2026-06-10
Simplify CreateProfile, fix bug filling list of profiles

evankanderson:
# Summary

I was chasing a different bug, and noticed that CreateProfile was attempting to name unnamed profiles uniquely, but actually discarded the map of profile names.  Testing for membership in a map will also be faster (and more correct) than repeated `string.Contains`.  It turns out that some of the tests were also incorrect (wrong number of characters chopped), so fixed that as well.

# Testing

Verified via unit tests.

### DIFF
--- a/pkg/profiles/service.go
+++ b/pkg/profiles/service.go
@@ -135,10 +135,8 @@ func (p *profileService) CreateProfile(
 
 	profileMap := MergeDatabaseListIntoProfiles(existingProfiles)
 
-	existingProfileNames := make([]string, 0, len(profileMap))
-
 	// Derive the profile name from the profile display name
-	name := DeriveProfileNameFromDisplayName(profile, existingProfileNames)
+	name := DeriveProfileNameFromDisplayName(profile, profileMap)
 
 	// if empty use the name
 	if displayName == "" {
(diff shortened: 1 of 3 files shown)


dakshhhhh16 (approved):

krrish175-byte (approved):
LGTM! Clean fix for a real bug.

The old code allocated `existingProfileNames` as an empty slice and never populated it, so `DeriveProfileNameFromDisplayName` always saw zero existing names and the collision check was a no-op.

Switching from `strings.Contains(strings.Join(...))` to a direct map lookup also fixes the false-positive substring matching problem (e.g. a derived name `foo` would have incorrectly collided with an existing profile named `foobar`).

The test string corrections (`_when` -> `_whe`) make sense since the 63-char limit truncates one character earlier than the old tests assumed.

intelligent-ears:
LGTM. The map lookup is correct and eliminates both the unpopulated slice bug and the substring false-positive. The test corrections for the 63-char truncation boundary make sense — the old tests were asserting wrong behavior

intelligent-ears (approved):

evankanderson:
One additional note for reviewers -- the map is _probably_ faster than checking `string.Contains`, though you'd need benchmarking to be sure of it.  (`string.Contains` probably benefits from linear memory layout, but suffers from repeated branch checks / prediction in comparison with the Go map implementation ([swiss tables](https://go.dev/blog/swisstable)).)

PR #6500, opened 2026-06-11
Clean up profile handling, avoid doubling eval status in ListEvaluationResults

evankanderson:
# Summary

It turns out a profile which contains rules for multiple entity types will end up duplicating the evaluation results due to the (previous) logic handling profiles from ListProfilesByProjectIDAndLabel, which actually returns a row for each `entity_profiles` row.  `entity_profiles` separates out the repeated contextual ruletypes by entity, so a profile which applies to two entity types will produce two output rows.  The other callers of this query feed it through MengeDatabaseListIntoProfiles, which ends up building a map of profiles by (string) UUID, but ListEvaluationResults only needs a subset of data, and did not do this.

There was also a lot of convoluted passing of arrays and maps in the ListEvaluationResults function, which I've hopefully simplified.

# Testing

Unit testing and a small manual test.

### DIFF
--- a/internal/controlplane/handlers_evalstatus.go
+++ b/internal/controlplane/handlers_evalstatus.go
@@ -4,6 +4,7 @@
 package controlplane
 
 import (
+	"cmp"
 	"context"
 	"database/sql"
 	"encoding/base64"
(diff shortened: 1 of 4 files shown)


krrish175-byte (approved):
LGTM -- solid cleanup. Two bugs fixed and a lot of complexity removed.

**Bug 1 -- Duplicate evaluation results:** Profiles with rules for multiple entity types (e.g. repos + PRs) produced duplicate rows from `ListProfilesByProjectIDAndLabel` (one per `entity_profiles` row). The old `sortEntitiesEvaluationStatus` iterated over these raw rows, so a profile touching 2 entity types would double its evaluation results. Fixing this by building the `profileStatuses` map upfront in `filterProfileLists` (keyed by UUID, so duplicates are naturally deduplicated) and then iterating over that map is clean.

**Bug 2 -- `MergeDatabaseListIntoProfiles` keying by name vs ID:** Switching from `profiles[p.GetProfile().Name]` to `profiles[profileID]` is the correct fix. The UUID is the true identity; profile names are user-facing but not guaranteed unique across the merge window. This also avoids the repeated map lookups by building the profile into a local `newProfile` variable first.

**Other observations:**
- `buildProfileStatus` deleted -- its logic is now inlined in `filterProfileLists`, which is the only place it was called. Good removal.
- `cmp.Or(p.Profile.DisplayName, p.Profile.Name)` is a nice simplification over the old `if displayName == ""  { ... }` pattern.
- The entity-not-found special case removal in `sortEntitiesEvaluationStatus` (lines 413-419 removed) is fine since the generic error log + continue that follows handles it the same way.
- The commented-out line `// profileProtos := profiles.MergeDatabaseListIntoProfiles(profileList)` at L552 looks like a leftover from development -- might want to remove it before merge.
- The test fix adding `slices.SortFunc` is correct since `ListProfiles` now sorts after dedup, so the test order may differ from insertion order.

One nit: the `//nolint:staticcheck` for the DeMorgan's law suppression on the filter condition is a reasonable call -- `!(a || b)` reads better than `!a && !b` for this particular logic.

JAORMX (approved):

intelligent-ears (approved):

PR #6501, opened 2026-06-11
fix(actions): treat EvalStatusError as remediation off

codejasleen:
# Summary

Treat `EvalStatusError` the same as `EvalStatusSuccess` in `shouldRemediate().`

When rule evaluation returns an error, remediation is now turned off instead of returning `ActionCmdDoNothing`. This aligns the implementation with the existing remediation-off semantics described in the state machine comments and the maintainer discussion around evaluation errors.

# Testing

Updated remediation state-machine tests in `internal/engine/actions/actions_test.go` to verify:

- eval error, previous success -> off

- eval error, previous error -> off

- eval error, previous failure -> off

Verified with:

go test ./internal/engine/actions/...

# Lifecycle Semantics

This change implements the remediation behavior discussed in the maintainer thread:

`fail -> error -> fail`

fail
  -> remediation ON

error
  -> remediation OFF

OFF action
  -> ErrActionSkipped
  -> RemediationStatusSkipped persisted

fail
  -> remediation ON

`fail -> error -> success`

After:

fail
  -> remediation ON

error
  -> remediation OFF

success
  -> remediation remains OFF

This aligns with the expectation that remediation should be turned off when rule evaluation is unavailable, while still allowing remediation to be re-enabled if the rule later evaluates as failing again.

### DIFF
--- a/internal/engine/actions/actions.go
+++ b/internal/engine/actions/actions.go
@@ -214,9 +214,7 @@ func shouldAlert(
 
 	// Proceed with use cases where the evaluation changed
 	switch evalStatus {
-	case EvalStatusError:
-		return engif.ActionCmdDoNothing
-	case EvalStatusFailure:
+	case EvalStatusError, EvalStatusFailure:
 		// Case 3 - Evaluation changed from something else to ERROR -> Alert should be ON
 		// Case 4 - Evaluation has changed from something else to FAILED -> Alert should be ON
 		// The Alert should be on (if it wasn't already)
--- a/internal/engine/actions/actions.go
+++ b/internal/engine/actions/actions.go
@@ -160,9 +160,7 @@ func shouldRemediate(prevEval *previousEval, evalStatus EvalStatus) engif.Action
 
 	// Proceed with use cases where the evaluation changed
 	switch evalStatus {
-	case EvalStatusError:
-		return engif.ActionCmdDoNothing
-	case EvalStatusSuccess:
+	case EvalStatusError, EvalStatusSuccess:
 		// Case 2 - Evaluation changed from something else to ERROR -> Remediation should be OFF
 		// Case 3 - Evaluation changed from something else to PASSING -> Remediation should be OFF
 		// The Remediation should be OFF (if it wasn't already)
--- a/internal/engine/actions/actions_test.go
+++ b/internal/engine/actions/actions_test.go
@@ -63,14 +63,12 @@ func TestShouldRemediate(t *testing.T) {
 			expected:   engif.ActionCmdDoNothing,
 		},
 		{
-			// NOTE: EvalStatusTypesError has an empty case body in shouldRemediate,
-			// so eval errors fall through to the default DoNothing. This may be a bug
-			// (see the comment on cases Error/Success in shouldRemediate).
-			name:       "eval error, prev success -> do nothing",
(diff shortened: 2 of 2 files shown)


internal/engine/actions/actions_test.go
```diff
@@ -63,14 +63,12 @@ func TestShouldRemediate(t *testing.T) {
 			expected:   engif.ActionCmdDoNothing,
 		},
 		{
-			// NOTE: EvalStatusTypesError has an empty case body in shouldRemediate,
-			// so eval errors fall through to the default DoNothing. This may be a bug
-			// (see the comment on cases Error/Success in shouldRemediate).
-			name:       "eval error, prev success -> do nothing",
+			// EvalStatusError now follows the same remediation-off behavior as EvalStatusSuccess
```
evankanderson: I'm not sure the "now" here helps -- think of reading this 2 years down the road.

Maybe "Cancel remediation actions if rule evaluation has an error"?

internal/engine/actions/actions_test.go
```diff
@@ -80,11 +78,18 @@ func TestShouldRemediate(t *testing.T) {
 			expected:   engif.ActionCmdDoNothing,
 		},
 		{
-			name:       "eval error, prev error -> do nothing",
+			name:       "eval error, prev error -> off",
 			prevStatus: RemediationStatusError,
 			hasPrev:    true,
 			evalErr:    errors.New("random error"),
-			expected:   engif.ActionCmdDoNothing,
+			expected:   engif.ActionCmdOff,
```
evankanderson, line 85: Looking at the implementation of `ActionCmdOff` in `internal/engine/actions`, this appears to be safe, because all the actions check for the existence of an in-flight remediation (PR, security advisory, comment, etc) and report "off" or "skipped" when there's nothing to do.

internal/engine/actions/actions.go
```diff
@@ -160,9 +160,7 @@ func shouldRemediate(prevEval *previousEval, evalStatus EvalStatus) engif.Action
 
 	// Proceed with use cases where the evaluation changed
 	switch evalStatus {
-	case EvalStatusError:
-		return engif.ActionCmdDoNothing
-	case EvalStatusSuccess:
+	case EvalStatusError, EvalStatusSuccess:
```
evankanderson, line 163: "shouldAlert" (below) has this same logic -- adjust it there, too?

evankanderson (commented):
Thanks for picking this up!  There's one more place to adjust (and I've looked at the downstream logic, and this appears _safe_ at a minimum).

internal/engine/actions/actions.go
codejasleen: Thanks for pointing that out. I've updated `shouldAlert()` to treat evaluation errors the same as evaluation failures, updated the corresponding test, and verified the actions test suite passes.

internal/engine/actions/actions_test.go
codejasleen: good point, thanks. I've updated the comment to use the suggested wording.

intelligent-ears:
Both `shouldRemediate` and `shouldAlert` are now consistent. Test cases cover all three previous-state combinations for eval errors. The comment update is clear. LGTM.

evankanderson (approved):

Issue #6504, opened 2026-06-14
[LFX] Rule Testing Framework: Core Starlark engine + eval() builtin


krrish175-byte:
Part of the rule testing framework implementation (parent issue: #6503 ).

## Goal
Build the core bridge between Starlark and Minder's rule engine. This is the foundational piece — a single Starlark test file + a single rules YAML file, evaluated through the Minder rules engine without any providers/ingest.

## Scope
- Set up a basic `go.starlark.net` execution environment in `pkg/ruletest`
- Implement `eval(rule, entity, mocks)` builtin that:
  - Reads the associated `*.yaml` rule definition from the filesystem
  - Wires it into `pkg/engine.Evaluate()`
  - Returns the evaluation result (status + violations) back to Starlark
- No providers/ingest wiring at this stage — focus on surfacing any "big rocks" early

## Acceptance Criteria
- [x] A minimal Starlark test file can call `eval()` with a rule and entity
- [x] The eval result (pass/fail/error + violations) is accessible in Starlark
- [x] Code lives in `pkg/ruletest`
- [x] Unit tests covering the core eval flow

krrish175-byte:
Update: We decided to split the work for this issue into smaller, easier-to-review PRs.

The foundational piece -- setting up the basic `go.starlark.net` execution environment, test discovery, and failure reporting in `pkg/ruletest` has been implemented and is currently in PR #6508.

krrish175-byte:
Update: The next piece of the implementation -- adding the `eval()` built-in to evaluate rules against profiles and entities, and integrating the standard `assert` module for robust test assertions has been completed and merged in PR #6521.

krrish175-byte:
Update: The final remaining items for the initial Starlark rule testing framework have been completed and merged in PR #6524! This PR adds comprehensive Go unit tests for the core `eval()` flow to ensure the evaluation logic is fully tested in isolation, and cleans up the framework by removing the custom `fail()` built-in in favor of the standard `assert.fail`.

krrish175-byte:
All goals and acceptance criteria for this issue have now been successfully completed and merged

Summary of completed work:

1. Starlark Execution Environment: Established a clean `go.starlark.net` execution environment isolated within `pkg/ruletest`.
[...]

PR #6508, opened 2026-06-16
feat(ruletest): add Starlark-based test runner for rule types

krrish175-byte:
## Summary

Adds `pkg/ruletest`, a Starlark-based test runner that lets rule authors write and execute tests for Minder rule types without needing a running server or live credentials.

The runner discovers `*.star` files under a directory, scans for `test_*` functions, and executes each one in a sandboxed Starlark interpreter. A `fail()` builtin records failures without aborting (expect semantics), so a single test can report multiple issues. Results integrate with Go's `testing.T` for standard `go test` workflows.

This is Phase 1 of the rule testing framework described in the design doc (PR #6492). Future phases will add engine integration, mocking, and the `mindev test` CLI command.

New dependency: `go.starlark.net` for the Starlark interpreter.

Fixes a part of #6504 

## Testing

Unit tests added in pkg/ruletest/runner_test.go covering:

- File discovery across nested directories
- Execution of passing and failing test_* functions
- fail() builtin behavior and failure collection
- Skipping non-test globals and functions with parameters

Run with:

`go test ./pkg/ruletest/...`

### DIFF
--- a/pkg/ruletest/runner.go
+++ b/pkg/ruletest/runner.go
@@ -1,0 +1,203 @@
+// SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package ruletest provides a Starlark-based test runner for Minder rule types.
+package ruletest
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"path/filepath"
+	"sort"
+	"strings"
+	"testing"
+
+	"go.starlark.net/starlark"
+	"go.starlark.net/syntax"
+)
+
+const threadFailuresKey = "ruletest.failures"
+
+// TestResult holds the outcome of a single Starlark test function.
+type TestResult struct {
+	Name     string
+	Failures []string
+}
+
+// Passed returns true if the test had no failures.
+func (tr *TestResult) Passed() bool {
+	return len(tr.Failures) == 0
+}
+
+// Runner loads and executes Starlark test files.
+type Runner struct {
+	predeclared starlark.StringDict
+}
+
+// NewRunner creates a new test runner with the default set of predeclared builtins.
+func NewRunner() *Runner {
+	return &Runner{
+		predeclared: starlark.StringDict{
+			"fail": starlark.NewBuiltin("fail", builtinFail),
+		},
+	}
+}
+
+func builtinFail(
+	thread *starlark.Thread,
+	b *starlark.Builtin,
+	args starlark.Tuple,
+	kwargs []starlark.Tuple,
+) (starlark.Value, error) {
+	var msg string
+	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &msg); err != nil {
+		return nil, err
+	}
+
+	appendFailure(thread, msg)
+	return starlark.None, nil
+}
+
+// RunFile executes a single Starlark test file and returns the results
+// for each test_* function found in it.
+// src may be nil, or a string, []byte, or io.Reader containing the file source.
+func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
+	thread := &starlark.Thread{
+		Name:  "ruletest",
+		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
+	}
+
+	globals, err := starlark.ExecFileOptions(&syntax.FileOptions{}, thread, filename, src, r.predeclared)
+	if err != nil {
+		if evalErr, ok := errors.AsType[*starlark.EvalError](err); ok {
+			return nil, fmt.Errorf("loading %s: %w\n%s", filename, err, evalErr.Backtrace())
+		}
+		return nil, fmt.Errorf("loading %s: %w", filename, err)
+	}
+
+	testFns := make(map[string]*starlark.Function)
+	for _, name := range globals.Keys() {
+		if !strings.HasPrefix(name, "test_") {
+			continue
+		}
+		fn, ok := globals[name].(*starlark.Function)
+		if !ok {
+			return nil, fmt.Errorf("expected %s to be a function, got %s", name, globals[name].Type())
+		}
+		if fn.NumParams() != 0 {
+			return nil, fmt.Errorf("expected %s to have no parameters, got %d", name, fn.NumParams())
+		}
+		testFns[name] = fn
+	}
+
+	var results []TestResult
+	for name, fn := range testFns {
+		result := runOneTest(name, fn)
+		results = append(results, result)
+	}
+
+	return results, nil
+}
+
+func runOneTest(name string, fn *starlark.Function) TestResult {
+	thread := &starlark.Thread{
+		Name:  name,
+		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
+	}
+
+	result := TestResult{Name: name}
+
+	_, err := starlark.Call(thread, fn, nil, nil)
+	if err != nil {
+		if evalErr, ok := errors.AsType[*starlark.EvalError](err); ok {
+			result.Failures = append(result.Failures, evalErr.Backtrace())
+		} else {
+			result.Failures = append(result.Failures, err.Error())
+		}
+	}
+
+	result.Failures = append(result.Failures, getFailures(thread)...)
+
+	return result
+}
+
+// DiscoverFiles walks the given directory tree and returns the paths of
+// all *.star files found, in sorted order.
+func DiscoverFiles(root string) ([]string, error) {
+	var files []string
+	err := filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if !d.IsDir() && strings.HasSuffix(d.Name(), ".star") {
+			files = append(files, path)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("walking %s: %w", root, err)
+	}
+	sort.Strings(files)
+	return files, nil
+}
+
+// RunDir discovers and executes all *.star test files under the given
+// directory, reporting results through t.
+func (r *Runner) RunDir(t *testing.T, dir string) {
+	t.Helper()
+
+	files, err := DiscoverFiles(dir)
+	if err != nil {
+		t.Fatalf("discovering test files: %v", err)
+	}
+
+	if len(files) == 0 {
+		t.Logf("no *.star test files found in %s", dir)
+		return
+	}
+
+	for _, file := range files {
+		rel, err := filepath.Rel(dir, file)
+		if err != nil {
+			t.Fatalf("failed to compute relative path for %s: %v", file, err)
+		}
+
+		t.Run(rel, func(t *testing.T) {
+			results, err := r.RunFile(file, nil)
+			if err != nil {
+				t.Fatalf("running %s: %v", file, err)
+			}
+
+			for _, result := range results {
+				t.Run(result.Name, func(t *testing.T) {
+					for _, msg := range result.Failures {
+						t.Error(msg)
+					}
+				})
+			}
+		})
+	}
+}
+
+func appendFailure(thread *starlark.Thread, msg string) {
+	ptr := thread.Local(threadFailuresKey)
+	if ptr == nil {
+		failures := make([]string, 0, 1)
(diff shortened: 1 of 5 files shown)


pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+	Message string
+}
+
+const threadFailuresKey = "ruletest.failures"
+
+// TestResult holds the outcome of a single Starlark test function.
+type TestResult struct {
+	Name     string
+	Passed   bool
+	Failures []string
```
evankanderson: Rather than recording both `Passed` and `Failures`, why not define an empty set of Failures as passed?  That way, you can avoid having to decide what `Passed: true, Falures: []string{"what?"}` means.

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+	if err := starlark.UnpackPositionalArgs(b.Name(), args, kwargs, 1, &msg); err != nil {
+		return nil, err
+	}
+
+	appendFailure(thread, msg)
+	return starlark.None, nil
+}
+
+// RunFile executes a single Starlark test file and returns the results
+// for each test_* function found in it.
```
evankanderson, line 63: Given that `src` is an `any`, you should document what it is.

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
+	thread := &starlark.Thread{
+		Name:  "ruletest",
+		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
+	}
+
+	globals, err := starlark.ExecFileOptions(&syntax.FileOptions{}, thread, filename, src, r.predeclared)
```
evankanderson, line 71: Can the first argument be `nil`?  Just trying to reduce the amount of "magic setup" that we do.

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
```
evankanderson, line 72: I'm not sure if you want to check for `starlark.EvalError` and print a backtrace if so.

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+		return nil, fmt.Errorf("loading %s: %w", filename, err)
+	}
+
+	type namedFn struct {
+		name string
+		fn   *starlark.Function
+	}
```
evankanderson: Why use an array of pairs rather than a map?

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+	var testFns []namedFn
+	for _, name := range globals.Keys() {
+		if !strings.HasPrefix(name, "test_") {
+			continue
+		}
+		fn, ok := globals[name].(*starlark.Function)
+		if !ok {
+			continue
```
evankanderson: Do you want to warn about `test_` things that aren't functions?  (I don't know -- asking!)

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+		if fn.NumParams() != 0 {
+			continue // test functions must be no-arg
+		}
+		testFns = append(testFns, namedFn{name: name, fn: fn})
+	}
+
+	sort.Slice(testFns, func(i, j int) bool {
+		return testFns[i].name < testFns[j].name
+	})
```
evankanderson: Do we want to guarantee execution order?  I'm _assuming_ that the evaluation will be side-effect-free, but I'm not sure if it is.

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+	for _, tf := range testFns {
+		result := r.runOneTest(thread, tf.name, tf.fn)
+		results = append(results, result)
+	}
+
+	return results, nil
+}
+
+func (*Runner) runOneTest(thread *starlark.Thread, name string, fn *starlark.Function) TestResult {
+	resetFailures(thread)
```
evankanderson: Do we want to use a separate Thread for each eval?

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+
+	result := TestResult{Name: name, Passed: true}
+
+	_, err := starlark.Call(thread, fn, nil, nil)
+	if err != nil {
+		result.Passed = false
+		if evalErr, ok := err.(*starlark.EvalError); ok {
```
evankanderson: Use `errors.AsType[*starlark.EvalError](err)` rather than a type cast here.

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+	}
+
+	failures := getFailures(thread)
+	if len(failures) > 0 {
+		result.Passed = false
+		for _, f := range failures {
+			result.Failures = append(result.Failures, f.Message)
+		}
+	}
```
evankanderson: If you use `len(result.Failures) > 0` as "Passed" (and change the return type on `getFailures`), this can just be:

```suggestion
	result.Failures = append(result.Failures, getFailures(thread)...)
```

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+	if len(files) == 0 {
+		t.Logf("no *.star test files found in %s", dir)
+		return
+	}
+
+	for _, file := range files {
+		rel, _ := filepath.Rel(dir, file)
+		if rel == "" {
+			rel = file
+		}
```
evankanderson: It's odd to ignore the error here and then skip the `Rel` step sometimes.  Do we need this check at all?

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+				t.Fatalf("running %s: %v", file, err)
+			}
+
+			for _, result := range results {
+				t.Run(result.Name, func(t *testing.T) {
+					if !result.Passed {
+						for _, msg := range result.Failures {
+							t.Error(msg)
+						}
+					}
```
evankanderson: If we use the length thing, I think we could just remove the `if !result.Passed` here, right?

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,213 @@
...
+				})
+			}
+		})
+	}
+}
+
+func resetFailures(thread *starlark.Thread) {
+	failures := make([]testFailure, 0)
+	thread.SetLocal(threadFailuresKey, &failures)
```
evankanderson: Documentation suggests we shouldn't call `SetLocal` after execution begins, but I'm guessing that's actually _concurrent_ with execution.

pkg/ruletest/runner_test.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		t.Fatalf("RunFile failed: %v", err)
+	}
+
+	if len(results) != 3 {
+		t.Fatalf("expected 3 test functions to be discovered and run, got %d", len(results))
+	}
+
+	// Because runner.go sorts the names, the order is:
+	// test_exception, test_failing, test_passing
```
evankanderson: I'd be inclined to sort the results in the test, rather than having the framework sort the tests.  In practice, having the test function execution order be undefined seems helpful for finding any sorts of dependencies or non-determinism that may creep in.

evankanderson (commented):
Looks like a good start.  I don't know if you want to pin `src` down to being e.g. an `io.Reader` or something like that.

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,219 @@
...
+// NewRunner creates a new test runner with the default set of predeclared builtins.
+func NewRunner() *Runner {
+	r := &Runner{}
+	r.predeclared = starlark.StringDict{
+		"fail": starlark.NewBuiltin("fail", r.builtinFail),
+	}
+	return r
+}
+
+func (*Runner) builtinFail(
```
evankanderson: Why is this a struct method?  It does seem like it does anything with the Runner.

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,219 @@
...
+		return nil, err
+	}
+
+	appendFailure(thread, msg)
+	return starlark.None, nil
+}
+
+// RunFile executes a single Starlark test file and returns the results
+// for each test_* function found in it.
+// src may be nil, or a string, []byte, or io.Reader containing the file source.
```
evankanderson, line 64: I think this is _okay_, but I'd encourage you to think about a tighter contract here than Skylark requires.

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,219 @@
...
+func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
+	thread := &starlark.Thread{
+		Name:  "ruletest",
+		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
+	}
+
+	globals, err := starlark.ExecFileOptions(&syntax.FileOptions{}, thread, filename, src, r.predeclared)
+	if err != nil {
+		var evalErr *starlark.EvalError
+		if errors.As(err, &evalErr) {
```
evankanderson: Go 1.26 provides `errors.AsType`, which avoids needing a separate declaration, and scopes the variable to the "if" block:

```suggestion
		if evalErr, ok := errors.AsType[*starlark.EvalError](err); ok {
```

krrish175-byte:
Pushed an update addressing these -

- **`Passed` field -> `Passed()` method:** `TestResult` no longer has a `Passed` bool -- it just checks `len(Failures) == 0` via a method, so there's no ambiguous state to worry about.
- **Map instead of slice:** Switched from the `namedFn` struct slice to a `map[string]*starlark.Function`since we are not sorting anymore.
- **Free functions:** `builtinFail` and `runOneTest` don't use any `Runner` state, so they're free functions now.
- **Removed `resetFailures`:** Since each test gets a fresh `starlark.Thread`, the thread-local starts nil and `appendFailure` handles lazy init -- `resetFailures` was dead code.
- **Removed `testFailure` struct:** It only had one `Message string` field, so the thread-local just stores `*[]string` directly now.
- **Strict validation on `test_` globals:** If something is prefixed with `test_` but isn't a no-arg function, we return an error instead of silently skipping.
- **`errors.As` for `EvalError`:** Using `errors.As` instead of type assertion, and including backtrace in the error message on file-load failures.
- Documented `src` parameter on `RunFile`.
- Test ordering: Removed sorting from the framework -- execution order is intentionally undefined. Tests sort results themselves before asserting.
- Re: `&syntax.FileOptions{}` -- passing `nil` panics in the current version of `go.starlark.net`, so keeping the empty struct for now.

cc @evankanderson

pkg/ruletest/runner.go
```diff
@@ -0,0 +1,202 @@
...
+	failures := ptr.(*[]string)
+	*failures = append(*failures, msg)
+}
+
+func getFailures(thread *starlark.Thread) []string {
+	ptr := thread.Local(threadFailuresKey)
+	if ptr == nil {
+		return nil
+	}
+	return *ptr.(*[]string)
```
evankanderson, line 202: This could cause a `panic`.  Given the _current_ situation, I think that's acceptable, but I'd generally encourage documenting that this function could panic, given that it's not a normal Go idiom.

pkg/ruletest/runner_test.go
```diff
@@ -0,0 +1,71 @@
...
+	if len(results[1].Failures) == 0 || results[1].Failures[0] != "this test failed intentionally" {
+		t.Errorf("unexpected failures for test_failing: %v", results[1].Failures)
+	}
+
+	if results[2].Name != "test_passing" {
+		t.Errorf("expected test_passing, got %s", results[2].Name)
+	}
+	if !results[2].Passed() {
+		t.Error("expected test_passing to pass")
+	}
```
evankanderson: If you add more cases, I'd be inclined to refactor this into a data table:

```go
	tc = []struct{name string, failures []string}{{
		name: "test_exception",
		failures: []string{"..."},
	}, {
		name: "test_failing",
		failures: []string{"this test failed intentionally"},
	}, {
		// ...
	}}
```

evankanderson (dismissed):
I'm okay merging this as is and consider the following for future iterations:

evankanderson (approved):

PR #6520, opened 2026-06-18
fix(oci): resolve artifact created time from image config

### LINKED ISSUE #6490: Fetch OCI image config to reliably determine artifact creation dates instead of relying on time.Now() fallback
### Describe the issue

Currently, when the OCI provider fetches artifact versions (`GetArtifactVersions`), it relies on the `org.opencontainers.image.created` annotation in the image manifest to determine the artifact's creation date. 

However, many container images are not built with this specific annotation. When it is missing, Minder falls back to assigning `time.Now()` as the creation date (explicitly marked as a `// FIXME: This is a hack` in `internal/providers/oci/oci.go`).

**Impact:**
If a container image is missing the creation annotation, Minder reports it as being created *at the exact moment the query was executed*. This creates unreliable data and breaks age-based artifact filtering logic, lifecycle policies, and analytics metrics regarding artifact creation.

**Proposed Solution:**
Instead of relying solely on the manifest annotations and falling back to a hack, the provider should fetch the actual image configuration blob (`application/vnd.oci.image.config.v1+json`). According to the [OCI Image Format Specification](https://github.com/opencontainers/image-spec/blob/main/config.md), the config blob natively contains a `created` timestamp property. By fetching the config blob when the annotation is missing, we can reliably determine the true creation date of the artifact.

### To Reproduce

1. Register a container repository (OCI provider) in Minder.
2. Push a container image to the registry that does *not* contain the `org.opencontainers.image.created` annotation in its manifest.
3. Have Minder fetch the artifact versions for that container.
4. Observe that the `CreatedAt` timestamp for the artifact is set to the exact time the fetch occurred (i.e. `time.Now()`), rather than the actual time the image was built.

main

somaz94:
# Summary

`GetArtifactVersions` fell back to `time.Now()` when an image manifest lacked the
`org.opencontainers.image.created` annotation (a `// FIXME: This is a hack`), so
artifacts without that annotation were recorded as "created right now", silently
breaking age-based filtering and lifecycle policies.

This reads the real build time from the image config blob instead: it prefers the
manifest annotation and, when absent, falls back to the image config's `Created`
field (`img.ConfigFile().Created`). A zero or Unix-epoch timestamp is a legitimate
reproducible-build value and is preserved as-is rather than overwritten with the
current time.

The created-date resolution is extracted into a pure helper (`resolveCreatedAt`) so
it is unit-testable without a registry, and a `getImage` helper lets the same fetched
image serve both the manifest and the config lookup.

Note: for a multi-arch image index lacking the annotation, the config now resolves to
the default platform's build time (or surfaces a config-read error) where the old hack
always returned `time.Now()`.

Fixes #6490

# Testing

- Added `internal/providers/oci/oci_test.go`: table-driven unit test for `resolveCreatedAt`
  (annotation present / invalid / absent→config / epoch preserved / zero preserved / config error).
- `go test ./internal/providers/oci/...` — pass
- `go vet ./internal/providers/oci/...` — clean
- `golangci-lint run ./internal/providers/oci/...` — 0 issues

### DIFF
--- a/internal/providers/oci/oci.go
+++ b/internal/providers/oci/oci.go
@@ -126,9 +126,8 @@ func (o *OCI) GetReferrer(ctx context.Context, contname, tag, artifactType strin
 	return refer, nil
 }
 
-// GetManifest returns the manifest for the given tag of the given container in the given namespace
-// for the OCI provider. It returns the manifest as a golang struct given the OCI spec.
-func (o *OCI) GetManifest(ctx context.Context, contname, tag string) (*v1.Manifest, error) {
+// getImage returns the remote image for the given tag of the given container.
+func (o *OCI) getImage(ctx context.Context, contname, tag string) (v1.Image, error) {
 	ref, err := o.getReference(contname, tag)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get reference: %w", err)
@@ -137,6 +136,17 @@ func (o *OCI) GetManifest(ctx context.Context, contname, tag string) (*v1.Manife
 	img, err := remote.Image(ref, remote.WithContext(ctx), remote.WithUserAgent(constants.ServerUserAgent))
 	if err != nil {
 		return nil, fmt.Errorf("failed to get image: %w", err)
+	}
+
+	return img, nil
+}
+
+// GetManifest returns the manifest for the given tag of the given container in the given namespace
+// for the OCI provider. It returns the manifest as a golang struct given the OCI spec.
+func (o *OCI) GetManifest(ctx context.Context, contname, tag string) (*v1.Manifest, error) {
+	img, err := o.getImage(ctx, contname, tag)
+	if err != nil {
+		return nil, err
 	}
 
 	man, err := img.Manifest()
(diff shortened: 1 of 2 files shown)


krrish175-byte:
This is a much better approach than the previous `time.Now()` fallback. I really like how you extracted `resolveCreatedAt` to be testable without a registry, and passing `img.ConfigFile` as a getter to defer the config blob network call until it's actually needed is a very clean, idiomatic design. LGTM. 

cc @evankanderson

somaz94:
Thanks for the review @krrish175-byte! 🙏

Heads-up on CI for whoever picks this up: the `test / Unit testing` failure is unrelated to this change. The `internal/providers/oci` package (the only one this PR touches) passes cleanly — the 4 failures are pre-existing flakes elsewhere:

- `internal/events/nats` → `TestNatsChannel`: timing-based (`timeout waiting for message 2`, `expected 4 messages, got 2`)
- `internal/providers/github/webhook` → `TestHandleGitHubWebHook/team_removed_from_repository`: `Expected value not to be nil`

A re-run of that job should go green. Happy to rebase if preferred. cc @evankanderson

internal/providers/oci/oci.go
```diff
@@ -231,6 +233,33 @@ func (ov *OCI) GetArtifactVersions(
 	return out, nil
 }
 
+// resolveCreatedAt determines the creation time of an artifact version. It
+// prefers the org.opencontainers.image.created manifest annotation and, when
+// that annotation is absent, falls back to the Created field of the image
+// configuration, which records the real build time. Unlike the previous
+// time.Now() fallback, the resolved value always reflects the image itself: a
```
evankanderson: Generally, it doesn't make sense to have comments referencing code that no longer exists after this PR.

The one exception (which this does not fall into) is if the _current_ code has to maintain a certain behavior when working with data that might have been produced by old code.  An example of this might be `// older code did not fill in Foo, even though that's now enforced by Validate()`

internal/providers/oci/oci.go
```diff
@@ -231,6 +233,33 @@ func (ov *OCI) GetArtifactVersions(
...
+func resolveCreatedAt(man *v1.Manifest, configFile func() (*v1.ConfigFile, error)) (time.Time, error) {
+	if strcreated, ok := man.Annotations[imgspecv1.AnnotationCreated]; ok {
+		createdAt, err := time.Parse(time.RFC3339, strcreated)
+		if err != nil {
+			return time.Time{}, fmt.Errorf("parsing created annotation %q: %w", strcreated, err)
+		}
+		return createdAt, nil
+	}
+
+	cfg, err := configFile()
```
evankanderson, line 253: The previous code had comments about handling multi-architecture images. Have you tested this with an image like `ghcr.io/mindersec/minder/server:latest`?

internal/providers/oci/oci.go
somaz94: Good point, removed. The doc comment no longer references the old time.Now() fallback, it just describes the current behavior (annotation first, then the image config Created field, with zero/epoch preserved). Pushed in 30c3425e8.

internal/providers/oci/oci.go
somaz94: Tested with ghcr.io/mindersec/minder/server:latest, which is a 2-platform index (linux/amd64 + linux/arm64).

remote.Image resolves the index down to a single-platform image (amd64 by default), so img.Manifest() and img.ConfigFile() both run against that concrete child image, not the index. That is the per-architecture config the old FIXME was after.

For that image the created annotation is present, so it returns 2026-04-23T11:01:57Z from the annotation. With the annotation missing it would read Created from the resolved child config. Same behavior on alpine:latest.

evankanderson (approved):
Thanks for checking that codepath!  It looks like we'll probably throw an error for multi-architecture images that _don't_ include `amd64/linux`, but I think that still puts us in a better position than today.  (And we can find this conversation in the commit history if we wonder why we stopped at this point....)

evankanderson:
The security scan is failing on unrelated (new) Go `x/crypto` CVEs fixed in #6537, I'll ignore (bypass) that check.

Nice fix, very clean, just waiting for the rest of CI to go green.

PR #6521, opened 2026-06-18
feat(ruletest): implement starlark eval built-in and assert module

krrish175-byte:
## Summary

This PR introduces the `eval` built-in function to the ruletest Starlark environment, allowing rule execution directly from test scripts using `eval(rule="...", profile={...}, entity={...})`. It also implements robust type conversion (`starlarkValueToGo`) to map Starlark dictionaries and lists back to native Go types, enabling proper execution of the underlying rule logic.

Additionally, while exploring the Starlark ecosystem to implement the `kwargs` design, I discovered and integrated the official `starlarktest` module (`go.starlark.net/starlarktest`). This replaces our manual `if/fail()` checks with the idiomatic `assert.eq()` and `assert.true()` testing pattern, making the Starlark test scripts much cleaner and more maintainable.

Adds 2nd part of #6504 

## Testing

- Added comprehensive table-driven unit tests in `starlark_util_test.go` to verify all type conversions (strings, ints, floats, bools, nested dicts, and lists).
- Updated `eval.star` and `sample.star` test fixtures to utilize the new `assert` module.
- Ran `go test -v ./pkg/ruletest/...` ensuring all 14 tests and sub-tests pass.
- Verified that Starlark thread errors (from failed assertions) are correctly surfaced as Go test failures via a custom `threadReporter`.

### DIFF
--- a/pkg/ruletest/eval.go
+++ b/pkg/ruletest/eval.go
@@ -1,0 +1,100 @@
+// SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ruletest
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"path/filepath"
+
+	"go.starlark.net/starlark"
+
+	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+	"github.com/mindersec/minder/pkg/engine/v1/interfaces"
+	"github.com/mindersec/minder/pkg/engine/v1/rtengine"
+	"github.com/mindersec/minder/pkg/fileconvert"
+)
+
+func builtinEval(
+	thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple,
+) (starlark.Value, error) {
+	var rulePath string
+	var entityDict *starlark.Dict
+	var profileDict *starlark.Dict
+
+	err := starlark.UnpackArgs("eval", args, kwargs,
+		"rule", &rulePath, "entity?", &entityDict, "profile?", &profileDict)
+	if err != nil {
+		return nil, err
+	}
+
+	if !filepath.IsAbs(rulePath) {
+		callerFrame := thread.CallFrame(1)
+		if callerFile := callerFrame.Pos.Filename(); callerFile != "" {
+			rulePath = filepath.Join(filepath.Dir(callerFile), rulePath)
+		}
+	}
+
+	decoder, closer := fileconvert.DecoderForFile(rulePath)
+	if decoder == nil {
+		return nil, fmt.Errorf("error opening file: %s", rulePath)
+	}
+	defer closer.Close()
+
+	rt, err := fileconvert.ReadResourceTyped[*minderv1.RuleType](decoder)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse rule type: %w", err)
+	}
+
+	profileMap, err := dictToGoMap(profileDict)
+	if err != nil {
+		return nil, fmt.Errorf("invalid profile argument: %w", err)
+	}
+
+	entityMap, err := dictToGoMap(entityDict)
+	if err != nil {
+		return nil, fmt.Errorf("invalid entity argument: %w", err)
+	}
+
+	ctx := context.Background()
+	evaluator, err := rtengine.NewRuleEvaluator(ctx, rt, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize rule evaluator: %w", err)
+	}
+
+	ingested := &interfaces.Ingested{Object: entityMap}
+	_, evalErr := evaluator.Eval(ctx, profileMap, nil, ingested)
+
+	return formatEvalResult(evalErr), nil
+}
+
+func formatEvalResult(evalErr error) *starlark.Dict {
+	result := starlark.NewDict(2)
+	status, msg := "", ""
+
+	switch {
+	case evalErr == nil:
+		status = "pass"
+	case errors.Is(evalErr, interfaces.ErrEvaluationFailed):
+		status = "fail"
+		msg = evalErr.Error()
+		var details interfaces.EvalError
+		if errors.As(evalErr, &details) {
+			msg = fmt.Sprintf("%s: %s", msg, details.Details())
+		}
+	case errors.Is(evalErr, interfaces.ErrEvaluationSkipped):
+		status = "skip"
+		msg = evalErr.Error()
+	default:
+		status = "error"
+		msg = evalErr.Error()
+	}
+
+	// We ignore the error from SetKey because result is a new Dict we just created
+	// and we know it's not frozen, and the keys are valid strings.
+	_ = result.SetKey(starlark.String("status"), starlark.String(status))
+	_ = result.SetKey(starlark.String("message"), starlark.String(msg))
+	return result
+}
--- a/pkg/ruletest/runner.go
+++ b/pkg/ruletest/runner.go
@@ -14,6 +14,7 @@ import (
 	"testing"
 
 	"go.starlark.net/starlark"
+	"go.starlark.net/starlarktest"
 	"go.starlark.net/syntax"
 )
 
@@ -37,10 +38,21 @@ type Runner struct {
 
 // NewRunner creates a new test runner with the default set of predeclared builtins.
 func NewRunner() *Runner {
+	assertMod, err := starlarktest.LoadAssertModule()
+	if err != nil {
+		panic(fmt.Errorf("failed to load starlarktest assert module: %w", err))
+	}
+
+	predeclared := starlark.StringDict{
+		"fail": starlark.NewBuiltin("fail", builtinFail),
+		"eval": starlark.NewBuiltin("eval", builtinEval),
+	}
+	for k, v := range assertMod {
+		predeclared[k] = v
+	}
+
 	return &Runner{
-		predeclared: starlark.StringDict{
-			"fail": starlark.NewBuiltin("fail", builtinFail),
-		},
+		predeclared: predeclared,
 	}
 }
 
@@ -63,10 +75,7 @@ func builtinFail(
 // for each test_* function found in it.
 // src may be nil, or a string, []byte, or io.Reader containing the file source.
 func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
-	thread := &starlark.Thread{
-		Name:  "ruletest",
-		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
-	}
+	thread := r.newThread("ruletest")
 
 	globals, err := starlark.ExecFileOptions(&syntax.FileOptions{}, thread, filename, src, r.predeclared)
 	if err != nil {
@@ -93,18 +102,24 @@ func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
 
 	var results []TestResult
 	for name, fn := range testFns {
-		result := runOneTest(name, fn)
+		result := r.runOneTest(name, fn)
 		results = append(results, result)
 	}
 
 	return results, nil
 }
 
-func runOneTest(name string, fn *starlark.Function) TestResult {
+func (*Runner) newThread(name string) *starlark.Thread {
 	thread := &starlark.Thread{
 		Name:  name,
 		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
 	}
+	starlarktest.SetReporter(thread, threadReporter{thread})
+	return thread
+}
+
+func (r *Runner) runOneTest(name string, fn *starlark.Function) TestResult {
+	thread := r.newThread(name)
 
 	result := TestResult{Name: name}
 
@@ -189,6 +204,14 @@ func appendFailure(thread *starlark.Thread, msg string) {
 	}
 	failures := ptr.(*[]string)
 	*failures = append(*failures, msg)
+}
+
+type threadReporter struct {
+	thread *starlark.Thread
+}
+
+func (r threadReporter) Error(args ...any) {
+	appendFailure(r.thread, fmt.Sprint(args...))
 }
 
 // getFailures retrieves the accumulated failure messages from the thread-local
--- a/pkg/ruletest/runner_test.go
+++ b/pkg/ruletest/runner_test.go
@@ -16,13 +16,20 @@ func TestDiscoverFiles(t *testing.T) {
 		t.Fatalf("DiscoverFiles failed: %v", err)
 	}
 
-	if len(files) != 1 {
-		t.Fatalf("expected 1 file, got %d", len(files))
+	found := make(map[string]bool)
+	for _, f := range files {
+		found[f] = true
 	}
 
-	expected := filepath.Join("testdata", "sample.star")
-	if files[0] != expected {
-		t.Errorf("expected %s, got %s", expected, files[0])
+	expected := []string{
+		filepath.Join("testdata", "eval.star"),
+		filepath.Join("testdata", "sample.star"),
+	}
+
+	for _, exp := range expected {
+		if !found[exp] {
+			t.Errorf("expected to find %s, but it was missing", exp)
+		}
 	}
 }
 
@@ -81,5 +88,32 @@ func TestRunFile(t *testing.T) {
 				t.Errorf("results[%d] (%s): failure[%d] = %q, want %q", i, tt.name, j, results[i].Failures[j], want)
 			}
 		}
+	}
+}
+
+func TestRunEvalFile(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		file string
+	}{
+		{name: "eval", file: "eval.star"},
+	}
+
+	r := NewRunner()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			results, err := r.RunFile(filepath.Join("testdata", tt.file), nil)
+			if err != nil {
+				t.Fatalf("RunFile failed for %s: %v", tt.file, err)
+			}
+			for _, res := range results {
+				if !res.Passed() {
+					t.Errorf("test %s failed: %v", res.Name, res.Failures)
+				}
+			}
+		})
 	}
 }
--- a/pkg/ruletest/starlark_util.go
+++ b/pkg/ruletest/starlark_util.go
@@ -1,0 +1,73 @@
+// SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ruletest
+
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+)
+
+// starlarkValueToGo converts a starlark.Value into its Go equivalent.
+func starlarkValueToGo(val starlark.Value) (any, error) {
+	switch v := val.(type) {
+	case starlark.NoneType:
+		return nil, nil
+	case starlark.Bool:
+		return bool(v), nil
+	case starlark.Int:
+		i, ok := v.Int64()
+		if !ok {
+			return nil, fmt.Errorf("starlark Int out of range")
+		}
+		return i, nil
+	case starlark.Float:
+		return float64(v), nil
+	case starlark.String:
+		return string(v), nil
+	case *starlark.List:
+		result := make([]any, v.Len())
+		for i := range v.Len() {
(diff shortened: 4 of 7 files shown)


pkg/ruletest/testdata/eval.star
```diff
@@ -0,0 +1,16 @@
+def test_eval_success():
+    res = eval(
+        rule="../../../cmd/cli/app/ruletype/fixture/rule_type_sample.yaml",
```
evankanderson: Let's just add a rule in the testdata, rather than pulling a file from far away.

pkg/ruletest/testdata/eval.star
```diff
@@ -0,0 +1,16 @@
...
+    )
+    assert.eq(res["status"], "pass")
+
+def test_eval_fail():
+    res = eval(
+        rule="../../../cmd/cli/app/ruletype/fixture/rule_type_sample.yaml",
+        profile={"required_reviews": 2},
+        entity={"required_pull_request_reviews": {"required_approving_review_count": 1}}
+    )
+    assert.eq(res["status"], "fail")
```
evankanderson, line 18: I don't know if we want to go this way, as it seems like it makes the return value of `eval` different from the Starlark built-in types.  We could make the result a new type, and have `status` and `message` fields, rather than having `res` be a dictionary?  Writing this as:

```python
assert.eq(res.status, "fail")
```

Feels slightly easier to read, but may not be worth the effort and difference (e.g. needing to define iteration, etc on the type).

This is me thinking out loud, not asking for a specific change.

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,114 @@
...
+
+	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+	"github.com/mindersec/minder/pkg/engine/v1/interfaces"
+	"github.com/mindersec/minder/pkg/engine/v1/rtengine"
+)
+
+func builtinEval(
+	thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple,
+) (starlark.Value, error) {
+	var rulePath string
```
evankanderson, line 23: This is temporary, right?  Long-term plan is to take a rule name, and load all the rules from the directory, right?

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,114 @@
...
+	var profileDict *starlark.Dict
+	var entityDict *starlark.Dict
+
+	err := starlark.UnpackArgs("eval", args, kwargs,
+		"rule", &rulePath, "profile?", &profileDict, "entity?", &entityDict)
```
evankanderson: I'd suggest switching the order of `entity` and `profile` -- based on usage in `minder-rules-and-profiles`, it seems more likely / common to pull entity details than to set profile definitions.

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if !filepath.IsAbs(rulePath) {
+		callerFrame := thread.CallFrame(1)
+		if callerFile := callerFrame.Pos.Filename(); callerFile != "" {
+			rulePath = filepath.Join(filepath.Dir(callerFile), rulePath)
+		}
+	}
```
evankanderson, line 38: I haven't thought deeply about this, but what's the desired behavior in the following case:

**base.star**
```python
def evalhelper(rulename):
  return eval(rulename + ".yaml", entity=DEFAULT_ENTITY)
```

**test.star**
```python
load("../base.star", "evalhelper")
def test_foo():
  res = evalhelper("my_rule")
```

I guess this shouldn't matter too much because of my other comment about automatically loading all ruletypes and using the ruletype name as the first argument eventually.

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,114 @@
...
+	f, err := os.Open(filepath.Clean(rulePath))
+	if err != nil {
+		return nil, fmt.Errorf("failed to open rule file: %w", err)
+	}
+	defer f.Close()
+
+	rt := &minderv1.RuleType{}
+	if err := minderv1.ParseResource(f, rt); err != nil {
+		return nil, fmt.Errorf("failed to parse rule type: %w", err)
+	}
```
evankanderson: Prefer the `fileconvert` methods (I think) -- I'm trying to funnel most reads and writes through those methods so it's easier to refactor how rules are presented.  (i.e. rules could also be `*.rego` with the rest of the fields as yaml-format comments at the top)

```suggestion
	decoder, closer := fileconvert.DecoderForFile(rulePath)
	if decoder == nil {
		return nil, fmt.Errorf("error opening file")
	}
	defer closer.Close()
	rt, err := fileconvert.ReadResourceTyped[*minderv1.RuleType](decoder)
	if err != nil {
		return nil, fmt.Errorf("failed to parse rule type: %w", err)
	}
```

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil, fmt.Errorf("invalid profile argument: %w", err)
+	}
+
+	var entityMap any
+	if entityDict != nil {
+		entityMap, err = starlarkValueToGo(entityDict)
+		if err != nil {
+			return nil, fmt.Errorf("invalid entity argument: %w", err)
+		}
+	}
```
evankanderson: Why does this look different from `profileMap`?

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,114 @@
...
+
+	ctx := context.Background()
+	evaluator, err := rtengine.NewRuleEvaluator(ctx, rt, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize rule evaluator: %w", err)
+	}
+
+	ingested := &interfaces.Ingested{Object: entityMap}
+	_, evalErr := evaluator.Eval(ctx, profileMap, nil, ingested)
```
evankanderson, line 68: Collecting and returning the output data (first argument) is a TODO, right?

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,114 @@
...
+// an empty map when d is nil.
+func dictToGoMap(d *starlark.Dict) (map[string]any, error) {
+	if d == nil {
+		return map[string]any{}, nil
+	}
+	v, err := starlarkValueToGo(d)
+	if err != nil {
+		return nil, err
+	}
+	return v.(map[string]any), nil
```
evankanderson: This can panic if `v` is not of the correct type, right?  Let's handle it.

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func formatEvalResult(evalErr error) *starlark.Dict {
+	result := starlark.NewDict(2)
+	status, msg := "pass", ""
+
+	if evalErr != nil {
```
evankanderson: Why not make `evalErr == nil` one of the switch cases?

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,114 @@
...
+		case errors.Is(evalErr, interfaces.ErrEvaluationSkipped):
+			status = "skip"
+			msg = evalErr.Error()
+		default:
+			status = "error"
+			msg = evalErr.Error()
+		}
+	}
+
+	_ = result.SetKey(starlark.String("status"), starlark.String(status))
```
evankanderson, line 97: Add a comment for why we're ignoring the error return result here (I'm assuming it's because it's unlikely to happen in practice, and it makes the rest of the code worse).

pkg/ruletest/runner.go
```diff
@@ -37,10 +38,21 @@ type Runner struct {
 
 // NewRunner creates a new test runner with the default set of predeclared builtins.
 func NewRunner() *Runner {
+	assertMod, err := starlarktest.LoadAssertModule()
+	if err != nil {
+		panic(fmt.Errorf("failed to load starlarktest assert module: %w", err))
+	}
+
+	predeclared := starlark.StringDict{
+		"fail": starlark.NewBuiltin("fail", builtinFail),
```
evankanderson, line 47: Do we want `fail` in addition to `assert`?

pkg/ruletest/runner.go
```diff
@@ -105,6 +118,7 @@ func runOneTest(name string, fn *starlark.Function) TestResult {
 		Name:  name,
 		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
 	}
+	starlarktest.SetReporter(thread, threadReporter{thread})
```
evankanderson, line 117: I'm wondering if you want to make a helper for creating threads?

pkg/ruletest/runner_test.go
```diff
@@ -16,13 +16,20 @@ func TestDiscoverFiles(t *testing.T) {
...
+	sort.Strings(files)
+
+	expected0 := filepath.Join("testdata", "eval.star")
+	if files[0] != expected0 {
+		t.Errorf("expected %s, got %s", expected0, files[0])
+	}
+
+	expected1 := filepath.Join("testdata", "sample.star")
+	if files[1] != expected1 {
+		t.Errorf("expected %s, got %s", expected1, files[1])
```
evankanderson: Again, you may want to use a data-table style over a list of assertions.

evankanderson: (This would also let you factor out the addition of the "testdata/" path.)

pkg/ruletest/starlark_util.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"fmt"
+
+	"go.starlark.net/starlark"
+)
+
+// starlarkValueToGo converts a starlark.Value into its Go equivalent.
+func starlarkValueToGo(val starlark.Value) (any, error) {
+	switch v := val.(type) {
```
evankanderson, line 14: Can we put these in [the same order as the starlark docs](https://pkg.go.dev/go.starlark.net/starlark#Value)?

* None
* Bool
* Bytes (not handled, which is fine)
* Int
* Float
* String
* *List
* *Tuple (not handled)
* *Dict
* *Set (not handled)

pkg/ruletest/starlark_util.go
```diff
@@ -0,0 +1,60 @@
...
+			if err != nil || !found {
+				return nil, fmt.Errorf("failed to get dictionary value for key %q", keyStr)
+			}
+			goVal, err := starlarkValueToGo(val)
+			if err != nil {
+				return nil, fmt.Errorf("key %q: %w", keyStr, err)
+			}
+			result[string(keyStr)] = goVal
+		}
+		return result, nil
```
evankanderson, line 52: This may be simpler using `Entries`:

```suggestion
		result := make(map[string]any, v.Len())
		for k, v := range v.Entries() {
			keyStr, ok := k.(starlark.String)
			if !ok {
				return nil, fmt.Errorf("dictionary key must be a string, got %s", key.Type())
			}
			goVal, err := starlarkValueToGo(v)
			if err != nil {
				return nil, fmt.Errorf("key %q: %w", keyStr, err)
			}
			result[string(keyStr)] = goVal
		}
		return result, nil
```

pkg/ruletest/starlark_util_test.go
```diff
@@ -0,0 +1,125 @@
...
+				d := starlark.NewDict(2)
+				_ = d.SetKey(starlark.String("key1"), starlark.String("val1"))
+				_ = d.SetKey(starlark.String("key2"), starlark.MakeInt(2))
+				return d
+			}(),
+			expected: map[string]any{
+				"key1": "val1",
+				"key2": int64(2),
+			},
+			err:      false,
```
evankanderson: gci thinks this file needs reformatting.

pkg/ruletest/starlark_util_test.go
```diff
@@ -0,0 +1,125 @@
...
+		name     string
+		input    starlark.Value
+		expected any
+		err      bool
+	}{
+		{
+			name:     "none",
+			input:    starlark.None,
+			expected: nil,
+			err:      false,
```
evankanderson: Since `err` is a boolean, you can _omit_ the value and it will default to `false`.  This means you can have `name`, `input`, and _either_ `expected` or `err` in the table.

pkg/ruletest/testdata/eval.star
krrish175-byte: Done, copied the rule YAML into testdata/rule_type_sample.yaml and updated the references.

pkg/ruletest/testdata/eval.star
krrish175-byte: Yeah, I think that would be cleaner to read. I will keep it as a dict for now since, as you mentioned, defining iteration and other type methods might not be worth the effort yet. But it could be a nice improvement once the eval API stabilizes more.

pkg/ruletest/eval.go
krrish175-byte: Yes, exactly. The plan is to eventually load all rule types from the directory automatically and let the test script refer to them by name rather than file path.

pkg/ruletest/eval.go
krrish175-byte: Makes sense -- swapped the order so entity comes before profile now.

pkg/ruletest/eval.go
krrish175-byte: Right, once we switch to loading rules by name from the directory, the relative path resolution won't matter. For now the caller-relative resolution should handle the common case fine.

pkg/ruletest/eval.go
krrish175-byte: I have replaced os.Open + ParseResource with fileconvert.DecoderForFile and fileconvert.ReadResourceTyped.

pkg/ruletest/eval.go
krrish175-byte: I have refactored this so that entityMap now uses the same dictToGoMap parsing as profileMap.

pkg/ruletest/eval.go
krrish175-byte: Yes, collecting and returning the output data is indeed a TODO for a future PR!

pkg/ruletest/eval.go
krrish175-byte: Fixed. Added proper ok-check logic inside dictToGoMap.

pkg/ruletest/eval.go
krrish175-byte: evalErr == nil is now handled cleanly as the first switch case in formatEvalResult.

pkg/ruletest/eval.go
krrish175-byte: Added a comment explaining that we ignore the SetKey error because the Dict is newly created, unfrozen, and the keys are valid strings.

pkg/ruletest/runner.go
krrish175-byte: We included fail since assert.fails() catches errors from the executed starlark code, but sometimes a test author might want to explicitly fail("message") if a complex condition isn't met and standard assert.eq isn't enough. Happy to remove it if you think assert covers all use cases

pkg/ruletest/runner.go
krrish175-byte: Extracted thread creation logic into a new newThread(name string) helper method on the Runner struct.

pkg/ruletest/runner_test.go
krrish175-byte: TestRunEvalFile has been refactored to use data-tables for readability.

pkg/ruletest/starlark_util.go
krrish175-byte: The starlarkValueToGo switch cases have been reordered to match the Starlark documentation.

pkg/ruletest/starlark_util.go
krrish175-byte: I used Items() in starlarkValueToGo (Entries() is not available on the go.starlark.net/starlark.Dict type, Items() is the equivalent method there).

pkg/ruletest/starlark_util_test.go
krrish175-byte: Ran gci to fix the third-party import grouping across the tests.

pkg/ruletest/starlark_util_test.go
krrish175-byte: Removed err: false fields from test definitions since it's the zero value.

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,120 @@
...
+		return nil, fmt.Errorf("failed to parse rule type: %w", err)
+	}
+
+	profileMap, err := dictToGoMap(profileDict)
+	if err != nil {
+		return nil, fmt.Errorf("invalid profile argument: %w", err)
+	}
+
+	var entityMap map[string]any
+	if entityDict != nil {
```
evankanderson: Why do we have a `nil` check here?  It seems like `dictToGoMap` handles the `nil` case internally (as we can see in the `profileMap` code).

pkg/ruletest/eval.go
```diff
@@ -0,0 +1,120 @@
...
+
+	ingested := &interfaces.Ingested{Object: entityMap}
+	_, evalErr := evaluator.Eval(ctx, profileMap, nil, ingested)
+
+	return formatEvalResult(evalErr), nil
+}
+
+// dictToGoMap converts a starlark.Dict to map[string]any, defaulting to
+// an empty map when d is nil.
+func dictToGoMap(d *starlark.Dict) (map[string]any, error) {
```
evankanderson: Why is this here but `starlarkValueToGo` is in `starlark_util`?

pkg/ruletest/runner.go
evankanderson: I was trying to figure out the difference between `fail` and `assert.true(false)`.  (There may be one -- I'm trying to pull out the difference between the two.)

pkg/ruletest/runner_test.go
evankanderson: I was asking about TestDiscoverFiles, which seems like it will continue to have `*.star` files added.

pkg/ruletest/starlark_util.go
evankanderson: I thought [`Dict.Entries` would return an `iter.Seq2[Value, Value]`](https://pkg.go.dev/go.starlark.net/starlark#Dict.Entries), which might avoid some allocations.

Items is fine too, though.

pkg/ruletest/starlark_util_test.go
```diff
@@ -0,0 +1,116 @@
...
+		{
+			name: "dict with non-string key",
+			input: func() starlark.Value {
+				d := starlark.NewDict(1)
+				_ = d.SetKey(starlark.MakeInt(1), starlark.String("val"))
+				return d
+			}(),
+			expected: nil,
+			err:      true,
+		},
```
evankanderson, line 99: Add an (error) test case for bytes?

evankanderson (dismissed):
A few minor cleanups, but LGTM!  Nice to see this starting to come together!

pkg/ruletest/runner.go
krrish175-byte: The `assert` module has `assert.fails()`, which catches exceptions from an expression, and `assert.true()`, which fails the test if a condition is false. However, standard Starlark testing often includes a simple `fail('msg')` function to explicitly throw an error (which fails the test) without needing a dummy boolean condition. We included it for flexibility, but we can drop it if you would prefer a tighter API.

pkg/ruletest/starlark_util.go
krrish175-byte: Good catch. The `go.starlark.net/starlark` package's `Dict.Entries` does return an iterator in newer Go versions, but `Items()` achieves the same goal for now. I will stick with `Items()` as it's fully supported, but we can keep an eye on `Entries()` for optimization down the road.

pkg/ruletest/runner.go
evankanderson: It looks like [the `assert.star` module](https://github.com/google/starlark-go/blob/master/starlarktest/assert.star#L52) maps `assert.fail` to the built-in `error` that is loaded through `LoadAssertModule`.

I'm inclined to reduce the number of names we export until we're actually writing tests and discover something that's awkward in the assert primitives.

evankanderson (approved):

pkg/ruletest/runner.go
krrish175-byte: Makes sense. I see now that `assert.fail()` already covers this by leveraging the underlying error built-in provided by the module.

Since this PR just merged, I will go ahead and remove our custom `fail` built-in from the environment in the next PR to keep the exported names minimal as you suggested.

PR #6523, opened 2026-06-19
Improve debugging of clock drift in GitHub Auth dance

evankanderson:
# Summary

I had a machine with a few seconds clock drift running a Minder server, and provider registration would consistently fail with errors from GitHub on the token exchange.  It turns out that 10 minutes is an _absolute maximum_ future expiry time for GitHub App tokens.  Additionally, "verification" was misspelled in the error message, which was bugging me.

# Testing

Tested manually in a run-docker environment (on the machine with the drifted clock).

### DIFF
--- a/internal/controlplane/handlers_oauth.go
+++ b/internal/controlplane/handlers_oauth.go
@@ -44,8 +44,9 @@ import (
 // and the state to be used for the callback. It accepts a provider string
 // and a boolean indicating whether the client is a CLI or web client
 // nolint:gocyclo
-func (s *Server) GetAuthorizationURL(ctx context.Context,
-	req *pb.GetAuthorizationURLRequest) (*pb.GetAuthorizationURLResponse, error) {
+func (s *Server) GetAuthorizationURL(
+	ctx context.Context, req *pb.GetAuthorizationURLRequest,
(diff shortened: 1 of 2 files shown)


internal/controlplane/handlers_oauth.go
```diff
@@ -44,8 +44,9 @@ import (
 // and the state to be used for the callback. It accepts a provider string
 // and a boolean indicating whether the client is a CLI or web client
 // nolint:gocyclo
-func (s *Server) GetAuthorizationURL(ctx context.Context,
-	req *pb.GetAuthorizationURLRequest) (*pb.GetAuthorizationURLResponse, error) {
+func (s *Server) GetAuthorizationURL(
+	ctx context.Context, req *pb.GetAuthorizationURLRequest,
+) (*pb.GetAuthorizationURLResponse, error) {
 	entityCtx := engcontext.EntityFromContext(ctx)
```
evankanderson, line 50: `gofmt` from the most recent Go language server wanted to do this, and it seemed fine.  (In particular, it didn't like the hanging argument followed by the return types.)

krrish175-byte (approved):
LGTM! The 9-minute buffer for clock drift makes perfect sense. Thanks for fixing the typo too!

PR #6530, opened 2026-06-21
feat(ruletest): add read_file and txtar built-ins

### LINKED ISSUE #6505: [LFX] Rule Testing Framework: Core Starlark builtins (read_file, txtar, check)

## Goal
Implement the pure-logic Starlark builtins that test authors will use to set up test data and make assertions.

## Scope
- `read_file(path)`: Reads a file from an `io/fs.FS` rooted at the test file's directory
- `txtar(string)`: Parses a txtar-formatted string into a Starlark dictionary
- `check` module with "expect semantics":
  - `check.eq(got, want)`: Equality assertion
  - `check.contains(haystack, needle)`: Containment assertion
  - Errors are collected in Starlark thread-local storage rather than aborting immediately

## Acceptance Criteria
- [x] `read_file` can load files relative to the test's directory
- [x] `txtar` correctly parses txtar format into a Starlark dict
- [x] `check.eq` and `check.contains` collect failures without aborting
- [ ] All failures are reported at the end of the test function
- [ ] Unit tests for each builtin

krrish175-byte:
## Summary

This PR implements the test data setup built-ins for the Starlark rule testing framework, addressing the file utilities portion of Issue #6505.

Per feedback from @evankanderson, we are dropping the custom `check` module from the scope of this issue and will instead rely on Starlark's standard `assert` module for assertions. Therefore, this PR focuses solely on file setup.

Changes included:
*   Added `read_file(path)` built-in: Safely reads a local file relative to the running Starlark test, explicitly rejecting absolute paths and directory traversals (`../`).
*   Added `txtar(string)` built-in: Parses Go text archives (`golang.org/x/tools/txtar`) from a Starlark string payload into a Starlark dictionary (`Dict`), making it easy to define complex multi-file structures inline.
*   Injected both built-ins into the Starlark global environment in `runner.go`.

Fixes #6505

## Testing

*   Added comprehensive Go unit tests in `pkg/ruletest/builtins_file_test.go`.
*   Verified that `txtar` parses strings correctly into Starlark dictionaries.
*   Verified that `read_file` successfully reads relative files and securely blocks directory traversal and absolute paths.


pkg/ruletest/builtins_file.go
```diff
@@ -0,0 +1,75 @@
...
+	args starlark.Tuple,
+	kwargs []starlark.Tuple,
+) (starlark.Value, error) {
+	var path string
+	if err := starlark.UnpackArgs(b.Name(), args, kwargs, "path", &path); err != nil {
+		return nil, err
+	}
+
+	// Prevent directory traversal
+	if filepath.IsAbs(path) || strings.Contains(path, "..") {
```
evankanderson: There are a bunch of scenarios that this doesn't cover.  The simplest is:

```
ln -s .. up
echo "out" > up/escaped.txt
```

pkg/ruletest/builtins_file.go
```diff
@@ -0,0 +1,75 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"go.starlark.net/starlark"
+	"golang.org/x/tools/txtar"
+)
+
+// builtinReadFile reads a file relative to the current Starlark test file.
+func builtinReadFile(
```
evankanderson: Does it make sense to have a struct that tracks execution of a starlark file (module) which include the file's location and/or a filesystem rooted at that directory?  If this is the Runner, you might do it as:

```
func (r *Runner) builtinReadFile(...) {starlark.Value, error) {
	var path string
	if err := starlark.UnpackArgs(b.Name(), args, kwargs, "path", &path); err != nil {
		return nil, err
	}
	data, err := util.ReadFile(r.fs, path)
	return starlark.String(string(data)), err
}
```

pkg/ruletest/builtins_file.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// builtinTxtar parses a txtar string and returns a Starlark dictionary.
+func builtinTxtar(
+	_ *starlark.Thread,
+	b *starlark.Builtin,
+	args starlark.Tuple,
+	kwargs []starlark.Tuple,
+) (starlark.Value, error) {
+	var content string
```
evankanderson, line 51: We should document (somewhere) that these methods only support `string` contents, not bytestreams.  I think that's okay, but there are some bytestreams that aren't valid UTF-8 strings, and so can't be handled.  I don't *think* that will be a problem for testing, but the other option would be to have these two builtins use `[]byte` for the file contents (and maybe have to convert in Starlark).

pkg/ruletest/builtins_file_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.starlark.net/starlark"
+	"go.starlark.net/syntax"
+)
+
+func TestBuiltinReadFile(t *testing.T) {
```
evankanderson: I'd be more inclined to test these via actual Starlark tests and show the utility and full wiring over unit tests.

Unit tests can be good for achieving edge case coverage that can be hard to set up, but I don't feel like we're at that point with this framework.  (Possibly when we get into the Providers we'll need some of that.)

pkg/ruletest/builtins_file_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	v, found, err := dict.Get(starlark.String("file1.txt"))
+	require.NoError(t, err)
+	require.True(t, found)
+	assert.Equal(t, "hello\n", v.(starlark.String).GoString())
+
+	v2, found, err := dict.Get(starlark.String("file2.txt"))
+	require.NoError(t, err)
+	require.True(t, found)
+	assert.Equal(t, "world\n", v2.(starlark.String).GoString())
```
evankanderson: Again, I feel like once you start repeating blocks like `require.NoError(), require.True(), assert.Equal(...)`, you may be better off using a table-driven test for _clarity_ -- it helps humans understand that you're doing the same steps twice or N times.  Note that you don't necessarily need to use the `t.Run()` sub-test style to have tables driving the test.

pkg/ruletest/builtins_file.go
krrish175-byte: I have refactored the thread context to hold an `fs.FS` (created via `os.DirFS(baseDir)`) and swapped `os.ReadFile` with `fs.ReadFile`. This naturally handles the symlink escapes and properly ties the execution to a virtual filesystem.

pkg/ruletest/builtins_file.go
krrish175-byte: This is exactly what I did. I updated the `threadContext` to use `fs.FS` (created via `os.DirFS`), which takes care of this perfectly as well as the symlink escapes.

pkg/ruletest/builtins_file.go
krrish175-byte: Good point. I have added documentation comments to both `builtinReadFile`and `builtinTxtar` noting that they specifically only support text files containing valid UTF-8 strings.

pkg/ruletest/builtins_file_test.go
krrish175-byte: I have completely removed the Go unit tests for these built-ins and replaced them with `builtins_test.star` to test everything end-to-end via Starlark.

pkg/ruletest/builtins_file_test.go
krrish175-byte: Since we completely switched to Starlark tests for these, these repeated blocks are gone now. We just use `assert.fails` directly in the `.star` file.

pkg/ruletest/testdata/builtins_test.star
```diff
@@ -0,0 +1,19 @@
+def test_read_file():
+    data = read_file("builtins_test.star")
+    assert.true("def test_read_file():" in data)
```
evankanderson, line 3: ❤️   I love the recursive test here.

pkg/ruletest/testdata/builtins_test.star
```diff
@@ -0,0 +1,19 @@
...
+
+def test_read_file_traversal():
+    assert.fails(lambda: read_file("../data.txt"), "invalid argument")
```
evankanderson: What happens if you don't put this in an `assert.fails`?  I'm assuming the whole test fails?

What do you think about expecting tests named `test_fail_...` to error out?

pkg/ruletest/testdata/builtins_test.star
```diff
@@ -0,0 +1,19 @@
...
+
+def test_txtar():
+    archive = txtar("""-- file1.txt --
+hello
+-- file2.txt --
+world
+""")
+    assert.eq(len(archive), 2)
+    assert.eq(archive["file1.txt"], "hello\n")
+    assert.eq(archive["file2.txt"], "world\n")
```
evankanderson, line 19: Do you want to add an error-case test for `txtar`?

pkg/ruletest/runner.go
```diff
@@ -59,7 +77,22 @@ func NewRunner() *Runner {
 // for each test_* function found in it.
 // src may be nil, or a string, []byte, or io.Reader containing the file source.
 func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
-	thread := r.newThread("ruletest")
+	baseDir := ""
+	if filename != "" {
+		baseDir = filepath.Dir(filename)
```
evankanderson: `filepath.Dir("")` produces `.`, as you'd want.  https://pkg.go.dev/path/filepath#example-Dir

pkg/ruletest/runner.go
```diff
@@ -59,7 +77,22 @@ func NewRunner() *Runner {
...
+		if !filepath.IsAbs(baseDir) {
```
evankanderson: Why bother with this normalization step?  The library should do it when you go to use it anyway.

pkg/ruletest/runner.go
```diff
@@ -18,7 +19,22 @@ import (
...
+}
+
+func getThreadContext(thread *starlark.Thread) *threadContext {
+	ctx, ok := thread.Local(threadContextKey).(*threadContext)
+	if !ok {
+		ctx = &threadContext{}
+		thread.SetLocal(threadContextKey, ctx)
+	}
+	return ctx
+}
```
evankanderson, line 59: This "feels" weird, especially as it seems somewhat parallel to the go `context.Context`.  Why not use bound functions with a struct, e.g.


```suggestion
type testRunner struct {
	thread      *starlark.Thread
	fs          fs.FS
	predeclared starlark.StringDict
	failures    []string
}

func newRunner(filename string) *testRunner {
	tr := testRunner{
		fs: os.DirFS(filepath.Dir(filename))
		predeclared: starlark.StringDict{}
	}
	tr.thread := &starlark.Thread{
		Name:  filepath.Base(filename),
		Print: func(_ *starlark.Thread, msg string) { fmt.Println(msg) },
	}
	starlarktest.SetReporter(tr.thread, &tr)  // testRunner implements the reporter interface and stores things in failures
	tr.predeclared["eval"] = builtinEval
	tr.predeclared["read_file"] = tr.ReadFile
	tr.predeclared["txtar" = builtinTxTar
	// also copy the assert module
	return &tr
}}

func (tr *testRunner) ReadFile(
	thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple,
) (starlark.Value, error) {
	var path string
	if err := starlark.UnpackArgs(b.Name(), args, kwargs, "path", &path); err != nil {
		return nil, err
	}

	data, err := tr.fs.ReadFile(ctx.fs, path)

	return starlark.String(string(data)), err
}

func (tr *testRunner) appendFailure(msg string) {
	tr.failures = append(tr.failures, msg)
}
```

evankanderson (commented):
If you make "Runner" or "testRunner" a controlling object, you might end up wanting a "Child" or "Clone" method to be used for running each test (which could inherit the FS object and possibly manage its own list of failures which could be aggregated by the parent test object).

pkg/ruletest/runner.go
```diff
@@ -33,35 +66,37 @@ func (tr *TestResult) Passed() bool {
...
 
 // RunFile executes a single Starlark test file and returns the results
 // for each test_* function found in it.
 // src may be nil, or a string, []byte, or io.Reader containing the file source.
 func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
-	thread := r.newThread("ruletest")
+	baseDir := ""
+	if filename != "" {
+		baseDir = filepath.Dir(filename)
+	}
```
evankanderson: [`filepath.Dir` will return `.` for an empty string](https://pkg.go.dev/path/filepath#Dir), which seems fine?

```suggestion
	baseDir = filepath.Dir(filename)
```

pkg/ruletest/runner.go
```diff
@@ -33,35 +66,37 @@ func (tr *TestResult) Passed() bool {
...
+
+	var fileSystem fs.FS
+	if baseDir != "" {
+		fileSystem = os.DirFS(baseDir)
```
evankanderson: If `baseDir` is empty-string, it seems like we'd end up with a `nil` filesystem?

pkg/ruletest/runner.go
```diff
@@ -33,35 +66,37 @@ func (tr *TestResult) Passed() bool {
...
+	}
+
+	tr := r.newTestCaseRunner("ruletest", fileSystem)
```
evankanderson: Maybe use `filepath.Base(filename)` for the name of the test case runner?

pkg/ruletest/runner.go
```diff
@@ -33,35 +66,37 @@ func (tr *TestResult) Passed() bool {
...
 
-	globals, err := starlark.ExecFileOptions(&syntax.FileOptions{}, thread, filename, src, r.predeclared)
+	globals, err := starlark.ExecFileOptions(&syntax.FileOptions{}, tr.thread, filename, src, tr.predeclared)
```
evankanderson: What do you think about moving this into `testCaseRunner.RunFile(filename, src)`?

pkg/ruletest/runner.go
```diff
@@ -116,7 +141,15 @@ func (r *Runner) runOneTest(name string, fn *starlark.Function) TestResult {
...
-	result.Failures = append(result.Failures, getFailures(thread)...)
+	result.Failures = append(result.Failures, tr.failures...)
+
+	if strings.HasPrefix(name, "test_fail_") {
+		if len(result.Failures) == 0 {
+			result.Failures = []string{"expected test to fail, but it succeeded"}
+		} else {
+			result.Failures = nil // test failed as expected
+		}
+	}
```
evankanderson: I think I was unclear -- I'd meant to put this into `runner_test.go`, not into the runner code itself.

pkg/ruletest/runner_test.go
evankanderson, line 123: So, put the "expect `test_fail_` to not pass" check here...

pkg/ruletest/builtins_file.go
```diff
@@ -0,0 +1,68 @@
...
+		return nil, fmt.Errorf("read_file: %w", err)
+	}
+
+	return starlark.String(string(data)), nil
+}
+
+// builtinTxtar parses a txtar string and returns a Starlark dictionary.
+// Note: This builtin only supports text files containing valid UTF-8 strings.
+// Bytestreams that aren't valid UTF-8 strings are not supported.
+func (_ *testCaseRunner) builtinTxtar(
```
evankanderson: You have a lint error here.

pkg/ruletest/eval.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/mindersec/minder/pkg/fileconvert"
 )
 
-func builtinEval(
+func (_ *testCaseRunner) builtinEval(
```
evankanderson: (and the same lint issue here)

You don't need to make all these methods part of testCaseRunner if they don't need the internal state.

evankanderson (commented):
This is starting to look a lot better (IMO).  One more round of comments, but I think this is getting close.

pkg/ruletest/testdata/builtins_test.star
```diff
@@ -0,0 +1,22 @@
...
+
+def test_fail_txtar_invalid_arg():
+    txtar(123)
```
evankanderson, line 22: Re-reading the txtar format, it appears that the following is a valid zero-entry txtar with a "comment":

```
    # txtar treats this file contents as a comment:
    txtar("""{
  "name": "my-package",
  "version": "0.0.1",
  "engines": {
    "node": ">22.0"
  }
}
""")
```

I'm not sure how I feel about that from an anti-error perspective, but this test is a good error-handling check.

krrish175-byte:
I have addressed the latest feedback. `builtinTxtar` and `builtinEval` have been decoupled from `testCaseRunner` since they don't require the internal state, and I have fixed the linting error regarding the exported method. PTAL

pkg/ruletest/builtins_file.go
```diff
@@ -0,0 +1,68 @@
...
+	var path string
+	if err := starlark.UnpackArgs(b.Name(), args, kwargs, "path", &path); err != nil {
+		return nil, err
+	}
+
+	if tr.fs == nil {
+		return nil, fmt.Errorf("read_file: no file system available in context")
+	}
+
+	// fs.ReadFile enforces unrooted, valid paths (preventing typical directory traversal)
```
evankanderson, line 33: This is actually the `os.DirFs` (or [`os.Root`](https://pkg.go.dev/os#Root)) implementation, not `ReadFile` per se.

(This is mostly a nit, but given the comment was here, I wanted to make clear that the actual enforcement you're talking about is a property of line 93 of `runner.go`.)

pkg/ruletest/builtins_file.go
```diff
@@ -0,0 +1,68 @@
...
+	kwargs []starlark.Tuple,
+) (starlark.Value, error) {
+	var path string
+	if err := starlark.UnpackArgs(b.Name(), args, kwargs, "path", &path); err != nil {
+		return nil, err
+	}
+
+	if tr.fs == nil {
+		return nil, fmt.Errorf("read_file: no file system available in context")
+	}
```
evankanderson, line 31: If you enforce the invariant that `tr.fs` is always set (e.g. via constructors), then you can avoid this check.  It's not a big deal, just pointing out some ways that you may find other people structuring their programs.

pkg/ruletest/builtins_file.go
```diff
@@ -0,0 +1,68 @@
...
+	var content string
+	if err := starlark.UnpackArgs(b.Name(), args, kwargs, "content", &content); err != nil {
+		return nil, err
+	}
+
+	archive := txtar.Parse([]byte(content))
+	dict := starlark.NewDict(len(archive.Files))
+
+	for _, f := range archive.Files {
+		key := starlark.String(strings.TrimSpace(f.Name))
```
evankanderson, line 60: txtar already trims the whitespace around the file names: https://pkg.go.dev/golang.org/x/tools/txtar#hdr-Txtar_format

pkg/ruletest/runner.go
```diff
@@ -33,35 +71,34 @@ func (tr *TestResult) Passed() bool {
...
 // for each test_* function found in it.
 // src may be nil, or a string, []byte, or io.Reader containing the file source.
 func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
-	thread := r.newThread("ruletest")
+	baseDir := filepath.Dir(filename)
+	fileSystem := os.DirFS(baseDir)
 
-	globals, err := starlark.ExecFileOptions(&syntax.FileOptions{}, thread, filename, src, r.predeclared)
+	name := "ruletest"
+	if filename != "" {
```
evankanderson, line 96: An empty filename is probably an error.  (`filepath.Base("")` produces ".", which doesn't really help you.)

evankanderson (approved):
A few thoughts for future code, but this looks good, merging!

PR #6531, opened 2026-06-21
feat(alert): allow line comments for PR alert type (#5118)

### LINKED ISSUE #5118: Allow line comments for PR alert type
Enhance the PR alert type to allow Minder to leave comments on specific lines.

Out of scope:
- Reading the output of the evaluation
- Gitlab

MahekPatel-2403:
# Summary

Adds support for line comments in PR alert reviews.

Changes include:
- Added line comment support to the PR alert workflow.
- Updated protobuf definitions and generated files.
- Added validation and rendering logic for single-line and multi-line comments.
- Refactored PR comment action handling into smaller helper functions to reduce cyclomatic complexity.
- Added comprehensive tests covering line comment scenarios.

Fixes #5118

# Testing

- Added unit tests for PR comment handling.
- Verified single-line and multi-line comment rendering.
- Verified review creation and update flows.

### DIFF
--- a/internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
+++ b/internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
@@ -6,6 +6,7 @@
 package pull_request_comment
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
 	"errors"
@@ -355,9 +356,6 @@ func (alert *Alert) getParamsForPRComment(
 
 	result.RuleName = params.GetRule().Name
 
-	action := alert.reviewCfg.GetAction()
-	if action == "" {
-		action = "comment"
 	action := cmp.Or(alert.reviewCfg.GetAction(), "comment")
 	if strings.ToLower(action) == "request_changes" {
 		result.Event = "REQUEST_CHANGES"
--- a/internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
+++ b/internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
@@ -57,15 +57,16 @@ type PrCommentTemplateParams struct {
 }
 
 type paramsPR struct {
-	Owner      string
-	Repo       string
-	CommitSha  string
-	Number     int
-	Comment    string
-	RuleName   string
-	Event      string
-	Metadata   *alertMetadata
-	prevStatus *db.ListRuleEvaluationsByProfileIdRow
+	Owner        string
+	Repo         string
+	CommitSha    string
+	Number       int
+	Comment      string
+	LineComments []*github.DraftReviewComment
+	RuleName     string
+	Event        string
+	Metadata     *alertMetadata
+	prevStatus   *db.ListRuleEvaluationsByProfileIdRow
 }
 
 type alertMetadata struct {
@@ -160,11 +161,29 @@ func (alert *Alert) run(ctx context.Context, params *paramsPR, cmd interfaces.Ac
 			if _, err := alert.gh.UpdateReview(ctx, params.Owner, params.Repo, params.Number, reviewID, params.Comment); err != nil {
 				return nil, fmt.Errorf("error updating PR review: %w, %w", err, enginerr.ErrActionFailed)
 			}
-			logger.Info().Int64("review_id", reviewID).Msg("PR review updated")
+			if len(params.LineComments) > 0 {
+				// The GitHub UpdateReview API only accepts a new body; it cannot add new
+				// line-level comments to an already-submitted review. Line comments are
+				// only posted when the review is first created.
+				logger.Warn().Int64("review_id", reviewID).
+					Int("line_comments", len(params.LineComments)).
+					Msg("PR review updated but line comments were not re-posted (GitHub API limitation)")
+			} else {
+				logger.Info().Int64("review_id", reviewID).Msg("PR review updated")
+			}
 		} else {
 			req := &github.PullRequestReviewRequest{
 				Body:  github.String(params.Comment),
 				Event: github.String(params.Event),
+			}
+			// CommitID is required by GitHub for reviews that include line comments;
+			// it tells GitHub which commit's diff to annotate.
+			if params.CommitSha != "" {
+				req.CommitID = github.String(params.CommitSha)
+			}
+			// Attach line-specific draft comments if any were configured
+			if len(params.LineComments) > 0 {
+				req.Comments = params.LineComments
 			}
 			review, err := alert.gh.CreateReview(ctx, params.Owner, params.Repo, params.Number, req)
 			if err != nil {
@@ -172,7 +191,9 @@ func (alert *Alert) run(ctx context.Context, params *paramsPR, cmd interfaces.Ac
 			}
 			reviewID = review.GetID()
 			existingReview = review
-			logger.Info().Int64("review_id", reviewID).Msg("PR review created")
+			logger.Info().Int64("review_id", reviewID).
+				Int("line_comments", len(params.LineComments)).
+				Msg("PR review created")
 		}
 
 		now := time.Now()
@@ -244,6 +265,9 @@ func (alert *Alert) runDry(ctx context.Context, params *paramsPR, cmd interfaces
 		body := github.String(params.Comment)
 		logger.Info().Msgf("dry run: create a PR comment on PR %d in repo %s/%s with the following body: %s",
 			params.Number, params.Owner, params.Repo, *body)
+		if len(params.LineComments) > 0 {
+			logger.Info().Msgf("dry run: would also post %d line comment(s)", len(params.LineComments))
+		}
 		return nil, nil
 	case interfaces.ActionCmdOff:
 		if params.Metadata == nil || params.Metadata.ReviewID == "" {
@@ -327,6 +351,31 @@ func (alert *Alert) getParamsForPRComment(
 
 	// Add magic comment to identify Minder reviews for this rule
 	result.Comment = fmt.Sprintf("%s\n\n<!-- minder-rule: %s -->", comment, result.RuleName)
+
+	// Render and collect line-specific draft comments
+	for i, lc := range alert.reviewCfg.GetLineComments() {
+		commentBody := lc.GetComment()
+		lineTmpl, err := util.NewSafeHTMLTemplate(&commentBody, fmt.Sprintf("line_comment_%d", i))
+		if err != nil {
+			return nil, fmt.Errorf("cannot parse line comment %d template: %w", i, err)
+		}
+		rendered, err := lineTmpl.Render(ctx, tmplParams, PrCommentMaxLength)
+		if err != nil {
+			return nil, fmt.Errorf("cannot render line comment %d: %w", i, err)
+		}
+
+		draftComment := &github.DraftReviewComment{
+			Path:     github.String(lc.GetFilepath()),
+			Body:     github.String(rendered),
+			Line:     github.Int(int(lc.GetLineEnd())),
+			StartLine: github.Int(int(lc.GetLineStart())),
+		}
+		// If the comment spans only one line, omit StartLine (GitHub requires start_line < line)
+		if lc.GetLineStart() == lc.GetLineEnd() {
+			draftComment.StartLine = nil
+		}
+		result.LineComments = append(result.LineComments, draftComment)
+	}
 
 	// Unmarshal the existing alert metadata, if any
 	if metadata != nil {
--- a/internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
+++ b/internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
@@ -6,7 +6,6 @@
 package pull_request_comment
 
 import (
-	"cmp"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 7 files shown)


internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
```diff
@@ -318,7 +355,10 @@ func (alert *Alert) getParamsForPRComment(
 
 	result.RuleName = params.GetRule().Name
 
-	action := cmp.Or(alert.reviewCfg.GetAction(), "comment")
+	action := alert.reviewCfg.GetAction()
+	if action == "" {
+		action = "comment"
+	}
```
krrish175-byte: Why did you replace `cmp.Or` here? Using `cmp.Or` is our preferred standard pattern in Go for falling back to default values. Let's revert this unless there's a specific reason to avoid it.
```suggestion
	action := cmp.Or(alert.reviewCfg.GetAction(), "comment")
```

krrish175-byte (commented):
A bunch of comments, but this is looking pretty good. Adding line-level comment support to our PR action is a great feature and the refactor of `Alert.run()` makes it much cleaner. Once you've made the adjustments from these comments, we can merge.

pkg/api/protobuf/go/minder/v1/validators.go
```diff
@@ -398,6 +398,29 @@ func (comment *RuleType_Definition_Alert_AlertTypePRComment) Validate() error {
...
 	}
 
+	for i, lc := range comment.GetLineComments() {
+		if lc.GetFilepath() == "" {
+			return fmt.Errorf("%w: line_comment %d: filepath cannot be empty", ErrInvalidRuleTypeDefinition, i)
+		}
+		if lc.GetLineStart() <= 0 {
+			return fmt.Errorf("%w: line_comment %d: line_start must be greater than 0", ErrInvalidRuleTypeDefinition, i)
+		}
+		if lc.GetLineEnd() <= 0 {
```
krrish175-byte, line 408: Since `line_start` and `line_end` default to `0` in protobufs, requiring both to be `> 0` means a user *must* specify both, even for a single-line comment. I see you handle `line_start == line_end` gracefully in `getParamsForPRComment` by omitting `StartLine` for the GitHub API.

*This is me thinking out loud:* Should we allow `line_start` to be omitted (i.e. `0`) and default to `line_end` for single-line comments in the profile validation logic, so the user only has to specify `line_end` for a single-line comment? Pick an option. If we keep the current behavior, just make sure to document that both are required.

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
```diff
@@ -139,83 +139,117 @@ func (alert *Alert) Do(
...
+	case interfaces.ActionCmdOn:
+		return alert.handleActionOn(ctx, params)
+	// Dismiss the review
+	case interfaces.ActionCmdOff:
+		return alert.handleActionOff(ctx, params)
+	case interfaces.ActionCmdDoNothing:
+		// Return the previous alert status.
+		return alert.runDoNothing(ctx, params)
+	}
+	return nil, enginerr.ErrActionSkipped
```
krrish175-byte, line 155: Nice extraction of `handleActionOn` and `handleActionOff`, it definitely cleans up `run()`.

internal/engine/actions/alert/pull_request_comment/pull_request_comment_test.go
```diff
@@ -241,6 +241,166 @@ func TestPullRequestCommentAlert(t *testing.T) {
 	}
 }
 
+func TestPullRequestLineCommentAlert(t *testing.T) {
```
krrish175-byte, line 244: The success test cases look great. However, can we add an (error) test case for when the line comment template rendering fails? Edge cases matter, and we want to ensure we aren't silently failing or panicking if the template is malformed.

pkg/api/protobuf/go/minder/v1/validators.go
MahekPatel-2403: I kept both line_start and line_end required for consistency and explicitness in the validation logic. Since the GitHub API supports deriving a single-line comment when line_start == line_end, I thought requiring both fields would make the configuration less ambiguous. I'm happy to adjust the validation to allow omitting line_start if you think that provides a better user experience.

proto/minder/v1/minder.proto
```diff
@@ -2972,6 +2972,38 @@ message RuleType {
...
+                        },
+                        (google.api.field_behavior) = REQUIRED
+                    ];
+                    // line_end is the ending line number.
+                    int32 line_end = 3 [
+                        (buf.validate.field).int32 = {
+                            gt: 0,
+                        },
+                        (google.api.field_behavior) = REQUIRED
+                    ];
```
evankanderson, line 2997: Is `line_end` required to be greater than or equal to `line_start`?

For GitLab, I'm assuming you'd use the [thread in merge request diff API](https://docs.gitlab.com/api/discussions/#create-a-new-thread-in-the-merge-request-diff), as that's the only one that seems to support -multi-line comments.  You might look at the APIs for both GitLab and [GitHub](https://docs.github.com/en/rest/pulls/comments#create-a-review-comment-for-a-pull-request) to see what kinds of parameters they accept.  For example, both seem to have the ability comment on either the "old" or "new" side (LEFT or RIGHT in GitHub).

proto/minder/v1/minder.proto
```diff
@@ -2972,6 +2972,38 @@ message RuleType {
...
+                    // comment is the comment to post on those lines.
+                    string comment = 4 [
+                        (buf.validate.field).string = {
+                            max_len: 65536,
+                        },
+                        (google.api.field_behavior) = REQUIRED
+                    ];
+                }
+                repeated PullRequestLineComment line_comments = 3;
```
evankanderson, line 3006: Practically, how does using `line_comments` like this work?  I can see how a single PR comment with template expansion could address finding anywhere from 1-100 items (using a loop in the template to repeat contents), but I don't see how this format could work for something that might produce more than one finding.

pkg/api/protobuf/go/minder/v1/validators.go
```diff
@@ -398,6 +398,29 @@ func (comment *RuleType_Definition_Alert_AlertTypePRComment) Validate() error {
...
+			return fmt.Errorf("%w: line_comment %d: line_end must be greater than 0", ErrInvalidRuleTypeDefinition, i)
+		}
+		if lc.GetLineStart() > lc.GetLineEnd() {
+			return fmt.Errorf("%w: line_comment %d: line_start (%d) must be <= line_end (%d)",
+				ErrInvalidRuleTypeDefinition, i, lc.GetLineStart(), lc.GetLineEnd())
+		}
```
evankanderson, line 414: Does this mean that we can't have single-line comments?

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
```diff
@@ -139,83 +139,117 @@ func (alert *Alert) Do(
...
+}
```
evankanderson, line 156: It's helpful to do this kind of refactoring in a separate PR as it makes it easier to check that existing behavior has been preserved.  You don't need to do that this time, but consider it a note for next time.

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
```diff
@@ -139,83 +139,117 @@ func (alert *Alert) Do(
...
+		if _, err := alert.gh.UpdateReview(ctx, params.Owner, params.Repo, params.Number, reviewID, params.Comment); err != nil {
+			return nil, fmt.Errorf("error updating PR review: %w, %w", err, enginerr.ErrActionFailed)
+		}
+		if len(params.LineComments) > 0 {
+			// The GitHub UpdateReview API only accepts a new body; it cannot add new
+			// line-level comments to an already-submitted review. Line comments are
+			// only posted when the review is first created.
+			logger.Warn().Int64("review_id", reviewID).
+				Int("line_comments", len(params.LineComments)).
+				Msg("PR review updated but line comments were not re-posted (GitHub API limitation)")
```
evankanderson, line 182: If the action proposes line comments that are different from the previous, should we leave a _new_ comment with the new line comments?

(I'm wondering what would be useful to users, not what's simplest to implement.)

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
```diff
@@ -318,6 +355,9 @@ func (alert *Alert) getParamsForPRComment(
 
 	result.RuleName = params.GetRule().Name
 
+	action := alert.reviewCfg.GetAction()
+	if action == "" {
+		action = "comment"
```
evankanderson: What's going on here?  Indentation seems off.

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
```diff
@@ -328,6 +368,31 @@ func (alert *Alert) getParamsForPRComment(
...
+			return nil, fmt.Errorf("cannot render line comment %d: %w", i, err)
+		}
+
+		draftComment := &github.DraftReviewComment{
+			Path:      github.String(lc.GetFilepath()),
+			Body:      github.String(rendered),
+			Line:      github.Int(int(lc.GetLineEnd())),
+			StartLine: github.Int(int(lc.GetLineStart())),
+		}
+		// If the comment spans only one line, omit StartLine (GitHub requires start_line < line)
```
evankanderson, line 387: Validation doesn't allow this to happen, does it?

evankanderson (changes requested):
Sorry that I didn't look at the spec in #5118 closely enough when you started -- it's not clear that the `line_comments` as specified are particularly useful for PR comments, as (for example) the number and position of the comments are fixed.  This _might_ make sense if you wanted to comment on line 1 of a particular file that the contents should be a specific value, but a lot of common checks (e.g. the Chainguard "boilerplate" checks that we run to check for license headers) have at least some variance in terms of either filename or position in the file.

I think you may need to go back to #5118 and work out a few use cases more clearly to get a better spec for the API than was described in the issue.  (To explain myself -- I'd read the title and said "yeah, we need line-level comments" and didn't actually read the implementation suggestion.  🤦 )

Sorry to have sent you in the wrong direction!

proto/minder/v1/minder.proto
MahekPatel-2403: Thanks for taking another look at the spec. That makes sense .  I implemented the API based on the requirements described in #5118, but I can see how fixed line positions may not be flexible enough for many real-world checks.

Regarding line_start and line_end, my assumption was:

- Single-line comments: line_start == line_end
- Multi-line comments: line_end >= line_start

I'll revisit the use cases in #5118 and review the GitHub and GitLab review comment APIs to better understand the positioning models before making further changes.
Thanks for the feedback and clarification.

proto/minder/v1/minder.proto
MahekPatel-2403: That was my assumption as well — each entry in `line_comments` would map to a separate review comment tied to a specific line or line range.

I agree that this could become noisy for checks that produce a large number of findings. A single PR comment with template expansion would definitely scale better in those cases.

pkg/api/protobuf/go/minder/v1/validators.go
MahekPatel-2403: Single-line comments should still be allowed. My intention was only to prevent invalid ranges where `line_start > line_end`.

For a single-line comment, `line_start == line_end`, while multi-line comments would have `line_end > line_start`.

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
MahekPatel-2403: Thanks, that's a good point. I'll keep feature changes and refactors separate in future PRs where possible.

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
MahekPatel-2403: My initial assumption was that line comments would only be posted when the review is first created, and subsequent updates would continue updating the review body rather than creating additional line comments.

That said, I haven't thought through the user experience deeply yet. If findings change between runs, creating new line comments might make the updated findings more visible, but it could also lead to noise and duplicate discussions.

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
MahekPatel-2403: I ran gofmt locally and it didn't make any changes. Is there a specific indentation issue you're seeing?

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
MahekPatel-2403: Good catch. With the current validation, single-line comments would still have both line_start and line_end set, so this path should only be reached when line_start == line_end. I'll double-check whether this special handling is still needed.

pkg/api/protobuf/go/minder/v1/validators.go
evankanderson: Whoops, I read the conditional backwards.

pkg/api/protobuf/go/minder/v1/validators.go
evankanderson: I think omitting `line_end` might be more natural.

proto/minder/v1/minder.proto
evankanderson: As a practical example, it would be great to be able to replace the [boilerplate license checks](https://github.com/mindersec/minder/blob/main/.github/workflows/boilerplates.yaml) with a Minder PR check / remediation.

This might be a good working example for some of the features needed.

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
evankanderson: Issue #5118 doesn't really provide a lot of motivation, so you may want to lay out some use cases.

Here are a couple:

1. Minder currently has a [check for license boilerplate at the top of many file types](https://github.com/mindersec/minder/blob/main/.github/workflows/boilerplates.yaml).  This burns a lot of actions time due to the implementation (it needs to build a go binary and then run it once for each file type), and adds a bunch of dependencies to the CI process.
2. Minder has [static (built-in) checks](https://github.com/mindersec/minder/tree/main/internal/engine/eval/homoglyphs/application) for various unicode attacks from the [Trojan Source paper](https://trojansource.codes/).  It would be great if Minder were flexible enough that these could instead be encoded as standard alert rules.

I suspect that we'll need to do some design work to figure out how to pass the data from the rule evaluation to the remediation to get these to work -- our typical mechanism for doing this is [the `output` parameter from the rule evaluation](https://docs.mindersec.dev/ref/rule_evaluation_details#alert-inputs), but you may need to define some mechanism to take structured output and convert it into a set of line comments, rather than defining the line comments statically in the rule definition.

evankanderson (commented):
I've suggested a couple sample "test cases" you might be able to use while developing this feature, but feel free to come up with your own.

pkg/api/protobuf/go/minder/v1/validators.go
MahekPatel-2403: That's a good suggestion. I agree that omitting line_end for single-line comments would be a more natural API. I'll keep that in mind while revisiting the design.

proto/minder/v1/minder.proto
MahekPatel-2403: That's a helpful example. I'll use the boilerplate license checks as a concrete use case and work through how the API would handle multiple findings before making further changes.

internal/engine/actions/alert/pull_request_comment/pull_request_comment.go
MahekPatel-2403: These examples help clarify the problem space a lot. I can see how checks like the license boilerplate and Trojan Source detection would need a more dynamic way of generating review comments from evaluation results. I'll spend some time working through these use cases and see how they fit with the current design.

MahekPatel-2403:
Hi! Just checking in on this PR. I've addressed the review comments and replied to the discussions. Let me know if there are any further changes you'd like me to make. Thanks!

evankanderson:
I'm concerned about the API surface for PR line comments.  The current design seems to be limited to fixed lines and files, which seems unlikely to be generally useful.  In particular, I don't want to merge the current API because it will commit us to supporting that API even if we come up with a much better solution in the future (for example, specifying a rule output key pointing to either [SARIF](https://sarifweb.azurewebsites.net/) or a custom JSON format value).

MahekPatel-2403:
Thanks for the clarification. I understand the concern about committing to an API that may be too restrictive. I'll spend some time thinking about how a more flexible approach, such as driving review comments from structured rule output, could fit into Minder before making further changes.

PR #6534, opened 2026-06-23
docs: GitLab provider enrollment and repository registration guides

intelligent-ears:
- Adds getting-started docs for the GitLab provider, mirroring the existing GitHub flow. Fixes an invalid `profile` OAuth scope in the GitLab config doc (GitLab rejects it; `read_user` is correct). Makes `repository_registration.md` provider-agnostic and cross-links GitHub/GitLab enrollment docs.
- Tested locally end-to-end: enrollment, repo listing, and the expected webhook timeout when running against localhost (documented as a known limitation).


docs/docs/getting_started/enroll_gitlab_provider.md
```diff
@@ -0,0 +1,43 @@
...
+Minder to inspect and manage your repository configuration. You will be
+prompted to grant Minder access.
+
+## Prerequisites
+
+Before you can enroll the GitLab provider, you must:
+
+- [log in to Minder using the CLI](login)
+- have a [GitLab OAuth application configured](../run_minder_server/config_gitlab_oauth)
+  on your Minder server
```
evankanderson: Since this is a "getting started" guide, let's orient this towards new users, and assume that the server they are using is configured sufficiently.


```suggestion
```

docs/docs/getting_started/enroll_gitlab_provider.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Enrolling the GitLab provider
+sidebar_position: 45
+---
+
+Once you have authenticated to Minder, you'll need to enroll your GitLab
+credentials to allow Minder to manage your GitLab repositories. This allows
```
evankanderson: Since this is a "getting started" guide, **I think it's fine to point people towards the OAuth flow here** (it's easiest to set up), but it has some substantial downsides over the GitHub App or GitLab [service account with PAT](https://docs.mindersec.dev/integrations/provider_integrations/gitlab#authorization-methods) methods. We should link to the GitLab provider integration docs for a more "production" setup after the user has done the "getting started" flow.

The downsides of the OAuth approach:

1. Minder is acting _as_ the enrollee.  This makes MRs harder (you're reviewing "your own" MR when Minder makes a suggestion) and general attribution of changes more difficult.
2. It's hard to limit what Minder has access to if the user has access to multiple projects.
3. If a user leaves a project, Minder configuration for that project will stop working.

For short term or personal use, OAuth works fine -- for larger projects or projects with multiple administrators, the service account is more work but ultimately a better option.

docs/docs/getting_started/register_repos_gitlab.md
```diff
@@ -0,0 +1,56 @@
...
+
+:::note
+Webhook registration requires GitLab to be able to reach your Minder
+server's webhook endpoint. If you're running Minder locally (e.g. against
+`localhost`), GitLab cannot reach it, and repository registration will fail
+with a `context deadline exceeded` error. To test repository registration
+locally, expose your server with a tool like `ngrok` and set
+`external_webhook_url` in your server config accordingly, or test against a
+publicly reachable Minder deployment.
+:::
```
evankanderson: One of the tricky things about technical writing (and writing in general) is crafting your content for your target audience.  In this case, we're writing to users who are unfamiliar with Minder and are trying to decide if it's worth it for them ("is the juice worth the squeeze" to quote a former colleague).  In this case, we're assuming that they'll use a hosted instance that is set up by someone else, rather than setting up and managing an instance themselves.

This belongs in something like the developer guide, but not the "getting started" guide.


```suggestion
```

docs/docs/getting_started/register_repos_gitlab.md
evankanderson, line 1: This doc looks _almost_ identical to the existing "Registering repositories" document.  What do you think about simply removing the "GitHub" wording from that document or replacing it with "GitHub or GitLab"?

docs/docs/getting_started/enroll_gitlab_provider.md
evankanderson, line 1: Since this is in the getting started flow (see also [this comment on audience](https://github.com/mindersec/minder/pull/6534/changes#r3461932926)), what do you think about folding this into the existing "Enrolling the GitHub provider" (and changing the name to "Enroll a provider")?

You can put the GitHub vs GitLab content in tabs using the format from `how-to/pr_reviews.mdx` (you'll need to make the file an MDX file):

```
import Tabs from '@theme/Tabs';
import TabItem from '@theme/TabItem';

<!-- ... -->
# Enrolling and granting access
<Tabs>
<TabItem value="github" label="GitHub" default>
....
</TabItem>
<TabItem value="gitlab" label="GitLab">
...
</TabItem>
```

Using tabs will allow the "next" and "previous" buttons at the bottom to flow smoothly, rather than needing to sometimes fork or skip over steps.

evankanderson (commented):
Thanks for adding this!  Since the "getting started" section is particularly aimed at getting users who haven't tried Minder up and running, I'm going to be particularly precise about writing for that audience.  In particular, beginners often don't know which parts are important, which is why [_tutorials_](https://diataxis.fr/start-here/#tutorials) need to be particularly clear.  (In general, I've found the https://diataxis.fr/ approach to documentation to be helpful reading, so I'm linking it here.)

intelligent-ears:
Thanks for the detailed feedback — this pushed the docs in a much better direction for the getting-started audience. Summary of changes made:

- Removed the OAuth-app-configuration prerequisite from the GitLab enrollment doc (assumes the server is already set up, per the beginner-audience framing)
- Added a note on OAuth tradeoffs (attribution/MR confusion, scoping, enrollment breaking if a user leaves a project), linking to the service account + PAT setup in the provider integration docs for production use
- Removed the local-dev webhook/`ngrok` note from the getting-started repo registration doc — that's developer-guide material, not beginner material
- Merged `enroll_provider.md` and `enroll_gitlab_provider.md` into a single tabbed `enroll_provider.mdx` (renamed to "Enroll a provider"), following the Tabs pattern from `pr_reviews.mdx`
- Merged `register_repos.md` and `register_repos_gitlab.md` into one provider-agnostic doc, since the content was nearly identical aside from "GitHub"/"GitLab" wording
- Fixed a stale link in `login.md` that pointed to the old `enroll_provider.md` path after the rename to `.mdx`

evankanderson (approved):

PR #6535, opened 2026-06-23
Improvements to ListEntities

evankanderson:
# Summary

This ended up being a few different changes to ListEntities; four are user-visible, and one is performance-related:

- Adds a proto field indicating the "originating" entity for resources like pull requests, which can help when building a tree of related entities.
- Reports the provider name associated with entities, which can also help with grouping
- Adds support for listing _all_ types of entities with a single API call
- Adds pagination support
- Converts the 1 + N database calls to fetch properties for entities to a single database call with additional data using JSON_OBJECT_AGG

# Testing

Manually tested as well as adding integration tests (assisted by Gemini 3 Flash, but adjusted by hand).

### DIFF
--- a/internal/entities/handlers/handler_test.go
+++ b/internal/entities/handlers/handler_test.go
@@ -6,6 +6,7 @@ package handlers
 import (
 	"encoding/json"
 	"errors"
+	"maps"
 	"testing"
 
 	watermill "github.com/ThreeDotsLabs/watermill/message"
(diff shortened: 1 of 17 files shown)


internal/entities/handlers/handler_test.go
```diff
@@ -6,14 +6,14 @@ package handlers
...
 	"errors"
+	"maps"
 	"testing"
 
 	watermill "github.com/ThreeDotsLabs/watermill/message"
 	"github.com/google/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
-	"golang.org/x/exp/maps"
```
evankanderson, line 16: This was fixing a lint error.

internal/entities/service/service.go
```diff
@@ -138,10 +148,12 @@ func (s *entityService) ListEntities(
 	qtx := s.store.GetQuerierWithTransaction(tx)
 
 	// TODO: Implement proper pagination with cursor
```
evankanderson:
```suggestion
```

JAORMX (approved):
Very nice!

PR #6541, opened 2026-06-23
Introduce issue provider facet and issue remediation configuration

### LINKED ISSUE #6532: Issue remediation: infrastructure
This issue tracks the infrastructure portion of the Issue Remediation design discussed in #6522 

## Goal

Introduce the infrastructure required to support Issue remediation.

## Scope

- Introduce the `IssuePublisher` provider facet.
- Implement the GitHub provider.
- Add the Issue remediation configuration.
- Update validation.

## Out of scope

- Runtime lifecycle.
- Metadata handling.
- Issue creation.
- Assignment workflow.
- Wiring the remediation type into the action factory (handled in the runtime lifecycle PR).

codejasleen:
# Summary

This PR introduces the initial infrastructure for issue based remediation support.

### Changes

* Introduce the `IssuePublisher` provider facet for issue operations.
* Implement the `IssuePublisher` interface for the GitHub provider with support for:

  * `CreateIssue`
  * `GetIssue`
  * `CloseIssue`
  * `ReopenIssue`
  * `AssignIssue`
* Add the `IssueRemediation` configuration message to the remediation protobuf schema.
* Regenerate protobuf, OpenAPI, and documentation artifacts.

This PR intentionally focuses on provider capabilities and configuration. Runtime integration (remediation implementation, lifecycle management, metadata handling, template interpolation, and engine integration) will follow in subsequent PRs.

Fixes: #6532

# Testing

* Verified the project builds successfully using:

  ```bash
  go build ./...
  ```

* Regenerated protobuf artifacts using:

  ```bash
  make buf
  ```

* Verified the generated protobuf, OpenAPI, and documentation files were updated successfully.

* Verified the GitHub provider compiles and satisfies the new `IssuePublisher` interface through compile time interface assertions.

### DIFF
--- a/pkg/providers/v1/providers.go
+++ b/pkg/providers/v1/providers.go
@@ -201,6 +201,43 @@ type ReviewPublisher interface {
 	GetPullRequest(ctx context.Context, owner, repo string, prNumber int) (*github.PullRequest, error)
 }
 
+// IssuePublisher is the interface for providers that can publish issues.
+type IssuePublisher interface {
+	Provider
+
+	// CreateIssue creates an issue in the given repository.
+	CreateIssue(
+		ctx context.Context,
+		owner, repo string,
+		title string,
+		body string,
+		labels []string,
+		assignees []string,
+	) (*github.Issue, error)
+
+	// GetIssue gets an issue from the given repository
+	GetIssue(
+		ctx context.Context,
+		owner, repo string,
+		number int,
+	) (*github.Issue, error)
+
+	// CloseIssue closes an existing issue
+	CloseIssue(
+		ctx context.Context,
+		owner, repo string,
+		number int,
+		comment string,
+	) (*github.Issue, error)
+
+	// ReopenIssue reopens an existing issue
+	ReopenIssue(
+		ctx context.Context,
+		owner, repo string,
+		number int,
+	) (*github.Issue, error)
+}
+
 // GitHub is the interface for interacting with the GitHub REST API
 // Add methods here for interacting with the GitHub Rest API
 type GitHub interface {
@@ -238,6 +275,11 @@ type GitHub interface {
 	CreatePullRequest(ctx context.Context, owner, repo, title, body, head, base string) (*github.PullRequest, error)
 	ClosePullRequest(ctx context.Context, owner, repo string, number int) (*github.PullRequest, error)
 	ListPullRequests(ctx context.Context, owner, repo string, opt *github.PullRequestListOptions) ([]*github.PullRequest, error)
+	CreateIssue(ctx context.Context, owner, repo string, title string, body string, labels []string,
+		assignees []string) (*github.Issue, error)
+	GetIssue(ctx context.Context, owner, repo string, number int) (*github.Issue, error)
+	CloseIssue(ctx context.Context, owner, repo string, number int, comment string) (*github.Issue, error)
+	ReopenIssue(ctx context.Context, owner, repo string, number int) (*github.Issue, error)
 	GetUserId(ctx context.Context) (int64, error)
 	GetName(ctx context.Context) (string, error)
 	GetLogin(ctx context.Context) (string, error)
--- a/proto/minder/v1/minder.proto
+++ b/proto/minder/v1/minder.proto
@@ -1,95 +1,95 @@
 // SPDX-FileCopyrightText: Copyright 2023 The Minder Authors
 // SPDX-License-Identifier: Apache-2.0
 
-
 syntax = "proto3";
 package minder.v1;
+
+import "buf/validate/validate.proto";
 import "google/api/annotations.proto";
 import "google/api/field_behavior.proto";
 import "google/protobuf/descriptor.proto";
-import "google/protobuf/timestamp.proto";
-import "google/protobuf/struct.proto";
 import "google/protobuf/field_mask.proto";
-import "buf/validate/validate.proto";
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
 
 option go_package = "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1";
 
 enum ObjectOwner {
-    OBJECT_OWNER_UNSPECIFIED = 0;
-    OBJECT_OWNER_PROJECT = 2;
-    OBJECT_OWNER_USER = 3;
-    reserved 1; // deprecated OBJECT_OWNER_ORGANIZATION
+  OBJECT_OWNER_UNSPECIFIED = 0;
+  OBJECT_OWNER_PROJECT = 2;
+  OBJECT_OWNER_USER = 3;
+  reserved 1; // deprecated OBJECT_OWNER_ORGANIZATION
 }
 
 enum Relation {
-    RELATION_UNSPECIFIED = 0;
-    RELATION_CREATE = 1 [(name) = "create"];
-    RELATION_GET = 2 [(name) = "get"];
-    RELATION_UPDATE = 3 [(name) = "update"];
-    RELATION_DELETE = 4 [(name) = "delete"];
-    RELATION_ROLE_LIST = 5 [(name) = "role_list"];
-    RELATION_ROLE_ASSIGNMENT_LIST = 6 [(name) = "role_assignment_list"];
-    RELATION_ROLE_ASSIGNMENT_CREATE = 7 [(name) = "role_assignment_create"];
-    RELATION_ROLE_ASSIGNMENT_REMOVE = 8 [(name) = "role_assignment_remove"];
-    RELATION_REPO_GET = 9 [(name) = "repo_get"];
-    RELATION_REPO_CREATE = 10 [(name) = "repo_create"];
-    RELATION_REPO_UPDATE = 11 [(name) = "repo_update"];
-    RELATION_REPO_DELETE = 12 [(name) = "repo_delete"];
-    RELATION_ARTIFACT_GET = 13 [(name) = "artifact_get"];
-    RELATION_ARTIFACT_CREATE = 14 [(name) = "artifact_create"];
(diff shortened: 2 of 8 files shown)


proto/minder/v1/minder.proto
evankanderson: Do you need to change / update this?

(Yes, we should have used `oneof` instead of an enumeration and a set of optional fields....)

proto/minder/v1/minder.proto
```diff
@@ -2929,10 +2929,32 @@ message RuleType {
...
+                string title = 1 [
+                    (buf.validate.field).string = {
+                        min_len: 1,
+                        max_len: 75,
+                    }
+                ];
+                //the body of the issue
+                // This is not validated here as it will be validated by the repository provider, i.e. GitHub upon
+                // creation of the issue
+                string body = 2 [
```
evankanderson: It's probably worth indicating two things:

1. Is any form of substitution (e.g. templating) available in this field?
2. Is this a plain-text field, or is it some other type (such as markdown)?

proto/minder/v1/minder.proto
```diff
@@ -2929,10 +2929,32 @@ message RuleType {
...
+                // the title of the issue
+                // This is not validated here as it will be validated by the repository provider, i.e. GitHub upon
+                // creation of the issue
+                string title = 1 [
+                    (buf.validate.field).string = {
+                        min_len: 1,
+                        max_len: 75,
+                    }
+                ];
+                //the body of the issue
```
evankanderson:
```suggestion
                // the body of the issue
```

pkg/providers/v1/providers.go
```diff
@@ -201,6 +201,47 @@ type ReviewPublisher interface {
 	GetPullRequest(ctx context.Context, owner, repo string, prNumber int) (*github.PullRequest, error)
 }
 
+// IssuePublisher is the interface for providers that can publish issues.
+type IssuePublisher interface {
```
evankanderson, line 205: If these are copies of the GitHub interface methods, you should add them to that interface as well.

At some point, we'll need to switch to more neutral types, but I think it's fine to start with the GitHub forms until we figure out a "neutral" representation.

pkg/providers/v1/providers.go
```diff
@@ -201,6 +201,47 @@ type ReviewPublisher interface {
...
+	Provider
+
+	// CreateIssue creates an issue in the given repository.
+	CreateIssue(
```
evankanderson, line 209: Following the example of `CreatePullRequest`, we should take title and body as arguments, rather than an `IssueRequest` (which opens the door to functionality it will be hard to take back)

pkg/providers/v1/providers.go
```diff
@@ -201,6 +201,47 @@ type ReviewPublisher interface {
...
+		ctx context.Context,
+		owner, repo string,
+		number int,
+	) (*github.Issue, error)
+
+	// CloseIssue closes an existing issue
+	CloseIssue(
+		ctx context.Context,
+		owner, repo string,
+		number int,
```
evankanderson, line 229: We should consider adding a `comment string` argument here.

pkg/providers/v1/providers.go
evankanderson: Looking at the issues service, labels and assignees seem like the two GitHub items that would be most likely to be added.  GitLab has similar fields -- I'm assuming that milestones and epics probably change too much to want them to be part of a security policy, so I don't think we'd need those without a clear use case.

pkg/providers/v1/providers.go
```diff
@@ -201,6 +201,47 @@ type ReviewPublisher interface {
...
+
+	// ReopenIssue reopens an existing issue
+	ReopenIssue(
+		ctx context.Context,
+		owner, repo string,
+		number int,
+	) (*github.Issue, error)
+
+	// AssignIssue assigns users to an issue
+	AssignIssue(
```
evankanderson: Since CreateIssue on both GitHub and GitLab can accept an assignee, I'm not sure that we need a separate "AssignIssue" method (at least until we have a stronger lifecycle that also handles comments and such).

evankanderson (commented):
You may want to also add a comment in the proto that the issue remediation type is not yet implemented.  (We'll take the comment out at the end.)

Since it's not fully implemented, you may also want to add a flag-check in `internal/controlplane/handlers_ruletype.go` to prevent people from creating / updating a ruletype with an "issue" remediation.  (This is a best practice, but I don't want to overwhelm you with too many tasks at once.)

codejasleen:
> You may want to also add a comment in the proto that the issue remediation type is not yet implemented. (We'll take the comment out at the end.)
> 
> Since it's not fully implemented, you may also want to add a flag-check in `internal/controlplane/handlers_ruletype.go` to prevent people from creating / updating a ruletype with an "issue" remediation. (This is a best practice, but I don't want to overwhelm you with too many tasks at once.)

At the moment issue isn't included in the allowed remediation type values, so requests are rejected during proto validation before reaching handlers_ruletype.go. Because of that I didn't add the extra handler check. Once issue remediation is enabled by allowing type: issue, adding a feature gate there would make sense if the implementation is still incomplete.

proto/minder/v1/minder.proto
codejasleen: Since the runtime implementation isn't in place yet, I intentionally left it out. That way, issue remediation can't be selected until the feature is actually supported.

I also added a note to IssueRemediation documenting that it's reserved for future support and that "issue" isn't an allowed remediation type yet.

Does that approach make sense, or would you prefer enabling "issue" now and gating it elsewhere until the implementation is complete?

proto/minder/v1/minder.proto
codejasleen: Thanks! I updated the documentation for these fields to indicate that they support Minder's template interpolation syntax, and that the rendered body is interpreted as Markdown by repository providers that support it (e.g. GitHub).

pkg/providers/v1/providers.go
codejasleen: Done. I added the issue methods to the `GitHub` interface as well, while keeping the `IssuePublisher` interface for the remediation abstraction. That should make the provider API consistent with the existing GitHub interface until we introduce a more provider-neutral representation.

pkg/providers/v1/providers.go
codejasleen: Done. I updated the interface to take `title` and `body` directly instead of a `github.IssueRequest`, following the `CreatePullRequest` pattern. I also added `labels` and `assignees` as explicit arguments, since they seem like the most generally useful fields across providers while keeping the interface focused.

pkg/providers/v1/providers.go
codejasleen: Done. I added the `comment` parameter so the provider API can support that workflow in the future, but left the actual comment creation for the issue remediation implementation. I also added a TODO in the GitHub provider to make that follow up explicit.

pkg/providers/v1/providers.go
codejasleen: Done. I removed the separate `AssignIssue` method and folded assignees into `CreateIssue`, along with labels, since both are part of the create issue request and fit the initial API better.

codejasleen:
Hi @evankanderson, I pushed the lint fixes, but the CI is still failing on `TestNatsChannel`. I reran the test locally (`go test ./internal/events/nats -count=10`) and wasn't able to reproduce the failure, it passed consistently. It looks unrelated to the provider API changes in this PR, but I wanted to mention it in case it's a flaky CI test.

evankanderson:
> Hi @evankanderson, I pushed the lint fixes, but the CI is still failing on `TestNatsChannel`. I reran the test locally (`go test ./internal/events/nats -count=10`) and wasn't able to reproduce the failure, it passed consistently. It looks unrelated to the provider API changes in this PR, but I wanted to mention it in case it's a flaky CI test.

Yes, that test is flaky (dropped messages) _specifically_ on GitHub CI.  I suspect it's due to overloading the poor CPU on the GitHub ephemeral (hosted) runners, but I don't have a lot of visibility on reproducing it.

proto/minder/v1/minder.proto
evankanderson, line 1: What happened to this file?  If you want to run some sort of auto-formatter on it, please do it in a separate PR without other changes.  I can't wade through 3k lines trying to find intended vs unintended changes.

evankanderson (changes requested):
You'll need to undo the formatting on `minder.proto`, but otherwise this looks good.

proto/minder/v1/minder.proto
codejasleen: @evankanderson thanks for pointing this out, and sorry about the unnecessary formatting changes. I didn't realize I'd accidentally reformatted the proto file while working on the changes. I've restored the original formatting so the PR now only contains the intended IssueRemediation additions and the corresponding generated code. Thanks for catching it!

evankanderson: (I do actually human review all the changes going in to Minder, so I will catch things like this.  That also means that I'll ask people to do extra work sometimes to make the code reviews easier on me.)

proto/minder/v1/minder.proto
```diff
@@ -2929,10 +2929,35 @@ message RuleType {
                 optional ActionsReplaceTagsWithSha actions_replace_tags_with_sha = 5;
             }
 
+            // NOTE: Issue remediation is not yet implemented.
+            // This configuration is reserved for future support.
+            // "issue" is intentionally not a supported remediation type yet.
+            message IssueRemediation {
+                // the title of the issue
+                // Supports Minder's template interpolation syntax.
```
evankanderson, line 2937: FWIW, this is Go template syntax, but I think this documentation is fine.

evankanderson (approved):
Thanks, this looks good.  My comments are color / flavor only, merging.

PR #6550, opened 2026-06-24
feat(ruletest): implement HTTP mocking layer for rule testing framework

### LINKED ISSUE #6506: [LFX] Rule Testing Framework: HTTP mocking layer

## Goal
Implement the HTTP mocking mechanisms so that test authors can simulate API responses without hitting real endpoints.

## Scope
- Starlark builtins: `body(payload)` and `code(status)` for building mock responses
- Custom `http.RoundTripper` that:
  - Takes a dictionary of URL glob patterns mapped to mock responses
  - Intercepts outbound HTTP requests during rule evaluation
  - Returns the mocked response instead of making real network calls

## Acceptance Criteria
- [x] `body()` and `code()` produce valid mock response objects in Starlark
- [x] The custom `RoundTripper` correctly matches URL patterns and returns mocked responses
- [x] Unmatched URLs produce a clear error (not a silent failure or real network call)
- [x] Unit tests covering pattern matching, response construction, and edge cases

krrish175-byte:
## Summary

This PR introduces the HTTP mocking layer for the `mindev` Rule Testing Framework, allowing test authors to simulate API responses without hitting real endpoints during rule evaluations. 

Addressing feedback to ensure reusability across the system, the underlying mock transport logic was integrated directly into `pkg/testkit/v1` rather than being isolated to `ruletest`.

**Key Features:**
- **Starlark Built-ins:** Added `body(payload)` and `code(status)` to easily construct HTTP mock responses in test scripts.
- **`TestKit` Integration:** Introduced `MockRoundTripper` into `pkg/testkit/v1` to intercept outbound HTTP requests.
- **Pattern Matching:** `MockRoundTripper` matches intercepted requests against URL glob patterns, returning the configured mock responses.
- **Strict Network Isolation:** Unmatched URLs immediately produce a clear error, ensuring no silent failures or accidental real network calls occur during testing.

Part of the rule testing framework implementation.

Fixes #6506 

## Testing

- Added comprehensive unit tests in `pkg/ruletest` for Starlark response construction (`body()`, `code()`) and edge cases.
- Added robust unit tests in `pkg/testkit/v1` ensuring the `MockRoundTripper` correctly resolves URL glob patterns and blocks unmatched requests.
- Verified all components integrate correctly into the evaluation pipeline and successfully intercept HTTP calls without network leaks.


pkg/testkit/v1/testkit.go
```diff
@@ -38,6 +41,13 @@ func WithGitDir(dir string) Option {
 	}
 }
 
+// WithRoundTripper is a functional option to configure the TestKit to use a specific RoundTripper
+func WithRoundTripper(rt http.RoundTripper) Option {
```
evankanderson: This feels like it might be simpler as a `WithHandlerFunc(http.HandlerFunc)`, which could take something like an `http.ServeMux` for different URLs.  Any reason not to build on the standard library?

evankanderson: (WithHTTP could then implement a HandlerFunc, which might simplify the overall flow)

pkg/ruletest/mock_http.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (m *MockResponse) Hash() (uint32, error) {
+	return 0, errors.New("unhashable type: mock_response")
+}
+
+func (m *MockResponse) Attr(name string) (starlark.Value, error) {
+	switch name {
+	case "code":
+		return starlark.NewBuiltin("code", m.builtinCode), nil
```
evankanderson, line 54: Why not return a value here?

pkg/ruletest/mock_http.go
```diff
@@ -0,0 +1,137 @@
...
+func (m *MockResponse) builtinBody(thread *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	var payload string
+	if err := starlark.UnpackArgs("body", args, kwargs, "payload", &payload); err != nil {
+		return nil, err
+	}
+
+	return &MockResponse{
+		StatusCode: m.StatusCode,
+		Body:       payload,
+	}, nil
```
evankanderson: Why does this not return `m`?

pkg/ruletest/mock_http.go
```diff
@@ -0,0 +1,149 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"go.starlark.net/starlark"
+
+	tkv1 "github.com/mindersec/minder/pkg/testkit/v1"
+)
+
+// MockResponse represents a mocked HTTP response in Starlark.
```
evankanderson, line 14: Maybe add a comment that this is expected to the value for the dict in NewMockRoundTripper?

pkg/ruletest/mock_http.go
```diff
@@ -0,0 +1,149 @@
...
+	if err := starlark.UnpackArgs("body", args, kwargs, "payload", &payload); err != nil {
+		return nil, err
+	}
+	return &MockResponse{
+		StatusCode: 200, // default code
+		Body:       payload,
+	}, nil
+}
+
+// NewMockRoundTripper creates a new tkv1.MockRoundTripper from a Starlark dictionary.
```
evankanderson: Since Starlark is python-shaped, document what the expected types for the dictionary keys and values are.

pkg/ruletest/eval.go
```diff
@@ -92,9 +116,70 @@ func formatEvalResult(evalErr error) *starlark.Dict {
...
 
-	// We ignore the error from SetKey because result is a new Dict we just created
-	// and we know it's not frozen, and the keys are valid strings.
 	_ = result.SetKey(starlark.String("status"), starlark.String(status))
 	_ = result.SetKey(starlark.String("message"), starlark.String(msg))
 	return result
 }
+
+//nolint:gocyclo // this is a simple switch over many entity types
+func mapToProto(entityType string, entityMap map[string]any) (protoreflect.ProtoMessage, error) {
```
evankanderson: We do this wrong in a lot of places, but `proto.Message` is the intended public name, and is an alias of `protoreflect.ProtoMessage`, which says "proto" a lot more times (called "naming stutter" in Go).


```suggestion
func mapToProto(entityType string, entityMap map[string]any) (proto.Message, error) {
```

pkg/ruletest/eval.go
```diff
@@ -92,9 +116,70 @@ func formatEvalResult(evalErr error) *starlark.Dict {
...
+	unmarshaller := protojson.UnmarshalOptions{DiscardUnknown: true}
+	entEnum := minderv1.EntityFromString(entityType)
+
+	switch entEnum {
+	case minderv1.Entity_ENTITY_REPOSITORIES:
+		var repo minderv1.Repository
+		if err := unmarshaller.Unmarshal(b, &repo); err != nil {
+			return nil, err
+		}
+		return &repo, nil
```
evankanderson: You can potentially simplify this code (particularly the cyclomatic part) as:

```suggestion
	var msg *proto.Message
	var err error
	switch entEnum {
	case minderv1.Entity_ENTITY_REPOSITORIES:
		var repo minderv1.Repository
		unmarshaller.Unmarshal(b, &repo)
		msg = &repo
	case ...
	}
	if err != nil {
		return nil, err
	}
	return msg, nil
```

evankanderson (commented):
I sent some comments offline on Slack -- this is looking promising, but I'd love to see an even smaller interface for `testkit` based on `http.HandleFunc` or `http.Handler` (it's easy to convert one to the other, so whichever is better for you -- you could have MockResponse implement `http.Handler`, which might be elegant...

pkg/ruletest/testdata/eval.star
```diff
@@ -4,16 +4,22 @@
...
 
 def test_eval_fail():
     res = eval(
         rule="rule_type_sample.yaml",
-        entity={"required_pull_request_reviews": {"required_approving_review_count": 1}},
+        entity={"owner": "test", "name": "repo"},
         profile={"required_reviews": 2},
+        mock_http={
+            "/repos/test/repo/branches/main/protection": body('{"required_pull_request_reviews": {"required_approving_review_count": 1}}')
+        }
```
evankanderson, line 22: Can we add a test with a response code?

pkg/ruletest/eval.go
```diff
@@ -58,18 +63,41 @@ func builtinEval(
...
+		return nil, fmt.Errorf("invalid mock_http configuration: %w", err)
+	}
+
 	ctx := context.Background()
-	evaluator, err := rtengine.NewRuleEvaluator(ctx, rt, nil)
+
+	opts := []tkv1.Option{}
+	if mockHandler != nil {
+		opts = append(opts, tkv1.WithHandlerFunc(mockHandler.ServeHTTP))
+	}
```
evankanderson: If you make `buildMockHTTPHandler` always return, you don't need this `nil` check.

pkg/ruletest/mock_http.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// buildMockHTTPHandler creates a new http.Handler from a Starlark dictionary.
+// The dictionary is expected to have string keys (representing the HTTP URL pattern)
+// and mock_response values (created via body() or code() built-ins).
+func buildMockHTTPHandler(mockDict *starlark.Dict) (http.Handler, error) {
+	if mockDict == nil {
+		return nil, nil
+	}
+
```
evankanderson, line 121: Why not return `mux` for a nil dict?

evankanderson (dismissed):
Two little refinements, but I think this can go in as-is.

pkg/ruletest/mock_http.go
```diff
@@ -0,0 +1,145 @@
...
+		if !ok {
+			return nil, fmt.Errorf("mock endpoints must have string keys, got %s", key.Type())
+		}
+
+		mockResp, ok := val.(*MockResponse)
+		if !ok {
+			return nil, fmt.Errorf("mock endpoint %q must be mapped to a mock_response, got %s", keyStr, val.Type())
+		}
+
+		mux.HandleFunc(keyStr, func(w http.ResponseWriter, r *http.Request) {
```
evankanderson: Actually, I'm just noticing that lint is complaining about the `r` parameter being unused, and it needs to be renamed to `_`.

pkg/ruletest/testdata/eval.star
```diff
@@ -4,16 +4,34 @@
...
+    assert.eq(res["status"], "fail")
+    assert.true(res["message"] != "")
+
+def test_eval_error_404():
+    res = eval(
+        rule="rule_type_sample.yaml",
+        entity={"owner": "test", "name": "repo"},
+        profile={"required_reviews": 2},
+        mock_http={
+            "/repos/test/repo/branches/main/protection": body('{"message": "Not found"}').code(404)
```
evankanderson, line 33: I think this is fine; you could also make `body` immutable (frozen) after creation, and have it take a `status` keyword argument.

evankanderson (approved):

evankanderson:
Only the NATS test failed, merging.

PR #6551, opened 2026-06-27
feat: implement mindev test cli command and rule loading by name

### LINKED ISSUE #6507: [LFX] Rule Testing Framework: CLI scaffolding + test discovery (mindev test)

## Goal
Wire everything together into the `mindev` CLI so users can run `mindev test` from the command line.

## Scope
- Register the `mindev test` cobra command in `cmd/dev/app/`
- Implement a directory walker that discovers all `*.star` test files
- For each test file, load it via the Starlark environment and scan globals for functions starting with `test_`
- Invoke each test function and collect/report results
- Output pass/fail summary with clear error messages

## Acceptance Criteria
- [x] `mindev test ./path/to/tests/` discovers and runs all `*.star` files
- [x] Only functions prefixed with `test_` are executed
- [x] Clear pass/fail output for each test function
- [x] Non-zero exit code on any test failure
- [x] Works end-to-end with the engine, builtins, and mocking from previous phases

krrish175-byte:
## Summary

This PR implements the CLI Scaffolding for the `mindev` Rule Testing Framework, wiring up the underlying `pkg/ruletest` engine into an accessible CLI command and making the framework fully CI/CD ready.

Addressing deferred tech debt from Phase 2, this PR also introduces "Rule Loading by Name", eliminating the need to hardcode file paths in Starlark scripts by dynamically pre-loading all YAML rule definitions from the target directory into an in-memory registry.

**Key Features**:

- New `mindev test` CLI Command: Added a `test` subcommand to `mindev`. Developers can now execute their Starlark test files across entire directories by running `mindev test [directories...]`.
- Rule Loading by Name: The `eval()` builtin has been updated to seamlessly look up and load rules directly by name from the pre-loaded YAML registry.
- Graceful Fallback: If `eval()` cannot find the specified rule by name in the registry, it safely falls back to resolving it as a file path, ensuring backward compatibility.
- CI/CD Readiness: The CLI aggregates the results of all Starlark tests executed. If any test fails, `mindev test` exits with a non-zero exit code (`1`), making it perfectly suited for automated CI environments.
- Part of the rule testing framework implementation.

Fixes #6507

## Testing

- Built the developer tooling locally (`make build`) and verified the new `test` cobra command registers properly.
- Ran the newly created command against the sample test data (`./mindev test ./pkg/ruletest/testdata`).
- Verified that the CLI correctly parses and aggregates all test execution results.
- Ensured that intentionally failing tests properly trigger the `1` exit code and bubble up the `one or more tests failed` error for CI environments.


cmd/dev/app/test/test.go
```diff
@@ -0,0 +1,62 @@
+// SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package test
```
evankanderson, line 5: Lint wants a comment on the package declaration.

cmd/dev/app/test/test.go
```diff
@@ -0,0 +1,62 @@
...
+		Long:  `Run Starlark-based tests for Minder rules. If no directories are provided, tests the current directory.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) == 0 {
+				args = []string{"."}
+			}
+
+			runner := ruletest.NewRunner()
+			hasFailures := false
+
+			for _, dir := range args {
```
evankanderson: Can we accept files or directories?

Also, this feels like it might be a reasonable function in runner.go (where you could test it more easily).

cmd/dev/app/test/test.go
```diff
@@ -0,0 +1,62 @@
...
+				fmt.Printf("Running tests in %s...\n", dir)
```
evankanderson: use `cmd.Printf()` so that we can override the `cmd.Output` and get this output from the command.

cmd/dev/app/test/test.go
```diff
@@ -0,0 +1,62 @@
...
+				results, err := runner.RunDir(dir)
+				if err != nil {
+					return fmt.Errorf("error running tests in %s: %w", dir, err)
+				}
```
evankanderson: Not sure -- do we want to exit-early here, or collect errors from remaining directories?

pkg/ruletest/runner.go
```diff
@@ -90,10 +94,13 @@ func NewRunner() *Runner {
...
 }
 
-// RunFile executes a single Starlark test file and returns the results
-// for each test_* function found in it.
-// src may be nil, or a string, []byte, or io.Reader containing the file source.
+// RunFile executes a single Starlark test file. If src is non-nil, it is
+// used as the file contents.
 func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
+	return r.runFileWithRules(filename, src, nil)
+}
```
evankanderson: Why not just add the `ruleTypes` argument to `RunFile`?

pkg/ruletest/runner.go
```diff
@@ -102,11 +109,11 @@ func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
 	fileSystem := os.DirFS(baseDir)
 
 	name := filepath.Base(filename)
-	tr := r.newTestCaseRunner(name, fileSystem)
+	tr := r.newTestCaseRunner(name, fileSystem, ruleTypes)
 
 	globals, err := tr.runFile(filename, src)
 	if err != nil {
-		if evalErr, ok := errors.AsType[*starlark.EvalError](err); ok {
+		if evalErr, ok := err.(*starlark.EvalError); ok {
```
evankanderson: Why change `errors.AsType` here?  AsType respects error wrapping, unlike a direct cast.

pkg/ruletest/eval.go
```diff
@@ -21,36 +21,47 @@ import (
 	tkv1 "github.com/mindersec/minder/pkg/testkit/v1"
 )
 
-func builtinEval(
+func (tr *testCaseRunner) builtinEval(
 	thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple,
 ) (starlark.Value, error) {
-	var rulePath string
+	var ruleNameOrPath string
```
evankanderson: Not sure I like "name or path" -- can we choose one?

evankanderson: Context - multi-typed arguments tend to produce more complexity (combinations) inside the function body, which makes it more difficult to test those combinations completely.

pkg/ruletest/eval_test.go
```diff
@@ -100,7 +100,8 @@ func TestBuiltinEval_InvalidArgs(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			_, err := builtinEval(thread, starlark.NewBuiltin("eval", builtinEval), tt.args, tt.kwargs)
+			tr := &testCaseRunner{}
```
evankanderson, line 126: I wonder if you want testCaseRunner to have a "load directory" function for loading all the rules files in the directory.

pkg/ruletest/runner.go
```diff
@@ -174,11 +181,63 @@ func DiscoverFiles(root string) ([]string, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("globbing yaml files: %w", err)
+	}
+	for _, yf := range yamlFiles {
+		decoder, closer := fileconvert.DecoderForFile(yf)
+		if decoder == nil {
+			return nil, fmt.Errorf("error opening file: %s", yf)
+		}
+		rt, err := fileconvert.ReadResourceTyped[*minderv1.RuleType](decoder)
+		closer.Close()
```
evankanderson: I think it's okay to not handle errors from Close in this case, but the linter does not, so you'll need to do something to make it okay with that.

evankanderson: Also, I'd prefer the `defer Close()` pattern, just in case extra logic or returns get added in later.

evankanderson: (Yes, we could be deferring a bunch of filehandle closes, but those should be fairly cheap)

pkg/ruletest/runner.go
```diff
@@ -129,15 +136,15 @@ func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
...
+		result := r.runOneTest(name, fn, fileSystem, ruleTypes)
 		results = append(results, result)
 	}
 
 	return results, nil
 }
 
-func (r *Runner) runOneTest(name string, fn *starlark.Function, fileSystem fs.FS) TestResult {
-	tr := r.newTestCaseRunner(name, fileSystem)
+func (r *Runner) runOneTest(name string, fn *starlark.Function, fileSystem fs.FS, ruleTypes map[string]*minderv1.RuleType) TestResult {
```
evankanderson: The linter wants you to fold the line at 130 characters.

pkg/ruletest/runner.go
```diff
@@ -174,11 +181,63 @@ func DiscoverFiles(root string) ([]string, error) {
...
+	}
+
+	files, err := DiscoverFiles(dir)
+	if err != nil {
+		return nil, fmt.Errorf("discovering test files: %w", err)
+	}
+
+	var allResults []TestResult
+	for _, file := range files {
+		results, err := r.runFileWithRules(file, nil, ruleTypes)
```
evankanderson: Rather than carrying around `ruleTypes`, I'd look to see if there's an object it might naturally live with that could encompass "test files in a directory".

pkg/ruletest/runner.go
```diff
@@ -174,11 +181,63 @@ func DiscoverFiles(root string) ([]string, error) {
...
+// directory, reporting results through t. It also loads *.yaml rules.
+func (r *Runner) TestDir(t *testing.T, dir string) {
 	t.Helper()
 
+	ruleTypes, err := loadRulesFromDir(dir)
+	if err != nil {
+		t.Fatalf("loading rules: %v", err)
+	}
+
 	files, err := DiscoverFiles(dir)
```
evankanderson: This looks like `RunDir`.  Why different?

evankanderson (commented):
You've added a bunch of code in `pkg/ruletest` -- can we restructure the existing rule execution tests to cover that code as well?

krrish175-byte:
**Key Updates:**
- **Added `mindev test` CLI command**: Implemented a new `test` subcommand in `mindev` (via `cmd/dev/app/test`) that accepts multiple file or directory paths.
- **Enhanced Test Runner (`pkg/ruletest/runner.go`)**: 
  - Added the `RunPaths` function to allow executing tests across a list of directories and files seamlessly.
  - Implemented automatic rule loading (`loadRulesFromDir`). The test runner now dynamically loads `rule` definitions alongside the tests, resolving panics that occurred when tests referenced rule names that were missing from the context.
- **CLI Output Formatting**: `mindev test` now aggregates test results and outputs a clean, structured list of `PASS`/`FAIL` results with nested failure messages for easy debugging.
- **Test Coverage**: Updated `runner_test.go` to maintain full test coverage and explicitly test execution behavior without panicking on uninitialized framework pointers.

*(Note: The filesystem mocking features are staged locally and will be pushed as a separate commit once this foundational work is settled.)*

pkg/ruletest/runner.go
```diff
@@ -7,13 +7,16 @@ package ruletest
...
 	"path/filepath"
 	"sort"
 	"strings"
 	"testing"
 
+	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+	"github.com/mindersec/minder/pkg/fileconvert"
 	"go.starlark.net/starlark"
 	"go.starlark.net/starlarktest"
 	"go.starlark.net/syntax"
```
evankanderson: GCI wants the minder imports _after_ the other third-party imports, in a separate section.


```suggestion
	"go.starlark.net/starlark"
	"go.starlark.net/starlarktest"
	"go.starlark.net/syntax"

	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
	"github.com/mindersec/minder/pkg/fileconvert"

```

pkg/ruletest/runner.go
evankanderson: Why the change from `errors.AsType`?

pkg/ruletest/runner.go
```diff
@@ -174,11 +183,100 @@ func DiscoverFiles(root string) ([]string, error) {
...
+		}
+	}
+	return ruleTypes, nil
+}
+
 // RunDir discovers and executes all *.star test files under the given
-// directory, reporting results through t.
-func (r *Runner) RunDir(t *testing.T, dir string) {
+// directory. It also discovers and loads any *.yaml rule files in the directory.
+func (r *Runner) RunDir(dir string) ([]TestResult, error) {
```
evankanderson: Not sure if you want to make this function private or roll it into `RunPaths`

pkg/ruletest/runner.go
evankanderson: Actually, why not simply call `RunPaths`, and then assert on the errors and results?

pkg/ruletest/runner_test.go
```diff
@@ -125,3 +126,17 @@ func TestRunEvalFile(t *testing.T) {
...
+	dir := t.TempDir()
+	content := []byte(`
+def test_pass():
+    assert.eq(1, 1)
+`)
+	if err := os.WriteFile(filepath.Join(dir, "pass.star"), content, 0644); err != nil {
+		t.Fatal(err)
+	}
+	r.TestDir(t, dir)
+}
```
evankanderson, line 48: What would it take to change all the above tests that use `RunFile` to have a single invocation of `TestDir` (or `RunPaths`) and then have specific tests which check for specific results?

I suspect that `results` might want to become something more sophisticated than a string (e.g. a file name + test name + output), but I think you might find that gives you benefits when it comes to creating e.g. XML output.

evankanderson (commented):
A few comments -- I feel like your structure is close to optimal, but you can probably remove some near-duplicate code and get better test coverage with a few more tweaks.

pkg/ruletest/eval.go
```diff
@@ -121,6 +118,51 @@ func formatEvalResult(evalErr error) *starlark.Dict {
...
+
+func (tr *testCaseRunner) loadRuleTypeFallback(ruleName string, thread *starlark.Thread) (*minderv1.RuleType, error) {
+	if tr.ruleTypes != nil {
+		if ruleType, ok := tr.ruleTypes[ruleName]; ok {
+			return ruleType, nil
+		}
+	}
+
+	rulePath := ruleName
+	if !filepath.IsAbs(rulePath) {
```
evankanderson: Why this check?  What would happen if the rulename started with `/` on Unix?

pkg/ruletest/eval.go
```diff
@@ -121,6 +118,51 @@ func formatEvalResult(evalErr error) *starlark.Dict {
...
+		return nil, fmt.Errorf("rule %q not found in loaded rule types and no file found at path %s", ruleName, rulePath)
+	}
+	defer closer.Close()
+
+	rt, err := fileconvert.ReadResourceTyped[*minderv1.RuleType](decoder)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse rule type: %w", err)
+	}
+
+	return rt, nil
```
evankanderson: I'm worried about this code being hit unexpectedly, and things "working" when the primary path is broken.

Alternatively, I'm worried that we have no coverage for this path, and that it's code that seems like it does something but doesn't.

Is there a good reason to keep this code vs removing it and updating the test cases?

pkg/ruletest/eval_test.go
```diff
@@ -95,12 +95,36 @@ func TestBuiltinEval_InvalidArgs(t *testing.T) {
...
+					starlark.String("mock_fs"),
+					func() *starlark.Dict {
+						d := starlark.NewDict(1)
+						_ = d.SetKey(starlark.MakeInt(1), starlark.String("content"))
+						return d
+					}(),
+				},
+			},
+			wantErr: "mock_fs keys and values must be strings",
+		},
```
evankanderson, line 120: It's good to have unit tests for edge cases that might be hard to hit otherwise, but I'd also like to see the mock FS actually being used in one of the integration test cases (which might require adding another ruletype to the test directory).

pkg/ruletest/runner.go
```diff
@@ -90,10 +96,9 @@ func NewRunner() *Runner {
 	}
 }
 
-// RunFile executes a single Starlark test file and returns the results
-// for each test_* function found in it.
-// src may be nil, or a string, []byte, or io.Reader containing the file source.
-func (r *Runner) RunFile(filename string, src any) ([]TestResult, error) {
+// RunFile executes a single Starlark test file. If src is non-nil, it is
+// used as the file contents.
```
evankanderson: The old comment was better.  (It explained the testing action more clearly, and better explained the values that `src` can take.)

pkg/ruletest/runner.go
```diff
@@ -174,39 +184,112 @@ func DiscoverFiles(root string) ([]string, error) {
...
 
-// RunDir discovers and executes all *.star test files under the given
-// directory, reporting results through t.
-func (r *Runner) RunDir(t *testing.T, dir string) {
-	t.Helper()
+// loadRulesFromDir finds and parses all *.yaml files in the given directory
+// into a map of RuleTypes keyed by rule name.
+func loadRulesFromDir(dir string) (map[string]*minderv1.RuleType, error) {
+	ruleTypes := make(map[string]*minderv1.RuleType)
+	yamlFiles, err := filepath.Glob(filepath.Join(dir, "*.yaml"))
```
evankanderson, line 181: #6558 will mean you'll need to expand this a little bit.  Feel free to adjust e.g. `ResourcesFromPaths` to have a non-recursive version, or a recursion boolean (actually, it looks like it is `ExpandFileArgs` that would need adjustment).

pkg/ruletest/runner.go
```diff
@@ -174,39 +184,112 @@ func DiscoverFiles(root string) ([]string, error) {
...
+		}
+	}
+	return ruleTypes, nil
+}
 
-	files, err := DiscoverFiles(dir)
+// runDir discovers and executes all *.star test files under the given
+// directory. It also discovers and loads any *.yaml rule files in the directory.
+func (r *Runner) runDir(dir string) ([]TestResult, error) {
+	ruleTypes, err := loadRulesFromDir(dir)
```
evankanderson: Do you want to error if `len(ruleTypes) == 0` (or `len(files) == 0`)?

evankanderson: I suppose you could also push that back to the caller to check that `len([]TestResult) > 0`

pkg/ruletest/runner.go
```diff
@@ -174,39 +184,112 @@ func DiscoverFiles(root string) ([]string, error) {
...
 
-	if len(files) == 0 {
-		t.Logf("no *.star test files found in %s", dir)
-		return
+	files, err := DiscoverFiles(dir)
+	if err != nil {
+		return nil, fmt.Errorf("discovering test files: %w", err)
 	}
 
+	var allResults []TestResult
```
evankanderson: I feel like you might want to gather TestResults under the `*.star` filename that they occurred in, either by adding a `filename` field to TestResult, or by making this a `map[string][]TestResult` (or doing something else equivalent).

pkg/ruletest/runner.go
```diff
@@ -174,39 +184,112 @@ func DiscoverFiles(root string) ([]string, error) {
...
+				errs = append(errs, fmt.Errorf("error running directory %s: %w", p, err))
 			}
+			allResults = append(allResults, res...)
+		} else {
+			ruleTypes, err := loadRulesFromDir(filepath.Dir(p))
+			if err != nil {
+				errs = append(errs, fmt.Errorf("loading rules for file %s: %w", p, err))
+				continue
+			}
+			res, err := r.RunFile(p, nil, ruleTypes)
```
evankanderson: If `runDir` took a glob pattern argument, could you re-use it both above (`"*.star"`) and here (`p`)?

pkg/ruletest/runner.go
```diff
@@ -174,39 +184,112 @@ func DiscoverFiles(root string) ([]string, error) {
...
+		}
+	}
+	return allResults, errors.Join(errs...)
+}
+
+// TestDir discovers and executes all *.star test files under the given
+// directory, reporting results through t. It also loads *.yaml rules.
+func (r *Runner) TestDir(t *testing.T, dir string) {
+	t.Helper()
+	t.Parallel()
```
evankanderson: `t.Parallel()` here causes a panic for calling `Parallel()` twice.

pkg/ruletest/runner.go
```diff
@@ -174,39 +184,112 @@ func DiscoverFiles(root string) ([]string, error) {
...
-					for _, msg := range result.Failures {
-						t.Error(msg)
-					}
-				})
+	for _, result := range results {
+		result := result
+		t.Run(result.Name, func(t *testing.T) {
+			t.Parallel()
+			for _, msg := range result.Failures {
+				t.Error(msg)
```
evankanderson: Did we lose the check on `strings.HasPrefix(result.name, "test_fail_")`?

pkg/ruletest/runner.go
evankanderson: Actually, `len(ruleTypes) == 0` could be valid for testing just the starlark engine, so maybe this is a bad idea.

pkg/ruletest/runner_test.go
evankanderson: Still hoping that this becomes:

```golang
r := NewRunner()
r.TestDir(t, "testdata")
```

And that we can get good coverage from just that and the file contents in testdata.

evankanderson: (We might even be able to delete some of the other tests!)

pkg/testkit/v1/testkit.go
```diff
@@ -35,6 +38,13 @@ func WithGitDir(dir string) Option {
 	}
 }
 
+// WithMockFS is a functional option to configure the TestKit with a mocked filesystem
+func WithMockFS(fs map[string]string) Option {
+	return func(tk *TestKit) {
+		tk.mockFS = fs
+	}
+}
```
evankanderson, line 59: Can you adjust testkit as follows (rather than having separate paths for the `WithGitDir` and this?  Also, I think "WithGitFiles" might be a better name here, to parallel the "Dir" option -- I don't feel strongly, but both are providing a mock FS?

The way I'd do this would be:

1. rename `tk.gitDir` to `tk.gitFS`, and change the type to `billy.Filesystem`
2. WithGitDir would more-eagerly create the `osfs.New(dir)`
3. This method would create a memfs, populate it, and assign it to `tk.gitFS`.  Note that we only need to return a `billy.Filesystem` here, we don't need a full Git storage by using the Ingester interface.

cmd/dev/app/test/test.go
```diff
@@ -0,0 +1,60 @@
...
+			if len(results) == 0 {
+				cmd.Printf("No tests found\n")
+				return err
+			}
+
+			hasFailures := false
+			for _, res := range results {
+				if len(res.Failures) > 0 {
+					hasFailures = true
+					cmd.Printf("FAIL: %s\n", res.Name)
```
evankanderson: I think you'll want the filename here as well as the test name, particularly if the same test name could occur in multiple `*.star` files.

cmd/dev/app/test/test.go
```diff
@@ -0,0 +1,60 @@
...
+
+			runner := ruletest.NewRunner()
+			results, err := runner.RunPaths(args)
+			if err != nil {
+				cmd.PrintErrf("Error(s) running tests:\n%v\n", err)
+			}
+
+			if len(results) == 0 {
+				cmd.Printf("No tests found\n")
+				return err
```
evankanderson: `err` is nil here, so you could just return `nil`.

cmd/dev/app/test/test.go
```diff
@@ -0,0 +1,60 @@
...
+				} else {
+					cmd.Printf("PASS: %s\n", res.Name)
+				}
+			}
+
+			if hasFailures {
+				return fmt.Errorf("one or more tests failed")
+			}
+
+			return err
```
evankanderson: `err` is also non-nil here (was last assigned on 28, then checked on 29)

evankanderson (commented):
Sorry for the delay; I got sniped by #6558 and several life things (like getting a bike repaired).

This is looking better, but I'm going to push for even more integration testing in the `pkg/ruletest` to make sure we don't break the CLI by accident.

pkg/ruletest/eval.go
```diff
@@ -121,12 +116,33 @@ func formatEvalResult(evalErr error) *starlark.Dict {
...
+		}
+	}
+	return mockFSMap, nil
+}
+
+func (tr *testCaseRunner) lookupRuleType(ruleName string) (*minderv1.RuleType, error) {
+	if tr.ruleTypes != nil {
+		if ruleType, ok := tr.ruleTypes[ruleName]; ok {
+			return ruleType, nil
+		}
```
evankanderson: Given that you're storing `*minderv1.RuleType`, you don't need the `,ok =` formulation, since a `nil` pointer could just be the error case below:


```suggestion
		if ruleType := tr.ruleTypes[ruleName]; ruleType != nil {
			return ruleType, nil
		}
```

pkg/ruletest/eval_test.go
```diff
@@ -95,12 +95,36 @@ func TestBuiltinEval_InvalidArgs(t *testing.T) {
 			kwargs:  []starlark.Tuple{{starlark.String("entity"), starlark.String("not a dict")}},
 			wantErr: "got string, want dict",
 		},
+		{
+			name: "invalid mock_fs type",
+			args: starlark.Tuple{starlark.String("rule.yaml")},
```
evankanderson: (I know we don't really use this here, but this should be `"fs_check"`, right?)

pkg/testkit/v1/testkit_provider.go
```diff
@@ -83,9 +86,46 @@ func (*TestKit) PropertiesToProtoMessage(_ minderv1.Entity, _ *properties.Proper
...
+			Name:  "TestKit",
+			Email: "testkit@minder.test",
+			When:  time.Now(),
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return repo, nil
```
evankanderson: Do we still need this?  I don't think we do.

evankanderson (commented):
There are a few lint errors hanging around, but this is looking close.

I'd still like to see `runner_test.go` call `r.TestDir(t, "testdata")` if that works -- it feels like it would make adding new test cases really simple.

pkg/ruletest/runner.go
```diff
@@ -154,59 +173,121 @@ func (r *Runner) runOneTest(name string, fn *starlark.Function, fileSystem fs.FS
...
+	for _, yf := range yamlFiles {
+		rt, err := loadSingleRule(yf)
 		if err != nil {
-			return err
+			continue // skip files that aren't valid rule types
 		}
-		if !d.IsDir() && strings.HasSuffix(d.Name(), ".star") {
-			files = append(files, path)
+		if rt != nil && rt.Name != "" {
+			ruleTypes[rt.Name] = rt
```
evankanderson, line 194: (for a subsequent PR)

We should return an error if `ruletypes[rt.Name]` already exists.

pkg/ruletest/runner.go
```diff
@@ -154,59 +173,121 @@ func (r *Runner) runOneTest(name string, fn *starlark.Function, fileSystem fs.FS
...
+
+	var allResults []TestResult
+	var errs []error
+
+	// Ensure deterministic execution order by sorting directories
+	var dirs []string
+	for dir := range filesByDir {
+		dirs = append(dirs, dir)
 	}
+	sort.Strings(dirs)
```
evankanderson: This looks like a use-case for [the Go generic map functions introduced a year or so ago](https://pkg.go.dev/maps#Keys):


```suggestion
	dirs = slices.Sorted(maps.Keys(filesByDir))
```

pkg/ruletest/runner.go
```diff
@@ -154,59 +173,121 @@ func (r *Runner) runOneTest(name string, fn *starlark.Function, fileSystem fs.FS
...
+			allResults = append(allResults, res...)
+		}
+	}
+	return allResults, errors.Join(errs...)
+}
+
+// TestDir discovers and executes all *.star test files under the given
+// directory, reporting results through t. It also loads *.yaml rules.
+//nolint:tparallel // TestDir acts as a test runner helper, caller is responsible for t.Parallel
+func (r *Runner) TestDir(t *testing.T, dir string) {
```
evankanderson: The linter doesn't like the `//nolint:` directive right after the other docs.  You might just move this into `runner_test`, not export it, or export with a name that doesn't start with "Test*"

pkg/ruletest/runner_test.go
evankanderson, line 1: This diff is beautiful, IMO.

pkg/ruletest/runner.go
```diff
@@ -154,59 +173,121 @@ func (r *Runner) runOneTest(name string, fn *starlark.Function, fileSystem fs.FS
...
+	if len(results) == 0 {
+		t.Logf("no *.star test files found in %s", dir)
+		return
+	}
+
+	for _, result := range results {
+		result := result
+		name := result.Filename + "/" + result.Name
+		t.Run(name, func(t *testing.T) {
+			if strings.HasPrefix(result.Name, "test_fail_") {
```
evankanderson: GCI wants _this anonymous function_ to call `t.Parallel()` after calling `t.Run()` to produce a subtest.

evankanderson (commented):
Looks good except for the lint errors.  I'm particularly pleased at how `runner_test.go` was able to be simplified but still have good coverage.

evankanderson (approved):

PR #6554, opened 2026-06-28
feat: warn when accepting Rego V0 policies

### LINKED ISSUE #6555: Add feature-flagged Rego V0 deprecation warnings
Part of #5262

## Problem

Minder currently accepts both Rego V0 and V1 policies, but users receive no indication when a rule type still depends on the deprecated V0 syntax.

Before V0 policies can be rejected, Minder needs a warn-but-accept stage that gives users time to migrate without breaking existing workflows.

## Proposed changes

- Add non-fatal warnings to `CreateRuleType` and `UpdateRuleType` responses.
- Add a `rego_v1_warn_v0` feature flag.
- When enabled, return an actionable warning for accepted V0 policies.
- Display server warnings in:
  - `minder ruletype create`
  - `minder ruletype apply`
- Preserve existing behavior when the flag is disabled.
- Add control-plane and CLI test coverage.

Suggested warning:

> This rule type uses Rego V0 syntax. Please migrate to V1 using `opa fmt --v0-v1`. V0 support will be removed in a future release.

## Acceptance criteria

- V0 policies remain accepted.
- V0 policies produce a client-visible warning when the flag is enabled.
- V1 policies do not produce the warning.
- No warnings are returned when the flag is disabled.
- Create and update paths are covered by tests.
- Generated API and reference documentation are updated.

dakshhhhh16:
Part of #5262
Resolves #6555 
## Summary

Introduces the warn-but-accept stage of the Rego v1 migration. Rego V0 rule types continue to work, but clients can now receive a clear migration warning when the new feature flag is enabled.

## What changed

- Added non-fatal `warnings` fields to CreateRuleType and UpdateRuleType responses.
- Added the `rego_v1_warn_v0` feature flag.
- Return an actionable warning when an accepted rule type uses Rego V0.
- Display server warnings in `minder ruletype create` and `minder ruletype apply`.
- Added control-plane and CLI coverage for V0 warnings and V1 policies without warnings.
- Regenerated protobuf, OpenAPI, and reference documentation.

The warning recommends migrating with:

```sh
opa fmt --v0-v1
```

No policies are rejected by this change.

## Testing

```sh
go test ./cmd/cli/app/ruletype ./internal/controlplane ./pkg/api/protobuf/go/minder/v1 ./pkg/flags
go run github.com/bufbuild/buf/cmd/buf@v1.70.0 lint
```

### DIFF
--- a/internal/controlplane/handlers_ruletype.go
+++ b/internal/controlplane/handlers_ruletype.go
@@ -32,7 +32,9 @@ import (
 	"github.com/mindersec/minder/pkg/ruletypes"
 )
 
-const regoV0DeprecationWarning = "This rule type uses Rego V0 syntax. Please migrate to V1 using `opa fmt --v0-v1`. V0 support will be removed in a future release."
+const regoV0DeprecationWarning = "This rule type uses Rego V0 syntax. " +
+	"Please migrate to V1 using `opa fmt --v0-v1`. " +
+	"V0 support will be removed in a future release."
 
 var (
 	maxReadableStringSize = 3 * 1 << 10 // 3kB
--- a/internal/controlplane/handlers_ruletype.go
+++ b/internal/controlplane/handlers_ruletype.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/microcosm-cc/bluemonday"
+	"github.com/open-policy-agent/opa/v1/ast"
 	"github.com/yuin/goldmark"
 	"github.com/yuin/goldmark/extension"
 	"github.com/yuin/goldmark/parser"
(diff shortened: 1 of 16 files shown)


internal/controlplane/handlers_ruletype.go
```diff
@@ -230,9 +238,31 @@ func (s *Server) UpdateRuleType(
...
+		return nil
+	}
+
+	eval := ruleType.GetDef().GetEval()
+	if eval.GetType() != regoeval.RegoEvalType || eval.GetRego().GetDef() == "" {
+		return nil
+	}
+
+	version := ast.RegoV0
+	if flags.Bool(ctx, s.featureFlags, flags.RegoV1DualParse) {
```
evankanderson, line 256: Do you need this flag check?  If the administrator somehow set `rego_v1_warn_v0` and didn't set `rego_v1_dual_parse`, then they'll get warnings on every single rule storage, which seems okay.

evankanderson (dismissed):
Very nice change!  I'm happy to merge this as-is, or you can address the one comment I had.

internal/controlplane/handlers_ruletype.go
dakshhhhh16: Good call. The warning path now only checks rego_v1_warn_v0 and independently detects V0-only Rego, so it works even if rego_v1_dual_parse is disabled. I added a focused test for that case.

dakshhhhh16:
@evankanderson 
It's good to go now ig

evankanderson:
https://github.com/mindersec/minder/pull/6556 for the lint error, which seems unrelated.

evankanderson (approved):

PR #6557, opened 2026-06-29
refactor: encapsulate project creation logic into the ProjectCreator interface

### LINKED ISSUE #5002: Refactor Project Creation into internal/projects Module
### Describe the issue

Currently, project creation logic is split between `internal/controlplane` and `internal/projects`. Specifically, in `internal/controlplane/handlers_projects.go`, we have code handling project creation directly, which leads to code duplication and a less modular structure.

To improve code maintainability and modularity, we should move all project creation logic into the internal/projects module. This change will:

- Centralise project creation processes.
- Reduce code duplication.
- Simplify the `internal/controlplane` module by offloading project creation responsibilities.

-------------------------------
Action Items

- Move the project creation code from `internal/controlplane/handlers_projects.go` to the `internal/projects` module.
- Ensure that all project creation (including sub-projects) is handled exclusively within internal/projects.
- Update any references and imports affected by this change.
- Adjust unit tests and documentation to reflect the new structure.

CC: @ethomson @evankanderson 

### To Reproduce

_No response_

_No response_

ShantKhatri:
# Summary

The PR centralizes project creation logic into `internal/projects`, removing duplication between the control plane and the projects module.

Before, CreateProject in `handlers_projects.go` directly managed the DB transaction, authz check, and feature flag validation before delegating to `ProvisionChildProject / ProvisionSelfEnrolledProject`.

Now, a new `ProvisionProject` method on the ProjectCreator interface owns the full creation orchestration. The handler is now a thin gRPC adapter.

Changes as, `internal/projects/creator.go` is the core change. `ProvisionProject` is added to the `ProjectCreator` interface with `db.Store` and `flags.Interface` injected as dependencies, so the module now owns its own transaction boundaries, authz checks, and feature flag gates. As a result, `CreateProject` in `internal/controlplane/handlers_projects.go` is reduced to a single delegate call with no business logic. `internal/service/service.go` is updated to pass store and featureFlagClient to NewProjectCreator to satisfy the new dependencies. Test call sites in `creator_test.go` and `handlers_user_test.go` are updated accordingly to supply the two new constructor arguments.

The flaggable intentional design made here is to `makeProjectFactory` in `service.go` still calls `ProvisionSelfEnrolledProject` directly with an existing `db.ExtendQuerier`. GitHub App webhook flows arrive with an in-flight transaction that must not be re-wrapped; routing those through ProvisionProject would start a nested transaction.

Fixes #5002

# Testing
1. Install required tools
`make bootstrap`

2. Generate example rules/profiles used by some tests
`make init-examples`
image

3. Lint
`make lint`
image

4. Run tests for affected packages
`go test ./internal/projects/... ./internal/controlplane/... ./internal/service/...`
image

5. Full suite
`go test ./...`
image


internal/projects/creator.go
```diff
@@ -204,3 +220,75 @@ func (p *projectCreator) ProvisionChildProject(
...
+func (p *projectCreator) ProvisionProject(
+	ctx context.Context,
+	projectName string,
+	parentProjectID uuid.UUID,
+) (*db.Project, error) {
+	var project *db.Project
+
+	if parentProjectID != uuid.Nil {
+		// Verify permissions if we have a parent
+		if err := p.authzClient.Check(ctx, "create", parentProjectID); err != nil {
```
intelligent-ears: The handler version of this call used `relationAsName(minderv1.Relation_RELATION_CREATE)` rather than a literal string. Can you confirm these are guaranteed equivalent, and either:

- import/reuse the same derivation so this can't drift if the relation enum changes, or
- if there's a layering reason internal/projects shouldn't depend on minderv1 directly, add a short comment pinning "create" to the enum it's meant to match, so a future rename doesn't silently break authz here.

This is permission-checking code moving one layer down the stack, so I'd rather we're explicit about why it's safe than assume it.

intelligent-ears (commented):
Nice work on this, @ShantKhatri  — this is a clean shape for the refactor and matches what the issue asked for. A couple of things I'd like addressed before merging, the authz one being the one I'd consider blocking.

internal/projects/creator.go
ShantKhatri: Thanks @intelligent-ears! The concern was valid. I have added the `RelationCreate`  constant with pinning it `RELATION_CREATE = 1 [(name) = "create"]` in the proto. 
Now the project creator uses `authz.RelationCreate`.

So after the updated changes, internal/projects does not import minderv1 directly, as the constant lives in internal/authz, which is already a dependency of that package.

internal/authz/interface.go
```diff
@@ -32,6 +32,11 @@ const (
 	RolePermissionsManager Role = "permissions_manager"
 )
 
+const (
+	// RelationCreate maps to Relation_RELATION_CREATE [(name) = "create"].
+	RelationCreate = "create"
+)
```
evankanderson: Can we hoist `relationAsName` to somewhere that it can be universally depended on?

I'd suggest `pkg/api/protobuf/go/minder/v1/api.go`, as we already have a bunch of hand-written code there.

internal/projects/creator.go
```diff
@@ -28,6 +31,13 @@ import (
 // 1. Move the delete operations into this interface
 // 2. The interface is very GitHub-specific. It needs to be made more generic.
 type ProjectCreator interface {
+	// ProvisionProject orchestrates the creation of a project, checking permissions,
+	// managing the database transaction, and calling the appropriate provisioning logic.
+	ProvisionProject(
+		ctx context.Context,
+		projectName string,
+		parentProjectID uuid.UUID,
+	) (*db.Project, error)
```
evankanderson: What do you think about using the proto (external) version of project as the return value, rather than the database row layout?

evankanderson: (I'd prefer for the representation details to be part of the service, rather than leaking database row abstractions up farther.)

internal/projects/creator.go
evankanderson: Do these two methods still need to be public and/or part of the interface?

intelligent-ears (commented):
The existing tests were updated for the new constructor signature (`mockStore`, `&flags.FakeClient{}`), but I don't see new tests targeting `ProvisionProject` itself. Since this method now owns permission checks, the feature-flag gate, and transaction boundaries — logic that previously lived in the handler — could we get coverage for:

- child-project creation when `authzClient.Check` denies
- child-project creation when `ProjectAllowsProjectHierarchyOperations` is false
- top-level creation when `flags.ProjectCreateDelete` is disabled
- top-level creation with no identity in context
- the `project == nil` post-creation guard

Even a few table-driven cases covering these branches would close the gap left by the handler logic moving here.

evankanderson (commented):
I realize that we didn't have existing coverage for this code, but it might be good to add test coverage while you're here.  Other than that, a couple comments on the service interface between the handlers in controlplane and the rest of the system -- this is somewhere that we were pretty sloppy at first, but we've been trying to clean things up over time.

internal/projects/creator.go
ShantKhatri: I appreciate the question, `ProvisionSelfEnrolledProject` is still needed outside the struct, `makeProjectFactory` in `service.go` calls it directly with an existing `qtx` (the GitHub App webhook flow is already inside a transaction, so it can't go through `ProvisionProject` which opens its own). We could keep it as an exported method on the concrete type but remove it from the interface, and pass it as a function closure at wiring time in `service.go`. `ProvisionChildProject` can go fully private since nothing external calls it directly. Happy to make that change if that's the direction you'd prefer.

internal/authz/interface.go
ShantKhatri: Makes sense. I'll move it to `api.go` as an exported `RelationAsName` function, update `handlers_authz.go` to use it, and drop the `RelationCreate` constant from authz, which keeps it clean and reduces the risk of drift.

internal/projects/creator.go
ShantKhatri: Agreed, the DB row shouldn't be leaking up that far. I'll do the mapping inside `ProvisionProject` and return the proto type. The lower-level helpers (`ProvisionSelfEnrolledProject`, `ProvisionChildProject`) stay on `*db.Project` since they're internal and the factory path needs it.

ShantKhatri:
> The existing tests were updated for the new constructor signature (`mockStore`, `&flags.FakeClient{}`), but I don't see new tests targeting `ProvisionProject` itself. Since this method now owns permission checks, the feature-flag gate, and transaction boundaries — logic that previously lived in the handler — could we get coverage for:
> 
> * child-project creation when `authzClient.Check` denies
> * child-project creation when `ProjectAllowsProjectHierarchyOperations` is false
> * top-level creation when `flags.ProjectCreateDelete` is disabled
> * top-level creation with no identity in context
> * the `project == nil` post-creation guard
> 
> Even a few table-driven cases covering these branches would close the gap left by the handler logic moving here.

Fair, will add table-driven tests covering as Check denied, hierarchy flag off, top-level flag disabled, missing identity, and the nil project guard. Will be straightforward since the logic is now centralized.

ShantKhatri:
> I realize that we didn't have existing coverage for this code, but it might be good to add test coverage while you're here. Other than that, a couple comments on the service interface between the handlers in controlplane and the rest of the system -- this is somewhere that we were pretty sloppy at first, but we've been trying to clean things up over time.

Thanks for the review. For sure, I'll be adding the test coverage as described by @intelligent-ears.

internal/projects/creator.go
evankanderson: I think just removing ProvisionChildProject would make sense.  Keeping the other in the interface for now makes it somewhat clear that the functionality is still in use elsewhere.

ShantKhatri: Perfect. I'll make the changes accordingly and will update soon.

internal/authz/interface.go
ShantKhatri: Done! I've hoisted RelationAsName to pkg/api/protobuf/go/minder/v1/api.go as an exported function and updated handlers_authz.go to use it. I also dropped the RelationCreate constant from authz/interface.go to keep things clean and reduce drift as discussed.

internal/projects/creator.go
ShantKhatri: Done. ProvisionProject now handles the mapping internally and returns the `*minderv1.Project ` proto type directly. I've also updated CreateProject in `handlers_projects.go` so it no longer has to deal with the DB row mapping.

internal/projects/creator.go
ShantKhatri: Done! ProvisionChildProject has been removed from the ProjectCreator interface and made unexported (provisionChildProject). I also removed the direct tests for it in creator_test.go and replaced them with the table-driven TestProvisionProject tests that exercise it through the public

ShantKhatri:
> The existing tests were updated for the new constructor signature (`mockStore`, `&flags.FakeClient{}`), but I don't see new tests targeting `ProvisionProject` itself. Since this method now owns permission checks, the feature-flag gate, and transaction boundaries — logic that previously lived in the handler — could we get coverage for:
> 
> * child-project creation when `authzClient.Check` denies
> * child-project creation when `ProjectAllowsProjectHierarchyOperations` is false
> * top-level creation when `flags.ProjectCreateDelete` is disabled
> * top-level creation with no identity in context
> * the `project == nil` post-creation guard
> 
> Even a few table-driven cases covering these branches would close the gap left by the handler logic moving here.

Done! I've added a comprehensive table-driven test (`TestProvisionProject`) in `creator_test.go` that covers all of these branches: child-project creation (when `authzClient.Check` denies and when `ProjectAllowsProjectHierarchyOperations` is false), top-level creation (when `ProjectCreateDelete` is disabled and with no identity in context), as well as the successful happy path. CC: @evankanderson

internal/authz/interface.go
```diff
@@ -32,6 +32,7 @@ const (
 	RolePermissionsManager Role = "permissions_manager"
 )
 
+
```
evankanderson: You somehow added a blank line to this file, which introduces a lint / gofmt error.

internal/projects/creator.go
```diff
@@ -204,3 +212,80 @@ func (p *projectCreator) ProvisionChildProject(
...
+func (p *projectCreator) ProvisionProject(
+	ctx context.Context,
+	projectName string,
+	parentProjectID uuid.UUID,
+) (*minderv1.Project, error) {
+	var project *db.Project
+
+	if parentProjectID != uuid.Nil {
+		// Verify permissions if we have a parent
+		if err := p.authzClient.Check(ctx, "create", parentProjectID); err != nil {
```
evankanderson:
```suggestion
		if err := p.authzClient.Check(ctx, minderv1.RelationAsName(minderv1.Relation_RELATION_CREATE), parentProjectID); err != nil {
```

evankanderson (dismissed):
test coverage looks good, thanks!

Need to fix the lint error, and then should be good to merge

internal/authz/interface.go
ShantKhatri: Sorry, it was a mistake. Updating.

ShantKhatri: Done!

intelligent-ears (approved):
LGTM !

evankanderson (approved):

PR #6558, opened 2026-07-01
Feature: support ruletype definitions as rego + metadata

evankanderson:
# Summary

I've talked about this for several months, but some of the recent Rego v1 migrations made me pull the trigger.

This allows Rego ruletypes to be defined as standard `.rego` files (so that `opa fmt`, etc will work on them), and includes the Minder Ruletype metadata via the [OPA metadata convention](https://www.openpolicyagent.org/docs/policy-language#metadata).

# Testing

Added integration tests for the various Minder commands.

Still TODO: documentation for the different file formats supported by Minder (DataSources, Profiles, and two filetypes for RuleTypes).


cmd/cli/app/ruletype/common.go
```diff
@@ -20,51 +20,51 @@ import (
...
-	if err := minderv1.ParseResource(reader, r); err != nil {
-		return fmt.Errorf("error parsing rule type: %w", err)
+		return err
 	}
 
-	// Override the YAML specified project with the command line argument
-	if proj != "" {
-		if r.Context == nil {
-			r.Context = &minderv1.Context{}
+	for _, resource := range resources {
```
dakshhhhh16, line 41: Could we return an error when an explicitly provided file contains no resources? `ReadFromPath` returns an empty slice on EOF, so an empty YAML or Rego file currently makes `ruletype create/apply` exit successfully without doing anything. That feels especially confusing for automation.

cmd/cli/app/ruletype/fixture/applied_rule.rego
```diff
@@ -0,0 +1,34 @@
...
+#     ingest:
+#       type: rest
+#       rest:
+#         endpoint: '/repos/{{.Entity.Owner}}/{{.Entity.Name}}'
+#         parse: 'json'
+#
+#     eval:
+#       rego:
+#         type: 'deny-by-default'
+package minder
```
dakshhhhh16, line 25: This second `package minder` makes the fixture invalid Rego; `opa fmt --fail` reports `unexpected package` here. Could we move the metadata before the original package declaration and keep only one package statement? The duplicate fixture under `cmd/cli/app/fixture` has the same problem.

pkg/fileconvert/testdata/directory/ruletype.rego
```diff
@@ -0,0 +1,37 @@
...
+    input.creator == "banned"
+}
+
+violations [{"msg": msg}] {
+    some comment in minder.datasource.ghapi_comments.pr_comment({
+        "owner": input.properties["github/repo_owner"],
+        "repo": input.properties["github/repo_name"],
+        "pr": input.properties["github/pr_number"],
+    })
+    comment contains "badword"
```
dakshhhhh16: This fixture still fails `opa check`. With `import rego.v1`, the rule on line 30 needs the v1 `contains ... if` form, and line 36 should use `contains(comment, "badword")`. Could we make the fixture valid so the successful conversion test represents a rule the server can actually accept?

cmd/cli/app/ruletype/fixture/applied_rule.rego
evankanderson: Yeah, it turns out that the metadata comment _after_ the package makes the annotation technically associated with the `import` line rather than the `package` line.

I ran `opa check` and `opa fmt` on all the sample rego files.

pkg/fileconvert/testdata/directory/ruletype.rego
evankanderson: Yes on making the fixture valid.  I don't really want the client to actually _check_ the Rego with an embedded toolchain, because that may end up blocking valid future rules if the server is upgraded, but the user is using an old client.  (As someone who has multiple gardens of tools on different machines, I often end up with out-of-date tools _somewhere_.)

cmd/cli/app/ruletype/common.go
evankanderson: Yes, good point!

Added a test for this case as well.

dakshhhhh16:
It's good to go I think

dakshhhhh16:
@evankanderson 
There is a conflict between the head and base, the pr looks good to me.

dakshhhhh16 (approved):

dakshhhhh16:
@JAORMX You can have a final look into this

JAORMX (approved):

PR #6575, opened 2026-07-04
feat: introduce datasources mocking in ruletest framework

### LINKED ISSUE #6573: [LFX] Rule Testing Framework: Datasources Mocking
Part of the rule testing framework implementation (parent issue: #6503).

### Goal
Provide a mechanism for test authors to mock external Data Sources during rule evaluation. This is necessary to test rules that rely on the data source engine to fetch external information (e.g., API calls to third-party services) before making a policy decision.

### Scope
- Introduce a mechanism in the Starlark testing environment (e.g., a `mock_datasource` argument) to allow users to define mock responses for specific data source queries.
- Integrate the mocked data sources into the evaluation engine so that the rule runner intercepts data source lookups and returns the mocked data instead of executing the actual queries.
- Add unit and integration tests to verify that rules using data sources evaluate successfully against the mocked data.

### Context
We have successfully implemented mocking for the `rest` ingester (HTTP) and the `git` ingester (Filesystem). However, rules that declare and depend on external Data Sources currently have no way to mock those queries, meaning they cannot be evaluated reliably in an isolated local environment without network access.

### Acceptance Criteria
- [x] Test authors can specify mock data source responses within their `*.star` test scripts.
- [x] The engine correctly intercepts queries to the Data Source provider and returns the mocked data.
- [x] Un-mocked data source queries are handled gracefully (e.g., returning an appropriate error or falling back safely).
- [x] Comprehensive unit tests are added for the data source mocking layer.

krrish175-byte:
# Summary

This PR introduces support for mocking external Data Sources during rule evaluation in the Starlark testing framework.

By using the new `data_sources` argument within the `eval()` built-in, test authors can now define predefined, offline responses for data source queries. This bypasses the need for the rule engine to hit actual external endpoints, guaranteeing deterministic and reliable test outcomes. 

**Changes Made:**
- Implemented a mock `DataSource` and builder function in `pkg/ruletest/mock_datasource.go` to convert a Starlark dictionary into a valid `DataSourceRegistry`.
- Updated the Starlark `eval()` built-in in `pkg/ruletest/eval.go` to parse the optional `data_sources` parameter and pass it into the rule evaluator initialization.

Fixes #6573 

# Testing

Tested by adding unit tests for the Starlark testing framework that exercise both passing and failing conditions using mocked data sources. 

**Steps to verify:**
1. Check out this branch locally.
2. Run the tests in `pkg/ruletest/` using the following command:
   ```bash
   make test


evankanderson:
I thought that we were going to implement datasources reading from the existing http mocks?  We'd need some sort of mechanism to load the datasource definitions (from a yaml file) and then override the http RoundTripper to return static results.

krrish175-byte:
@evankanderson : I have just pushed a massive update that implements exactly this.

Here's a breakdown of the new changes:

- Loading Data Sources from YAML: The Starlark `eval()` function now accepts a `data_sources` argument (a list of file paths). It reads these YAML files, parses them into protobuf definitions, and registers them into the `DataSourceRegistry` during evaluation.
- HTTP Mocking via RoundTripper: I modified the `datasources` factory builder to accept a `WithTestOnlyTransport()` option. In` pkg/ruletest`, we now inject a custom `http.RoundTripper` that passes requests directly to the test kit's `mockHTTPHandler`.
- Seamless Mocking: Rules can call data sources naturally using `minder.datasource.name.get_val({})`. The HTTP request goes through the mocked RoundTripper, gets intercepted by the `mock_http` dictionary definitions provided in the Starlark test, and returns the mocked response body and status code.

The `pkg/ruletest` tests for data sources are now all passing.

internal/datasources/rest/rest.go
```diff
@@ -51,6 +70,11 @@ func NewRestDataSource(rest *minderv1.RestDataSource, provider provinfv1.Provide
 		return nil, errors.New("rest data source definition is nil")
 	}
 
+	rOpts := &options{}
+	for _, opt := range opts {
+		opt(rOpts)
+	}
```
evankanderson: Consider adding the definition of Option as `func(*DataSource)` to the public API, and then storing the http transport as a field in the DataSource, rather than a separate buildOption / option struct (in two different packages).

https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis has a good writeup of this pattern, though you seem to have grasped most of it.  (We could also make the provider an option function if we decide we want, particularly since it's actually just used for the RoundTripper / Do function.)

internal/datasources/factory.go
```diff
@@ -25,11 +44,16 @@ func BuildFromProtobuf(ds *minderv1.DataSource, provider provinfv1.Provider) (v1
...
+	for _, opt := range opts {
+		opt(bOpts)
+	}
+
 	switch ds.GetDriver().(type) {
 	case *minderv1.DataSource_Structured:
 		return structured.NewStructDataSource(ds.GetStructured())
 	case *minderv1.DataSource_Rest:
-		return rest.NewRestDataSource(ds.GetRest(), provider)
+		return rest.NewRestDataSource(ds.GetRest(), provider, rest.WithTestOnlyTransport(bOpts.testOnlyTransport))
```
evankanderson: If you define the public Option interface, you can then just pass `opts...` to this function.

pkg/ruletest/eval.go
```diff
@@ -13,7 +13,14 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 
+	"github.com/mindersec/minder/internal/datasources"
+	"github.com/mindersec/minder/pkg/fileconvert"
```
evankanderson: The `github.com/mindersec/...` imports need to be in the next section with the other module imports (lint error).

pkg/ruletest/eval.go
```diff
@@ -136,6 +151,62 @@ func parseMockFSDict(mockFSDict *starlark.Dict) (map[string]string, error) {
 	return mockFSMap, nil
 }
 
+type testKitRoundTripper struct {
+	tk *tkv1.TestKit
+}
+
+func (t *testKitRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	return t.tk.Do(req.Context(), req)
+}
```
evankanderson: What about adding this definition to TestKit so that TestKit is a RoundTripper?

pkg/ruletest/eval.go
```diff
@@ -136,6 +151,62 @@ func parseMockFSDict(mockFSDict *starlark.Dict) (map[string]string, error) {
...
+
+func buildDataSourceRegistry(datasourcesList *starlark.List, tk *tkv1.TestKit) (*v1datasources.DataSourceRegistry, error) {
+	registry := v1datasources.NewDataSourceRegistry()
+	if datasourcesList == nil {
+		return registry, nil
+	}
+
+	for i := 0; i < datasourcesList.Len(); i++ {
+		val := datasourcesList.Index(i)
```
evankanderson: You can write this as:


```suggestion
	for i, val := range(datasourcesList) {
```

Actually, since you don't use `i` after this, it could just be `for _, val := ...`

pkg/ruletest/eval.go
```diff
@@ -136,6 +151,62 @@ func parseMockFSDict(mockFSDict *starlark.Dict) (map[string]string, error) {
...
+
+		path := string(pathStr)
+
+		ds, err := loadDataSource(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load data source %q: %w", path, err)
+		}
+
+		// Use the testKitRoundTripper to intercept HTTP requests
+		transport := &testKitRoundTripper{tk: tk}
```
evankanderson: Can you share the RoundTripper across all the DataSources, rather than creating a separate one for each?

pkg/ruletest/eval.go
```diff
@@ -136,6 +151,62 @@ func parseMockFSDict(mockFSDict *starlark.Dict) (map[string]string, error) {
...
+}
+
+func loadDataSource(path string) (*minderv1.DataSource, error) {
+	decoder, closer := fileconvert.DecoderForFile(path)
+	if decoder == nil {
+		return nil, fmt.Errorf("error opening file: %s", path)
+	}
+	defer func(c io.Closer) {
+		_ = c.Close()
+	}(closer)
```
evankanderson: You can simply capture `closer`, like so:

```golang
defer func() {
    closer()
}
```

I think you can also just `defer closer()`, though you may need to silence the linter at that point.

evankanderson (commented):
This looks good!  Very exciting to see this getting hooked up!

One or two small comments (in particular, the lint error needs fixing), but this looks pretty close to good to go.  Nice to see it's not a huge amount of code!

krrish175-byte:
I have addressed your latest comments:

- In `internal/datasources/factory.go`, I updated the iteration logic to just use `range` on `datasourcesList` instead of iterating manually over its keys.
- I created a new `pkg/datasources/v1/options.go` file to hold the `BuildOption` type so it can be shared universally between `rest`,` structured`, and `builtin` datasources without cyclical imports.
- I updated the `ruletest/eval.go` implementation to initialize a single `http.RoundTripper` across multiple datasources, avoiding the creation of an HTTP client repeatedly per rule type if multiple are listed. I also updated it to use `defer closer.Close()` to ensure resources are cleaned up safely.
- Fixed a couple of unused import and Starlark Iter-related build errors that cropped up after the refactoring.

pkg/datasources/v1/options.go
```diff
@@ -0,0 +1,21 @@
...
+
+import "net/http"
+
+// Option is a functional option for configuring a data source
+type Option func(*Options)
+
+// Options contains configuration for a data source
+type Options struct {
+	TestOnlyTransport http.RoundTripper
+}
```
evankanderson, line 14: Because I hadn't actually looked at the code in `pkg/datasources` before -- we're introducing an `Options` struct with public fields because the public package only defines _interfaces_ for DataSources.  It does provide a concrete registry implementation, but that's not helpful for our initialization process.

Given that, this seems reasonable, though it's slightly annoying that the implementations are only available in `internal` and can't be re-used elsewhere.  (But that's not a fight to pick in this PR.)

pkg/ruletest/mock_http.go
```diff
@@ -130,6 +131,11 @@ func buildMockHTTPHandler(mockDict *starlark.Dict) (http.Handler, error) {
 			return nil, fmt.Errorf("mock endpoints must have string keys, got %s", key.Type())
 		}
 
+		parsedURL, err := url.Parse(keyStr)
+		if err == nil && parsedURL.Host != "" {
+			keyStr = parsedURL.Host + parsedURL.Path
+		}
```
evankanderson, line 137: I'm assuming we're doing this to normalize the host and path and avoid e.g. mismatches due to providing query parameters, etc?  (The normalization might be worth a comment in general -- in particular, why is it okay to skip the normalization if `url.Parse()` fails?)

pkg/ruletest/eval.go
```diff
@@ -136,6 +148,53 @@ func parseMockFSDict(mockFSDict *starlark.Dict) (map[string]string, error) {
...
+	registry := v1datasources.NewDataSourceRegistry()
+	if datasourcesList == nil {
+		return registry, nil
+	}
+
+	iter := datasourcesList.Iterate()
+	defer iter.Done()
+	var val starlark.Value
+	for iter.Next(&val) {
+		pathStr, ok := val.(starlark.String)
```
evankanderson, line 161: If you use `Elements` with go 1.23 iterators, this can simply be called with a range loop, per https://pkg.go.dev/go.starlark.net/starlark#List.Elements:


```suggestion
	for val := range datasourcesList.Elements():
		pathStr, ok := val.(starlark.String)
```

evankanderson (approved):
A couple comments that can be handled in a follow-up if needed, but happy to merge this now.

PR #6577, opened 2026-07-06
feat(dev): add JUnit XML output to mindev test and ruletest cleanups

### LINKED ISSUE #6576: [LFX] Rule Testing Framework: JUnit XML Output
Part of the rule testing framework implementation (parent issue: #6503).

## Goal

Provide a mechanism for `mindev test` to output rule evaluation results in standard JUnit XML format. This is necessary to allow CI/CD systems (like GitHub Actions) to natively parse, analyze, and display test reports, failures, and summaries without relying on custom plain-text parsing.

## Scope

* Introduce a new `--output` flag (shorthand `-o`) to the `mindev test` command.
* Implement JUnit XML serialization structs that map Minder rule test results (`ruletest.TestResult`) into standard JUnit tags (`<testsuites>`, `<testsuite>`, `<testcase>`, `<failure>`).
* Ensure that the XML payload is strictly written to `stdout`, while any underlying execution logs or errors continue to route safely to `stderr` to prevent corrupting the XML file.

## Context

We have successfully built the core local rule testing framework (`mindev test`) which runs Starlark tests and outputs a human-readable text summary of passes and failures. However, when these tests run in automated CI environments, developers need a structured, standardized way to consume the results. JUnit XML is the universally supported format for test reporting across almost all CI providers.

## Acceptance Criteria

- [ ] Users can pass `--output=junit` or `-o junit` to `mindev test`.
- [ ] The command successfully outputs a well-formed, standard JUnit XML report containing all evaluated test suites, test cases, and aggregate failure messages.
- [ ] Normal CLI logging output is not mixed into the XML `stdout` stream.
- [ ] Existing output behavior is preserved when the `--output` flag is not provided.

krrish175-byte:
# Summary

This PR introduces a new CLI capability for `mindev test` to produce parsable JUnit XML output for CI integration. This allows standard CI tools (like GitHub Actions) to natively parse and display `mindev test` results. Additionally, it addresses follow-up cleanups from #6575.

**Changes:**
* `cmd/dev/app/test/`: Added an `--output` (`-o`) flag to `mindev test`. By passing `--output=junit`, the tool natively outputs standard JUnit XML to `stdout` instead of the traditional plain-text format.
* `pkg/ruletest/eval.go`: Refactored to use Go 1.23 iterator `range` loops with `datasourcesList.Elements()` instead of manual iteration logic.
* `pkg/ruletest/mock_http.go`: Added comments to clarify why URL normalization is done for mock endpoints and why we safely fall back to the raw string if parsing fails.

Fixes #6576 

# Testing

* Built the updated `mindev` CLI locally:
  ```bash
  go build ./cmd/dev/...
  ```
* Ran rule tests with the new JUnit output flag:
  ```bash
  ./dev test -o junit
  ```
* Verified that the standard output produces valid, well-formatted XML conforming to the JUnit schema, with test suites, test cases, and nested failure details correctly serialized.
* Verified that error logs and general info continue to output to `stderr` so they do not corrupt the XML payload when redirecting stdout to a file (e.g. `./dev test -o junit > results.xml`).
* Ran `make test` to ensure no existing tests were broken by the `ruletest` cleanups.

### DIFF
--- a/cmd/dev/app/test/test.go
+++ b/cmd/dev/app/test/test.go
@@ -5,7 +5,9 @@
 package test
 
 import (
+	"encoding/xml"
 	"errors"
+	"fmt"
 
 	"github.com/spf13/cobra"
 
@@ -14,12 +16,18 @@ import (
 
 // CmdTest returns the test cobra command
 func CmdTest() *cobra.Command {
+	var outputFormat string
+
 	cmd := &cobra.Command{
 		Use:   "test [paths...]",
 		Short: "Run Minder rule tests",
 		Long: "Run Starlark-based tests for Minder rules. Each path may be a file or directory. " +
 			"If no paths are provided, tests the current directory.",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if outputFormat != "text" && outputFormat != "junit" {
+				return fmt.Errorf("unsupported output format %q: must be \"text\" or \"junit\"", outputFormat)
+			}
+
 			if len(args) == 0 {
 				args = []string{"."}
 			}
@@ -31,7 +39,25 @@ func CmdTest() *cobra.Command {
 			}
 
 			if len(results) == 0 {
-				cmd.Printf("No tests found\n")
+				if outputFormat == "text" {
+					cmd.Printf("No tests found\n")
+				}
+				return nil
+			}
+
+			if outputFormat == "junit" {
+				suites := ruletest.AsJUnit(results)
+				bytes, fmtErr := xml.MarshalIndent(suites, "", "  ")
+				if fmtErr != nil {
+					return fmtErr
+				}
+				cmd.Println(xml.Header + string(bytes))
+
+				for _, res := range results {
+					if len(res.Failures) > 0 {
+						return errors.New("one or more tests failed")
+					}
+				}
 				return nil
 			}
 
@@ -55,6 +81,8 @@ func CmdTest() *cobra.Command {
 			return nil
 		},
 	}
+
+	cmd.Flags().StringVarP(&outputFormat, "output", "o", "text", "Output format (text, junit)")
 
 	return cmd
 }
--- a/pkg/ruletest/junit.go
+++ b/pkg/ruletest/junit.go
@@ -1,0 +1,80 @@
+// SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ruletest
+
+import (
+	"encoding/xml"
+	"maps"
+	"slices"
+	"strings"
+)
+
+// JUnitTestSuites is the root element of a JUnit XML report.
+type JUnitTestSuites struct {
+	XMLName    xml.Name         `xml:"testsuites"`
+	TestSuites []JUnitTestSuite `xml:"testsuite"`
+}
+
+// JUnitTestSuite represents a single test suite in a JUnit XML report.
+type JUnitTestSuite struct {
+	Name      string          `xml:"name,attr"`
+	Tests     int             `xml:"tests,attr"`
+	Failures  int             `xml:"failures,attr"`
+	TestCases []JUnitTestCase `xml:"testcase"`
+}
+
+// JUnitTestCase represents a single test case in a JUnit XML report.
+type JUnitTestCase struct {
+	Name      string        `xml:"name,attr"`
+	ClassName string        `xml:"classname,attr"`
+	Failure   *JUnitFailure `xml:"failure,omitempty"`
+}
+
+// JUnitFailure represents a failure within a test case.
+type JUnitFailure struct {
+	Message string `xml:"message,attr"`
+	Type    string `xml:"type,attr,omitempty"`
+	Body    string `xml:",chardata"`
+}
+
+// AsJUnit converts test results into a JUnitTestSuites structure.
+func AsJUnit(results []TestResult) JUnitTestSuites {
+	suitesMap := make(map[string]*JUnitTestSuite)
+
+	for _, res := range results {
+		suiteName := res.Filename
+		if _, ok := suitesMap[suiteName]; !ok {
+			suitesMap[suiteName] = &JUnitTestSuite{
+				Name: suiteName,
+			}
+		}
+
+		suite := suitesMap[suiteName]
+		suite.Tests++
+
+		tc := JUnitTestCase{
+			Name:      res.Name,
+			ClassName: suiteName,
+		}
+
+		if len(res.Failures) > 0 {
+			suite.Failures++
+			tc.Failure = &JUnitFailure{
+				Message: "Test failed",
+				Body:    strings.Join(res.Failures, "\n"),
+			}
+		}
+
+		suite.TestCases = append(suite.TestCases, tc)
+	}
+
+	var suitesList []JUnitTestSuite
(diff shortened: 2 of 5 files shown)


cmd/dev/app/test/junit.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// JUnitFailure represents a failure within a test case.
+type JUnitFailure struct {
+	Message string `xml:"message,attr"`
+	Type    string `xml:"type,attr,omitempty"`
+	Body    string `xml:",chardata"`
+}
+
+func formatJUnitXML(results []ruletest.TestResult) (string, error) {
```
evankanderson: What do you think about changing this to `AsJUnit(results []ruletest.TestResult) (JUnitTestSuites, error)`, and putting it in `pkg/ruletest`, so other implementations could also use it?

(That might also make it easier to add some tests)

cmd/dev/app/test/junit.go
```diff
@@ -0,0 +1,88 @@
...
+
+		if len(res.Failures) > 0 {
+			suite.Failures++
+			var failureMsg string
+			for i, f := range res.Failures {
+				if i > 0 {
+					failureMsg += "\n"
+				}
+				failureMsg += f
+			}
```
evankanderson: Is this `strings.Join(res.Failures, "\n")`?

cmd/dev/app/test/junit.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		suite.TestCases = append(suite.TestCases, tc)
+	}
+
+	suites := JUnitTestSuites{}
+	for _, suite := range suitesMap {
+		suites.TestSuites = append(suites.TestSuites, *suite)
+	}
```
evankanderson: Can this be:

```suggestion
	suites := JUnitTestSuites{
		TestSuites: slices.Collect(maps.Values(suitesMap))
	}
```

cmd/dev/app/test/test.go
```diff
@@ -31,20 +36,34 @@ func CmdTest() *cobra.Command {
...
 				return nil
 			}
 
+			if outputFormat == "junit" {
+				xmlOut, fmtErr := formatJUnitXML(results)
+				if fmtErr != nil {
+					return fmtErr
+				}
+				fmt.Fprintln(os.Stdout, xmlOut)
+			}
```
evankanderson, line 62: This feels like it should be an "if/else" (or an early return), rather than an "if" and then a bunch of point-escapes from printing later on.

cmd/dev/app/test/test.go
```diff
@@ -14,11 +16,14 @@ import (
...
 // CmdTest returns the test cobra command
 func CmdTest() *cobra.Command {
+	var outputFormat string
+
 	cmd := &cobra.Command{
 		Use:   "test [paths...]",
 		Short: "Run Minder rule tests",
 		Long: "Run Starlark-based tests for Minder rules. Each path may be a file or directory. " +
 			"If no paths are provided, tests the current directory.",
+		SilenceUsage: true,
```
evankanderson: You never want to print usage?  What we usually do is print usage if the command doesn't make sense (for example, a nonsense output format, or pointing to a directory with no tests).

evankanderson: Right now, `--output=junk` doesn't produce an error.

cmd/dev/app/test/junit.go
```diff
@@ -0,0 +1,24 @@
...
+		return "", err
+	}
+
+	bytes, err := xml.MarshalIndent(suites, "", "  ")
+	if err != nil {
+		return "", err
+	}
+
+	return xml.Header + string(bytes), nil
+}
```
evankanderson: Given the size of this helper, we'd normally put it in `test/test.go` rather than a separate file.

cmd/dev/app/test/test.go
```diff
@@ -31,7 +39,18 @@ func CmdTest() *cobra.Command {
...
+				}
+				return nil
+			}
+
+			if outputFormat == "junit" {
+				xmlOut, fmtErr := formatJUnitXML(results)
+				if fmtErr != nil {
+					return fmtErr
+				}
+				fmt.Fprintln(os.Stdout, xmlOut)
```
evankanderson: Do you want to have a non-zero exit status if the test summary has failures?

evankanderson: You probably want to use `cmd.Println()` for this, so you can do output capture if we do integration tests.

pkg/ruletest/junit.go
```diff
@@ -0,0 +1,82 @@
...
+
+// JUnitFailure represents a failure within a test case.
+type JUnitFailure struct {
+	Message string `xml:"message,attr"`
+	Type    string `xml:"type,attr,omitempty"`
+	Body    string `xml:",chardata"`
+}
+
+// AsJUnit converts test results into a JUnitTestSuites structure.
+func AsJUnit(results []TestResult) (JUnitTestSuites, error) {
```
evankanderson: Do you need to return `error` at all?

pkg/ruletest/junit.go
```diff
@@ -0,0 +1,82 @@
...
+	for _, suite := range slices.Collect(maps.Values(suitesMap)) {
+		suitesList = append(suitesList, *suite)
+	}
+
+	suites := JUnitTestSuites{
+		TestSuites: suitesList,
+	}
+
+	return suites, nil
+}
```
evankanderson, line 80: Do you want to add a test for this logic?  I can see a few cases:

1. 3 passing results. 2 from one file, one from another (should produce 2 suites)
2. 2 failing, 1 passing result.  Ensure that aggregation happens correctly
3. empty input ==> empty output

evankanderson (commented):
One last set of changes, and I think this should be good to go.

cmd/dev/app/test/junit.go
krrish175-byte: Moved `formatJUnitXML` inline into` test.go` and deleted the separate `junit.go` file.

cmd/dev/app/test/test.go
krrish175-byte: Yes, good catch. Updated the junit branch to return "one or more tests failed" if any result has failures, same as the text path.

krrish175-byte: Done

pkg/ruletest/junit.go
krrish175-byte: You're right, it can never fail. Simplified the signature to `AsJUnit(results []TestResult) JUnitTestSuites`.

pkg/ruletest/junit.go
krrish175-byte: Added all these

cmd/dev/app/test/test.go
```diff
@@ -31,7 +39,25 @@ func CmdTest() *cobra.Command {
...
+				if fmtErr != nil {
+					return fmtErr
+				}
+				cmd.Println(xml.Header + string(bytes))
+
+				for _, res := range results {
+					if len(res.Failures) > 0 {
+						return errors.New("one or more tests failed")
+					}
+				}
```
evankanderson, line 60: JUnit has summary statistics that you could use here, but this seems fine.

evankanderson (approved):

Issue #6533, opened 2026-06-22
Issue remediation: runtime lifecycle


codejasleen:
This issue tracks the runtime lifecycle implementation for the Issue Remediation design discussed in #6522 

## Goal

Implement the runtime behavior for Issue remediation.

## Scope

- Wire issue remediation into the action factory.
- Create issues.
- Create labels.
- Assign issues.
- Persist remediation metadata.
- Close resolved issues.
- Template interpolation.
- Tests.

## Future considerations

The implementation should remain extensible for future enhancements such as comments, and accepted-risk ("Won't Fix") workflows.

## Implementation plan

To keep review size manageable, I currently expect this issue to be implemented across multiple PRs:

- [x] #6578 
- [x] #6609 
- [x] #6631

PR #6578, opened 2026-07-07
Implement runtime support for issue remediation

codejasleen:
# Summary

This PR completes the runtime implementation for Issue Remediation, building on the previously introduced provider interface and remediation configuration.

## Changes

- Add the Issue remediation engine.

- Implement runtime support for:

     - creating GitHub issues when remediation is enabled
     - closing previously created issues when remediation is disabled

- Persist issue metadata to avoid creating duplicate issues while a remediation remains active.

- Wire the issue remediation type into the remediation engine.

- Enable type: issue in remediation validation.

- Add validation for IssueRemediation configuration.

- Add unit tests covering:

     - issue creation
     - issue creation failures
     - existing issue reuse
     - issue closure
     - issue closure without stored metadata

- Regenerate protobuf, OpenAPI, and documentation after enabling the new remediation type.

This makes issue a functional remediation type end-to-end.

Related to #6533 

## Testing

Verified with:

go test ./internal/engine/actions/remediate/issue
go test ./internal/engine/actions/remediate/...
go build ./...

### DIFF
--- a/internal/engine/actions/remediate/issue/issue.go
+++ b/internal/engine/actions/remediate/issue/issue.go
@@ -42,6 +42,7 @@ type issueMetadata struct {
 	Number int `json:"issue_number,omitempty"`
 }
 
+// Remediator implements the issue remediation engine.
 type Remediator struct {
 	issueCli   provifv1.IssuePublisher
 	actionType interfaces.ActionType
@@ -97,8 +98,8 @@ func NewIssueRemediate(
 	}, nil
 }
 
-// IssueTemplateParams is the parameters for the Issue templates
-type IssueTemplateParams struct {
+// TemplateParams is the parameters for the Issue templates
+type TemplateParams struct {
 	// Entity is the entity being evaluated.
 	Entity any
 	// Profile contains the profile definition.
@@ -183,7 +184,7 @@ func (r *Remediator) getParamsForIssueRemediation(
 		return nil, fmt.Errorf("expected repository, got %T", ent)
 	}
 
-	tmplParams := &IssueTemplateParams{
+	tmplParams := &TemplateParams{
 		Entity:  ent,
 		Profile: params.GetRule().Def,
 		Params:  params.GetRule().Params,
@@ -363,7 +364,7 @@ func (r *Remediator) runOff(
 
 func (r *Remediator) getIssueBodyText(
 	ctx context.Context,
-	tmplParams *IssueTemplateParams,
+	tmplParams *TemplateParams,
 ) (string, error) {
 
 	body := new(bytes.Buffer)
--- a/internal/engine/actions/remediate/issue/issue.go
+++ b/internal/engine/actions/remediate/issue/issue.go
@@ -1,0 +1,401 @@
+// SPDX-FileCopyrightText: Copyright 2023 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package issue provides the issue remediation engine.
+package issue
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/rs/zerolog"
+	"google.golang.org/protobuf/proto"
+
+	dbadapter "github.com/mindersec/minder/internal/adapters/db"
+	"github.com/mindersec/minder/internal/db"
+	"github.com/mindersec/minder/internal/engine/interfaces"
+	"github.com/mindersec/minder/internal/util"
+	pb "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+	enginerr "github.com/mindersec/minder/pkg/engine/errors"
+	"github.com/mindersec/minder/pkg/profiles/models"
+	provifv1 "github.com/mindersec/minder/pkg/providers/v1"
+)
+
+const (
+	// RemediateType is the type of the Issue remediation engine.
+	RemediateType = "issue"
+)
+
+// Keep these limits in sync with the proto validation constraints.
+const (
+	// TitleMaxLength is the maximum number of bytes for the title.
+	TitleMaxLength = 75
+
+	// BodyMaxLength is the maximum number of bytes for the body.
+	BodyMaxLength = 65536
+)
+
+type issueMetadata struct {
+	Number int `json:"issue_number,omitempty"`
+}
+
+type Remediator struct {
+	issueCli   provifv1.IssuePublisher
+	actionType interfaces.ActionType
+	setting    models.ActionOpt
+
+	issueCfg *pb.RuleType_Definition_Remediate_IssueRemediation
+
+	titleTemplate *util.SafeTemplate
+	bodyTemplate  *util.SafeTemplate
+}
+
+type paramsIssue struct {
+	repo       *pb.Repository
+	title      string
+	body       string
+	labels     []string
+	assignees  []string
+	metadata   *issueMetadata
+	prevStatus *db.ListRuleEvaluationsByProfileIdRow
+}
+
+// NewIssueRemediate creates a new Issue remediation engine.
+func NewIssueRemediate(
+	actionType interfaces.ActionType,
+	issueCfg *pb.RuleType_Definition_Remediate_IssueRemediation,
+	issueCli provifv1.IssuePublisher,
+	setting models.ActionOpt,
+) (*Remediator, error) {
+	err := issueCfg.Validate()
+	if err != nil {
+		return nil, fmt.Errorf("issue remediation config is invalid: %w", err)
+	}
+
+	titleTmpl, err := util.NewSafeHTMLTemplate(&issueCfg.Title, "title")
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse title template: %w", err)
+	}
+
+	bodyTmpl, err := util.NewSafeHTMLTemplate(&issueCfg.Body, "body")
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse body template: %w", err)
+	}
+
+	return &Remediator{
+		issueCli: issueCli,
+		issueCfg: issueCfg,
+
+		actionType: actionType,
+		setting:    setting,
+
+		titleTemplate: titleTmpl,
+		bodyTemplate:  bodyTmpl,
+	}, nil
+}
+
+// IssueTemplateParams is the parameters for the Issue templates
+type IssueTemplateParams struct {
+	// Entity is the entity being evaluated.
+	Entity any
+	// Profile contains the profile definition.
+	Profile map[string]any
+	// Params contains the rule instance parameters.
+	Params map[string]any
+	// EvalResultOutput contains the evaluation output.
+	EvalResultOutput any
+}
+
+// Class returns the action type of the remediation engine.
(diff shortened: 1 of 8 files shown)


pkg/api/protobuf/go/minder/v1/validators.go
```diff
@@ -510,6 +510,33 @@ func (prRem *RuleType_Definition_Remediate_PullRequestRemediation) Validate() er
 	return nil
 }
 
+// Validate validates an issue remediation definition.
+func (issueRem *RuleType_Definition_Remediate_IssueRemediation) Validate() error {
+	if issueRem == nil {
+		return fmt.Errorf("%w: issue remediation is nil", ErrInvalidRuleTypeDefinition)
+	}
```
evankanderson, line 521: Do we have a check somewhere that this is working correctly?  I think you might need to hook this into the ruletype definition validation function.

internal/engine/actions/remediate/issue/issue.go
```diff
@@ -0,0 +1,402 @@
...
+		return r.runOn(ctx, p)
+
+	case interfaces.ActionCmdOff:
+		return r.runOff(ctx, p)
+
+	case interfaces.ActionCmdDoNothing:
+		return r.runDoNothing(ctx, p)
+	}
+
+	return nil, enginerr.ErrActionSkipped
```
evankanderson: This should probably be an "unimplemented", rather than "skipped", as I don't think we expect to hit this case.

internal/engine/actions/remediate/issue/issue.go
```diff
@@ -0,0 +1,402 @@
...
+	if params.GetEvalResult() != nil {
+		tmplParams.EvalResultOutput = params.GetEvalResult().Output
+	}
+
+	title, err := r.titleTemplate.Render(ctx, tmplParams, TitleMaxLength)
+	if err != nil {
+		return nil, fmt.Errorf("cannot execute title template: %w", err)
+	}
+
+	body, err := r.getIssueBodyText(ctx, tmplParams)
```
evankanderson: Why use `Execute` in a helper for the body text, but `Render` for the title?

(Given that both are going into a JSON struct, I'd use `Render` for both -- `Execute` is handy if you have an existing `io.Writer` that you're looking to fill in.)

codejasleen:
HII @evankanderson I'd appreciate a review when you have time. Thanks!

internal/engine/actions/remediate/issue/issue.go
```diff
@@ -0,0 +1,402 @@
...
+) (json.RawMessage, error) {
+	logger := zerolog.Ctx(ctx).Info().Str("repo", p.repo.String())
+
+	// Process the command
+	switch cmd {
+	case interfaces.ActionCmdOn:
+		logger.Msgf("title:\n%s\n", p.title)
+		logger.Msgf("body:\n%s\n", p.body)
+		logger.Msgf("labels:\n%v\n", p.labels)
+		logger.Msgf("assignees:\n%v\n", p.assignees)
```
evankanderson: Rather than using `Msgf`, can you use `Str` and produce a single log line?

internal/engine/actions/remediate/issue/issue.go
```diff
@@ -0,0 +1,402 @@
...
+			// We cannot do anything without an issue number, so we assume that closing this is a success.
+			return nil, fmt.Errorf("no issue number provided: %w", enginerr.ErrActionSkipped)
+		}
+
+		logger.Msgf(
+			"would close issue #%d in %s/%s",
+			p.metadata.Number,
+			p.repo.GetOwner(),
+			p.repo.GetName(),
+		)
```
evankanderson: Similarly, use named fields in the output rather than %f formatting.

internal/engine/actions/remediate/issue/issue.go
```diff
@@ -0,0 +1,402 @@
...
+	logger := zerolog.Ctx(ctx).With().
+		Str("repo", p.repo.String()).
+		Logger()
+
+	// If we already have an issue recorded in the remediation metadata,
+	// don't create another one.
+	if p.metadata != nil && p.metadata.Number != 0 {
+		logger.Info().
+			Int("issue_number", p.metadata.Number).
+			Msg("issue already exists")
```
evankanderson: We probably don't need this at info resolution; it can end up firing pretty often.

internal/engine/actions/remediate/issue/issue.go
```diff
@@ -0,0 +1,402 @@
...
+	if err != nil {
+		return nil, fmt.Errorf(
+			"error marshalling issue remediation metadata json: %w",
+			err,
+		)
+	}
+
+	logger.Info().
+		Int("issue_number", issue.GetNumber()).
+		Msg("issue remediation completed")
```
evankanderson: I think "completed" here is ambiguous -- does it mean the issue was closed, or does it mean that the issue (in this case) created?

internal/engine/actions/remediate/issue/issue_test.go
```diff
@@ -0,0 +1,278 @@
...
+		},
+		nil,
+		nil,
+		&ratecache.NoopRestClientCache{},
+		credentials.NewGitHubTokenCredential("token"),
+		clients.NewGitHubClientFactory(telemetry.NewNoopMetrics()),
+		properties.NewPropertyFetcherFactory(),
+		"",
+	)
+}
```
evankanderson: Is this actually running against github?  We should use a mock or a fake for this provider.

internal/engine/actions/remediate/issue/issue_test.go
```diff
@@ -0,0 +1,278 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			ctrl := gomock.NewController(t)
+			defer ctrl.Finish()
+
+			mockClient := mockghclient.NewMockGitHub(ctrl)
+
+			provider, err := testGithubProvider()
```
evankanderson: We should just be able to pass in the mockClient to NewIssueRemediate, rather than creating and discarding a potentially-live GitHub client.

internal/engine/actions/remediate/issue/issue_test.go
evankanderson: I can see where you copied this from `pull_request_test.go`, but I don't think it's needed there, either.

internal/engine/actions/remediate/issue/issue_test.go
```diff
@@ -0,0 +1,278 @@
...
+					Return(
+						&github.Issue{
+							Number: github.Int(42),
+						},
+						nil,
+					)
+			},
+			expectedErr:      errors.ErrActionSkipped,
+			expectedMetadata: nil,
+		},
```
evankanderson, line 187: Can we add a test for "attempt to close an already-closed issue"?

I'm imagining a case where Minder opens an issue, someone does work to close it (e.g. maybe via a pull request), marks the issue closed, and then Minder comes along and discovers that the policy no longer needs remediation, and tries to close the already-closed issue.

evankanderson (commented):
This is looking promising!  A few comments, but the tests are looking pretty good, in my opinion.

internal/engine/actions/remediate/issue/issue_test.go
codejasleen: I tested this against the GitHub API. Closing an already-closed issue returns 200 OK and no error, just like closing an open issue. Since runOff() only checks for an error and doesn't branch on the returned issue state, the existing "close an issue" test already exercises the same code path. Unless you'd prefer to explicitly document the GitHub behavior in a separate test, I believe the current test already covers this case. Am I understanding this correctly or are you thinking of different scenario?

evankanderson: I think it's with documenting the 200 behavior, particularly as other code forges may not have the same behavior and may need adjustment code.

codejasleen:
@evankanderson I've pushed the requested changes. Could you please take a look when you have a chance and let me know if they look right?

internal/engine/actions/remediate/issue/issue.go
```diff
@@ -0,0 +1,385 @@
...
+	"github.com/mindersec/minder/pkg/profiles/models"
+	provifv1 "github.com/mindersec/minder/pkg/providers/v1"
+)
+
+const (
+	// RemediateType is the type of the Issue remediation engine.
+	RemediateType = "issue"
+)
+
+// Keep these limits in sync with the proto validation constraints.
```
evankanderson: These don't need to be in sync -- the proto limits are the length of the go templates (e.g. `(( if ... }}xyz{{end}}`, while these limits are limits on the length of the _rendered_ output that goes to GitHub/GitLab.

internal/engine/actions/remediate/issue/issue.go
```diff
@@ -0,0 +1,385 @@
...
+
+	var remErr error
+	switch r.setting {
+	case models.ActionOptOn:
+		return r.run(ctx, cmd, p)
+
+	case models.ActionOptDryRun:
+		return r.dryRun(ctx, cmd, p)
+
+	case models.ActionOptOff, models.ActionOptUnknown:
```
evankanderson, line 166: Do we want a default case here as well?

evankanderson: (if you have a default case, you don't need to pre-declare remErr and then return it after the case statement)

internal/engine/actions/remediate/issue/issue_test.go
```diff
@@ -0,0 +1,277 @@
...
+				},
+			}
+
+			evalParams.SetEvalResult(&interfaces2.EvaluationResult{})
+
+			cmd := interfaces.ActionCmdOn
+			switch tt.name {
+			case "close an issue", "close already closed issue", "close issue without metadata":
+				cmd = interfaces.ActionCmdOff
+			}
```
evankanderson: It's better to make this another parameter (using default-false to only declare it for the two tests) rather than hard-coding test names in the body -- people won't expect or look for the names as strings in the body.

evankanderson (commented):
A few more items, but this is looking close!  Have you tested this end to end against GitHub with a rule yet?

codejasleen:
> A few more items, but this is looking close! Have you tested this end to end against GitHub with a rule yet?

Yes. I tested the runtime end to end against GitHub using a temporary rule and profile.

Verified behavior:

- Issue is created when remediation becomes active.

- Reconciliation while remediation remains active does not create duplicate issues.

- Issue is automatically closed when remediation is no longer required.

internal/engine/actions/remediate/issue/issue_test.go
codejasleen: Good suggestion. I refactored the test so the action command and remediation metadata are explicit test case parameters rather than deriving behavior from the test name. I considered using a boolean flag as suggested, but since the value being varied is ActionCmd, making it an explicit parameter felt a bit more direct and scalable?

codejasleen:
@evankanderson I also ran into an unrelated CI failure in the GitHub webhook tests. It passes locally, so it appears to be a flaky timing issue rather than being related to this PR.

evankanderson (approved):

PR #6591, opened 2026-07-12
Allow digit-prefixed repository owners

### LINKED ISSUE #6590: fix: GitLab repository registration fails for namespaces starting with a digit
## Summary

Minder fails to register GitLab repositories whose namespace (group or
username) starts with a digit, with the error:

```
validation error: repository.owner: does not match regex pattern `^[A-Za-z][-[:word:]]*$`
```

GitLab explicitly allows group and namespace names starting with digits
("Start with a letter, digit, emoji, or underscore" per GitLab's UI).
GitHub does not allow this, so the original regex was correct for GitHub
but is too restrictive for GitLab.

## Reproduction

1. Create a GitLab group with a digit-starting name (e.g. `2testminder`)
2. Create a repo inside it (e.g. `2testminder/2testminder`)
3. Enroll the GitLab provider in Minder
4. Run `minder repo register`

The repo appears correctly in the TUI list (discovery works) but fails
at registration with the regex validation error.

## Location

`proto/minder/v1/minder.proto`, `UpstreamRepositoryRef.owner` field (~line 1278):

```protobuf
string owner = 1 [
        pattern: "^[A-Za-z][-[:word:]]*$"  // too restrictive for GitLab
        max_len: 200,
    },
```

## Fix

```protobuf
pattern: "^[A-Za-z0-9][-[:word:]]*$"
```

## Impact

- Any GitLab group or namespace starting with a digit cannot have repos
  registered in Minder, despite being valid GitLab namespaces
- The repos appear in the TUI list but always fail registration —
  inconsistent UX
- Discovered during GitLab provider testing for OSPS level-1 rule porting

### To Reproduce

_No response_

_No response_

w3lld1:
# Summary

Allow `UpstreamRepositoryRef.owner` values to start with a digit so valid GitLab namespaces such as `2testminder` can be registered. The generated protobuf descriptor is updated, and a regression test covers letter-prefixed, digit-prefixed, and invalid punctuation-prefixed owners.

Fixes #6590

# Testing

- `go test ./pkg/api/protobuf/go/minder/v1 -run TestUpstreamRepositoryRef_Validate -count=1`
- `buf lint`
- `make gen`
- `git diff --check`


intelligent-ears (commented):
Thanks for the turnaround on this! The fix looks correct — `^[A-Za-z0-9][-[:word:]]*$` accurately reflects GitLab's actual namespace rules while remaining valid for GitHub (which doesn't allow digit-starting org names anyway).

A few notes:
- Proto change is minimal and correct
- `make gen` was clearly run — the `.pb.go` binary descriptor update matches the pattern change
- Test coverage is good with the three cases

One suggestion: it might be worth adding a test case for a digit-hyphen-word style name like `2-mindertest` (digit followed by hyphens and words), since that's the real-world pattern that triggered this issue. The current regex should handle it, but an explicit test case would make that clear.
Otherwise this looks good to me. Just needs the CLA to be signed before it can be merged.

w3lld1:
Added the suggested `2-mindertest` regression case in `2cb9a60f`.

Validation:
- `go test ./pkg/api/protobuf/go/minder/v1 -run '^TestUpstreamRepositoryRef_Validate$'`
- `git diff --check`

The CLA check is now green as well.

intelligent-ears (approved):

intelligent-ears:
LGTM !

evankanderson (approved):

PR #6593, opened 2026-07-13
fix: request license information from GitLab API

### LINKED ISSUE #6589: fix: GitLab project license property always empty due to missing ?license=true parameter
## Summary

The `gitlab/license` entity property (`RepoPropertyLicense`) is always empty for GitLab repositories because the project fetch in `internal/providers/gitlab/repository_properties.go` does not pass the `?license=true` query parameter required by the GitLab API to include license information in the response.

## Reproduction

Without `?license=true`:
```
GET /api/v4/projects/gitlab-org%2Fgitlab
→ "license": null
```
With `?license=true`:
```
GET /api/v4/projects/gitlab-org%2Fgitlab?license=true
→ "license": {"key": "mit", "name": "MIT License", ...}
```
## Fix

In `getGitLabProject()`, append `?license=true` to the project URL path:

```go
projectURLPath, err := url.JoinPath("projects", url.PathEscape(upstreamID))
// change to:
projectURLPath = projectURLPath + "?license=true"
```

## Impact

- `gitlab/license` property is always `""` for all GitLab repositories
- Any rule relying on `RepoPropertyLicense` or `.Entity.License` for GitLab
  will never get license data
- Discovered during OSPS level-1 rule porting work (#6435)

### To Reproduce

_No response_

_No response_

Anni0506:
# Summary

Fixes #(related issue)

# Testing

# Summary

This PR fixes an issue where the `gitlab/license` entity property (`RepoPropertyLicense`) is always empty for GitLab repositories. The fetch logic in `internal/providers/gitlab/repository_properties.go` has been updated to correctly pass the `?license=true` query parameter required by the GitLab API to include license details in the response.

Fixes #6589

# Testing

This is a straightforward API query parameter addition. The underlying `gitlab` provider package logic handles the mapping once the license payload is retrieved via the modified URL path.

### DIFF
--- a/internal/providers/gitlab/properties_test.go
+++ b/internal/providers/gitlab/properties_test.go
@@ -131,7 +131,6 @@ func Test_gitlabClient_FetchAllProperties(t *testing.T) {
 			},
 		},
 		{
-			name: "repository succeeds",
 			args: args{
 				ctx: context.TODO(),
 				getByProps: properties.NewProperties(map[string]any{
@@ -143,18 +142,29 @@ func Test_gitlabClient_FetchAllProperties(t *testing.T) {
 				properties.RepoPropertyIsPrivate:  true,
 				properties.RepoPropertyIsArchived: false,
 				properties.RepoPropertyIsFork:     false,
+				RepoPropertyLicense:               "mit", // Uses the correct constant
 			}),
 			wantErr: false,
-			gitLabServerMockFunc: func(w http.ResponseWriter, _ *http.Request) {
+			gitLabServerMockFunc: func(w http.ResponseWriter, r *http.Request) {
+				// Verify the query parameter you added is being sent
+				if r.URL.Query().Get("license") != "true" {
+					t.Errorf("expected query param license=true, got %s", r.URL.RawQuery)
+					w.WriteHeader(http.StatusBadRequest)
+					return
+				}
+
 				resp := &gitlab.Project{
-					ID:                1,
-					Name:              "project-1",
-					Description:       "project-1 description",
-					Visibility:        gitlab.PrivateVisibility,
-					Archived:          false,
-					ForkedFromProject: nil,
+					ID:                  1,
+					Name:                "project-1",
+					Description:         "project-1 description",
+					Visibility:          gitlab.PrivateVisibility,
+					Archived:            false,
+					ForkedFromProject:   nil,
 					Namespace: &gitlab.ProjectNamespace{
 						Path: "group",
+					},
+					License: &gitlab.ProjectLicense{
+						Name: "mit", // Changed Key to Name to match repository_properties.go
 					},
 				}
 
--- a/internal/providers/gitlab/properties_test.go
+++ b/internal/providers/gitlab/properties_test.go
@@ -407,3 +407,44 @@ func newTestGitlabProvider(endpoint string) *gitlabClient {
 		cli: &http.Client{},
 	}
 }
+
+
+func TestGitlabProjectToProperties_License(t *testing.T) {
+	t.Parallel()
+
+	// 1. Create a mock GitLab project with a license
+	mockProject := &gitlab.Project{
+		ID:   123,
+		Name: "test-repo",
+		Namespace: &gitlab.ProjectNamespace{
+			Path: "test-org",
+		},
+		License: &gitlab.ProjectLicense{
+			Name: "Apache-2.0",
+		},
+		Visibility:    gitlab.PublicVisibility,
+		Archived:      false,
+		DefaultBranch: "main",
+	}
+
+	// 2. Run the function we want to test
+	props, err := gitlabProjectToProperties(mockProject)
+	if err != nil {
+		t.Fatalf("unexpected error converting project to properties: %v", err)
+	}
+
+	// 3. Verify that the license was successfully populated and is not empty
+	licenseProp := props.GetProperty(RepoPropertyLicense)
+	if licenseProp == nil {
+		t.Fatal("expected license property to be populated, but got nil")
+	}
+
+	licenseVal, err := licenseProp.AsString()
+	if err != nil {
+		t.Fatalf("failed to read license property as string: %v", err)
+	}
+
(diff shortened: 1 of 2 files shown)


Anni0506:
I have read the CLA Document and I hereby sign the CLA

internal/providers/gitlab/repository_properties.go
```diff
@@ -50,6 +50,7 @@ func (c *gitlabClient) getGitLabProject(
 	ctx context.Context, upstreamID string,
 ) (*gitlab.Project, error) {
 	projectURLPath, err := url.JoinPath("projects", url.PathEscape(upstreamID))
+	projectURLPath = projectURLPath + "?license=true"
 	if err != nil {
```
intelligent-ears, line 53: The `?license=true` is appended before the error check — the code appends the query param before checking if `url.JoinPath` returned an error. If `JoinPath` fails, `projectURLPath` would be an empty string and the append would produce just `?license=true`. The correct order is:

```
projectURLPath, err := url.JoinPath("projects", url.PathEscape(upstreamID))
if err != nil {
    return nil, fmt.Errorf("failed to join URL path for project using upstream ID: %w", err)
}
projectURLPath = projectURLPath + "?license=true"
```

intelligent-ears (commented):
This PR doesn't add any tests verifying that the license field actually gets populated.There can be at least a unit test for `gitlabProjectToProperties` that confirms `license` is non-empty when the API response includes a license.

evankanderson:
> This PR doesn't add any tests verifying that the license field actually gets populated.There can be at least a unit test for `gitlabProjectToProperties` that confirms `license` is non-empty when the API response includes a license.

In particular, we should probably update `Test_gitlabClient_FetchAllProperties` in `properties_test.go` to properly emulate the server behavior, and complain / fail the test if the query param is not present.

internal/providers/gitlab/properties_test.go
```diff
@@ -131,7 +131,6 @@ func Test_gitlabClient_FetchAllProperties(t *testing.T) {
 			},
 		},
 		{
-			name: "repository succeeds",
```
evankanderson, line 134: Why change this name to empty-string?

internal/providers/gitlab/properties_test.go
```diff
@@ -143,19 +142,30 @@ func Test_gitlabClient_FetchAllProperties(t *testing.T) {
...
+					Name:                "project-1",
+					Description:         "project-1 description",
+					Visibility:          gitlab.PrivateVisibility,
+					Archived:            false,
+					ForkedFromProject:   nil,
 					Namespace: &gitlab.ProjectNamespace{
 						Path: "group",
 					},
+					License: &gitlab.ProjectLicense{
+						Name: "mit", // Changed Key to Name to match repository_properties.go
```
evankanderson: Will this comment make sense in 6 months?

internal/providers/gitlab/properties_test.go
```diff
@@ -397,3 +407,44 @@ func newTestGitlabProvider(endpoint string) *gitlabClient {
 		cli: &http.Client{},
 	}
 }
+
+
+func TestGitlabProjectToProperties_License(t *testing.T) {
```
evankanderson: What does this test cover that the other test doesn't? 

In general, extra tests add maintenance burden when refactoring code.  That burden pays off when the tests cover additional important code cases, but redundant tests add more maintenance cost without commensurate benefits.

evankanderson (commented):
Thanks for adding the tests!  The test implementation brought up a couple questions (this is the learning process, so it's totally normal to hit these sorts of things).

internal/providers/gitlab/properties_test.go
Anni0506: My mistake! Restored the test name back to \"repository succeeds\".

internal/providers/gitlab/properties_test.go
Anni0506: Good point—removed the temporary comment to keep the code clean!

internal/providers/gitlab/properties_test.go
Anni0506: Agreed! Since the table test already covers RepoPropertyLicense, I removed this redundant test to avoid duplicate coverage and extra maintenance.

internal/providers/gitlab/properties_test.go
```diff
@@ -143,9 +143,17 @@ func Test_gitlabClient_FetchAllProperties(t *testing.T) {
 				properties.RepoPropertyIsPrivate:  true,
 				properties.RepoPropertyIsArchived: false,
 				properties.RepoPropertyIsFork:     false,
+				properties.RepoPropertyLicense:    "mit",
 			}),
 			wantErr: false,
-			gitLabServerMockFunc: func(w http.ResponseWriter, _ *http.Request) {
+			gitLabServerMockFunc: func(w http.ResponseWriter, r *http.Request) {
+				// Verify the query parameter you added is being sent
```
evankanderson, line 150: The same comment here [as in a previous review](https://github.com/mindersec/minder/pull/6593#discussion_r3597006886) -- will this comment make sense when reading the code in 3 months?

internal/providers/gitlab/properties_test.go
```diff
@@ -143,9 +143,17 @@ func Test_gitlabClient_FetchAllProperties(t *testing.T) {
 				properties.RepoPropertyIsPrivate:  true,
 				properties.RepoPropertyIsArchived: false,
 				properties.RepoPropertyIsFork:     false,
+				properties.RepoPropertyLicense:    "mit",
```
evankanderson, line 146: I think this is just:


```suggestion
				RepoPropertyLicense:    "mit",
```

Since it's from _this_ package, not from the shared `properties` package.

evankanderson (commented):
Your test is failing to compile due to the incorrect package name for the constant.

intelligent-ears:
We've merged #6724, which landed the same fix with the test wired into place, so I'm closing this as this seems dead.

PR #6607, opened 2026-07-17
Complete Rego v1 migration safeguards

dakshhhhh16:
## Summary

This completes the remaining Rego v1 migration work before starting with the Phase 4.

Minder can already detect both Rego versions and warn users about legacy policies. This adds the final enforcement step, allowing deployments to reject Rego v0 policies through a feature flag while still providing clear migration guidance.

Part of #5262.

## What changed

- Added the `rego_v1_refuse_v0` feature flag.
- Rejects V0-only policies during rule type creation and updates when refusal is enabled.
- Returns actionable guidance recommending `opa fmt --v0-v1`.
- Updated `minder rule-type lint` to validate policies using Rego v1 syntax.
- Added tests covering V0 refusal, V1 acceptance, invalid policies, and CLI lint behavior.
- Migrated remaining documentation, protobuf test policies, CLI fixtures, and golden files to `import rego.v1`.
- Removed all remaining `future.keywords` imports.
- Preserved intentional V0 compatibility fixtures used to verify detection, warnings, and fallback behavior until Phase 4 removes that support.

## Rollout behavior

- With refusal disabled, existing Rego v0 policies continue to work.
- With `rego_v1_refuse_v0` enabled, V0-only policies are rejected.
- Rego v1 policies continue to be accepted.
- Users receiving a refusal are told how to migrate their policy with OPA's formatter.

## Testing

The complete repository test suite passes:

```bash
TMPDIR=/private/tmp go test ./...

### CODE REVIEW

**Comment by dakshhhhh16** [comment]
@evankanderson, 
I have completed the migration (i.e all four phases locally) after this pr is merged I'll open the final pr for the migration which will  close the parent issue #5262

File: pkg/ruletypes/service.go [hunk 1]
```diff
@@ -338,7 +339,10 @@ func (s *ruleTypeService) validateAndDetectRegoVersion(ctx context.Context, rule
 		}
 		return "v0", nil
```
evankanderson, line 351: It's slightly weird to have this case potentially return "v0" when the refusal flag is on, but it's probably a configuration error to have `rego_v1_dual_parse` false and `rego_v1_refuse_v0` true.

Maybe emit a sternly-worded (error-level) message to the operator in this case?

pkg/ruletypes/service_test.go
```diff
@@ -326,6 +326,36 @@ func TestRuleTypeService(t *testing.T) {
...
+			RuleType:      newRuleType(withBasicStructure, withV0OnlyRego),
+			ExpectedError: "opa fmt --v0-v1",
+			TestMethod:    update,
+			FeatureFlags: map[string]any{
+				string(flags.RegoV1DualParse): true,
+				string(flags.RegoV1RefuseV0):  true,
+			},
+		},
+		{
+			Name:       "CreateRuleType allows V1-only rego when refusal flag is on",
```
evankanderson, line 353: Can you have a test with refusal on and a dual-parsed ruletype, just to make sure that it goes through even without `import rego.v1`?

evankanderson (commented):
This looks very good!  Only two comments, one on a configuration edge case I'm not 100% sure how to solve, and one on an extra test case for peace of mind.

pkg/ruletypes/service.go
dakshhhhh16: Minder now emits an error-level operator message when `rego_v1_refuse_v0` is enabled without `rego_v1_dual_parse`. I also added a test confirming the invalid configuration is logged at error level.

pkg/ruletypes/service_test.go
dakshhhhh16: The new test uses a policy that parses under both Rego V0 and V1 without importing `rego.v1`. It confirms the policy is accepted and stored as V1 when dual parsing and V0 refusal are both enabled.

pkg/ruletypes/service.go
evankanderson, line 366: Okay, sorry I didn't see this earlier, but it seems like we might be able to simplify this (and save some effort) on the refuse-V0 path by storing the migration error instead of attempting to parse as V0 only to refuse it later:


```suggestion
func (s *ruleTypeService) validateAndDetectRegoVersion(ctx context.Context, ruleType *pb.RuleType) (string, error) {
	def := ruleType.GetDef().GetEval().GetRego().GetDef()
	if def == "" {
		return "v0", nil
	}

	dualParse := flags.Bool(ctx, s.featureFlags, flags.RegoV1DualParse)
	refuseV0 := flags.Bool(ctx, s.featureFlags, flags.RegoV1RefuseV0)
	if refuseV0 && !dualParse {
		zerolog.Ctx(ctx).Error().Msg(
			"invalid Rego feature flag configuration: rego_v1_refuse_v0 requires rego_v1_dual_parse")
	}

	type errV0 error
	if refuseV0 {
		errV0 = regoeval.V0MigrationMessage
	} else {
		errV0 := ast.ParseModuleWithOpts("minder-ruletype-def.rego", def,
ast.ParserOptions{RegoVersion: ast.RegoV0})
	}
	_, errV1 := ast.ParseModuleWithOpts("minder-ruletype-def.rego", def,
ast.ParserOptions{RegoVersion: ast.RegoV1})

	switch {
	case errV0 == nil && errV1 == nil:
		// Valid in both dialects — store as V1 when flag is on.
		if dualParse {
			return "v1", nil
		}
		return "v0", nil
	case errV0 == nil:
		// if V0 migration is enabled, we will always have a V0 error.
		return "v0", nil
	case errV1 == nil:
		// V1 only — gate behind feature flag.
		if !dualParse {
			return "", fmt.Errorf("rego definition uses V1-only syntax which is not yet enabled")
		}
		return "v1", nil
	default:
		// Invalid in both dialects — return V0 error as it's the more
		// likely intent for authors not yet on V1.
		return "", fmt.Errorf("rego definition is invalid: %w", errV0)
	}
}
```

pkg/ruletypes/service_test.go
```diff
@@ -381,6 +434,11 @@ func TestRuleTypeService(t *testing.T) {
 			ctrl := gomock.NewController(t)
 			defer ctrl.Finish()
 			ctx := context.Background()
+			var logOutput bytes.Buffer
+			if scenario.ExpectedLog != "" {
+				log := zerolog.New(&logOutput)
+				ctx = log.WithContext(ctx)
+			}
```
evankanderson, line 440: This is great and goes above and beyond, thanks!

evankanderson (dismissed):
Up to you whether you want to adjust `validateAndDetectRegoVersion` to the suggested one.  Otherwise, we can merge as-is.

pkg/ruletypes/service.go
dakshhhhh16: Applied in 6c1d51f0d. `validateAndDetectRegoVersion` now skips the V0 parse when refusal is enabled and carries the migration error through the existing version decision. I used a `var errV0 error` plus assignment to avoid declaration shadowing, and updated the invalid flag-configuration test for the resulting fail-closed behavior. The affected tests and golangci-lint pass.

evankanderson (approved):

PR #6622, opened 2026-07-22
Advance the Rego V1 migration while retaining V0 compatibility

### LINKED ISSUE #5262: Figure out a Rego V1 migration plan

dakshhhhh16:
## Summary

This PR completes Phase 4 of the Rego V1 migration and makes Rego V1 the primary supported policy dialect in Minder, while retaining support for existing stored V0 policies.

The earlier migration phases introduced version detection, migrated existing policies, and added safeguards for legacy Rego V0 definitions. Since the existing policies and test fixtures now use Rego V1, the temporary compatibility paths and rollout flags can be removed.

Closes #5262

## What changed

- Removed Rego V0 parsing and fallback behavior from rule type validation
- Configured the Rego evaluator to always compile policies using `ast.RegoV1`
- Removed runtime selection based on the stored Rego version
- Removed the `rego-v1-dual-parse` feature flag
- Removed the `rego-v1-warn-v0` feature flag
- Removed the `rego-v1-refuse-v0` feature flag
- Removed obsolete Rego V0 warnings from rule type create and update responses
- Updated `mindev ruletype lint` to validate policies exclusively as Rego V1
- Added actionable migration guidance when a policy cannot be parsed as Rego V1
- Updated rule type creation and updates to persist `v1` as the Rego version
- Removed version detection and conversion helpers that are no longer required
- Updated tests to cover permanent Rego V1 behavior
- Added migration documentation for external rule authors

## Behavior after this change

Rule types containing valid Rego V1 policies continue to be accepted and evaluated normally.

Policies using legacy Rego V0-only syntax are rejected during rule type creation or update. The returned validation error includes the command required to migrate the policy.

The evaluator no longer reads the stored Rego version to choose a parser. All Rego policies are evaluated using Rego V1 semantics.

## Migration guidance

External rule authors with Rego V0 policies can migrate them using OPA's formatter:

```bash
opa fmt --v0-v1 --write policy.rego
```

After placing the migrated policy back into the rule type YAML, the complete rule type can be validated with:

```bash
./bin/mindev ruletype lint -r path/to/rule-type.yaml
```

The documentation also explains how to migrate policies embedded in the `def.eval.rego.def` field.

## Testing

- Ran the focused rule type service tests
- Ran the Rego evaluator and runtime engine tests
- Ran the control-plane handler tests
- Ran the `mindev` rule type lint tests
- Ran the complete Go test suite with `go test ./...`
- Ran `golangci-lint run --new-from-rev upstream/main`
- Confirmed `go mod tidy` produces no changes


pkg/flags/constants.go
```diff
@@ -14,14 +14,4 @@ const (
...
-	// RegoV1DualParse enables dual-parse version detection (try V1, fall back
-	// to V0) when creating or updating rule types. When disabled, all rule
-	// types are treated as Rego V0.
-	RegoV1DualParse Experiment = "rego_v1_dual_parse"
-	// RegoV1WarnV0 adds a non-fatal warning to rule type create and update
-	// responses when the accepted policy is evaluated as Rego V0.
-	RegoV1WarnV0 Experiment = "rego_v1_warn_v0"
-	// RegoV1RefuseV0 rejects V0-only Rego when creating or updating rule
-	// types.
-	RegoV1RefuseV0 Experiment = "rego_v1_refuse_v0"
```
evankanderson, line 26: Can we keep the `refuse_v0` path at least for a while?  I haven't rolled out that PR to the Custcodian-hosted instance yet, so this feels like it may slam the door a bit faster than I'd like for folks who haven't rewritten their rules yet.

I've got a race this weekend, but I can try to get some statistics on how many rules are currently still in V0 mode sometime soon.

internal/controlplane/handlers_ruletype_test.go
```diff
@@ -86,58 +64,6 @@ func TestCreateRuleType(t *testing.T) {
...
-			),
-			features: map[string]any{
-				string(flags.RegoV1DualParse): true,
-				string(flags.RegoV1WarnV0):    true,
-			},
-			request: &minderv1.CreateRuleTypeRequest{
-				RuleType: ruleTypeWithRego(regoV0Definition),
-			},
-			expectedWarnings: []string{regoV0DeprecationWarning},
-		},
```
evankanderson, line 106: We should probably keep some of these tests, just remove the flag guards (e.g. we should still check that V0 gives a reasonable warning, just that we don't need flags to get it to happen).

dakshhhhh16:
@evankanderson
This is the final pr marking the close of the issue #5262

internal/engine/eval/rego/eval.go
```diff
@@ -140,13 +139,8 @@ func (e *Evaluator) Eval(
 	// this explicitly.
 	obj := res.Object
 
-	// Register options to expose functions
-	regoFuncOptions := []func(*rego.Rego){
-		rego.SetRegoVersion(e.regoVersion),
-	}
```
evankanderson, line 146: What happens if we have a rule stored in the database as v0 and we roll forward this change?

evankanderson (changes requested):
I'm not sure this is safe to do yet; I certainly haven't checked my instances to ensure that all V0 rules have been migrated to V1.

I'm happy to force the dual-parse and even the warning flag on, but I don't think we can get rid of the V0 code yet.

dakshhhhh16:
I agree that removing V0 support completely is premature until existing installations have been checked.

I have revised the PR so dual parsing and V0 warnings are always enabled, while V0 parsing and evaluation remain supported. The `rego_v1_refuse_v0` flag is also retained so operators can enable rejection when their installations are ready.

pkg/flags/constants.go
dakshhhhh16: Kept `rego_v1_refuse_v0` as requested. Only the dual-parse and warning flags have been removed because those behaviors are now always enabled.

V0-only policies remain accepted by default and are rejected with migration guidance only when the refusal flag is enabled.

internal/controlplane/handlers_ruletype_test.go
dakshhhhh16: Restored the warning coverage without the feature flag setup.

The tests now verify that creating or updating a V0 rule type always returns the migration warning, while a V1 rule type does not produce that warning.

internal/engine/eval/rego/eval.go
dakshhhhh16: Existing rules stored as `v0` will continue to work.

The runtime cache now always reads the stored `rego_version` and passes it to the evaluator through `WithRegoVersion`. A rule stored as `v0` is therefore still parsed and evaluated using `ast.RegoV0`, while records stored as `v1` use `ast.RegoV1`. The V0 evaluator regression and version conversion tests have also been retained.

evankanderson (approved):
Looks good, thanks for the changes!

evankanderson:
Security scan failures were due to newly-announced vulnerabilities, not introduced in this PR.

PR #6637, opened 2026-07-28
fix(mindev): exit with non-zero status on syntax errors

krrish175-byte:
Fixes a bug where `mindev test` exits with 0 even when test files fail to load due to syntax errors.

Additionally, this PR fixes a correctness issue where an empty `mock_fs` dict caused an evaluation error during testing, instead of successfully simulating an empty mocked filesystem. Tests were added to ensure `mock_fs` correctly supports being empty.

Adds support for profile `params` being applied to rules.  By default, `params` defaults to an empty map unless explicitly specified.

### DIFF
--- a/cmd/dev/app/test/test.go
+++ b/cmd/dev/app/test/test.go
@@ -41,7 +41,7 @@ func CmdTest() *cobra.Command {
 			}
 
 			if len(results) == 0 {
-				if outputFormat == "text" {
+				if outputFormat == "text" && loadErr == nil {
 					cmd.Printf("No tests found\n")
 				}
 				return loadErr
--- a/cmd/dev/app/test/test.go
+++ b/cmd/dev/app/test/test.go
@@ -33,16 +33,18 @@ func CmdTest() *cobra.Command {
 			}
 
 			runner := ruletest.NewRunner()
+			var loadErr error
 			results, err := runner.RunPaths(args)
 			if err != nil {
 				cmd.PrintErrf("Error(s) running tests:\n%v\n", err)
+				loadErr = errors.New("one or more test files failed to load")
 			}
 
 			if len(results) == 0 {
 				if outputFormat == "text" {
 					cmd.Printf("No tests found\n")
 				}
-				return nil
+				return loadErr
 			}
 
 			if outputFormat == "junit" {
@@ -58,7 +60,7 @@ func CmdTest() *cobra.Command {
 						return errors.New("one or more tests failed")
 					}
 				}
-				return nil
+				return loadErr
 			}
 
 			hasFailures := false
@@ -78,7 +80,7 @@ func CmdTest() *cobra.Command {
 				return errors.New("one or more tests failed")
 			}
 
-			return nil
+			return loadErr
 		},
 	}
 
--- a/cmd/dev/app/test/test_test.go
+++ b/cmd/dev/app/test/test_test.go
@@ -1,0 +1,33 @@
+// SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package test
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestCmdTest_ExitCodeOnLoadError(t *testing.T) {
+	t.Parallel()
+
+	tmpDir := t.TempDir()
+	brokenFile := filepath.Join(tmpDir, "broken_test.star")
+	err := os.WriteFile(brokenFile, []byte("this is invalid starlark syntax\n"), 0600)
+	if err != nil {
+		t.Fatalf("failed to write test file: %v", err)
+	}
+
+	cmd := CmdTest()
+	cmd.SetArgs([]string{brokenFile})
+
+	err = cmd.Execute()
+	if err == nil {
+		t.Fatal("expected Execute() to return an error, got nil")
+	}
+
+	if err.Error() != "one or more test files failed to load" {
+		t.Errorf("unexpected error message: got %q, want 'one or more test files failed to load'", err.Error())
+	}
+}
--- a/cmd/dev/app/test/test_test.go
+++ b/cmd/dev/app/test/test_test.go
@@ -9,25 +9,85 @@ import (
 	"testing"
 )
 
-func TestCmdTest_ExitCodeOnLoadError(t *testing.T) {
+func TestCmdTest_ExitCodes(t *testing.T) {
 	t.Parallel()
 
-	tmpDir := t.TempDir()
-	brokenFile := filepath.Join(tmpDir, "broken_test.star")
-	err := os.WriteFile(brokenFile, []byte("this is invalid starlark syntax\n"), 0600)
-	if err != nil {
-		t.Fatalf("failed to write test file: %v", err)
+	tests := []struct {
+		name         string
+		files        map[string]string // filename -> content
+		args         []string          // arguments to mindev test (e.g. "--output junit")
+		wantErrMsg   string            // expected error substring, or empty if success expected
+	}{
+		{
+			name: "only malformed returns non-zero",
+			files: map[string]string{
+				"broken_test.star": "this is invalid starlark syntax\n",
+			},
+			wantErrMsg: "one or more test files failed to load",
+		},
+		{
+			name: "malformed plus passing returns non-zero",
+			files: map[string]string{
+				"broken_test.star": "this is invalid starlark syntax\n",
+				"pass_test.star":   "def test_pass():\n  assert.eq(1, 1)\n",
+			},
+			wantErrMsg: "one or more test files failed to load",
+		},
+		{
+			name: "malformed with junit output returns non-zero",
+			files: map[string]string{
+				"broken_test.star": "this is invalid starlark syntax\n",
+			},
+			args:       []string{"--output", "junit"},
+			wantErrMsg: "one or more test files failed to load",
+		},
+		{
+			name: "valid passing files return zero",
+			files: map[string]string{
+				"pass_test.star": "def test_pass():\n  assert.eq(1, 1)\n",
+			},
+			wantErrMsg: "",
+		},
+		{
+			name: "valid failing files return non-zero",
+			files: map[string]string{
+				"fail_test.star": "def test_fail():\n  assert.eq(1, 2)\n",
+			},
+			wantErrMsg: "one or more tests failed",
+		},
 	}
 
-	cmd := CmdTest()
-	cmd.SetArgs([]string{brokenFile})
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			tmpDir := t.TempDir()
+			
+			var paths []string
+			for filename, content := range tt.files {
+				filePath := filepath.Join(tmpDir, filename)
+				if err := os.WriteFile(filePath, []byte(content), 0600); err != nil {
+					t.Fatalf("failed to write test file: %v", err)
+				}
+				paths = append(paths, filePath)
+			}
 
-	err = cmd.Execute()
-	if err == nil {
-		t.Fatal("expected Execute() to return an error, got nil")
-	}
+			cmd := CmdTest()
+			cmd.SetArgs(append(paths, tt.args...))
 
-	if err.Error() != "one or more test files failed to load" {
-		t.Errorf("unexpected error message: got %q, want 'one or more test files failed to load'", err.Error())
+			err := cmd.Execute()
+			if tt.wantErrMsg != "" {
+				if err == nil {
+					t.Fatalf("expected Execute() to return an error containing %q, got nil", tt.wantErrMsg)
+				}
+				if err.Error() != tt.wantErrMsg {
+					t.Errorf("unexpected error message: got %q, want %q", err.Error(), tt.wantErrMsg)
+				}
+			} else {
+				if err != nil {
+					t.Fatalf("expected Execute() to succeed, got error: %v", err)
+				}
+			}
+		})
 	}
 }
--- a/cmd/dev/app/test/test_test.go
+++ b/cmd/dev/app/test/test_test.go
@@ -13,10 +13,10 @@ func TestCmdTest_ExitCodes(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name         string
-		files        map[string]string // filename -> content
-		args         []string          // arguments to mindev test (e.g. "--output junit")
-		wantErrMsg   string            // expected error substring, or empty if success expected
+		name       string
+		files      map[string]string // filename -> content
+		args       []string          // arguments to mindev test (e.g. "--output junit")
+		wantErrMsg string            // expected error substring, or empty if success expected
 	}{
 		{
 			name: "only malformed returns non-zero",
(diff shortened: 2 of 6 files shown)


pkg/ruletest/eval.go
```diff
@@ -56,6 +57,11 @@ func (tr *testCaseRunner) builtinEval(
 		return nil, fmt.Errorf("invalid profile argument: %w", err)
 	}
 
+	paramsMap, err := dictToGoMap(paramsDict)
```
dakshhhhh16: `dictToGoMap(nil)` returns an empty, non-nil map. This means omitting `params` now causes `RuleTypeEngine.Eval` to validate `{}` against the rule's parameter schema, whereas previously `nil` skipped parameter validation.

I reproduced this with a rule containing a required parameter: the test passes on `main` when `params` is omitted, but returns an evaluation error on this branch.

If that behavior change is intentional, could we document it and add coverage? Otherwise, can we preserve `nil` when `paramsDict` is omitted and only convert it when the caller explicitly supplies `params`?

cmd/dev/app/test/test.go
```diff
@@ -33,16 +33,18 @@ func CmdTest() *cobra.Command {
 			}
 
 			runner := ruletest.NewRunner()
+			var loadErr error
```
dakshhhhh16, line 36: Could we add regression tests for the exit-code behavior introduced here? This package currently has no tests, so the reported bug could be reintroduced without CI detecting it.

Suggested cases:

- only a malformed `.star` file returns non-zero;
- malformed plus passing files still returns non-zero;
- malformed input with `--output junit` returns non-zero;
- valid passing files continue to return zero.

I verified locally that the malformed-only case exits `0` on `main` and `1` on this branch.

pkg/ruletest/runner.go
```diff
@@ -206,7 +206,22 @@ func loadSingleRule(path string) (*minderv1.RuleType, error) {
 	defer func(c io.Closer) {
 		_ = c.Close()
 	}(closer)
-	return fileconvert.ReadResourceTyped[*minderv1.RuleType](decoder)
+
```
dakshhhhh16: There is trailing whitespace on this line. `git diff --check` reports it, and the PR's golangci-lint check currently fails with `File is not properly formatted (gci)` at this location. Please run the repository's formatting command and commit the result.

cmd/dev/app/test/test.go
```diff
@@ -33,16 +33,18 @@ func CmdTest() *cobra.Command {
...
 			results, err := runner.RunPaths(args)
 			if err != nil {
 				cmd.PrintErrf("Error(s) running tests:\n%v\n", err)
+				loadErr = errors.New("one or more test files failed to load")
 			}
 
 			if len(results) == 0 {
 				if outputFormat == "text" {
 					cmd.Printf("No tests found\n")
```
dakshhhhh16, line 45: When every supplied test file has a syntax error, this prints `No tests found` immediately after reporting the load error. That output is misleading because tests were supplied but could not be parsed.

Could we print this message only when `loadErr == nil`, or replace it with wording that distinguishes an empty test run from files that failed to load?

pkg/ruletest/eval.go
```diff
@@ -74,7 +80,7 @@ func (tr *testCaseRunner) builtinEval(
 	ctx := context.Background()
 
 	tkOpts := []tkv1.Option{tkv1.WithHandlerFunc(mockHandler.ServeHTTP)}
-	if len(mockFSMap) > 0 {
+	if mockFSDict != nil {
```
dakshhhhh16, line 86: This changes the behavior of an explicitly empty `mock_fs={}`: it now installs an empty in-memory Git filesystem and overrides the rule's normal ingester. Previously, the override happened only when the map contained files.

That may be useful, but it is unrelated to the exit-code fix and currently has no test. Could we add coverage and explain this behavior in the PR description, or move it to a separate PR?

pkg/ruletest/runner.go
```diff
@@ -206,7 +206,22 @@ func loadSingleRule(path string) (*minderv1.RuleType, error) {
...
+	rt, err := fileconvert.ReadResourceTyped[*minderv1.RuleType](decoder)
+	if err != nil {
+		return nil, err
+	}
+
+	if rt != nil {
+		if rt.Context == nil {
+			rt.Context = &minderv1.Context{}
+		}
+		if rt.Context.Project == nil || *rt.Context.Project == "" {
```
dakshhhhh16: This silently injects a synthetic project into every loaded rule type that lacks one. Since the PR description only discusses syntax-error exit codes, could you explain why this additional behavior is required and add coverage for it? Otherwise, it would be clearer to move this change into a separate PR.

dakshhhhh16 (dismissed):

dakshhhhh16:
The core exit-code fix works: I verified locally that malformed test files now produce exit status 1, including mixed valid/invalid files and JUnit output.

Before merging, I think this needs:

1. regression coverage for the stated exit-code bug;
2. resolution or documentation of the omitted-`params` compatibility change;
3. the formatting issue in `runner.go` fixed.

The `params`, empty `mock_fs`, and default project-context changes also expand the scope beyond the PR description. Please document and test them, or split them into separate PRs.

cmd/dev/app/test/test.go
evankanderson: Do we want malformed + valid starlark files to exit as zero or non-zero? 

I'm imagining the typical use-case here is similar to go tests, where you'd say "run all the tests in this directory".  I'm also imagining that this might be running in CI, with people only looking at the exit status.  In that case, what might look like new coverage or coverage improvements (i.e. test cases in one file that introduce a subtle syntax error) could actually be reducing coverage if the `test` command has zero exit status with mixed inputs.

pkg/ruletest/eval.go
evankanderson: +1 to allowing users to omit params in the (many) common cases.  Params were probably an early optimization for a few edge cases (such as number of required reviewers), but we realized that a lot of parameter flexibility makes it hard to explain what a rule _does_.

pkg/ruletest/eval.go
evankanderson: I think this is a correctness fix based on attempted usage -- there are several actual tests for ruletypes where one of the test cases is "file does not exist".  It's natural to write this as `mock_fs={}`, but that previously meant that *no* filesystem ingest occurred.

Given the current state of all the codebase, I don't think we can simply _always_ ingest a filesystem, but that would probably be a little simpler.

pkg/ruletest/eval.go
```diff
@@ -93,11 +99,9 @@ func (tr *testCaseRunner) builtinEval(
 		rte.WithCustomIngester(tk)
 	}
 
-	res, err := rte.Eval(ctx, entityProto, profileMap, nil, &stubResultSink{})
-
-	// Because Eval returns the error, we pass that error to formatEvalResult
-	// We ignore res for now as we just want the error
```
evankanderson, line 99: There's a lot of tests you've added that check the message of the failure is non-nil.

Not sure if this is the problem for it, but we should probably start exposing the data from `res` for tests before we get too much farther on testing rule results. 

This is a place where we grew things over time and it got a little haphazard.  In addition to the plain evaluation, we have (at least) a string-templated message about the error and a JSON-flavored output (sometimes rendered as EvalResultOutput).

pkg/ruletest/runner.go
evankanderson: I'm guessing this is needed in actual use to allow for flag-feature evaluation, but it's worth writing that out here.

evankanderson (commented):
A little more detail in the PR description about the motivation for these changes (I'm guessing it's https://github.com/mindersec/minder-rules-and-profiles/pull/418 from memory) helps reviewers understand the context of these changes.  😁

Also, if you highlight the **intentional** semantic differences, it's easier for us to help flush unintentional differences.

cmd/dev/app/test/test.go
evankanderson: Actually, I think this exits with one error if there is at least one failing test, and with a different error if there are tests with syntax errors but no valid test fails. 

This is feeling somewhat complex due to the multiple outputs and logic paths. (Outputs: stdout, return value/error code.  Logic paths: aggregation of errors, test reporting style.) At the least, we probably need a comment to establish the desired invariants for this code block. 

If you want to try to tackle the problem more directly, you may want to build a test-output interface (junit and plain text) so that you can call `outputter.AddResult(...)` and `outputter.AddError(...)` or what-not without needing to branch the function return logic.

cmd/dev/app/test/test.go
evankanderson, line 56: Just re-reading this now, and this is a fairly inefficient way to do output.  The following is potentially much more efficient:


```suggestion
				cmd.Println(xml.Header, string(bytes))
```

The `xml.Header + string(bytes)` mode requires generating a temporary string that is O(N) bytes long (enough for both strings) and then copying the contents to the temporary before sending it to stdout.

In general for places where you're building bigger strings, you want to mix two approaches:

1. Avoid copies where possible, particularly for larger strings (e.g. > 1KB)
2. If you do need to copy (for example, to make a byte buffer), collect a bunch of small strings and then Join() them all together in one operation, rather than tacking each small string onto a growing one (which produces an N^2 amount of copies).

You don't need to fix this in this PR, but I wanted to call out the general issue pattern, since I seem to have missed it on the previous review.

krrish175-byte:
Thanks for the reviews! I have addressed all the feedback:

1. `params` are now preserved as `nil` when omitted.
2. 'No tests found' is only printed if there are no tests AND no load errors.
3. Added table-driven tests for exit codes covering all the requested cases.
4. Reverted the synthetic project context injection since it was not needed.
5. Updated the PR description with context around the `mock_fs` fix.

pkg/ruletest/eval.go
krrish175-byte: I have updated this to preserve `nil` when `paramsDict` is omitted so we don't unnecessarily validate an empty map against the rule schema. This now matches the original behavior

cmd/dev/app/test/test.go
krrish175-byte: Agreed, it was definitely misleading. Updated the logic to only print 'No tests found' if `loadErr == nil`.

pkg/ruletest/eval.go
krrish175-byte: I agree that exposing the res data would be very useful for testing moving forward. I can tackle exposing that data in a follow-up PR to keep this one scoped to the immediate bug fixes

cmd/dev/app/test/test.go
krrish175-byte: I have added comprehensive table-driven tests in `test_test.go` to explicitly cover all these edge cases. Currently, it properly exits with a non-zero code if _any_ test files are malformed, even if other files pass

pkg/ruletest/eval.go
evankanderson: Do we have a test that uses this?

cmd/dev/app/test/test_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestCmdTest_ExitCodes(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name         string
```
evankanderson: You have a lint error here (I think too many spaces).  `go fmt` should fix it.

cmd/dev/app/test/test_test.go
```diff
@@ -0,0 +1,93 @@
...
+			},
+			args:       []string{"--output", "junit"},
+			wantErrMsg: "one or more test files failed to load",
+		},
+		{
+			name: "valid passing files return zero",
+			files: map[string]string{
+				"pass_test.star": "def test_pass():\n  assert.eq(1, 1)\n",
+			},
+			wantErrMsg: "",
```
evankanderson, line 49: Not sure if it's too "cute", but you don't need this assignment, since empty-string is the default value.

cmd/dev/app/test/test_test.go
```diff
@@ -0,0 +1,93 @@
...
+			for filename, content := range tt.files {
+				filePath := filepath.Join(tmpDir, filename)
+				if err := os.WriteFile(filePath, []byte(content), 0600); err != nil {
+					t.Fatalf("failed to write test file: %v", err)
+				}
+				paths = append(paths, filePath)
+			}
+
+			cmd := CmdTest()
+			cmd.SetArgs(append(paths, tt.args...))
```
evankanderson, line 76: Why pass `paths` and not `tmpDir`?

cmd/dev/app/test/test_test.go
```diff
@@ -0,0 +1,93 @@
...
+		args         []string          // arguments to mindev test (e.g. "--output junit")
+		wantErrMsg   string            // expected error substring, or empty if success expected
+	}{
+		{
+			name: "only malformed returns non-zero",
+			files: map[string]string{
+				"broken_test.star": "this is invalid starlark syntax\n",
+			},
+			wantErrMsg: "one or more test files failed to load",
+		},
```
evankanderson, line 27: Tiny nit since you'll be in here anyway:

Prefer to put the "normal case" tests first.  In this case, I'd probably order it as:

1. valid passing files return zero  _(Hopefully, most common)_
2. valid failing files return non-zero  _(Again, hopefully next-most common)_
3. only malformed returns non-zero  _(A single scenario rather than a combination)_
4. malformed plus passing returns non-zero  _(Still varying the file inputs)_
5. malformed with junit output returns non-zero  _(Important other code paths)_

cmd/dev/app/test/test.go
evankanderson: This may be a case where it's simpler to use a _named_ return value `func(cmd *cobra.Command, args []string) (finalErr error) {`, but I think this is okay as-is.

There's a bit of subtlety here in the use of `cmd.PrintErrf` to send syntax errors to stdout while _some_ other diagnostics go to stdout along with the JUnit output.  I'm going to let it slide for now, but something to think about when we're back in the code again.

pkg/ruletest/runner.go
```diff
@@ -206,7 +206,18 @@ func loadSingleRule(path string) (*minderv1.RuleType, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if rt != nil {
+		if rt.Context == nil {
+			rt.Context = &minderv1.Context{}
+		}
+	}
+	return rt, nil
```
evankanderson, line 220: The last comment said this was reverted, but it still seems to be in the diff.  Intentional?

pkg/ruletest/eval.go
```diff
@@ -29,13 +29,14 @@ func (tr *testCaseRunner) builtinEval(
...
 	var profileDict *starlark.Dict
+	var paramsDict *starlark.Dict
 	var mockHttpDict *starlark.Dict
 	var mockFSDict *starlark.Dict
 	var datasourcesList *starlark.List
 
 	err := starlark.UnpackArgs("eval", args, kwargs,
 		"rule", &ruleName, "entity?", &entityDict,
-		"profile?", &profileDict, "mock_http?", &mockHttpDict,
+		"profile?", &profileDict, "params?", &paramsDict, "mock_http?", &mockHttpDict,
```
evankanderson, line 39: This is worth including the in the PR description, if it's not already.  (It was probably an oversight in the first place, but that happens to all of us.)

evankanderson (commented):
One lint error that needs fixing.  The rest is just me sort of puttering around, and could be ignored.

JAORMX (dismissed):

evankanderson (approved):
The security check is complaining about a separate CVE in `/tools` that's unrelated.

PR #6639, opened 2026-07-29
Enable test framework to test rego-format ruletypes

evankanderson:
# Summary

Enables the `mindev test` command to discover and load ruletypes stored as `.rego` files in addition to the existing support for `.yaml` files.

# Testing

Updated the existing tests to migrate one (of two) ruletype definitions from `.yaml` to `.rego`.


evankanderson:
I'd like to get this in alongside #6637 for a v0.3.0 client tools release (which would include the exit-code updates for `mindev test` as well as official support for `.rego`-format ruletypes).

pkg/ruletest/runner.go
```diff
@@ -182,7 +184,12 @@ func loadRulesFromDir(dir string) (map[string]*minderv1.RuleType, error) {
 	if err != nil {
 		return nil, fmt.Errorf("globbing yaml files: %w", err)
 	}
-	for _, yf := range yamlFiles {
+	regoFiles, err := filepath.Glob(filepath.Join(dir, "*.rego"))
+	if err != nil {
+		return nil, fmt.Errorf("globbing rego files: %w", err)
+	}
+	ruleFiles := append(yamlFiles, regoFiles...)
```
krrish175-byte: Minor nit: `append(yamlFiles, regoFiles...)` may reuse `yamlFiles`'s backing array if it has excess capacity, silently clobbering it. `filepath.Glob` returns a tightly-sized slice in practice so this is safe today, but it's a fragile assumption. Consider:
```go
ruleFiles := make([]string, 0, len(yamlFiles)+len(regoFiles))
ruleFiles = append(ruleFiles, yamlFiles...)
ruleFiles = append(ruleFiles, regoFiles...)
```

pkg/ruletest/runner.go
```diff
@@ -182,7 +184,12 @@ func loadRulesFromDir(dir string) (map[string]*minderv1.RuleType, error) {
...
+	for _, yf := range ruleFiles {
 		rt, err := loadSingleRule(yf)
 		if err != nil {
 			continue // skip files that aren't valid rule types
```
krrish175-byte: This silent `continue` was fine for YAML (where non-ruletype YAML files legitimately live in directories). For `.rego` files the assumption is much stronger -- any `.rego` file in a testdata directory is almost certainly *meant* to be a ruletype, so swallowing the parse error gives no signal when metadata is malformed. At minimum a `log.Printf` or zerolog warn here would save a lot of head-scratching.

pkg/ruletest/runner.go
```diff
@@ -212,11 +219,15 @@ func loadSingleRule(path string) (*minderv1.RuleType, error) {
...
-		if rt.Context == nil {
-			rt.Context = &minderv1.Context{}
-		}
+	// Ensure that ruletypes have a project context; the Minder CLI + API tooling
+	// will automatically apply the "current project" when creating a ruletype.
+	if rt.Context == nil {
+		rt.Context = &minderv1.Context{}
 	}
+	if rt.Context.GetProject() == "" {
+		rt.Context.Project = ptr.Ptr(uuid.UUID{}.String())
```
krrish175-byte: The zero UUID is a fine sentinel, but `uuid.UUID{}.String()` is a somewhat roundabout way to write it. It constructs a value type just to call `String()`. The literal `"00000000-0000-0000-0000-000000000000"` (used in the deleted YAML file) is clearer and avoids the `uuid` import entirely:
```go
rt.Context.Project = ptr.Ptr("00000000-0000-0000-0000-000000000000")
```

pkg/ruletest/runner.go
```diff
@@ -212,11 +219,15 @@ func loadSingleRule(path string) (*minderv1.RuleType, error) {
 		return nil, err
 	}
 
-	if rt != nil {
-		if rt.Context == nil {
-			rt.Context = &minderv1.Context{}
-		}
+	// Ensure that ruletypes have a project context; the Minder CLI + API tooling
+	// will automatically apply the "current project" when creating a ruletype.
+	if rt.Context == nil {
```
krrish175-byte, line 227: The previous code guarded this with `if rt != nil`. `ReadResourceTyped` can't actually return `(nil, nil)` today (it always returns a pointer), so removing the guard is technically correct -- but it's a non-obvious invariant. A short comment here would prevent someone from reflexively restoring the guard in a future cleanup:
```go
// rt is never nil when err is nil; ReadResource always returns a pointer value.
```

pkg/ruletest/testdata/mock_datasource_rule.rego
```diff
@@ -0,0 +1,35 @@
+# SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# METADATA
+#
```
krrish175-byte, line 5: Trailing space on a blank comment line (`# ` instead of `#`). The parser handles it fine, but OPA convention (and most editors) use a bare `#` for blank comment lines. Worth trimming.

krrish175-byte (commented):
One required change (stale doc comment) and one worth discussing (silent rego parse errors). Everything else is minor polish.

pkg/ruletest/runner.go
evankanderson: I'm trying to figure out if there's a place where the comment isn't "obvious" -- it's a fairly standard Go practice when returning `(type, error)` to have at least one of the two be non-nil.  `ReadResourceTyped` could return a non-pointer type, except that protobuf intentionally prevents accidental copying of values, so non-pointer proto objects can't be returned.

I guess I'll add that comment to `ReadResource` and `ReadResourceTyped`?

pkg/ruletest/runner.go
evankanderson: `"00000000-0000-0000-0000-000000000000"` is 38 characters vs `uuid.UUID{}.String()` 18 for the other way to write it, but your point is taken.  😁

(And yes, I did consider both, and picked the one with fewer characters, mostly because I felt lazy about counting the number of zeros.)

pkg/ruletest/runner.go
evankanderson: Renamed `yamlFiles` to `ruleFiles` to make it clear that I'm intentionally reusing the storage, since `yamlFiles` is a temporary that's not used after this point.

pkg/ruletest/runner.go
evankanderson: Alas!  I was too clever.  Thank you!

Now I'm trying to figure out where the boundary on this error reporting should be (e.g. is `type: rule-type` reportable?), but I think that's a task for another PR...

krrish175-byte (approved):
Two minor nits still open (trailing space on line 5 of `mock_datasource_rule.rego` and the stale doc comment on `loadRulesFromDir`) but neither blocks merge. LGTM.

PR #6666, opened 2026-07-31
feat(ruletest): Expose rule evaluation output to Starlark tests

krrish175-byte:
Currently, the rule testing framework evaluates rules and ignores the `EvaluationResult` output returned by the engine. This means developers writing tests only have access to a pass/fail/skip status and a string error message, making it difficult to debug why a rule failed, especially when dealing with complex JSON payloads (like those from jq or rego).\n\nThis PR exposes the `res.Output` data directly to the Starlark testing context under the key `'output'` so users can assert against it and view it.\n\nThis PR addresses feedback from Evan in PR #6637 and supports the testing infrastructure epic #6434.

### DIFF
--- a/pkg/ruletest/testdata/output.star
+++ b/pkg/ruletest/testdata/output.star
@@ -1,0 +1,31 @@
+# SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# These tests demonstrate that res["output"] is populated from the rule's rego
+# `output` expression, allowing test authors to assert on structured evaluation
+# data rather than just status/message.
+
+def test_output_present_on_fail():
+    res = eval(
+        rule="check_required_setting",
+        entity={"owner": "test", "name": "repo"},
+        profile={"required_value": "enabled"},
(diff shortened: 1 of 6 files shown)


evankanderson (commented):
Can you add an integration (`*.star`) test for this as well?  It will also serve as interim documentation. 😁

pkg/ruletest/testdata/output.star
```diff
@@ -0,0 +1,31 @@
...
+    res = eval(
+        rule="check_required_setting",
+        entity={"owner": "test", "name": "repo"},
+        profile={"required_value": "enabled"},
+        mock_http={
+            "/repos/test/repo/settings": body('{"value": "enabled"}')
+        }
+    )
+    assert.eq(res["status"], "pass")
+    assert.true("output" not in res)
```
evankanderson, line 31: That's interesting (not saying it's wrong) -- I'd expect that `output` would be present for both passing and failing rules, per the desired behavior in #6105

evankanderson (approved):
I want to investigate the output behavior, but that's not blocking this PR.  Approving.

Issue #6650, opened 2026-07-30
Add provider-trait gating for rule types (`provider_traits`)


intelligent-ears:
### Problem
Rule types have no declared, up-front way to say "I require an implementation of API Y." Today, provider-specific scoping either:
- doesn't exist (rule type just assumes a shape and may fail/behave oddly against a provider that doesn't implement the API it needs), or
- gets pushed into CEL profile selectors, which evaluate per-entity and return `ErrEvaluationSkipped` on mismatch. This isn't silent — it shows up in evaluation status/history — but it does mean every non-matching entity generates skip noise instead of the rule type simply not applying.

This is currently blocking `osps-ac-03-02` (branch deletion protection), since GitHub and GitLab branch-protection semantics/APIs differ enough that a single generic rule type can't cleanly cover both.

### Framing (important distinction)
The field declares "this rule type requires API Y," not "this rule type is for provider class X." In some cases (git, OCI) API Y is standardized and multiple providers can implement it. In other cases the API is specific to a single company/provider class (GitHub management API, Docker management API, various cloud provider APIs) — that looks like "this rule is for provider class X," but it's really "only this provider
currently implements API Y." The distinction matters for how the field should be understood and extended as more providers implement overlapping APIs.

### Proposed direction
Add a `provider_traits` field to the rule type schema, checked by rtengine before evaluation — the same way entity-type gating already works today — rather than relying on profile selectors. This reuses the existing `ProviderType`/`CanImplement` trait vocabulary (`pkg/providers/v1/providers.go`), so a rule type declares which trait(s) it needs (e.g. `github-api`, `git`, `oci`), not a specific named provider registration — a single rule type can apply across multiple provider instances that implement the same trait.

### Sub-issues
- [x] Schema: add `provider_traits` field to rule type proto/schema
- [x] Engine: rtengine check for `provider_traits` before eval
- [ ] Apply `provider_traits` to unblock `osps-ac-03-02`
- [x] Docs: document `provider_traits` for rule type authors

### Context
Design discussed with @evankanderson — reusing the existing `ProviderType`/`CanImplement` trait pattern rather than introducing a new capability abstraction, since GitHub/GitLab branch protection differ enough in shape that a boolean "supports X" capability flag would hide more than it clarifies. Naming settled on `provider_traits` over `required_traits` since the latter doesn't make clear who/what is doing the requiring (rule type vs. entity vs. policy). Open to a better name before this ships if one comes up.

evankanderson:
[...]

PR #6669, opened 2026-08-02
Add provider_trait field to RuleType.Definition

### LINKED ISSUE #6651: rule types: add `provider_traits` field to schema
Add a `provider_traits` field to the rule type definition (proto + YAML schema), listing the `ProviderType` trait(s) (e.g. `github-api`, `git`, `oci`) a rule type requires an implementation of. This mirrors the existing `supported_provider_types` pattern on `ProviderClassInfo`, but scoped to rule types rather than provider classes.

Field declares an API/interface requirement, not a specific provider class or named provider registration — see framing note in the parent issue.

**Open questions:**
- Single value or list (can a rule type require *any of* multiple traits)?
- Final field name — `provider_traits` is the current best option, open to revisiting before this is exposed/used.

intelligent-ears:
# Summary

Adds a `provider_trait` field to `RuleType.Definition`, letting a rule type declare an API/interface requirement (e.g. `github`, `git`, `oci`) rather than being pinned to a specific named provider registration. This reuses the existing `ProviderType` / `CanImplement` trait vocabulary already used elsewhere in the provider system, instead of introducing a
new capability abstraction.

This is currently blocking `osps-ac-03-02` (branch deletion protection), since GitHub and GitLab branch-protection semantics/APIs differ enough that a single generic rule type can't cleanly cover both. This PR lays the schema + validation groundwork; applying it to unblock `osps-ac-03-02` is a separate follow-up in `minder-rules-and-profiles`.

## Changes
- Added `provider_trait` (`ProviderType`, field 8) to `RuleType.Definition` in the proto, and regenerated bindings    (`pkg/api/protobuf/go/minder/v1/minder.pb.go`, `pkg/api/openapi/minder/v1/minder.swagger.json`, `docs/docs/ref/proto.mdx`).
- Wired validation into `pkg/profiles/validator.go`'s `validateEntities`,using `qtx.ListProvidersByProjectID` + `db.Provider.CanImplement` to check whether any provider in the project satisfies the rule type's declared trait. This follows the same static, up-front pattern as the existing `in_entity` check, rather than the noisier `ErrEvaluationSkipped` per-entity skip path used by profile selectors.
- Field is intentionally singular (not a list) — no existing rule type needs "trait A OR trait B"; rule types are already split per-provider (`github/` vs `gitlab/` in `minder-rules-and-profiles`) where the APIs genuinely diverge.

## Testing
Added unit tests in `pkg/profiles/validator_test.go` covering:
- A rule type's `provider_trait` is satisfied by a provider in the project → validation passes.
- No provider in the project satisfies the declared trait → validation fails with a clear error.
- Project has no providers at all, but the rule type requires a trait → validation fails.
- All pre-existing test scenarios (using rule types with an unset `provider_trait`) continue to pass unmodified, confirming the default (`PROVIDER_TYPE_UNSPECIFIED`) is not gated.

Ran: `go build ./pkg/profiles/...` and
`go test ./pkg/profiles/... -run TestValidatorScenarios -v` — all 16 scenarios pass.

## Scope
This covers schema + validation + tests only. Follow-ups:
- rtengine-level gating, if further design discussion determines it's needed beyond static profile validation
- Applying `provider_trait` to `osps-ac-03-02`
- Docs for rule type authors

Part of #6650, closes #6651


JAORMX (dismissed):

JAORMX:
seems there are linter and unit test issues to solve

pkg/profiles/validator.go
```diff
@@ -21,6 +21,7 @@ import (
 	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
 	"github.com/mindersec/minder/pkg/engine/selectors"
 	"github.com/mindersec/minder/pkg/ruletypes"
+	"github.com/mindersec/minder/internal/providers"
```
krrish175-byte: The `internal/providers` import needs to be placed above the `pkg/` imports in this block. `gci` expects module-local `internal/` packages to come before `pkg/` packages, alphabetically, and the current ordering is what's tripping up the lint CI check. Moving it up should be a one-line fix.

krrish175-byte (dismissed):
The design here is genuinely well thought out. Reusing the existing `ProviderType`/`CanImplement` vocabulary instead of spinning up a new capability abstraction is the right call, and doing the check up front at profile validation time (matching the `in_entity` pattern) is much cleaner than reaching for `ErrEvaluationSkipped`. The test coverage covers the meaningful cases.

One thing blocking merge right now: the `internal/providers` import in `validator.go` is in the wrong group, which is causing the `gci` lint failure in CI. Left an inline comment with details. It's a one-line fix.

Separately, the unit test failure in `TestHandleGitHubWebHook/create` ("no event received after waiting 10ms") is in a completely different package and looks like a pre-existing flaky test rather than a regression from this PR. Worth getting a clean re-run after fixing the import to confirm nothing else is red.

proto/minder/v1/minder.proto
```diff
@@ -2651,6 +2651,16 @@ message RuleType {
...
 
+        // provider_trait declares the provider trait this rule type
+        // requires an implementation of (e.g. "github", "git", "oci").
+        // This is an API/interface requirement, not a pin to a specific
+        // named provider registration: any provider whose implementation
+        // satisfies this trait (see ProviderType / CanImplement) may
+        // evaluate this rule type. If PROVIDER_TYPE_UNSPECIFIED (the
+        // zero value / default), the rule type is not gated by
+        // provider trait.
+        ProviderType provider_trait = 8;
```
evankanderson: Given the presence of datasources which allow pulling from additional APIs (e.g. `git` and `github` or `git` and `gitlab`), let's make this a list -- see my last comments on #6651 for a couple examples that check both repo contents and forge-specific APIs.

pkg/profiles/validator.go
```diff
@@ -314,6 +322,30 @@ func validateEntities(
...
+
+			if !satisfied {
+				return &RuleValidationError{
+					Err: fmt.Sprintf(
+						"rule type %s requires provider trait %s, but no provider in this project implements it",
+						ruleTypePB.Name, trait.String()),
+					RuleType: ruleTypePB.Name,
+				}
+			}
+		}
```
evankanderson: I'm not sure we want to block ruletype creation / management based on the set of currently-registered providers.

Two concerns about this semantic:

1. It prevents loading a set of ruletypes first into an empty-ish project, then adding resources to be managed by those ruletypes.  I'm not sure how common this is, but I can certainly imagine it for both automated testing and consulting-type roles.
2. As implemented, this protection is incomplete -- if the provider is later un-registered, the ruletype (and any profiles including it) will still remain.  Unwinding that pile of policy seems surprising, so I'd probably steer away from the semantic that "you must currently have providers registered matching a ruletype's `provider_traits` to manage that ruletype".

I think a _warning_ when editing a profile could be appropriate, but you'd need to add a bunch of plumbing to support collecting and returning warnings in addition to the rule execution (unlike errors, which can early-return in many cases).

evankanderson: With that said, I think we **do** want a validation that the `provider_traits` are in the set registered by Minder.

This should happen when the ruletype is created or updated, not when it is added to a profile.

evankanderson (commented):
This is a good start, but a couple comments for when you're back from break.

intelligent-ears:
Thanks both — makes sense on all counts.

@krrish175-byte the import ordering is fixed locally (gci), will push once I'm back.

@evankanderson agreed on both:
- Switching back to a list (`repeated ProviderType provider_traits`), but for AND semantics this time given datasources needing e.g. git + forge-specific API together — different case from the OR-across-vendors question we closed out on the issue.
- Agreed the current profile-time check against currently-registered providers has real problems (blocks preloading rule types, goes stale if a provider is removed). Will move validation to rule type create/update time instead, checking declared traits against the valid ProviderType set rather than the project's current providers, and drop the ListProvidersByProjectID check from validator.go entirely.

I'm on a break this week so will push the rework once I'm back — wanted to confirm I've got the plan right before I dig in. Will follow up on #6651 too.

pkg/profiles/validator.go
intelligent-ears: Done

pkg/ruletypes/provider_traits_test.go
evankanderson, line 1: Generally, I'd suggest adding these tests to `service_test.go`, rather than creating a new file.  It's not a big deal, but it makes it slightly easier to find the tests for code in `service.go` if you know it's generally in `service_test.go`.  (We do break this rule sometimes, particularly for testing code transitively through other functions.)

evankanderson (approved):

PR #6682, opened 2026-08-05
feat(ruletest): expose eval output to Starlark tests and fix CWD-sensitive path resolution

### LINKED ISSUE #6672: mindev test is sensitive to $CWD
### Describe the issue

I just noticed that (`cd pkg/ruletest; mindev test testdata`) passes, but `mindev test pkg/ruletest/testadata` fails.

Tests should locate other resources (such as datasource definitions) relative to the directory of the test file, not the current working directory of the `mindev` command.  (This will make tests easier to run and relocate reliably.)

### To Reproduce

_No response_

v0.3.0

krrish175-byte:
### Summary

This PR makes two improvements to the `mindev test` / `ruletest` framework:

**1. Expose rule evaluation output to Starlark tests** (`feat`)

The `eval()` builtin now returns an `output` key in its result dict, containing the structured output produced by the rule evaluator. This lets test authors assert on the data returned by rules, not just pass/fail status.

The `goToStarlarkValue` helper (which converts arbitrary Go values to Starlark values) has been extracted into `starlark_util.go` with its own unit tests.

**2. Fix CWD-sensitive datasource path resolution** (`fix`)

`mindev test pkg/ruletest/testdata` failed when run from the project root, while `(cd pkg/ruletest && mindev test testdata)` passed. Paths supplied via `data_sources=` are now resolved relative to the directory containing the `.star` test file rather than the process working directory.

Fixes #6672

### Testing:

- `go test ./pkg/ruletest/...` passes.
- New integration tests in `testdata/output.star` assert the `output` field
  is populated correctly for a passing and failing evaluation.
- `mock_datasources.star` updated to use bare filenames
  (`mock_datasource_def.yaml`) confirming relative-path resolution works.

### DIFF
--- a/pkg/ruletest/eval_test.go
+++ b/pkg/ruletest/eval_test.go
@@ -19,12 +19,23 @@ func TestFormatEvalResult(t *testing.T) {
 
 	tests := []struct {
 		name    string
+		res     *interfaces.EvaluationResult
 		err     error
 		wantSt  string
 		wantMsg string
+		wantOut any
 	}{
 		{
+			name:    "pass on nil error with output",
+			res:     &interfaces.EvaluationResult{Output: map[string]any{"foo": "bar"}},
+			err:     nil,
+			wantSt:  "pass",
+			wantMsg: "",
+			wantOut: map[string]any{"foo": "bar"},
+		},
+		{
 			name:    "pass on nil error",
+			res:     nil,
 			err:     nil,
 			wantSt:  "pass",
 			wantMsg: "",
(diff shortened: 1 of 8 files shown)


pkg/ruletest/eval_test.go
```diff
@@ -19,12 +19,23 @@ func TestFormatEvalResult(t *testing.T) {
...
+			name:    "pass on nil error with output",
+			res:     &interfaces.EvaluationResult{Output: map[string]any{"foo": "bar"}},
+			err:     nil,
+			wantSt:  "pass",
+			wantMsg: "",
+			wantOut: map[string]any{"foo": "bar"},
+		},
 		{
 			name:    "pass on nil error",
+			res:     nil,
```
evankanderson, line 38: FWIW (for what it's worth), you don't need to `nil`-initialize these test fields explicitly if you find it easier to read the short version.

pkg/ruletest/runner.go
```diff
@@ -116,7 +120,7 @@ func (r *Runner) RunFile(filename string, src any, ruleTypes map[string]*minderv
 	fileSystem := os.DirFS(baseDir)
 
 	name := filepath.Base(filename)
-	tr := r.newTestCaseRunner(name, fileSystem, ruleTypes)
+	tr := r.newTestCaseRunner(name, fileSystem, baseDir, ruleTypes)
```
evankanderson, line 123: Given that `fileSystem` is rooted at `baseDir`, could we use that to read the datasource definition?

evankanderson: Answer: no, because e.g. in minder-rules-and-profiles, the datasource definitions are several directories "up" from the ruletype definitions in an adjacent directory.  It's okay to break out of the test root for this case, because the datasources really are separate resources.

evankanderson (approved):

PR #6684, opened 2026-08-07
Retry remediations after transient errors

### LINKED ISSUE #4949: Remediation status does not change after being "error"
### Describe the issue

When a rule remediation fails, the remediation status is set to "error". If the cause of the error is fixed, then the remediation status should change, but it stays as "error".

For example, I had the rule type `scorecard_enabled` and initially it failed because I hadn't set a field that was required by the remediation. This was the YAML output:
```
      remediationDetails: 'cannot get PR remediation params: cannot create PR entries: cannot execute content template (index 0): template: Content[0]:26:36: executing "Content[0]" at <.Profile.sarif_file>: map has no entry for key "sarif_file"'
      remediationLastUpdated: "2024-11-12T12:41:55.941493Z"
      remediationStatus: error
```

Then, I set the `sarif_file` and the new YAML output was:
```
      remediationDetails: 'generic remediation error status: error'
      remediationLastUpdated: "2024-11-12T12:42:45.319731Z"
      remediationStatus: error
```

The same rule configuration does not produce a remediation error if applied from the start (i.e if there wasn't already an error in remediation)

### To Reproduce

_No response_

_No response_

maxi-maxima:
## Summary

- retry remediations whose previous attempt ended in an error after a fixed five-minute cooloff
- suppress immediate retry loops while leaving skipped, pending, and completed remediation behavior unchanged
- count suppressed retries with the `actions.remediation_cooloff_hits` OpenTelemetry counter
- cover both sides of the cooloff boundary with deterministic decision-table tests

## Root cause

`shouldRemediate` only returned `ActionCmdOn` when the previous remediation status was `skipped`. A transient or corrected remediation error therefore left the rule stuck in `error`. Retrying immediately, however, could busy-loop when remediation-triggered changes caused another evaluation.

The decision now uses the persisted `RemLastUpdated` timestamp: a failed evaluation inside the five-minute cooloff remains a no-op, while the boundary and later evaluations retry.

## Validation

- `go test ./internal/engine/actions`
- `go vet ./internal/engine/actions`
- `gofmt -d internal/engine/actions/actions.go internal/engine/actions/actions_test.go internal/engine/actions/types.go` (no output)

Closes #4949

### DIFF
--- a/internal/engine/actions/actions.go
+++ b/internal/engine/actions/actions.go
@@ -11,8 +11,12 @@ import (
 	"errors"
 	"fmt"
 	"runtime/debug"
+	"sync"
+	"time"
 
 	"github.com/rs/zerolog"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/metric"
 	"google.golang.org/protobuf/reflect/protoreflect"
 
 	"github.com/mindersec/minder/internal/engine/actions/alert"
@@ -24,6 +28,13 @@ import (
 	"github.com/mindersec/minder/pkg/engine/v1/interfaces"
 	"github.com/mindersec/minder/pkg/profiles/models"
 	provinfv1 "github.com/mindersec/minder/pkg/providers/v1"
+)
+
+const remediationRetryCooloff = 5 * time.Minute
+
+var (
+	remediationCooloffHits metric.Int64Counter
+	actionMetricsInit      sync.Once
 )
 
 // RuleActionsEngine is the engine responsible for processing all actions i.e., remediation and alerts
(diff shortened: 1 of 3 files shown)


evankanderson (commented):
This looks mostly right, but I'm concerned about the possibility of "remediation loops" mentioned in the previous comment.  Is there a way that we could either track the number of repeated errors (e.g. stop attempting after 3 error remediations) or implement some sort of exponential backoff on remediation attempts so that we don't end up potentially busy-looping on an error state which triggers enough of a change to cause re-evaluation, but not enough to resolve the state?

evankanderson (approved):

PR #6690, opened 2026-08-11
feat(mindev): add --junit-file flag to test command and add Starlark testing docs

krrish175-byte:
Addresses feedback to support writing JUnit XML to a specific file while preserving standard terminal output. Also adds Starlark testing documentation to the mindev how-to docs.


docs/docs/how-to/mindev.md
```diff
@@ -193,6 +193,73 @@ For more information on the rego print statement, the following blog post is a
 good resource:
 [Introducing the OPA print function](https://blog.openpolicyagent.org/introducing-the-opa-print-function-809da6a13aee)
 
+## Testing with Starlark
+
+In addition to evaluating rules against a single entity, `mindev test` allows you to run comprehensive Starlark-based tests against your rule types. This is the recommended way to verify rule behavior, as it lets you mock external systems like GitHub REST/GraphQL APIs and Git filesystems.
```
evankanderson: Two comments:

1. It's probably worth linking to the starlark language documentation, so people can understand what's supported.
2. I'd make a stronger statement than "recommended way" -- something like "`mindev test` is the supported mechanism for verifying rule behavior, and is designed for integration into CI/CD pipelines as well as standalone usage".

docs/docs/how-to/mindev.md
evankanderson, line 1: I think it's fine to add this usage here, but we should think about a standalone page describing "how to test rules".

This would cover both manual and automated (CI) testing, including referencing the action and examples of setup, as well as some hints on effective testing usage, and documentation of the custom functions added to the starlark runtime.

docs/docs/how-to/mindev.md
```diff
@@ -193,6 +193,73 @@ For more information on the rego print statement, the following blog post is a
...
+
+    # 3. Call the `eval` builtin to run the rule
+    result = eval(
+        rule="rule-types/github/branch_protection.yaml",
+        entity=entity,
+        mock_http=mock_http
+    )
+
+    # 4. Assert the result
+    assert result["status"] == "success", "Expected rule to pass"
```
evankanderson: In the expanded page, we should include information about what entries are in the "result" object, as well as the input arguments to "eval".

docs/docs/how-to/mindev.md
```diff
@@ -193,6 +193,73 @@ For more information on the rego print statement, the following blog post is a
...
+
+def test_failure_case():
+    # ... mock the API to return invalid configuration ...
+    result = eval(...)
+    assert result["status"] == "error", "Expected rule to fail"
+```
+
+The `eval` builtin takes several optional arguments:
```
evankanderson: Some of these arguments are not optional (at a minimum, the "rule" argument, and maybe the "entity" argument).

evankanderson (dismissed):
I'm probably going to hold off on cutting a release until we figure out if there are any other `mindev` changes you'll want for tests.

evankanderson (approved):
Looks good, thank you!

PR #6700, opened 2026-08-14
Make REST data source limits configurable

maxi-maxima:
# Summary

REST data source calls currently hard-code a five-second timeout in the HTTP client and a one-MiB response cap in the handler. That prevents server administrators from tuning these denial-of-service safeguards, and provider-backed REST calls do not receive the hard-coded client timeout.

This change:

- adds server-level `data_sources.rest.request_timeout` and `max_response_bytes` settings
- preserves the existing `5s` and `1048576` defaults
- wires the settings through the data source service into REST handlers
- applies the timeout through the request context so direct and provider-backed calls share the same limit
- adds regression coverage for config decoding/defaults, request deadlines, and response truncation

No new dependencies are required.

Related to #6275. This implements the maintainer-recommended server-controlled duration and payload-limit scope; authentication and fallback remain out of scope.

# Testing

- `go test ./internal/datasources/rest -run '^TestNewRestDataSourceAppliesServerLimits$' -count=1`
- `go test ./pkg/config/server -run '^(TestReadValidConfig|TestReadDefaultConfig)$' -count=1`
- `go test ./internal/datasources/service -count=1`
- `go vet ./internal/datasources/rest ./pkg/config/server ./internal/datasources/service ./internal/service`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --new-from-rev=origin/main --timeout=5m` — 0 issues
- `go mod tidy` — no content changes

The combined target-package test run also reaches an existing unrelated assertion in `Test_restHandler_HTTPCall/Invalid_URL`: the current URI template dependency reports column 36 while the unchanged test expects column 37. The other target packages pass, and the new REST limit regression test passes independently.

### DIFF
--- a/internal/datasources/rest/handler.go
+++ b/internal/datasources/rest/handler.go
@@ -35,6 +35,8 @@ import (
 )
 
 const (
+	// DefaultRequestTimeout is the default timeout for REST data source requests.
+	DefaultRequestTimeout = 5 * time.Second
 	// MaxBytesLimit is the maximum number of bytes to read from the response body
 	// We limit to 1MB to prevent abuse
 	MaxBytesLimit int64 = 1 << 20
@@ -54,10 +56,12 @@ type restHandler struct {
 	// used only to allow requests to localhost during tests
 	testOnlyTransport http.RoundTripper
 	// contains the request body or the key
-	body          string
-	bodyFromInput bool
-	headers       map[string]string
-	parse         string
+	body             string
+	bodyFromInput    bool
+	headers          map[string]string
+	parse            string
+	requestTimeout   time.Duration
+	maxResponseBytes int64
 	// TODO implement fallback
 	// TODO implement auth
 	provider interfaces.RESTProvider
@@ -82,6 +86,8 @@ func newHandlerFromDef(
 	def *minderv1.RestDataSource_Def,
 	provider provinfv1.Provider,
 	testOnlyTransport http.RoundTripper,
+	requestTimeout time.Duration,
+	maxResponseBytes int64,
 ) (*restHandler, error) {
 	if def == nil {
 		return nil, errors.New("rest data source handler definition is nil")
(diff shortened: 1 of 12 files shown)


pkg/config/server/datasources.go
```diff
@@ -0,0 +1,17 @@
...
+
+// DataSourceConfig contains server-controlled data source limits.
+type DataSourceConfig struct {
+	REST RESTDataSourceConfig `mapstructure:"rest"`
+}
+
+// RESTDataSourceConfig contains resource limits for REST data source requests.
+type RESTDataSourceConfig struct {
+	RequestTimeout   time.Duration `mapstructure:"request_timeout" default:"5s"`
+	MaxResponseBytes int64         `mapstructure:"max_response_bytes" default:"1048576"`
```
evankanderson: Can you add an accessor, switch this to a string, and use [`github.com/dustin/go-humanize`](https://pkg.go.dev/github.com/dustin/go-humanize) to parse?  It's much easier reading `1MiB` than `1048576`.

I'm okay with either a "fail to start" (big, noticeable error) or "Log error message and continue" for handling incomprehensible sizes.

internal/datasources/rest/handler.go
```diff
@@ -35,6 +35,8 @@ import (
 )
 
 const (
+	// DefaultRequestTimeout is the default timeout for REST data source requests.
+	DefaultRequestTimeout = 5 * time.Second
```
evankanderson, line 39: Should we have some way to keep this in sync with the value in `config.go`?  (Or default to 0s in the config and have zero mean "whatever the code says"?)

evankanderson (commented):
Very nice!  A couple small thoughts / questions, but this makes sense overall.

pkg/config/server/datasources.go
maxi-maxima: Addressed in `68632215e`. `max_response_bytes` is now a human-readable string (the example uses `1MiB`) and `GetMaxResponseBytes` parses it with `go-humanize`. Invalid or overflowing values now return a startup error from `AllInOneServerService`. The regression table covers empty/default, IEC, SI, invalid, and overflow cases.

internal/datasources/rest/handler.go
maxi-maxima: Addressed in `68632215e` by making the config defaults `0s`/empty. Those zero values flow through to the existing REST code fallbacks (`DefaultRequestTimeout` and `MaxBytesLimit`), so there is now one source of truth for each default. The example config remains explicit at `5s` and `1MiB`.

evankanderson (approved):

PR #6701, opened 2026-08-15
Gate rule type engine construction on provider_traits

### LINKED ISSUE #6652: rtengine: gate rule evaluation on `provider_traits`
Before evaluating a rule type against an entity, rtengine should check whether the entity's provider implements the trait(s) listed in the rule type's `provider_traits` (if set), the same way it already gates on entity type. If the provider doesn't implement the required trait, the rule type should not run against that entity — analogous to how entity-type mismatches are handled today, not via `ErrEvaluationSkipped`.

Depends on: schema field from sub-issue 1.

intelligent-ears:
## Summary

Widens the gating logic in `NewRuleTypeEngine` so a rule type carrying `provider_traits` (added in #6669) is rejected up front — at engine construction time — if the provider doesn't implement every listed trait. This mirrors the existing pattern used for entity-context mismatches (reject at construction, not via a runtime `ErrEvaluationSkipped`), per the issue.

The trait check is a package-local interface in `pkg/engine/v1/rtengine/engine.go`:

```go
type traitProvider interface {
}
```
rather than a new method on pkg/engine/v1/interfaces.Provider. The original design called for widening interfaces.Provider directly, but that creates an import cycle: pkg/engine/v1/interfaces can't import minderv1 because minderv1 (via its hand-written datasources.go) imports pkg/datasources/v1, which imports pkg/engine/v1/interfaces. Declaring traitProvider locally in
rtengine and type-asserting against it avoids the cycle; every concrete provider already implements CanImplement with this exact signature, so they satisfy it structurally with zero provider-side changes.

pkg/testkit/v1.TestKit.CanImplement was hardcoded to always return true; added a WithCanImplement functional option so tests can control it per case, defaulting to true so existing callers (including pkg/ruletest/eval.go) are unaffected.

Part of #6650, closes #6652.

Test plan

- [ ] go build ./pkg/engine/... ./internal/providers/...
- [ ] go test ./pkg/engine/v1/rtengine/... -v
- [ ] gofmt -l on all touched files (clean)

### DIFF
--- a/pkg/engine/v1/interfaces/provider.go
+++ b/pkg/engine/v1/interfaces/provider.go
@@ -14,6 +14,14 @@ import (
 
 // Provider is a slice of the github.com/mindersec/minder/pkg/providers/v1.Provider
 // interface which contains only the methods needed for engine evaluation. (currently none)
+//
+// Note: this interface intentionally does not depend on
+// github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1 (minderv1). That
+// package transitively imports pkg/datasources/v1, which imports this
+// package for interfaces.Ingested — so importing minderv1 here would create
+// an import cycle. Callers that need a minderv1-typed method (e.g.
+// CanImplement) should type-assert against a package-local interface
+// instead, the way rtengine does.
 type Provider interface {
 }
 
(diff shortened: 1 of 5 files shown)


intelligent-ears:
closing this as, just checked `go test ./pkg/engine/v1/rtengine/... -v` is failing

PR #6702, opened 2026-08-15
Gate rule type engine construction on provider_traits

### LINKED ISSUE #6652: rtengine: gate rule evaluation on `provider_traits`

intelligent-ears:
## Summary

Widens the gating logic in `NewRuleTypeEngine` so a rule type carrying `provider_traits` (added in #6669) is rejected up front — at engine construction time — if the provider doesn't implement every listed trait. This mirrors the existing pattern used for entity-context mismatches (reject at construction, not via a runtime `ErrEvaluationSkipped`), per the issue.

The trait check is a package-local interface in `pkg/engine/v1/rtengine/engine.go`:

```go
type traitProvider interface {
}
```
rather than a new method on pkg/engine/v1/interfaces.Provider. The original design called for widening interfaces.Provider directly, but that creates an import cycle: pkg/engine/v1/interfaces can't import minderv1 because minderv1 (via its hand-written datasources.go) imports pkg/datasources/v1, which imports pkg/engine/v1/interfaces. Declaring traitProvider locally in
rtengine and type-asserting against it avoids the cycle; every concrete provider already implements CanImplement with this exact signature, so they satisfy it structurally with zero provider-side changes.

pkg/testkit/v1.TestKit.CanImplement was hardcoded to always return true; added a WithCanImplement functional option so tests can control it per case, defaulting to true so existing callers (including pkg/ruletest/eval.go) are unaffected.

Part of #6650, closes #6652.

Test plan

- [ ] go build ./pkg/engine/... ./internal/providers/...
- [ ] go test ./pkg/engine/v1/rtengine/... -v
- [ ] gofmt -l on all touched files (clean)


pkg/engine/v1/interfaces/provider.go
```diff
@@ -14,6 +14,14 @@ import (
...
 // Provider is a slice of the github.com/mindersec/minder/pkg/providers/v1.Provider
 // interface which contains only the methods needed for engine evaluation. (currently none)
+//
+// Note: this interface intentionally does not depend on
+// github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1 (minderv1). That
+// package transitively imports pkg/datasources/v1, which imports this
+// package for interfaces.Ingested — so importing minderv1 here would create
+// an import cycle. Callers that need a minderv1-typed method (e.g.
+// CanImplement) should type-assert against a package-local interface
+// instead, the way rtengine does.
```
evankanderson: Hmm -- I feel like I'd rather have the protobuf package not depend on the `pkg/datasources/v1`, but have the dependency go the other way.

What do you think about moving the two constants to the protobuf package to avoid the loop (and then add the appropriate methods to this interface)?

pkg/engine/v1/rtengine/engine.go
```diff
@@ -65,6 +77,21 @@ func NewRuleTypeEngine(
 		return nil, fmt.Errorf("rule type context must have a project")
 	}
 
+	if traits := ruletype.GetDef().GetProviderTraits(); len(traits) > 0 {
```
evankanderson: I don't think you need the check for `len(traits) > 0`; if you simply write `range ruletype.GetDef().GetProviderTraits()`, the loop will execute 0 times if no traits were declared (and then will be no `CanImplement` checks).

pkg/engine/v1/rtengine/engine.go
```diff
@@ -65,6 +77,21 @@ func NewRuleTypeEngine(
...
+		tp, ok := provider.(traitProvider)
+		if !ok {
+			return nil, fmt.Errorf(
+				"provider does not support trait checks required by rule type %s", ruletype.Name)
+		}
+		for _, trait := range traits {
+			if !tp.CanImplement(trait) {
+				return nil, fmt.Errorf(
+					"provider does not implement required trait %s for rule type %s",
+					trait.String(), ruletype.Name)
```
evankanderson: What happens in a running Minder instance if this returns error repeatedly?  I suspect that's not a common case that we expect, and might cause looping or other evaluation problems.

Can you test the following:

A profile with three ruletypes:
1. The first declares no `provider_traits`
2. The second declares `provider_traits` that aren't satisfied by any provider (e.g. add a "GitLab" trait, and require both GitHub and GitLab)
3. The third declares a `provider_trait` satisfied by the provider.

And then register at least one repository with e.g. the GitHub or GitLab provider that should evaluate the profile.

What _should_ happen as I understand it is that we should get rule evaluation results for the first and third ruletypes, and no evaluation results (not listed, not "skipped" or "error") for the second ruletype.

evankanderson (commented):
Thanks for adding support in TestKit!  Can you add some `provider_traits` to the tests in `pkg/ruletest/testdata`, as that's where the `mindev test` integration tests live.

pkg/engine/v1/interfaces/provider.go
intelligent-ears: Agreed, and it turned out to be a one-line dependency — minderv1's hand-written `datasources.go` imported `pkg/datasources/v1` only so `GetDriverType()` could return those two constants. Moved them into minderv1 (d16e01e) with deprecated aliases left behind in `pkg/datasources/v1`, since that's a public package. With the cycle gone, `interfaces.Provider` now declares `CanImplement(minderv1.ProviderType)` bool directly and the package-local `traitProvider` interface plus its explanatory comment are deleted (da95a73).

One thing to flag: `interfaces.Provider` was previously empty, so this is its first method and technically a breaking change for any out-of-tree implementer. Everything in-tree already satisfies it, including `TestKit`.

pkg/engine/v1/rtengine/engine.go
intelligent-ears: Removed in da95a73 — you're right, the getters are nil-safe and `range` over a nil slice is a no-op, so the guard wasn't buying anything.

pkg/ruletest/testdata/rule_type_sample.yaml
```diff
@@ -16,6 +16,10 @@ guidance: |
   and set the 'Required approving reviews' count to 2 or higher.
 def:
   in_entity: repository
+  # RuleType uses plain encoding/json (not protojson) for these fields, so
+  # provider_traits must be the numeric enum value, not the string name:
+  # 1 = PROVIDER_TYPE_GITHUB.
+  provider_traits: [1]
```
evankanderson: Hmm, this doesn't feel very good.  I feel like I'd rather have this be a list of string than numeric enums.

pkg/ruletest/eval.go
```diff
@@ -34,15 +35,22 @@ func (tr *testCaseRunner) builtinEval(
...
 	var mockFSDict *starlark.Dict
 	var datasourcesList *starlark.List
+	var providerMissingTraitsList *starlark.List
 
 	err := starlark.UnpackArgs("eval", args, kwargs,
 		"rule", &ruleName, "entity?", &entityDict,
 		"profile?", &profileDict, "params?", &paramsDict, "mock_http?", &mockHttpDict,
-		"mock_fs?", &mockFSDict, "data_sources?", &datasourcesList)
+		"mock_fs?", &mockFSDict, "data_sources?", &datasourcesList,
+		"provider_missing_traits?", &providerMissingTraitsList)
```
evankanderson: I think I'd rather prefer to see a positive list of "provider traits present" in the test, with the default behavior still being to have all traits present if the argument is not set.

internal/engine/executor.go
```diff
@@ -176,6 +178,13 @@ func (e *executor) evaluateRule(
 	// retrieve the rule type engine from the cache
 	ruleEngine, err := ruleEngineCache.GetRuleEngine(ctx, inf.ProjectID, rule.RuleTypeID)
 	if err != nil {
+		if errors.Is(err, rtenginev1.ErrProviderTraitNotSatisfied) {
+			// This rule type doesn't apply to this entity's provider.
+			// Produce zero evaluation-status footprint: no error, no
+			// skip record, nothing — stricter than the SkipSilently
+			// path, which still logs at info level.
+			return nil
+		}
```
evankanderson: Rather than going through the engine cache and back with an error status, what about adding the provider check right above `GetRuleEngine`, or even in `EvalEntityEvent` before calling `evaluateRule`?

evankanderson (commented):
Thanks for updating this!

I know that we use special error status to return different types of evaluation results, but this feels like a place where we can simplify to regular methods that return a boolean, rather than needing the additional complexity.

pkg/ruletest/testdata/rule_type_sample.yaml
evankanderson: I think we still have time to go back and change the proto definition (even though it will technically be incompatible), since this feature hasn't really been documented or used.

intelligent-ears: Done — `provider_traits` is now `repeated string` (SHA1). Field 8 is reserved and the name is reused for field 9, so the number can never come back with a different meaning. Testdata reads `provider_traits: [github]` and the comment apologising for numeric enums is gone.

Two consequences worth flagging.

`buf breaking` needed a narrow `FIELD_NO_DELETE_UNLESS_NAME_RESERVED` exception scoped to `minder/v1/minder.proto`. WIRE_JSON wants both the number *and* the name reserved on deletion, which conflicts with reusing `provider_traits` for field 9. Reserving the number only seemed right, and the alternative — renaming the field — seemed worse. Happy to do it differently if you'd rather.

Dropping the enum also means the compiler stops catching typos, and under the boolean check in the other thread a misspelled trait would silently make a rule type never evaluate: no error, no skipped status, no row. That's the failure mode the enum was implicitly protecting against, so SHA3 adds validation at rule type creation that rejects unknown trait names and lists the valid ones, plus a warning log in `NewRuleTypeEngine` as a fallback for rule types already stored or any path that bypasses validation.

pkg/ruletest/eval.go
intelligent-ears: Renamed to `provider_traits_present` with the logic inverted (SHA4). Default is unchanged: when the argument isn't passed, every trait is present.

While inverting it I noticed the argument was taking full enum names (`PROVIDER_TYPE_GITHUB`) while rule type YAML takes short trait names, so it now takes the same short names as the YAML field — `provider_traits_present=["git", "rest"]`. An unrecognised name errors out and lists the valid ones rather than silently parsing as absent.

One divergence to call out: the harness reports `status: "skip"` for a rule type the test provider doesn't support, whereas the executor produces nothing at all for that case. A Starlark test needs `eval()` to return something assertable, so reproducing the executor's silent no-op didn't seem workable. There's a comment on `skippedResult` explaining it, but say the word if you'd rather it surfaced differently.

internal/engine/executor.go
intelligent-ears: Agreed, this is much simpler. The sentinel is gone entirely (SHA2).

`NewRuleTypeEngine` now computes trait satisfaction once at construction and exposes `SupportedByProvider() bool`, so it no longer returns an error for something that isn't really an error. The cache's `errors.Is`/`continue` is gone and every rule type gets cached again; `evaluateRule` checks the boolean and returns before `createOrUpdateEvalStatus`, so there's still no eval status row for an inapplicable rule — not listed, not skipped, not error.

On placement: I put it on the engine rather than above `GetRuleEngine` or in `EvalEntityEvent` because the check needs the rule type's `provider_traits`, so doing it earlier means either an extra DB lookup or reaching for the engine we'd be trying to avoid building. The trade-off is that engines now get constructed for rule types that will never evaluate, so we pay validator/ingester/evaluator setup for them once per cache build. That seemed like the cheaper side of the trade, but if you'd rather pay the lookup and skip construction entirely I'm happy to move it.

pkg/engine/v1/interfaces/provider.go
evankanderson: We don't guarantee API compatibility on `interfaces.Provider`, so anyone out of tree needs to be able to deal with this sort of API breakage.  Thanks for thinking of it, though!

pkg/ruletest/testdata/rule_type_sample.yaml
evankanderson: Since we haven't released a client with this definition, I feel good about disabling or ignoring the buf check for this particular incompatibility.

I'll take a look at the implementation -- I think we should be able to (on ingest) flag typo'ed provider traits based on the **server**'s knowledge of the trait universe.  This also fixes a problem where someone using a ruletype which requires a recent trait would need a recent _client_ in order to upload the ruletype to the server.  IMO, it's better for the client to be able to treat this as opaque data, and people to be able to expect that e.g. a client from 6 months ago can work reasonably as long as the server is fairly up to date.  (This matches, for example, my usage of the `gh` CLI and other cloud clients.)

proto/buf.yaml
```diff
@@ -9,6 +9,17 @@ deps:
...
+  # new field 9 of the same name but a different type (repeated ProviderType
+  # -> repeated string). Field 8 is reserved so its number can never be
+  # reused, but WIRE_JSON also requires the *name* to be reserved on
+  # deletion, which would conflict with reusing "provider_traits" for field
+  # 9. This is a deliberate, sanctioned exception: provider_traits has never
+  # been documented or used, so no real caller depends on the old field 8
+  # wire/JSON shape.
+  ignore_only:
+    FIELD_NO_DELETE_UNLESS_NAME_RESERVED:
+      - minder/v1/minder.proto
```
evankanderson, line 27: We should be able to delete this after this PR is merged, as the buf incompatibility check only compares between the PR base and current (as it should).

pkg/ruletypes/service.go
```diff
@@ -355,15 +356,39 @@ func (s *ruleTypeService) validateAndDetectRegoVersion(ctx context.Context, rule
 	}
 }
 
+// validProviderTraitNames lists the trait names a rule type's
+// provider_traits may declare: the (name) option of every defined
+// ProviderType value.
+var validProviderTraitNames = func() []string {
```
evankanderson: I'd prefer to explicitly use an `init` method to initialize this at startup, rather than implicitly through an anonymous function evaluation.  It seems a little more clear to me.

pkg/ruletypes/service.go
```diff
@@ -355,15 +356,39 @@ func (s *ruleTypeService) validateAndDetectRegoVersion(ctx context.Context, rule
...
+// provider_traits are known trait names. This checks against the static set
+// of provider types Minder defines, not against providers currently
+// registered in any project: gating on registration would block loading
+// rule types before providers exist, and would leave rule types silently
+// stale if a provider is later unregistered.
+//
+// This validation is the only safeguard against a typo in provider_traits:
+// an unrecognized string parses fine but matches no trait, silently making
+// the rule type never evaluate (see rtengine.RuleTypeEngine.SupportedByProvider),
+// with no error, skipped status, or eval status row to reveal the mistake.
```
evankanderson: In particular, I think this means that any stored ruletype with a `provider_trait` name that is later renamed would stop evaluating?  (Because we're using the JSON-serialized proto definition stored in the `definition` row.)

I think that's acceptable, but if we do another set of commits, you might want to add that comment to the proto definition for the provider types.

evankanderson: Changing the name of the provider traits would also break all the clients using `provider_trait`, so I think it's fair to consider that a fixed-ish interface.

pkg/api/protobuf/go/minder/v1/providers.go
```diff
@@ -10,6 +10,18 @@ func (provt ProviderType) ToString() string {
 	return enumToStringViaDescriptor(provt.Descriptor(), provt.Number())
 }
 
+// ProviderTypeFromString returns the ProviderType whose (name) option
+// matches s (e.g. "github", "git"), and true if one was found. It is the
+// inverse of ProviderType.ToString.
```
evankanderson: Given the cost of iterating, I'd prefer to have this use a map generated at init time, rather than called in a loop (like we do in `pkg/ruletypes/service.go`).

pkg/engine/v1/rtengine/engine.go
```diff
@@ -65,6 +70,31 @@ func NewRuleTypeEngine(
...
+			// or any path that bypasses it, would otherwise fail this
+			// exact same way with no signal anywhere that anything is
+			// wrong. Treat it as unsupported, same as a provider that
+			// doesn't implement the trait, but log it so it's visible.
+			zerolog.Ctx(ctx).Warn().
+				Str("rule_type", ruletype.GetName()).
+				Str("provider_trait", trait).
+				Msg("rule type declares an unknown provider trait; treating as unsupported")
+			supportedByProvider = false
+			break
```
evankanderson: If we mismatch here, can we make this return an error to the user?  I'd need to think about the mechanics, but it "feels" right, since the user can fix it by adjusting the ruletype definition.

internal/engine/executor.go
```diff
@@ -178,6 +178,13 @@ func (e *executor) evaluateRule(
 	if err != nil {
 		return fmt.Errorf("error creating rule type engine: %w", err)
 	}
+	if !ruleEngine.SupportedByProvider() {
+		// This rule type doesn't apply to this entity's provider.
+		// Produce zero evaluation-status footprint: no error, no
+		// skip record, nothing — stricter than the SkipSilently
+		// path, which still logs at info level.
+		return nil
+	}
```
evankanderson: If you can cache the "unknown provider trait" in the ruleEngine, you can use it here to record the error _only_ for ruletypes that have unrecognized provider traits.

(Sorry to move the goals on you -- your observation that a completely unknown provider_trait has a silent failure mode made me look to make that loud.)

pkg/datasources/v1/datasources.go
```diff
@@ -9,16 +9,24 @@ import (
...
+	// Deprecated: use minderv1.DataSourceDriverStruct. Kept here as an alias so
+	// existing callers of this package don't break; the canonical definition
+	// lives in minderv1 to avoid an import cycle (minderv1 -> this package ->
+	// pkg/engine/v1/interfaces).
+	DataSourceDriverStruct = minderv1.DataSourceDriverStruct
 	// DataSourceDriverRest is the driver type for a REST data source.
-	DataSourceDriverRest = "rest"
+	//
+	// Deprecated: use minderv1.DataSourceDriverRest. See DataSourceDriverStruct.
+	DataSourceDriverRest = minderv1.DataSourceDriverRest
```
evankanderson: IMO, since the package is pre-1.0, it would be okay to remove these entirely.

pkg/ruletest/eval.go
```diff
@@ -162,6 +201,47 @@ func parseMockFSDict(mockFSDict *starlark.Dict) (map[string]string, error) {
...
+	for v := range minderv1.ProviderType_name {
+		t := minderv1.ProviderType(v)
+		if t == minderv1.ProviderType_PROVIDER_TYPE_UNSPECIFIED {
+			continue
+		}
+		names = append(names, t.ToString())
+	}
+	sort.Strings(names)
+	return names
+}
```
evankanderson: Given that we use this a few different places, what do you think about creating this as a map in the proto package?

evankanderson (commented):
This is looking really good, by the way!  A few more comments, but only because I'm excited to see this working.  I'll also be running an integration test for you and have results probably tomorrow morning.

proto/buf.yaml
intelligent-ears: Agreed. Updated the comment in 82b6b46 to say so explicitly rather than removing it now, since deleting it before merge turns the check red. Happy to open the follow-up to drop it once this lands, or leave it to whoever's next in that file.

pkg/ruletypes/service.go
intelligent-ears: Done in c0a0f1a. It's an explicit `init()` now, and it moved to the minderv1 package — between your comment here and the one on `pkg/ruletest/eval.go`, it was clear this list had no business existing in three places. So this local copy is gone entirely and both callers use `minderv1.ValidProviderTraitNames()`.

pkg/ruletypes/service.go
intelligent-ears: Correct — the definition is stored as serialized proto, so a renamed trait name would stop matching for anything already in the database, and your second point is the sharper one: it'd break every client constructing or parsing those names too. That makes them a public interface whether or not we intended them as one.

Added that to the `ProviderType` enum in 82b6b46: the `(name)` values are a stable interface, renaming one silently stops it matching stored rule types as well as breaking clients, so add new values rather than renaming existing ones.

The comment you're commenting on is also now out of date — this validation is no longer the *only* safeguard. c2dd129 catches an unrecognized name at evaluation time too and reports it as an eval status error, so a rule type stored before this validation existed no longer fails silently. Updated the comment to say that instead.

pkg/api/protobuf/go/minder/v1/providers.go
intelligent-ears: Done in c0a0f1a — `providerTypeByName` is built once in `init()` and this is a map lookup now. `enumFromStringViaDescriptor` is deleted since it had no other callers; `enumToStringViaDescriptor` stays, it's still used by `ToString()` on several enums.

The same `init()` also builds the sorted valid-name list, exported as `ValidProviderTraitNames()` for error messages, which let me delete the duplicate copies in `pkg/ruletypes` and `pkg/ruletest` you flagged elsewhere. It returns a copy rather than the package slice, so a caller sorting or mutating it can't affect anyone else.

pkg/engine/v1/rtengine/engine.go
intelligent-ears: Agreed, and done in c2dd129 using the mechanics you sketched in the executor thread — the engine caches the unknown names rather than trying to report from here.

`NewRuleTypeEngine` now collects every unrecognized trait name into `unknownProviderTraits` and exposes it via `UnknownProviderTraits()`. The loop no longer `break`s on the first one, so someone with two typos sees both rather than fixing one and rediscovering the other on the next run. The two conditions stay distinct instead of both collapsing into `supportedByProvider = false`, since only one of them is something the user can act on.

Dropped the warning log — it was a consolation prize for not being able to tell the user, and now we can.

internal/engine/executor.go
intelligent-ears: No apology needed — I flagged the silent failure and then left it silent, so this is the better end of that. Done in c2dd129, exactly as you described.

There are three states now rather than two:

- Unknown trait name → `evalErr` is set to an error naming the rule type and every offending trait, which flows through the existing `SetEvalErr` → `DoActions` → `createOrUpdateEvalStatus` path, so it lands as a normal eval status error. No parallel reporting mechanism.
- Known trait the provider doesn't implement → unchanged, silent early return, no row at all.
- Everything satisfied → evaluates as before.

The silent return is guarded on `len(unknownTraits) == 0 && !SupportedByProvider()`, so the two can't collapse into each other. Worth noting the error is deliberately not wrapped in `ErrEvaluationFailed` or `ErrEvaluationSkipped`, so it classifies as `EvalStatusTypesError` rather than a rule failure — the rule didn't fail, the rule type is malformed.

pkg/datasources/v1/datasources.go
intelligent-ears: Removed entirely in 82b6b46. Turned out to be free — nothing referenced the aliases at all; every caller was already using the minderv1 constants directly, so there was no migration to do and the `minderv1` import in this file went with them.

pkg/ruletest/eval.go
intelligent-ears: Yes — this was the third copy, which is two too many. c0a0f1a builds `providerTypeByName` once in an `init()` in minderv1 and exports `ValidProviderTraitNames()` alongside it; this function and the one in `pkg/ruletypes/service.go` are both gone, and `ProviderTypeFromString` is a map lookup rather than a descriptor walk.

The accessor returns a copy rather than the package slice, so a caller sorting or mutating it can't affect anyone else.

pkg/api/protobuf/go/minder/v1/providers.go
```diff
@@ -3,13 +3,69 @@
...
+			continue
+		}
+		providerTypeByName[t.ToString()] = t
+	}
+
+	validProviderTraitNames = make([]string, 0, len(providerTypeByName))
+	for name := range providerTypeByName {
+		validProviderTraitNames = append(validProviderTraitNames, name)
+	}
+	sort.Strings(validProviderTraitNames)
```
evankanderson: I think this can be:


```suggestion
	validProviderTraitNames = slices.Sorted(maps.Keys(validProviderTraitNames))
```

pkg/api/protobuf/go/minder/v1/providers.go
```diff
@@ -3,13 +3,69 @@
...
+// ProviderTypeFromString returns the ProviderType whose (name) option
+// matches s (e.g. "github", "git"), and true if one was found. It is the
+// inverse of ProviderType.ToString.
+func ProviderTypeFromString(s string) (ProviderType, bool) {
+	t, ok := providerTypeByName[s]
+	if !ok {
+		return ProviderType_PROVIDER_TYPE_UNSPECIFIED, false
+	}
+
+	return t, true
```
evankanderson: Since unspecified is both the zero (`iota`) value and not in the map, I think this can just be:


```suggestion
func ProviderTypeFromString(s string) (ProviderType, bool) {
	t := providerTypeByName[s]
	return t, t == ProviderType_PROVIDER_TYPE_UNSPECIFIED
```

(or you could even remove the bool parameter at this point and hoist that comparison into the caller)

evankanderson (commented):
I haven't managed to test this yet -- I needed to set up a GitLab repository (more involved in my test environment), since there's no "gitlab" provider trait yet to exclude GitHub repos.

I'd like to be able to say:

```
def:
  provider_traits: ["rest", "github"]
```

To say that a rule uses the GitHub API, or `["rest", "gitlab"]` for the GitLab API.  I can do the former, but not the latter yet.  (That could be a follow-on PR, but it would also make this easier to test if added here.)

pkg/api/protobuf/go/minder/v1/providers.go
intelligent-ears: Taken, thanks — much tidier. One small fix on the way in: the suggestion takes `maps.Keys(validProviderTraitNames)`, i.e. keys of the slice being assigned, so it needed to be the map instead:

```go
validProviderTraitNames = slices.Sorted(maps.Keys(providerTypeByName))
```

`sort` is dropped from the imports, `maps` added.

pkg/api/protobuf/go/minder/v1/providers.go
intelligent-ears: Went with your parenthetical and dropped the bool entirely — the returned value already carries the information, so the second return was just restating it:

```go
func ProviderTypeFromString(s string) ProviderType {
	return providerTypeByName[s]
}
```

Callers compare against `PROVIDER_TYPE_UNSPECIFIED` themselves. Three of them, not two — `pkg/ruletypes/service.go`, `pkg/engine/v1/rtengine/engine.go`, and `pkg/ruletest/eval.go`.

Worth flagging on the intermediate suggestion: `return t, t == ProviderType_PROVIDER_TYPE_UNSPECIFIED` returns true on a *miss*, so it needed `!=`. Moot now that the bool is gone, but it would have quietly inverted trait validation — rejecting valid traits and accepting typos — without failing any existing test, so worth a mention.

internal/providers/providers.go
```diff
@@ -75,6 +75,13 @@ func PBToDBType(t minderv1.ProviderType) (db.ProviderType, bool) {
 		return db.ProviderTypeOci, true
 	case minderv1.ProviderType_PROVIDER_TYPE_IMAGE_LISTER:
 		return db.ProviderTypeImageLister, true
+	case minderv1.ProviderType_PROVIDER_TYPE_GITLAB:
+		// PROVIDER_TYPE_GITLAB exists as a rule type provider_traits value
+		// only; it is gated at evaluation time through
+		// provifv1.Provider.CanImplement, not through the providers table.
+		// There is deliberately no 'gitlab' value in the provider_type
+		// database enum, so it has no db.ProviderType to map to.
```
evankanderson, line 83: We probably want to fix this at some point with a migration to expand the enum, but it's fine not to include it in this PR, which is already fairly large.

evankanderson (approved):
This looks good, so I'm going to merge it; I expect we'll probably have a few other improvements on the client side next week, and then I'll cut a new Minder client binary which will interpret the new field and pass it to the server.

internal/providers/providers.go
intelligent-ears: Agreed — I'll open a follow-up for the enum migration plus a provifv1.GitLab interface so SupportedProviderTypes reports it too, which closes the minder provider get asymmetry at the same time.

PR #6703, opened 2026-08-17
feat(github): record remaining API quota gauges

### LINKED ISSUE #5124: Monitor remaining tokens when calling GitHub
Develop a way to monitor remaining tokens when calling GitHub, so it is possible to see when there is a problem. This is driven by the possibility of repository registration draining the token pool. 

See #3271 for a possible follow up effort to mitigate token starvation.

jeremychoi:
## Summary
- Record `github.Response.Rate.Remaining` and `Rate.Limit` as OpenTelemetry gauges (`github.api.ratelimit.remaining`, `github.api.ratelimit.limit`) before the GitHub provider returns those responses.
- Observe at `ListFiles`, `GitHub.Do`, and `ListAllRepositories` (OAuth and App). These already hold `*github.Response` and cover PR diffs, REST rules, and repository registration.
- Use gauges rather than counters because summing remaining quota across calls is not useful. The existing HTTP RoundTripper histogram (`github.http.ratelimit.remaining`) is unchanged.

Fixes #5124

Follow-ups if useful: `ClientServiceImplementation` (installations and membership) and other call sites that currently discard `*github.Response`.

## Test plan
- [x] `go test ./internal/providers/github/ ./internal/providers/github/clients/`
- [x] Confirm gauges appear after a `ListFiles` or repo-list call against a live GitHub token (remaining and limit last-value)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/providers/github/clients/app.go
+++ b/internal/providers/github/clients/app.go
@@ -257,6 +257,8 @@ func (g *GitHubAppDelegate) ListAllRepositories(ctx context.Context) ([]*minderv
 
 		repos, resp, err = g.client.Apps.ListRepos(ctx, listOpt)
 
+		github.ObserveResponseRate(ctx, resp)
+
 		if err != nil {
 			return ghcommon.ConvertRepositories(allRepos), fmt.Errorf("error listing repositories: %w", err)
 		}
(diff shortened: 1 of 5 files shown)


evankanderson (approved):
The security scan will be fixed with #6657 ; this otherwise looks good.

PR #6724, opened 2026-08-26
fix(gitlab): request project license from the API

### LINKED ISSUE #6589: fix: GitLab project license property always empty due to missing ?license=true parameter

YAshhh29:
Fixes #6589

`getGitLabProject()` builds the request path as `projects/<id>` with no query
string. GitLab only populates the `license` field on the project payload when
`license=true` is passed, so `RepoPropertyLicense` (`gitlab/license`) was always
empty.

This appends `license=true` to the project request. `NewRequest` -> `getParsedURL`
already parses path and query separately and copies `RawQuery` onto the final URL,
so no change is needed in the REST layer.

`getGitLabProject()` is also called from `pull_request_properties.go` and
`release_properties.go`, so those entity fetches pick up license data as well.

Adds a regression test asserting the outgoing request carries `license=true` and
that `RepoPropertyLicense` is populated from the response.

### DIFF
--- a/internal/providers/gitlab/properties_test.go
+++ b/internal/providers/gitlab/properties_test.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


evankanderson (approved):
Given that #6593 seems to have gone dead, I'm approving this version.

Issue #6667, opened 2026-07-31
Accepted Risk (Won't Fix) For Issue Remediation


codejasleen:
Currently, every rule violation eventually results in a remediation issue until the rule becomes compliant.

However, there are cases where a repository owner intentionally decides that a particular violation is an accepted risk and does not plan to remediate it. In those cases, repeatedly creating new remediation issues for the same violation isn't useful.

One possible approach would be to introduce an explicit Accepted Risk workflow for Minder-created remediation issues.

The current idea is to reuse GitHub's existing` wontfix` label. If a repository owner applies the wontfix label to a remediation `issue created by Minder`, Minder would treat that as an explicit Accepted Risk decision. It would record that state for the corresponding repository/rule, close the issue, and avoid creating new remediation issues for subsequent evaluations until that Accepted Risk state is removed.

This keeps the workflow explicit, avoids trying to infer intent from comments or issue state, and fits naturally into the existing GitHub issue workflow.

`Open question:` Is reusing GitHub's existing wontfix label the right trigger for Accepted Risk, or would a different workflow fit Minder better?
@evankanderson

evankanderson:
Remediation PRs have the same challenge; an explicit accepted-risk state feels like a reasonable solution. 

A few questions to clarify thinking (I don't know answers yet, but I'm hoping we can figure out a sketch for the answer alongside our solution to this):

1. How can organization owners track and review the accepted risks?
2. Can "accepted-risk" be undone?  If so, how (what API) and by whom (what permission on what resource)? 
3. Does "accepted-risk" change the rule evaluation status, or only remediation actions?

One idea I've played with in other contexts is to have policy exceptions be time-bounded; that is, you can't say "I accept the risk that this port is open" but instead you say "I accept the risk that this port is open for 2 quarters (while the team builds a cleaner solution / vendor migrates to other API / etc)".  After 2 quarters, if it's not done, you have to go ask for another time-bounded exception.

codejasleen:
> Remediation PRs have the same challenge; an explicit accepted-risk state feels like a reasonable solution.
> 
> A few questions to clarify thinking (I don't know answers yet, but I'm hoping we can figure out a sketch for the answer alongside our solution to this):
> 
> 1. How can organization owners track and review the accepted risks?
> 2. Can "accepted-risk" be undone?  If so, how (what API) and by whom (what permission on what resource)?
> 3. Does "accepted-risk" change the rule evaluation status, or only remediation actions?
> 
[...]

PR #6726, opened 2026-08-28
Add accepted risk foundation

codejasleen:
# Summary

Adds the database and service layer foundation for accepted risks.

## Changes

- Add `accepted_risks` database table with:
  - project and provider references
  - entity name
  - rule type reference
  - expiration timestamp
  - creation timestamp
- Add sqlc queries for creating, listing active risk, and deleting accepted risks
- Add accepted risk service with Create, List, and Delete operations
- Add generated mocks and unit tests for the service

Part Of #6667


database/migrations/000118_accepted_risks.up.sql
```diff
@@ -0,0 +1,12 @@
+-- SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+-- SPDX-License-Identifier: Apache-2.0
+
+CREATE TABLE accepted_risks (
+    id UUID NOT NULL DEFAULT gen_random_uuid() PRIMARY KEY,
+    project_id UUID NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
+    provider_id UUID NOT NULL REFERENCES providers(id) ON DELETE CASCADE,
+    entity_name TEXT NOT NULL,
```
evankanderson, line 8: I'm wondering whether we want to enable some way to have multiple provider_id + entity name combinations, or whether we want to actually resolve an entity id here, possibly with `ON DELETE CASCADE`.

The question is: what's the desired lifecycle here.  If I define an accepted risk for `orgX/repoY`, and then I delete something, what happens to the exception?
* If I delete the repo, does the exception disappear? If I re-register the repo, will I need to re-create the exception?
* If I delete the provider (and transitively the repo), what happens to the exception? If I re-register things, do I need to recreate the exception?

It feels to me like the answer to both questions should be the same -- either the accepted risk disappears when the thing it references goes away (by any means), or it sticks around as an independent policy until explicitly revoked.


I could see a different argument about the `rule_type_id` field (i.e. recreating a ruletype with the same name might not mean the same policy), since that's not related to entity lifecycle.

database/query/accepted_risks.sql
```diff
@@ -0,0 +1,30 @@
...
+    rule_type_id,
+    expires_at
+) VALUES (
+    sqlc.arg(project_id),
+    sqlc.arg(provider_id),
+    sqlc.arg(entity_name),
+    sqlc.arg(rule_type_id),
+    sqlc.arg(expires_at)
+)
+RETURNING *;
```
evankanderson, line 16: Can we have multiple overlapping accepted risks for the same entity and rule_type_id?  I'm assuming the practical result would be to take the longest, but I'm wondering if we want some sort of semantics to limit overlapping.

internal/acceptedrisk/service/service.go
```diff
@@ -0,0 +1,95 @@
...
+	EntityName string
+	RuleTypeID uuid.UUID
+	ExpiresAt  time.Time
+}
+
+// Service encapsulates logic related to accepted risks.
+type Service interface {
+	Create(ctx context.Context, req CreateRequest) (*db.AcceptedRisk, error)
+	List(ctx context.Context, projectID uuid.UUID) ([]db.AcceptedRisk, error)
+	Delete(ctx context.Context, id uuid.UUID, projectID uuid.UUID) error
```
evankanderson, line 32: Generally, I'd try to keep the database types bundled _inside_ the service layer, rather than leaking the database abstractions into the rest of the code (we didn't do this well at the start of the project).

internal/acceptedrisk/service/service.go
```diff
@@ -0,0 +1,95 @@
...
+type CreateRequest struct {
+	ProjectID  uuid.UUID
+	ProviderID uuid.UUID
+	EntityName string
+	RuleTypeID uuid.UUID
+	ExpiresAt  time.Time
+}
+
+// Service encapsulates logic related to accepted risks.
+type Service interface {
```
evankanderson, line 29: What does the interaction between the rule engine and accepted risks look like?

You don't need to write the queries now, but I want to understand what they look like.

evankanderson (commented):
This looks like a good start.  A few questions to make sure you're headed on the right track, and a couple style / layering learnings that we've had.

internal/acceptedrisk/service/service.go
codejasleen: Makes sense, i will correct this.

database/migrations/000118_accepted_risks.up.sql
codejasleen: I think we can keep entity_name for the semantic matching, but also store entity_id to tie the accepted risk to the entity lifecycle. If the entity is deleted, the accepted risk can be deleted through ON DELETE CASCADE. This also means if an entity is recreated with the same name, the old accepted risk wont automatically apply to the new entity.

database/query/accepted_risks.sql
codejasleen: I had thought about this earlier. If we would keep unique (entity_id, rule_type_id) then if a risk expires today and user wants to create a new one tomorrow, it wont be possible until scheduler runs again.
Yes, but creating duplicates is bad so, I can think of a possible approach:

- Lazy cleanup

When we create a new accepted risk, check if an existing record for the same scope exists?

1. If it is active then we can prevent the user to create new until the previous expire or can introduce update.

2. If is expired then we can delete the expired record as part of the create operation and then create the new one. means the cleanup job can be the eventual cleanup rather than something the API depends on

internal/acceptedrisk/service/service.go
codejasleen: I think Accepted Risk would fit between the rule evaluation and `actionEngine.DoActions()` . rule will run normally and give evaluation result then we can use the existing RuleTypeID and EntityID already available in evalstatusparam to check for an active accepted risk. If a matching risk exists, we would skip the remediation action, otherwise continue with DoActions() as normal.

This is where unique (entity_id, rule_type_id) becomes useful..This is how i am thinking about fitting it into the existing flow..Does this looks right to you?

evankanderson: Sorry, my question was slightly different -- once you have the rule evaluation result, what will the function calls look like to hook this up in the engine?  Will the engine use its own `db.Store` to look these up, or will it need to take an `acceptedrisk.Service` instance and call some functions on it?

Does the engine own matching acceptedrisks with rule evaluations, or is that a function of the acceptedrisk service?

database/query/accepted_risks.sql
evankanderson: A few possibilities:

1. We allow multiple AcceptedRisks for a given entity+ruletype.  They can be distinguished by ID, but we might also allow a textual note attached to each AcceptedRisk.  All AcceptedRisks need to be cleared before the evaluation can be triggered.
2. We only allow one AcceptedRisk, whose duration is equal to the _longest_ or _last_ creation of an AcceptedRisk.  The rationale for "longest" would be effectively stacking risk acceptances like in case 1, but without creating and expiring as many resources.  The rationale for "latest" would effectively be an "Upsert" capability.

I don't have a strong opinion here -- I'd just like us to have considered the design space before falling into the first idea that occurs to us.  (We might still use the first idea, but it shouldn't be because we didn't think of the second one.)

In terms of implementing approach 2, we could use an index to flag duplicates, and use [`ON CONFLICT ... DO UPDATE SET ... WHERE ...`](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT) to perform the update.

database/migrations/000118_accepted_risks.up.sql
evankanderson: Okay, let's consider that a design decision -- for exact-match AcceptedRisks, deleting the entity also deletes the acceptedrisk.  Do we want the same decision for the ruletype?

(I'm trying to think about how this applies in the future where you may be able to create a pattern-match AcceptedRisk, e.g. via CEL selectors)

codejasleen: Yes, i think we should have the same behavior for the rule type for the V1 exact-match case. I think we can keep this decision scoped to exact-match Accepted Risks for now, so that it doesn't necessarily constrain the lifecycle semantics of future pattern match Accepted Risks such as CEL selectors.

database/query/accepted_risks.sql
codejasleen: I doubt if we have a meaningful use case where two separate accepted risks for the exact same entity + rule type needs to coexist? and for that, option 2 makes more sense to me as updating with latest request.

internal/acceptedrisk/service/service.go
codejasleen: I think the engine should take an acceptedrisk.Service instance, similar to the other services it already receives. After the rule evaluation, the engine can call the accepted Rrsk service with the EntityID and RuleTypeID. The accepted risk service would own the matching logic and return whether there is an active matching risk. The engine would then use that result to decide whether to call actionEngine.DoActions().

this way engine will still own the decision of whether to run the actions while keeping the accepted risk lookup out of the engines db.Store logic.

evankanderson: What I'm wondering is whether the engine should call a service interface, and provide it with an entity (with properties) and a ruletype, and then any matching logic lives in the acceptedrisk.Service (exception.Service), rather than in the engine.  I'm trying to avoid ending up with a big-ball-of-mud in the engine.  😁

codejasleen: Yes, that makes sense. I was thinking about passing the EntityID and RuletypeID, but I agree that passing the entity (with properties) and rule type to the service makes the boundary cleaner. That also gives us more room for future matching such as CEL selectors without adding that logic to the engine.

evankanderson: > Yes, that makes sense. I was thinking about passing the EntityID and RuletypeID, but I agree that passing the entity (with properties) and rule type to the service makes the boundary cleaner. That also gives us more room for future matching such as CEL selectors without adding that logic to the engine.

Exactly!  (And without having this service go directly to the db for entity rows, which feels like a bit of a layering violation when interfacing with the engine.)

codejasleen:
hey @evankanderson, CI is failing because the merge environment has another (000118_gitlab_provider_type), which conflicts with our 000118_exceptions migration. The current main only goes up to 000117, so this mightbe be from another concurrent PR. I will rename our exception migration to 000119 to avoid the collision?

database/migrations/000118_exceptions.up.sql
```diff
@@ -0,0 +1,13 @@
+-- SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+-- SPDX-License-Identifier: Apache-2.0
+
+CREATE TABLE exceptions (
+    id UUID NOT NULL DEFAULT gen_random_uuid() PRIMARY KEY,
+    project_id UUID NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
+    entity_name TEXT NOT NULL,
```
evankanderson, line 7: What does "entity_name" represent?  It doesn't seem to reference anything else, so I'm wondering how the values are managed.

evankanderson: I think it's fine to either use a "name" property or an "Id" property to reference the entity -- if we're matching by name, then "name" would make sense, and would allow us to keep an exception even after the corresponding entity was deleted.  I think in this discussion, we agreed that deleting an entity deletes the corresponding direct exceptions, though: https://github.com/mindersec/minder/pull/6726#discussion_r3918313394

So I'd remove this, and keep the entity name only in public-facing APIs (where people may refer to things by name).

internal/exception/service/service.go
```diff
@@ -0,0 +1,127 @@
...
+	RuleTypeID uuid.UUID
+	ExpiresAt  time.Time
+}
+
+// Exception represents an exception.
+type Exception struct {
+	ID         uuid.UUID
+	ProjectID  uuid.UUID
+	EntityID   uuid.UUID
+	EntityName string
```
evankanderson, line 33: I think it's fine to get this, though we may want to include more information here.  Glancing at `EntityInstance`:

* provider
* name (from `entity_instances`)
* type
* properties

At the least, name + type + provider may be needed to ensure uniqueness.

internal/exception/service/service.go
```diff
@@ -0,0 +1,127 @@
...
+	RuleTypeID uuid.UUID
+	ExpiresAt  time.Time
+	CreatedAt  time.Time
+}
+
+// Service encapsulates logic related to exceptions.
+type Service interface {
+	Create(ctx context.Context, req CreateRequest) (*Exception, error)
```
evankanderson, line 41: Why not have `Create` take an `Exception` struct, with certain fields documented to be ignored (ID, CreatedAt, etc)?  Alternatively, we might limit the arguments to only those that can't be determined otherwise (for example, we might want to pull ProjectID from `GetProjectID(ctx context.Context)`).

database/query/exceptions.sql
```diff
@@ -0,0 +1,33 @@
...
+    project_id,
+    entity_name,
+    entity_id,
+    rule_type_id,
+    expires_at
+) VALUES (
+    sqlc.arg(project_id),
+    sqlc.arg(entity_name),
+    sqlc.arg(entity_id),
+    sqlc.arg(rule_type_id),
```
evankanderson, line 13: Can we somewhere validate that `entity_id` and `rule_type_id` belong to `project_id`?  I worry about passing around too many opaque IDs and accidentally introducing another https://github.com/mindersec/minder/security/advisories/GHSA-9c2x-rm59-4335 or https://github.com/mindersec/minder/security/advisories/GHSA-v627-69v2-xx37

evankanderson (commented):
This is looking on the right track; just a few questions based on the earlier discussions.

evankanderson:
> hey @evankanderson, CI is failing because the merge environment has another (000118_gitlab_provider_type), which conflicts with our 000118_exceptions migration. The current main only goes up to 000117, so this mightbe be from another concurrent PR. I will rename our exception migration to 000119 to avoid the collision?

Yes, this was another PR from @intelligent-ears that went in while this PR was in flight.  You'll need to merge the main branch and move your migration to the 000119 number.

database/migrations/000118_exceptions.up.sql
codejasleen: Agreed. I will remove entity_name from the persisted exception and keep the entity information at the API boundary.

database/query/exceptions.sql
codejasleen: Yes, I agree we shouldn't accept combinations of opaque ids. I checked the existing patterns: EntityService.GetEntityByID already validates that the entity belongs to the supplied project, and GetRuleTypeByID already scopes the lookup by project. we can derive the project from the request context and reuse these project scoped lookups to validate both resources before creating the exception.

internal/exception/service/service.go
codejasleen: That makes sense. I will keep the exception keyed by entity_id and look at the existing EntityInstance representation to see what entity information we should expose in the service/API response. I'll keep the evaluation side property changes for the exception matching work in the next part.

internal/exception/service/service.go
codejasleen: That makes sense. I think it would be cleaner to have Create take the exception struct and treat fields like ID and CreatedAt as output only. I will also derive ProjectID from the request context rather than passing it separately.

internal/exception/service/service.go
codejasleen: @evankanderson For the API response, we can expose the entity information (provider, name, type, ..) as part of the EntityInstance representation, while keeping only entity_id in the exceptions table as the entity ID already identifies the exact EntityInstance. Do you think the same?

codejasleen: provider/type/name combination is used to uniqely identify the EntityInstance, but once the EntityInstance exists, its UUID uniquely identifies that exact entity

PR #6729, opened 2026-08-30
Add fallback support for REST data source handler

shreyaabaranwal:
## What
Wires up `RestDataSource_Def.fallback` and `expected_status` - already
defined in the proto, previously unused in `handler.go`.

- Checks the response status against `expected_status` (default: `200` only).
- If unexpected and a matching `Fallback` is configured, returns its
  `body`/`http_status` instead of the real response.
- If unexpected and no fallback matches, falls through to existing
  behavior (parses/returns the real response) ,  zero behavior change
  for datasources that don't opt in.

## Why
Related to #6275. #6700 already covers the timeout/response-limit part
of that issue; this covers the fallback part.

## Open design question
See discussion on #6275 , when no fallback matches, should we keep
today's fall-through behavior or return a structured error instead?
Opening this as a **draft** pending that discussion - will mark ready
for review once we align.

## Testing
- New table-driven tests in `handler_fallback_test.go` (unit tests +
  full `Call()`-path tests via `httptest.NewServer`).
- All existing `internal/datasources/rest` tests still pass, no
  regressions.
- `gofmt -l` clean, `golangci-lint run ./internal/datasources/rest/...`
   0 issues.


krrish175-byte (dismissed):
The description says this moves to ready-for-review once #6275 settles whether an unmatched status should keep falling through or return a structured error. This PR keeps the current fall-through behavior unchanged either way, and it's now marked ready. Did that discussion land on a decision? If not I'd rather hold this as draft,  the answer could mean revisiting this diff. Otherwise solid, well-tested work.

internal/datasources/rest/handler.go
```diff
@@ -233,16 +241,53 @@ func (h *restHandler) doRequest(dofunc func(*http.Request) (*http.Response, erro
 
 	recordMetrics(req.Context(), resp, start)
 
+	if !h.isExpectedStatus(resp.StatusCode) {
+		if fb, ok := h.matchFallback(resp.StatusCode); ok {
+			zerolog.Ctx(req.Context()).Debug().
```
krrish175-byte: Nit: this masks a real upstream failure -- feels more like Info than Debug so it's visible without cranking verbosity.

internal/datasources/rest/handler.go
```diff
@@ -233,16 +241,53 @@ func (h *restHandler) doRequest(dofunc func(*http.Request) (*http.Response, erro
...
+				Int("status_code", resp.StatusCode).
+				Int32("fallback_status_code", fb.GetHttpStatus()).
+				Msg("unexpected status code from datasource, using configured fallback")
+			return buildRestOutput(int(fb.GetHttpStatus()), fb.GetBody()), nil
```
krrish175-byte: This bypasses `parseResponseBody`, so `fb.GetBody()` comes back as a raw string even when `parse: "json"` is set on this datasource. The non-fallback path (line 259) JSON-decodes `bout` when `h.parse == "json"`; this path doesn't. A rule reading `.body.field` will work on a real 200 and silently break the moment a fallback fires. Can you route the fallback body through the same parse-mode logic as `parseResponseBody` (e.g. a shared `parseBody(raw string, mode string) (any, error)` both paths call) so the output shape is guaranteed regardless of which path produced it?

internal/datasources/rest/handler.go
```diff
@@ -233,16 +241,53 @@ func (h *restHandler) doRequest(dofunc func(*http.Request) (*http.Response, erro
...
+		if int(s) == statusCode {
+			return true
+		}
+	}
+
+	return false
+}
+
+// matchFallback returns the configured fallback for statusCode, if any.
+func (h *restHandler) matchFallback(statusCode int) (*minderv1.RestDataSource_Def_Fallback, bool) {
```
krrish175-byte: Nit: duplicate http_status entries in the fallback list silently take the first match. Might be worth rejecting duplicates at config-load time in newHandlerFromDef so a copy-paste config mistake doesn't fail silently.

internal/datasources/rest/handler_fallback_test.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+}
+
+// Test_restHandler_HTTPCall_Fallback exercises the fallback behavior through
+// the full Call path, following the same server/table pattern as
+// Test_restHandler_HTTPCall.
+func Test_restHandler_HTTPCall_Fallback(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
```
krrish175-byte: None of these cases set `parse`, so they only exercise the zero-value/raw-string path on both branches and don't catch the type mismatch when `parse: "json"` is set (see comment on handler.go:250). Can you add a case with `parse: "json"` on the handler + a triggered fallback, asserting the fallback body comes back decoded, not as a string?

internal/datasources/rest/handler.go
```diff
@@ -62,7 +62,13 @@ type restHandler struct {
...
 	requestTimeout   time.Duration
 	maxResponseBytes int64
-	// TODO implement fallback
+	// fallback holds the per-status-code fallback responses configured on
+	// this data source definition. It is consulted in doRequest whenever the
+	// upstream response status is not in expectedStatus.
+	fallback []*minderv1.RestDataSource_Def_Fallback
+	// expectedStatus is the set of status codes treated as a normal
+	// response. If empty, only http.StatusOK is treated as expected.
+	expectedStatus []int32
```
evankanderson: Why use `int32` here?  I'd generally prefer `int` unless there are specific size requirements.

evankanderson: (I'm guessing this is because we're copying from the proto array)

internal/datasources/rest/handler.go
```diff
@@ -233,16 +241,53 @@ func (h *restHandler) doRequest(dofunc func(*http.Request) (*http.Response, erro
...
+func (h *restHandler) isExpectedStatus(statusCode int) bool {
+	if len(h.expectedStatus) == 0 {
+		return statusCode == http.StatusOK
+	}
+
+	for _, s := range h.expectedStatus {
+		if int(s) == statusCode {
+			return true
+		}
+	}
```
evankanderson: Is this [`slices.Contains`](https://pkg.go.dev/slices#Contains)?

If you store `http.StatusOK` in `h.expectedStatus` during construction if the slice is empty, you can avoid needing this method at all, and just say:

```
if slices.Contains(h.expectedStatus, resp.StatusCode) {
  // ...
```

internal/datasources/rest/handler.go
```diff
@@ -233,16 +241,53 @@ func (h *restHandler) doRequest(dofunc func(*http.Request) (*http.Response, erro
...
+
+// matchFallback returns the configured fallback for statusCode, if any.
+func (h *restHandler) matchFallback(statusCode int) (*minderv1.RestDataSource_Def_Fallback, bool) {
+	for _, fb := range h.fallback {
+		if int(fb.GetHttpStatus()) == statusCode {
+			return fb, true
+		}
+	}
+
+	return nil, false
```
evankanderson: Again, this may be simpler using [`slices.IndexFunc`](https://pkg.go.dev/slices#IndexFunc), or sorting the list and using [`slices.BinarySearchFunc`](https://pkg.go.dev/slices#BinarySearchFunc).  Since I think we're constructing this per evaluation and using it few times, `IndexFunc` is probably faster.


```suggestion
	i := slices.IndexFunc(h.fallback, func(fb *minderv1.RestDataSource_Def_Fallback) bool {
		return fb.GetHttpStatus() == statusCode
	})
	if i < 0 {
		return nil, false
	}
	return h.fallback[i], true
```

internal/datasources/rest/handler_fallback_test.go
evankanderson, line 1: Can you add these tests for code in `handler.go` to `handler_test.go` instead of creating a new file?

evankanderson (commented):
This looks pretty good, just a few small questions / nits.

internal/datasources/rest/handler.go
shreyaabaranwal: Good catch. `int32` was just carried over from the proto since `expected_status` is a repeated `int32`. 

I’ve changed it to `[]int` and do the conversion once in `newHandlerFromDef`. 
I also moved the defaulting there, so it’s set to `[http.StatusOK]` upfront instead of checking for an empty value on every request.

internal/datasources/rest/handler.go
shreyaabaranwal: Done. Removed `isExpectedStatus` entirely. `expectedStatus` now defaults to `[http.StatusOK]` during construction when it isn’t set, so `doRequest` can just use `slices.Contains(h.expectedStatus, resp.StatusCode)` directly.

internal/datasources/rest/handler.go
shreyaabaranwal: Applied this, thanks! 

I made one small change from the suggested diff though , I kept the explicit `int(fb.GetHttpStatus())` cast in the comparison. `GetHttpStatus()` returns an `int32`, while `statusCode` is an `int`, so the comparison in the suggested version wouldn’t compile as-is.

I can also change `statusCode` to `int32` if you’d prefer to avoid the cast here.

internal/datasources/rest/handler_fallback_test.go
shreyaabaranwal: Moved the tests into `handler_test.go` and removed the separate test file.

 I also dropped the `isExpectedStatus`-specific test since that method no longer exists, and replaced it with a construction-level test to verify that `newHandlerFromDef` defaults `expectedStatus` to `[http.StatusOK]` when it’s unset.

evankanderson (approved):

krrish175-byte (approved):

PR #6733, opened 2026-09-01
Fix panic recovery logging and OAuth callback race

### LINKED ISSUE #6732: Controlplane: panic recovery event is never logged; OAuth callback panics when provider creation loses a race
### Describe the issue

Two silent-failure bugs in internal/controlplane, found while reading the code.

**1. The gRPC panic recovery handler never emits its log event.**

recoveryHandler in internal/controlplane/server.go builds a zerolog event with the panic value and stack trace, but never calls a terminator:

	func recoveryHandler(ctx context.Context, p any) error {
		log.Ctx(ctx).Error().
			Str("msg", "recovered from panic").
			Any("panic", p).
			Bytes("stack", debug.Stack())
		return status.Errorf(codes.Internal, "%s", p)
	}

zerolog only writes an event when .Msg()/.Send() runs, so recovered panics are invisible in server logs - the client gets its Internal error, but the panic value and stack trace are discarded. The Str("msg", ...) field suggests the Msg() call got lost at some point.

**2. The OAuth callback panics on a nil provider when provider creation loses a race.**

In processOAuthCallback (internal/controlplane/handlers_oauth.go), when CreateProviderFromSessionState returns a unique violation the handler logs "Provider already exists" and falls through - but the returned provider is nil on that path, so the next line panics:

	logger.BusinessRecord(ctx).ProviderID = p.ID

The unique violation only escapes when two callbacks for the same provider race: the session service checks GetByName first and creates on not-found (internal/providers/session/service.go), so losing the create race is the one way to hit this branch. When it happens the callback panics, and the losing flow's access token is never stored, since the service returns before UpsertAccessToken.

### To Reproduce

1. Panic inside any gRPC handler: the request fails with codes.Internal but nothing about the panic appears in the logs.
2. Reproduced in a unit test by having the store mocks return not-found on the provider lookup and a pq unique violation on create - the handler panics.

I have fixes with regression tests for both and will open a PR.

### To Reproduce

_No response_

_No response_

NAME-ASHWANIYADAV:
# Summary

Two small fixes in internal/controlplane, one commit each.

`recoveryHandler` built a zerolog event with the panic value and stack trace but never called a terminator, so zerolog discarded the event and recovered panics were invisible in server logs. The message now goes through `Msg()` so the event is actually emitted.

In the OAuth callback, when two enrollment callbacks for the same provider raced, the loser got a unique violation plus a nil provider from `CreateProviderFromSessionState`, logged "Provider already exists", then panicked on `p.ID`. The losing flow's access token was also never stored. On a unique violation the handler now retries the call once: the provider exists by
then, so the retry takes the fetch path, stores this flow's token, and returns the existing provider.

Fixes #6732

# Testing

- Added `TestRecoveryHandlerEmitsLogEvent`: attaches a buffer-backed zerolog  logger to the context, calls `recoveryHandler`, and asserts the emitted event contains the message, panic value, and stack, and that the returned error is `codes.Internal`. Without the fix the buffer stays empty.
- Added the "Provider creation race retries and succeeds" case to TestProviderCallback`: the store mocks return not-found on the provider lookup and a pq unique violation on create, then expose the provider on the retry. Without the fix this case panics with a nil pointer dereference; with the fix the callback completes with the 307 redirect and stores the token.
- `go test ./internal/controlplane/` passes for both suites; `go vet` is clean.


internal/controlplane/handlers_oauth.go
```diff
@@ -349,9 +349,13 @@ func (s *Server) processOAuthCallback(
...
 	p, err := s.sessionService.CreateProviderFromSessionState(ctx, db.ProviderClass(provider), &encryptedToken, state)
 	if db.ErrIsUniqueViolation(err) {
-		// todo: update config?
-		zerolog.Ctx(ctx).Info().Str("provider", provider).Msg("Provider already exists")
-	} else if errors.As(err, &errConfig) {
+		// We lost the creation race against a concurrent callback for the
+		// same provider. The provider exists now, so a retry takes the
+		// fetch path and still stores this flow's access token.
+		zerolog.Ctx(ctx).Info().Str("provider", provider).Msg("Provider already exists, fetching it")
+		p, err = s.sessionService.CreateProviderFromSessionState(ctx, db.ProviderClass(provider), &encryptedToken, state)
```
evankanderson: I'm not sure that storing the second provider's token is necessary / correct in this race.  I'd consider changing the contract of `CreateProviderFromSessionState` to return the provider alongside the uniqueness error, so the client here can simply fall through the uniqueness case with a comment.

(I'm also noticing that `CreateProviderFromSessionState` uses the old pre-generics `errors.As` style rather than `errors.AsType`, which would be a bit easier to read.)

evankanderson (commented):
I'm assuming that you found both of these via LLM-assisted flows?  I'm willing to merge this, but I think I have a slight preference for avoiding the additional database writes in the race condition case.

I'm _also_ noticing that `CreateProviderFromSessionState` isn't really making use of database transactions, which are generally supported (and useful) with our Postgres dependency.

On the third hand, it looks like passing around a `db.Store` or appropriate sub-type might be annoying given the implementation of getters and managers in the implementation, so I'd be amenable to a "yes, that would be better, but it's too hard" argument.

internal/controlplane/handlers_oauth.go
NAME-ASHWANIYADAV: Done in 1eef981 - contract changed as you suggested and the retry is gone. The test now also asserts the loser doesn't write a second token.

evankanderson (approved):
Looks good.  I have no problem with using LLMs to find or implement fixes, but I wanted to understand if this was a problem that had been observed in the wild, or discovered by inspection.  (Observed-in-the-wild problems might be worth broader notification about the discovered bug.)

PR #6747, opened 2026-09-03
Surface the HTTP status when REST remediation fails

### LINKED ISSUE #6746: REST remediation reports 'remediation failed: %!s(<nil>)' and drops the HTTP status
### Describe the issue

In internal/engine/actions/remediate/rest/rest.go, run() translates a non-2xx response into an action failure like this:

	if engerrors.HTTPErrorCodeToErr(resp.StatusCode) != nil {
		return engerrors.NewErrActionFailed("remediation failed: %s", err)
	}

At that point err is guaranteed nil (the earlier `if err != nil` branch already returned), so the user-visible remediation detail becomes:

	action failed: remediation failed: %!s(<nil>)

and both the mapped error from HTTPErrorCodeToErr and the status code are discarded, leaving no way to tell why the remediation failed.

The path is reachable with providers whose Do returns non-2xx responses without an error - the generic REST provider and the GitLab provider
pass the response through as-is. The GitHub provider is unaffected since go-github's Do already returns an error for non-2xx.

### To Reproduce

Reproduced in a unit test that drives run() through a provider whose Do returns the response as-is, with a handler returning 403 - the resulting
error is exactly the nil-format message above.

I have a fix with a regression test ready and will open a PR.

NAME-ASHWANIYADAV:
# Summary

When a REST remediation gets a non-2xx response from a provider whose Do returns the response without an error (the generic REST provider and GitLab pass responses through as-is), the failure path formatted the request error - which is guaranteed nil at that point - so the stored remediation detail read `remediation failed: %!s(<nil>)` and the mapped status error was discarded. The failure now includes the status code and the translated error, e.g. `remediation failed: HTTP status 403: forbidden`.

Fixes #6746 

# Testing

- Added a regression case to TestRestRemediate that drives run() through a provider whose Do returns the response as-is (a small wrapper over  the test REST provider that accepts the JSON body run() produces).  Without the fix it fails with the nil-format message above; with the
  fix the error carries the status.
- go test ./internal/engine/actions/remediate/rest/ passes; go vet is clean.

### DIFF
--- a/internal/engine/actions/remediate/rest/rest.go
+++ b/internal/engine/actions/remediate/rest/rest.go
@@ -189,6 +189,13 @@ func (r *Remediator) run(ctx context.Context, method string, endpoint string, bo
 	}
 
 	resp, err := r.cli.Do(ctx, req)
+	if resp != nil {
+		defer func() {
+			if closeErr := resp.Body.Close(); closeErr != nil {
+				log.Printf("cannot close response body: %v", closeErr)
+			}
+		}()
+	}
 	if err != nil {
 		var respErr *github.ErrorResponse
 		if errors.As(err, &respErr) {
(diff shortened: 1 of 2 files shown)


evankanderson (commented):
I'm wondering whether we should change the REST contract (and GitHub, etc providers) to return `err` on non-2xx responses.  It feels janky that there is one codepath that specifically deals with github-shaped errors, and another for dealing with other providers.  (Alternatively, maybe the GitHub-shaped one is the funky one, but it feels like handling errors as explicit allows the provider to map different types of return codes more explicitly.)

Thank you for highlighting these inconsistencies, by the way!  A bunch of this code got put together on the fly / for specific demos, and it didn't always get a second pass to figure out the best, most consistent expression.

With that said, I think handling non 2xx codes as error makes sense, though I'd probably prefer to have it in the same block with the other error handling.  (Maybe move the `defer` up, and then combine the two error handlers?)

evankanderson (approved):
I'm still slightly concerned that the rest provider `Do` contract isn't clear about error handling, but this seems to fix the immediate problem.

PR #6748, opened 2026-09-04
Add gitlab to the provider_type enum and report it in SupportedProviderTypes

intelligent-ears:
Follow-up to #6702, discussed with @evankanderson.

#6702 added a `gitlab` provider trait so rule types can declare `provider_traits: ["rest", "gitlab"]`. Rule gating works, because it goes through `provifv1.Provider.CanImplement`, which the GitLab client implements. But the persisted side was left behind: `db.Provider.Implements` is populated from `ProviderClassInfo.SupportedProviderTypes`, which `ProviderTypesFromImpl` derives by reflection against `providerTypeMap` — a map of `ProviderType` to Go interface. With no `provifv1.GitLab` interface, there was no entry, so `minder provider get` never listed `gitlab` under Implements even though gating honoured it.

This closes that gap:

- Adds `provifv1.GitLab` (`Provider` + `RepoLister` + `REST` + `Git`, plus `GetCredential() GitLabCredential`). The credential method is what makes it discriminating — no other provider client returns `GitLabCredential`, so nothing else accidentally satisfies the interface.
- Adds the `providerTypeMap` entry and a compile-time assertion on `gitlabClient`.
- Migration `000118` adds `gitlab` to the `provider_type` Postgres enum, following the pattern of `000002` and `000055`. The down migration is a no-op, as those are — Postgres has no `DROP VALUE`.
- `PBToDBType` now maps the trait to `db.ProviderTypeGitlab` instead of returning no DB equivalent, and `DBToPBType` gains the symmetric case so a persisted `gitlab` trait isn't dropped on read.

### Not included: backfill

A new enum value doesn't retroactively update existing `providers.implements` rows, so GitLab providers registered before this will still show `{git, rest, repo-lister}`. I didn't write a backfill migration because nothing currently queries `implements` for `gitlab` — every `GetByTraitInHierarchy` / `BulkInstantiateByTrait` caller is hardcoded to `github` or `repo-lister`, and rule-type gating uses the live provider instance rather than this column. So the only effect today is `provider get` under-reporting for pre-existing rows. Happy to add one if you'd rather.

### Test plan

- [x] `make sqlc` — `db.ProviderTypeGitlab` generated
- [x] `go build ./...`
- [x] `go test ./pkg/providers/... ./internal/providers/... ./internal/engine/... ./pkg/engine/... ./internal/controlplane/...`
- [ ] Migration applied against a local Postgres

### DIFF
--- a/database/migrations/000118_gitlab_provider_type.down.sql
+++ b/database/migrations/000118_gitlab_provider_type.down.sql
@@ -1,0 +1,6 @@
+-- SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Postgres can't remove a value for an enum type. So, we can't really
+-- do a down migration. Instead, we'll just leave this here as a
+-- reminder that we can't remove this value.
--- a/database/migrations/000118_gitlab_provider_type.up.sql
+++ b/database/migrations/000118_gitlab_provider_type.up.sql
@@ -1,0 +1,8 @@
+-- SPDX-FileCopyrightText: Copyright 2026 The Minder Authors
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Add `gitlab` provider type, so that a registered GitLab provider's
+-- `implements` column can record that it supports the `gitlab` trait,
+-- matching what `provifv1.Provider.CanImplement` already gates at
+-- evaluation time.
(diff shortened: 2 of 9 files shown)


evankanderson (commented):
`make gen`  seemed to think that there were more mocks that needed to be created, but I'm going to re-run, just in case.

evankanderson (approved):

PR #6749, opened 2026-09-06
Add coverage tracking for `mindev test`

evankanderson:
# Summary

Tracks which rules in a directory were executed by `mindev test`, and includes a report of uncovered rules in the JUnit XML as well as in the text output.

For JUnit output, uncovered rules are reported in a new TestSuite that includes the rule names along with a coverage percentage of 0% for that TestSuite.  Existing TestSuites are marked as 100% coverage.

# Testing

Manual testing and added unit tests, particularly for JUnit output.

### DIFF
--- a/cmd/dev/app/test/test.go
+++ b/cmd/dev/app/test/test.go
@@ -68,10 +68,12 @@ func CmdTest() *cobra.Command {
 				}
 			}
 
-			for _, res := range results {
-				if len(res.Failures) > 0 {
-					finalErr = errors.New("one or more tests failed")
-					break
+			for _, run := range results {
+				for _, res := range run.Results {
+					if len(res.Failures)+len(res.Errors) > 0 {
+						finalErr = errors.New("one or more tests failed")
+						break
+					}
 				}
 			}
 			return finalErr
@@ -80,27 +82,44 @@ func CmdTest() *cobra.Command {
 
 	cmd.Flags().StringVarP(&outputFormat, "output", "o", "text", "Output format (text, junit)")
 	cmd.Flags().StringVar(&junitFile, "junit-file", "", "File to write JUnit report to (in addition to standard output)")
+	cmd.Flags().Bool("coverage", false, "Show uncovered rules in the output")
 
 	return cmd
 }
 
-func formatFailuresHuman(cmd *cobra.Command, results []ruletest.TestResult) {
+func formatFailuresHuman(cmd *cobra.Command, results []ruletest.TestRun) {
 	if len(results) == 0 {
 		cmd.Printf("No tests found\n")
 	}
-	for _, res := range results {
-		if len(res.Failures) > 0 {
-			cmd.Printf("FAIL: %s/%s\n", res.Filename, res.Name)
-			for _, f := range res.Failures {
-				cmd.Printf("  - %s\n", f)
+	coverage, _ := cmd.Flags().GetBool("coverage")
+	for _, run := range results {
+		cmd.Printf("Test run in %s:\n", run.BaseDir)
+		for _, res := range run.Results {
+			if len(res.Failures) > 0 {
+				cmd.Printf("FAIL: %s/%s\n", res.Filename, res.Name)
+				for _, f := range res.Failures {
+					cmd.Printf("  - %s\n", f)
+				}
+			} else if len(res.Errors) > 0 {
+				cmd.Printf("ERROR: %s/%s\n", res.Filename, res.Name)
+				for _, e := range res.Errors {
+					cmd.Printf("  - %s\n", e)
+				}
+			} else {
+				cmd.Printf("PASS: %s/%s\n", res.Filename, res.Name)
 			}
-		} else {
-			cmd.Printf("PASS: %s/%s\n", res.Filename, res.Name)
+		}
+		// TODO: hide this behind a flag
+		if uncovered := run.UncoveredRules(); coverage && len(uncovered) > 0 {
+			cmd.Printf("UNCOVERED RULES:\n")
+			for _, rule := range uncovered {
+				cmd.Printf("  - %s\n", rule)
+			}
 		}
 	}
 }
 
-func writeJUnit(w io.Writer, results []ruletest.TestResult) error {
+func writeJUnit(w io.Writer, results []ruletest.TestRun) error {
 	suites := ruletest.AsJUnit(results)
 	if _, err := fmt.Fprint(w, xml.Header); err != nil {
 		return fmt.Errorf("failed to write XML header: %w", err)
--- a/cmd/dev/app/test/test.go
+++ b/cmd/dev/app/test/test.go
@@ -71,8 +71,7 @@ func CmdTest() *cobra.Command {
 			for _, run := range results {
 				for _, res := range run.Results {
 					if len(res.Failures)+len(res.Errors) > 0 {
-						finalErr = errors.New("one or more tests failed")
-						break
+						return errors.New("one or more tests failed")
 					}
 				}
 			}
@@ -109,7 +108,6 @@ func formatFailuresHuman(cmd *cobra.Command, results []ruletest.TestRun) {
 				cmd.Printf("PASS: %s/%s\n", res.Filename, res.Name)
 			}
 		}
-		// TODO: hide this behind a flag
 		if uncovered := run.UncoveredRules(); coverage && len(uncovered) > 0 {
 			cmd.Printf("UNCOVERED RULES:\n")
 			for _, rule := range uncovered {
--- a/pkg/ruletest/junit.go
+++ b/pkg/ruletest/junit.go
@@ -6,29 +6,49 @@ package ruletest
 import (
 	"encoding/xml"
 	"maps"
+	"path/filepath"
 	"slices"
 	"strings"
 )
 
 // JUnitTestSuites is the root element of a JUnit XML report.
 type JUnitTestSuites struct {
-	XMLName    xml.Name         `xml:"testsuites"`
+	XMLName  xml.Name `xml:"testsuites"`
+	Tests    int      `xml:"tests,attr,omitempty"`
+	Errors   int      `xml:"errors,attr,omitempty"`
+	Failures int      `xml:"failures,attr,omitempty"`
+	Skipped  int      `xml:"skipped,attr,omitempty"`
+	Disabled int      `xml:"disabled,attr,omitempty"`
+
 	TestSuites []JUnitTestSuite `xml:"testsuite"`
 }
 
 // JUnitTestSuite represents a single test suite in a JUnit XML report.
 type JUnitTestSuite struct {
-	Name      string          `xml:"name,attr"`
-	Tests     int             `xml:"tests,attr"`
-	Failures  int             `xml:"failures,attr"`
+	Name     string `xml:"name,attr"`
+	Tests    int    `xml:"tests,attr"`
+	Errors   int    `xml:"errors,attr"`
+	Failures int    `xml:"failures,attr"`
+	Skipped  int    `xml:"skipped,attr,omitempty"`
+	Disabled int    `xml:"disabled,attr,omitempty"`
+
+	File string `xml:"file,attr,omitempty"`
+
+	Properties *[]Property `xml:"properties>property,omitempty"`
+
 	TestCases []JUnitTestCase `xml:"testcase"`
 }
 
 // JUnitTestCase represents a single test case in a JUnit XML report.
 type JUnitTestCase struct {
-	Name      string        `xml:"name,attr"`
-	ClassName string        `xml:"classname,attr"`
+	Name      string `xml:"name,attr"`
+	ClassName string `xml:"classname,attr"`
+
 	Failure   *JUnitFailure `xml:"failure,omitempty"`
+	Error     *JUnitFailure `xml:"error,omitempty"`
+	Skipped   *JUnitFailure `xml:"skipped,omitempty"`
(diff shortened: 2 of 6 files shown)


pkg/ruletest/junit.go
```diff
@@ -38,43 +59,99 @@ type JUnitFailure struct {
...
-	var suitesList []JUnitTestSuite
-	for _, suite := range slices.Collect(maps.Values(suitesMap)) {
-		suitesList = append(suitesList, *suite)
-	}
+	res := JUnitTestSuites{}
 
-	return JUnitTestSuites{
-		TestSuites: suitesList,
+	for _, suite := range slices.SortedFunc(
+		slices.Collect(maps.Values(suitesMap)),
```
dakshhhhh16: I think the extra slices.Collect here is the culprit behind the CI failures. SortedFunc expects the sequence returned by maps.Values, but Collect has already converted it into a slice. Could we either pass maps.Values(suitesMap) directly, or collect first and then use slices.SortFunc?

pkg/ruletest/junit.go
```diff
@@ -38,43 +59,99 @@ type JUnitFailure struct {
...
+		func(a, b *JUnitTestSuite) int {
+			return cmp.Compare(a.Name, b.Name)
```
dakshhhhh16: Could we use File as the tie-breaker here, or simply sort by File? Suites from different directories can have the same name, the test already has both foo/file_b.star and bar/file_b.star. Since these values come from a map, equal names leave their final order unpredictable and could make the XML test flaky.

pkg/ruletest/junit.go
```diff
@@ -38,43 +59,99 @@ type JUnitFailure struct {
...
+					{
+						Name:  "coverage.statements.pct",
+						Value: "0",
+					},
+				},
+			}
+			suite := suitesMap[run.BaseDir]
+			for _, rule := range run.UncoveredRules() {
+				// Represent each uncovered rule as a test case with a failure
+				suite.TestCases = append(suite.TestCases, JUnitTestCase{
```
dakshhhhh16: Could we keep the suite counts in sync with these entries? We add one <testcase> for every uncovered rule, but suite.Tests remains zero, so the report can say tests="0" while still containing test cases. Some JUnit consumers trust that attribute and may hide or ignore these entries.

pkg/ruletest/runner_test.go
```diff
@@ -29,20 +29,21 @@ func testDir(t *testing.T, r *Runner, dir string) {
...
+					}
+					return
 				}
-				return
-			}
-			for _, msg := range result.Failures {
-				t.Error(msg)
-			}
-		})
+				for _, msg := range result.Failures {
```
dakshhhhh16, line 43: Looks like we only surface assertion failures here. Now that execution problems are stored separately in result.Errors, a normal test that hits a Starlark runtime error can still appear successful to this Go test. Could we report result.Errors here as well?

dakshhhhh16 (dismissed):

pkg/ruletest/junit.go
evankanderson: Yeah, I was trying to fix this with a remote shell, and didn't do a test run before pushing due to timing.  Fixed now, along with the test.

pkg/ruletest/junit.go
evankanderson: Yeah, switched to file (which the test made obvious).

pkg/ruletest/junit.go
evankanderson: That's interesting... we actually have zero test cases here, we're reporting rules (think Java classes) that have no tests.  Maybe a better way to represent would be to have a testsuite for each uncovered rule, with no files in it?  Right now, we don't track where we loaded the rule from, though.

pkg/ruletest/runner_test.go
evankanderson: Good catch.  done!

pkg/ruletest/junit.go
```diff
@@ -38,43 +59,99 @@ type JUnitFailure struct {
...
+					Message: "Test error",
+					Body:    strings.Join(res.Errors, "\n"),
+				}
 			}
-		}
 
-		suite.TestCases = append(suite.TestCases, tc)
+			suite.TestCases = append(suite.TestCases, tc)
+		}
+		if len(run.UncoveredRules()) > 0 {
```
krrish175-byte: `run.UncoveredRules()` is called twice here (line 119 and again line 131), which rebuilds the list via `slices.Clone` + `slices.DeleteFunc` both times. Store it once:

```go
uncovered := run.UncoveredRules()
if len(uncovered) > 0 {
    ...
    for _, rule := range uncovered {

File: pkg/ruletest/junit.go (hunk 3 above)
Comment (krrish175-byte) [reply pkg/ruletest/junit.go]: This is still the Tests=0 problem @dakshhhhh16 flagged. We create TestCases entries but never increment suite.Tests, so JUnit consumers that trust the tests attribute will see tests="0" and may hide or skip these entries entirely.

The simplest fix is to increment suite.Tests inside the loop:

```go
for _, rule := range uncovered {
    suite.Tests++
    suite.TestCases = append(suite.TestCases, JUnitTestCase{Name: rule})
}
```

Alternatively, if you go with the "one empty testsuite per uncovered rule" design you mentioned, that avoids the issue altogether, but needs the tracking of where each rule was loaded from.

pkg/ruletest/runner.go
```diff
@@ -72,20 +74,60 @@ func (tr *testCaseRunner) runFile(filename string, src any) (starlark.StringDict
...
+
 // TestResult holds the outcome of a single Starlark test function.
 type TestResult struct {
 	Filename string
 	Name     string
 	Failures []string
+	Errors   []string
+
+	// printOutput contains the output of any `print` statements executed during the test.
+	printOutput strings.Builder
```
krrish175-byte: `printOutput strings.Builder` is never written to, so `Output()` always returns `""`. Shipping an exported method on top of a permanently-empty field is misleading. I would suggest removing the field and the `Output()` method (lines 107–109) until the print-capture TODO is actually implemented. 

A `// TODO: implement print capture` comment on the struct is enough to signal intent without exposing a stub API.

krrish175-byte (changes requested):
Just a few comments from my side but overall looks good.

pkg/ruletest/runner.go
evankanderson: Done.

pkg/ruletest/junit.go
evankanderson: Thanks, following the rewrite suggestion below, we no longer need this "if"

pkg/ruletest/junit.go
evankanderson: I ended up doing the "one empty testsuite per uncovered rule"

cmd/dev/app/test/test.go
```diff
@@ -68,10 +68,12 @@ func CmdTest() *cobra.Command {
...
-			for _, res := range results {
-				if len(res.Failures) > 0 {
-					finalErr = errors.New("one or more tests failed")
-					break
+			for _, run := range results {
+				for _, res := range run.Results {
+					if len(res.Failures)+len(res.Errors) > 0 {
+						finalErr = errors.New("one or more tests failed")
+						break
+					}
```
krrish175-byte, line 75: This `break` only exits the inner `for _, res` loop, it doesn't stop iterating over `runs`. The old code used a single-level range and truly short-circuited on first failure. This change causes the loop to always scan all runs and all results, even after setting `finalErr`. Not a correctness bug (the error message is idempotent), but it lost the early-exit property and may surprise a future reader. Consider a labeled break (`outer:`) or restructuring as `goto` / checking `finalErr` in the outer loop condition.

cmd/dev/app/test/test.go
```diff
@@ -80,27 +82,44 @@ func CmdTest() *cobra.Command {
...
+				cmd.Printf("ERROR: %s/%s\n", res.Filename, res.Name)
+				for _, e := range res.Errors {
+					cmd.Printf("  - %s\n", e)
+				}
+			} else {
+				cmd.Printf("PASS: %s/%s\n", res.Filename, res.Name)
+			}
+		}
+		// TODO: hide this behind a flag
+		if uncovered := run.UncoveredRules(); coverage && len(uncovered) > 0 {
```
krrish175-byte: This `// TODO: hide this behind a flag` comment is stale, the condition is already gated on the `coverage` flag. Please remove it.

pkg/ruletest/runner.go
```diff
@@ -72,20 +74,54 @@ func (tr *testCaseRunner) runFile(filename string, src any) (starlark.StringDict
...
 }
 
+// Error is called by the starlarktest.Reporter interface when a test case fails.
 func (tr *testCaseRunner) Error(args ...any) {
 	tr.failures = append(tr.failures, fmt.Sprint(args...))
 }
 
+// UsedRules returns a set of rule names that were evaluated during the test case execution.
+func (tr *testCaseRunner) UsedRules() map[string]struct{} {
+	return maps.Clone(tr.usedRules)
```
krrish175-byte: `UsedRules` is exported (capital `U`) but `testCaseRunner` is unexported, so this method can't be called from outside the package anyway. Naming it `usedRules()` (lowercase) would match the unexported receiver and make intent clearer. The only caller is `runOneTest` within the same package.

pkg/ruletest/junit_test.go
```diff
@@ -5,90 +5,145 @@ package ruletest
...
-func TestAsJUnit_EmptyInput(t *testing.T) {
-	t.Parallel()
-	suites := AsJUnit(nil)
 
-	if len(suites.TestSuites) != 0 {
-		t.Errorf("expected 0 suites, got %d", len(suites.TestSuites))
+	coverage100 := []Property{{
+		Name:  "coverage.statements.pct",
+		Value: "100",
+	}}
```
krrish175-byte, line 18: `coverage.statements.pct` is a Cobertura/Codecov convention that tools interpret as line/statement-level code coverage. What this actually reports is "was this rule invoked by at least one test", a rule with one `eval()` call gets `100%` regardless of how many of its internal branches are exercised. This could confuse coverage dashboards (e.g., Codecov, SonarQube) that display it as "100% line coverage." Consider using a less ambiguous property name like `coverage.rules.pct` or adding an XML comment clarifying the semantics.

pkg/ruletest/runner.go
```diff
@@ -72,20 +74,54 @@ func (tr *testCaseRunner) runFile(filename string, src any) (starlark.StringDict
...
+
+// UncoveredRules returns a list of rule names that were loaded but not evaluated during the test run.
+func (tr *TestRun) UncoveredRules() []string {
+	uncovered := slices.Clone(tr.LoadedRules)
+	for _, res := range tr.Results {
+		uncovered = slices.DeleteFunc(uncovered, func(rule string) bool {
+			_, ok := res.EvaluatedRules[rule]
+			return ok
+		})
+	}
```
krrish175-byte, line 118: This is O(results × rules), for each test result it linearly scans `uncovered`. Since `EvaluatedRules` is already a map, you can merge all evaluated rules first, then do one `DeleteFunc`:

```go
 func (tr *TestRun) UncoveredRules() []string {
     evaluated := make(map[string]struct{})
     for _, res := range tr.Results {
         for rule := range res.EvaluatedRules {
             evaluated[rule] = struct{}{}
         }
     }
     return slices.DeleteFunc(slices.Clone(tr.LoadedRules), func(r string) bool {
         _, ok := evaluated[r]
         return ok
     })
 }
```

Not a problem at typical scale, but the current form is easy to misread as "each result independently computes uncovered" when it's actually accumulating deletions across results.

krrish175-byte (commented):
Once the stale TODO and the break scoping are fixed, this is ready. Rest LGTM

cmd/dev/app/test/test.go
evankanderson: I switched this to just be an early-return.

cmd/dev/app/test/test.go
evankanderson: Whoops, thanks!  That leaked through my implementation process (get it working, then flag-protect).

pkg/ruletest/runner.go
evankanderson: Actually, I removed this method, since it was only being used internally, and the lowercase name would collide with the field name.

pkg/ruletest/junit_test.go
evankanderson: I *did* want this to show up as coverage data in UI renderers.  It's not entirely accurate right now, but it seemed like having "is covered" and "is not covered" would be an improvement, and marking the rules as 0% line coverage could help folks figure out what rules they still needed to test / should prioritize.

I don't see an easy way to get coverage through our current stack, though that would be a nice future refinement.

pkg/ruletest/runner.go
evankanderson: I considered that, but:

1. I'm not sure it's more efficient.  If the number of loaded rules is large and the number of results is large, but few rules are covered, this will loop a lot.  If most rules are covered, `uncovered` will shrink fairly quickly, while doing a large merge of maps may be less efficient.  I'd considered the other pattern, but decided this was simpler given point 2:
2. This is called at most twice per directory per invocation (once for text output and once for JUnit).  Given reasonable sizes (less than 1000 rules and 1000 tests / directory), this is probably a few million instructions, which is a few milliseconds on a GHz CPU.

krrish175-byte (approved):
lgtm. @dakshhhhh16 what do you say?

PR #6751, opened 2026-09-06
fix(gitlab): paginate project listing

### LINKED ISSUE #6750: GitLab repo listing silently caps at the first page of projects
### Describe the issue

ListAllRepositories in internal/providers/gitlab/repo_lister.go makes a single request to the projects endpoint:

	glRESTGet(ctx, c, fmt.Sprintf("projects?min_access_level=%d", ...), &managedProjects)

GitLab paginates collection responses at per_page items (20 by default) and reports further pages via the X-Next-Page response header, which
glRESTGet discards. So any account with more than 20 accessible projects silently loses everything past the first page - the repo registration
flow just never sees those projects, with no error or warning.

### To Reproduce

1. Register a GitLab provider for an account with more than 20 projects  at the required access level.
2. List repositories for registration - only the first 20 appear.

Reproduced in a unit test with a two-page mock server: the current code requests no page parameters and returns only the first page's projects.

I have a fix with tests ready and will open a PR.

### To Reproduce

_No response_

_No response_

NAME-ASHWANIYADAV:
# Summary

ListAllRepositories made a single request to the projects endpoint, and GitLab caps unpaginated responses at 20 items, so accounts with more than 20 accessible projects silently lost the rest during repo registration.

This adds a paginated variant of the REST helper that follows the X-Next-Page response header at GitLab's maximum page size (100), preserves the caller's query parameters, and stops on an empty page so a misbehaving server can't cause an infinite loop. The 404 -> ErrEntityNotFound mapping matches the existing helper. The project listing now uses it; single-object callers of glRESTGet are unchanged.

Fixes #6750

# Testing

- New tests for the helper: multi-page traversal (asserts page=1 then page=2 with per_page=100 and preserved query params), single page, empty collection, a server that keeps reporting a next page with no items, 404 mapping, and an error on a later page failing the whole listing.
- New ListAllRepositories tests drive the real client against a two-page httptest server and assert all three projects come back in
  order. Without the fix this test fails - the old code sends no page parameters and returns only the first page.
- The full internal/providers/gitlab package passes; go vet is clean.

### DIFF
--- a/internal/providers/gitlab/gitlab_rest.go
+++ b/internal/providers/gitlab/gitlab_rest.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"strconv"
 
 	provifv1 "github.com/mindersec/minder/pkg/providers/v1"
 )
@@ -103,6 +104,91 @@ func glRESTGet[T any](ctx context.Context, cli genericRESTClient, path string, o
 	}
 
 	return nil
+}
+
+const (
+	// glPerPage is the page size requested from collection endpoints.
+	// 100 is the maximum GitLab allows.
+	glPerPage = 100
+
+	// glNextPageHeader is the response header GitLab uses to point at the
+	// next page of an offset-paginated collection. It is empty on the
+	// last page.
+	glNextPageHeader = "X-Next-Page"
+)
+
+// glRESTGetPaginated retrieves every page of a GitLab collection endpoint
+// and appends the items to out. GitLab caps responses at per_page items
+// (20 by default), so collection endpoints have to follow the X-Next-Page
+// response header to retrieve the full result set.
+func glRESTGetPaginated[T any](ctx context.Context, cli genericRESTClient, path string, out *[]T) error {
+	for page := "1"; page != ""; {
+		items, nextPage, err := glRESTGetPage[T](ctx, cli, path, page)
+		if err != nil {
+			return err
+		}
+		*out = append(*out, items...)
+
+		// An empty page means there is nothing left to fetch regardless
+		// of what the header says, which also guards against a server
+		// that keeps reporting a next page.
+		if len(items) == 0 {
+			break
+		}
+		page = nextPage
+	}
+
+	return nil
+}
+
+// glRESTGetPage retrieves a single page of a GitLab collection endpoint and
+// returns the items along with the next page reported by the server.
+func glRESTGetPage[T any](ctx context.Context, cli genericRESTClient, path, page string) ([]T, string, error) {
+	pagedPath, err := pathWithPagination(path, page)
+	if err != nil {
+		return nil, "", err
+	}
+
+	req, err := cli.NewRequest(http.MethodGet, pagedPath, nil)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to create request: %w", err)
+	}
+
+	resp, err := cli.Do(ctx, req)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to get resource '%s': %w", pagedPath, err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		if resp.StatusCode == http.StatusNotFound {
+			return nil, "", provifv1.ErrEntityNotFound
+		}
+		return nil, "", fmt.Errorf("failed to get resource '%s': %s", pagedPath, resp.Status)
+	}
+
+	var items []T
+	if err := json.NewDecoder(resp.Body).Decode(&items); err != nil {
+		return nil, "", fmt.Errorf("failed to decode response: %w", err)
+	}
+
+	return items, resp.Header.Get(glNextPageHeader), nil
+}
+
+// pathWithPagination sets the page and per_page query parameters on path,
+// preserving any query parameters it already carries.
+func pathWithPagination(path, page string) (string, error) {
+	u, err := url.Parse(path)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse path '%s': %w", path, err)
+	}
+
+	q := u.Query()
+	q.Set("page", page)
+	q.Set("per_page", strconv.Itoa(glPerPage))
+	u.RawQuery = q.Encode()
+
+	return u.String(), nil
 }
 
 func getParsedURL(endpoint, path string) (*url.URL, error) {
--- a/internal/providers/gitlab/gitlab_rest_test.go
+++ b/internal/providers/gitlab/gitlab_rest_test.go
@@ -6,6 +6,7 @@ package gitlab
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -18,6 +19,7 @@ import (
 	"github.com/mindersec/minder/internal/providers/credentials"
 	"github.com/mindersec/minder/internal/util/ptr"
 	minderv1 "github.com/mindersec/minder/pkg/api/protobuf/go/minder/v1"
+	provifv1 "github.com/mindersec/minder/pkg/providers/v1"
 )
 
 type mockGitlabClient struct {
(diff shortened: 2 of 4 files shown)


internal/providers/gitlab/gitlab_rest.go
```diff
@@ -105,6 +106,91 @@ func glRESTGet[T any](ctx context.Context, cli genericRESTClient, path string, o
...
+	// last page.
+	glNextPageHeader = "X-Next-Page"
+)
+
+// glRESTGetPaginated retrieves every page of a GitLab collection endpoint
+// and appends the items to out. GitLab caps responses at per_page items
+// (20 by default), so collection endpoints have to follow the X-Next-Page
+// response header to retrieve the full result set.
+func glRESTGetPaginated[T any](ctx context.Context, cli genericRESTClient, path string, out *[]T) error {
+	for page := "1"; page != ""; {
```
intelligent-ears, line 125: This loop has no upper bound. The only exits are an empty `X-Next-Page` and an empty page of items — a server that returns items on every page and always reports a next page will loop forever and grow `out` without limit. The endpoint is the user's enrolled GitLab instance, so this is remote-controlled input to the Minder server.

Parsing the header as an int and requiring it to advance also removes the need to feed an arbitrary header value back into the query string:

```go
// glMaxPages bounds the paginated fetch so a misbehaving server cannot
// make us loop indefinitely.
const glMaxPages = 1000

func glRESTGetPaginated[T any](ctx context.Context, cli genericRESTClient, path string, out *[]T) error {
	page := 1
	for range glMaxPages {
		items, nextPage, err := glRESTGetPage[T](ctx, cli, path, strconv.Itoa(page))
		if err != nil {
			return err
		}
		*out = append(*out, items...)
		if len(items) == 0 || nextPage == "" {
			return nil
		}
		next, err := strconv.Atoi(nextPage)
		if err != nil || next <= page {
			// Bogus or non-advancing next page; nothing more to fetch.
			return nil
		}
		page = next
	}
	return fmt.Errorf("too many pages listing '%s' (limit %d)", path, glMaxPages)
}
```
Returning an error at the cap rather than truncating matters — silently returning a partial result set is the bug this PR should be fixing.

internal/providers/gitlab/gitlab_rest.go
```diff
@@ -105,6 +106,91 @@ func glRESTGet[T any](ctx context.Context, cli genericRESTClient, path string, o
...
+		}
+		page = nextPage
+	}
+
+	return nil
+}
+
+// glRESTGetPage retrieves a single page of a GitLab collection endpoint and
+// returns the items along with the next page reported by the server.
+func glRESTGetPage[T any](ctx context.Context, cli genericRESTClient, path, page string) ([]T, string, error) {
```
intelligent-ears, line 146: This re-implements everything `glRESTGet` already does: request construction, `Do`, the status check, the 404 → `ErrEntityNotFound` mapping and the error strings. Two copies will drift. Could you pull the shared part into a helper that returns a checked response and have both call it?

internal/providers/gitlab/gitlab_rest.go
```diff
@@ -105,6 +106,91 @@ func glRESTGet[T any](ctx context.Context, cli genericRESTClient, path string, o
...
+	var items []T
+	if err := json.NewDecoder(resp.Body).Decode(&items); err != nil {
+		return nil, "", fmt.Errorf("failed to decode response: %w", err)
+	}
+
+	return items, resp.Header.Get(glNextPageHeader), nil
+}
+
+// pathWithPagination sets the page and per_page query parameters on path,
+// preserving any query parameters it already carries.
```
intelligent-ears, line 179: Small accuracy fix: `page` and `per_page` are overwritten rather than preserved if the caller already set them. Worth saying so, since a future caller may reasonably expect its own `per_page` to survive.

internal/providers/gitlab/repo_lister.go
```diff
@@ -23,7 +23,7 @@ var minAccessLevelControl = 25 // "Security Manager"
 
 func (c *gitlabClient) ListAllRepositories(ctx context.Context) ([]*minderv1.Repository, error) {
 	managedProjects := []*gitlab.Project{}
-	if err := glRESTGet(ctx, c, fmt.Sprintf("projects?min_access_level=%d", minAccessLevelControl), &managedProjects); err != nil {
+	if err := glRESTGetPaginated(ctx, c, fmt.Sprintf("projects?min_access_level=%d", minAccessLevelControl), &managedProjects); err != nil {
```
intelligent-ears, line 26:
```suggestion
	path := fmt.Sprintf("projects?min_access_level=%d", minAccessLevelControl)
	if err := glRESTGetPaginated(ctx, c, path, &managedProjects); err != nil {
```

internal/providers/gitlab/repo_lister_test.go
```diff
@@ -0,0 +1,79 @@
...
+		]`,
+		"2": `[
+			{"id": 3, "name": "three", "namespace": {"path": "group"}}
+		]`,
+	}
+
+	var requestedPages []string
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		q := r.URL.Query()
+		assert.Equal(t, "25", q.Get("min_access_level"), "query params from the caller must be preserved")
```
intelligent-ears, line 35: Please derive this from the constant so the test follows it if the access level changes: `strconv.Itoa(minAccessLevelControl)`.

internal/providers/gitlab/gitlab_rest_test.go
```diff
@@ -355,3 +357,122 @@ func Test_getParsedURL(t *testing.T) {
...
+		}))
+		defer ts.Close()
+
+		var out []string
+		err := glRESTGetPaginated(context.Background(), newPagingClient(ts), "test", &out)
+		require.NoError(t, err)
+		assert.Empty(t, out)
+	})
+
+	t.Run("stops when the server keeps reporting a next page with no items", func(t *testing.T) {
```
intelligent-ears, line 430: Good case to have, but it only covers the empty-page half. Could you add the non-empty version — a handler that always returns `["a"]` with `X-Next-Page: 2` — asserting the fetch terminates? That's the case that currently hangs.

intelligent-ears (changes requested):
Thanks for picking this up — the diagnosis is correct and the fix is in the right place. Following `X-Next-Page` is the right mechanism for GitLab's offset pagination, requesting at `per_page=100` is the sensible default, and pulling the single-page fetch into its own function so `defer resp.Body.Close()` doesn't accumulate inside the loop is the detail most first attempts at this get wrong. The tests genuinely exercise the behaviour — asserting the actual page sequence and that the caller's query params survive is more than I expected.

internal/providers/gitlab/repo_lister_test.go
evankanderson: This depends -- if the constant is something that we control, then it makes sense to avoid duplication.  If this constant is something that **GitLab** defines, then the using the constant as a string in the test helps to ensure that we're not deviating from the spec.

In this case, I think we could go either way -- we defined the constant based on GitLab's access levels (and I'm not sure it's correct -- I was trying to use a Security Manager service account with `repo register`, and it failed because [the role doesn't have permissions to configure webhooks on a project](https://docs.gitlab.com/user/permissions/#projects)).

internal/providers/gitlab/gitlab_rest.go
```diff
@@ -105,6 +106,91 @@ func glRESTGet[T any](ctx context.Context, cli genericRESTClient, path string, o
...
+		*out = append(*out, items...)
+
+		// An empty page means there is nothing left to fetch regardless
+		// of what the header says, which also guards against a server
+		// that keeps reporting a next page.
+		if len(items) == 0 {
+			break
+		}
+		page = nextPage
+	}
```
evankanderson, line 139: Can we put some kind of bound (say, 1000?) on this, so that we don't completely melt down if someone with a million group memberships tries calling Minder?  I think if you have 1 million memberships, any sort of enumeration is just going to break, but we should break in a controlled way.

internal/providers/gitlab/gitlab_rest.go
evankanderson, line 146: Do we need this to be prefixed with "gl", given that it's in the `gitlab` packge?  Here, and elsewhere.

Yeah, someone did some of that before, but let's stop.

internal/providers/gitlab/gitlab_rest.go
```diff
@@ -105,6 +106,91 @@ func glRESTGet[T any](ctx context.Context, cli genericRESTClient, path string, o
...
+	// next page of an offset-paginated collection. It is empty on the
+	// last page.
+	glNextPageHeader = "X-Next-Page"
+)
+
+// glRESTGetPaginated retrieves every page of a GitLab collection endpoint
+// and appends the items to out. GitLab caps responses at per_page items
+// (20 by default), so collection endpoints have to follow the X-Next-Page
+// response header to retrieve the full result set.
+func glRESTGetPaginated[T any](ctx context.Context, cli genericRESTClient, path string, out *[]T) error {
```
evankanderson, line 124: Why does this function use an out parameter, but `glRESTGetPage` uses a return value for the slice?

internal/providers/gitlab/gitlab_rest.go
evankanderson: +1 on extending `glRESTGet` (maybe combine with a rename to `restGet` or the like), even if it means touching all the call sites.

internal/providers/gitlab/gitlab_rest.go
```diff
@@ -105,6 +106,91 @@ func glRESTGet[T any](ctx context.Context, cli genericRESTClient, path string, o
...
+		return nil, "", fmt.Errorf("failed to decode response: %w", err)
+	}
+
+	return items, resp.Header.Get(glNextPageHeader), nil
+}
+
+// pathWithPagination sets the page and per_page query parameters on path,
+// preserving any query parameters it already carries.
+func pathWithPagination(path, page string) (string, error) {
+	u, err := url.Parse(path)
```
evankanderson, line 181: Rather than calling `url.Parse` on every page, what do you think of having the caller pass in a `url.URL` to `glRESTGetPage` (or any of the simplifications thereof)?  I think the rest of this could then live in-line in `glRESTGetPaginated`.

internal/providers/gitlab/gitlab_rest_test.go
```diff
@@ -355,3 +357,122 @@ func Test_getParsedURL(t *testing.T) {
...
+			w.WriteHeader(http.StatusNotFound)
+		}))
+		defer ts.Close()
+
+		var out []string
+		err := glRESTGetPaginated(context.Background(), newPagingClient(ts), "test", &out)
+		require.ErrorIs(t, err, provifv1.ErrEntityNotFound)
+	})
+
+	t.Run("error on a later page fails the whole listing", func(t *testing.T) {
```
evankanderson, line 461: Do we want an error on page 4 to prevent _any_ repos from being registered?  I feel like this may be a case to log an error and continue with partial data.

internal/providers/gitlab/repo_lister_test.go
evankanderson, line 1: This is optional, but I'm a fan (particularly within a given package) of larger tests that exercise things end-to-end over a lot of small unit tests.  You could fold the tests in `gitlab_rest_test.go` into the repo lister and still get equivalent coverage, I think.

evankanderson (commented):
Nice improvements, thank you!  A few comments, but this looks like a good idea.

Issue #6436, opened 2026-05-04
Roadmap: Expand OCI Registry Provider Coverage


evankanderson:
## This is a Roadmap Epic, and needs to be broken into smaller issues

In addition to adding support for Git Forges (where the Minder entity model is
well-tested), Minder has some support for artifact repositories (primarily OCI
repositories such as GHCR and DockerHub). This should be expanded, possibly
following the "rule-driven child entities" work in #6105 
to enable automatic generation of dependent entities given a known parent.

Additionally, further work should be done to enable managing OCI registry
security properties:

- Verify settings (immutable tags, MFA enabled, permissions, etc) for
  repositories and possibly hierarchical namespaces
- Establish best practices for credential handling for remote OCI registries
- Expand support for validating both
  [artifacts](https://oras.land/docs/glossary/#artifacts) and
  [config](https://oras.land/docs/glossary/#config) contents as well as via the
  [referrers API](https://oras.land/docs/glossary/#manifest-referrers-api)

Concrete OCI registries which may need specific implementations (particularly
around credential handling; possibly also enumeration of registries, namespaces,
and repositories) include:

- Major Cloud Providers (AWS ECR, Google GCR, Azure ACR, etc)
- Major OSS Providers (Quay.io, Harbor, GHCR, DockerHub)

### Solution Proposal

* Document configuration of OCI registry providers
* Add support for repositories as well as artifacts
   * Add support for configuring per-registry API settings such as immutable tags on a per-repository basis
* Add support for at least 3 more OCI registries
* Document (example rules + documentation) checking at least 3 additional artifact properties besides sigstore signatures on an artifact.  These should be checked using either the JQ or Rego engine, possibly using DataSource integration if additional data is needed (like the referrers API for SBOMs).  Possible sample rules:
   * All container images support x86 and ARM64 architectures
   * Tagged images have SBOMs which contain a common minimum set of elements
   * Image layers are below a size limit, with a max number of layers

shreyaabaranwal:
Hi @evankanderson , I just wrapped up #6729 and would love to help with this epic.

A few things I noticed while looking through the existing providers:

* `dockerhub.go` builds on the generic OCI client and adds DockerHub-specific auth/config handling.
* I think we could follow the same pattern for a **Quay.io or Harbor provider**.
[...]

PR #6753, opened 2026-09-08
Add Quay.io provider (first slice of #6436)

### LINKED ISSUE #6752: Add Quay.io provider (first slice of #6436)
First scoped slice of #6436 (Roadmap: Expand OCI Registry Provider Coverage).

## Scope
- New Quay.io provider following the existing DockerHub pattern (`oci.OCI` generic client + registry-specific wrapper).
- Auth: document setup of a Quay.io "Robot account" (non-human, org/repo-scoped credential) as the recommended credential, mirroring how GitLab's scoped service-account credential is documented , rather than a personal OAuth token.
- Entity-interface methods (SupportsEntity, RegisterEntity, etc.) left stubbed, matching DockerHub's current behavior , entity-management design is being tracked separately by @evankanderson.
- No new rule types / policy rules in this slice , goal is DockerHub-equivalent functionality.
- Short-lived credential exchange explicitly out of scope, deferred.

Discussed and scoped with @evankanderson in Slack.

shreyaabaranwal:
Fixes #6752

First slice of #6436 , Quay.io provider, mirroring the existing DockerHub pattern (`oci.OCI` generic client + registry-specific wrapper).

## Scope
- New `internal/providers/quay/` package following `dockerhub.go`'s structure
- Auth via Quay.io Robot accounts (documented, mirroring GitLab's scoped service-account pattern) , long-lived token, per discussion on #6436
- Entity-interface methods left stubbed, matching DockerHub's current behavior
- No new rule types in this slice

## Sequencing note
This PR's own manager.go now includes the OAuthManager fix directly (see the "Address review feedback" commit) ,  no longer blocked on a separate PR. The DockerHub-side companion fix is tracked separately in #6756.
Build/tests/lint all clean.


internal/providers/quay/auth.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/mindersec/minder/internal/providers/manager"
+	provv1 "github.com/mindersec/minder/pkg/providers/v1"
+)
+
+// NewOAuthConfig implements the providerClassOAuthManager interface.
+// Quay only supports the user-input (token) authorization flow, so there
+// is no OAuth2 authorization code flow to configure.
+func (*providerClassManager) NewOAuthConfig(_ db.ProviderClass, _ bool) (*oauth2.Config, error) {
+	return nil, fmt.Errorf("quay provider does not support the OAuth2 authorization code flow")
+}
```
evankanderson, line 22: Why do we need this method at all?  (I'm seeing that DockerHub does not include it.)

internal/providers/quay/auth.go
```diff
@@ -0,0 +1,45 @@
...
+// account token) rather than obtained via an OAuth2 exchange, so
+// validation is limited to a basic sanity check.
+func (*providerClassManager) ValidateCredentials(
+	_ context.Context, cred provv1.Credential, _ *manager.CredentialVerifyParams,
+) error {
+	switch c := cred.(type) {
+	case provv1.OAuth2TokenCredential:
+		if _, err := c.GetAsOAuth2TokenSource().Token(); err != nil {
+			return fmt.Errorf("cannot get token from credential: %w", err)
+		}
```
evankanderson: Why have this code if we're not doing an OAuth token?

internal/providers/quay/manager.go
```diff
@@ -0,0 +1,123 @@
...
+	if encToken.EncryptedAccessToken.Valid {
+		encryptedData, err = crypto.DeserializeEncryptedData(encToken.EncryptedAccessToken.RawMessage)
+		if err != nil {
+			return nil, err
+		}
+	} else if encToken.EncryptedToken.Valid {
+		encryptedData = crypto.NewBackwardsCompatibleEncryptedData(encToken.EncryptedToken.String)
+	} else {
+		return nil, fmt.Errorf("no secret found for provider %s", encToken.Provider)
+	}
```
evankanderson: Since this is new code, we should only need the `EncryptedAccessToken` path, not the `EncryptedToken` path.  (I suspect we can retire that code entirely at this point, though I'd want to do a database check before approving such a PR.)

internal/providers/quay/quay.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// Ensure that the Quay client implements the ImageLister interface
+var _ provifv1.ImageLister = (*quayImageLister)(nil)
+
+// New creates a new Quay client
+func New(cred provifv1.OAuth2TokenCredential, cfg *minderv1.QuayProviderConfig) (*quayImageLister, error) {
+	cli := oauth2.NewClient(context.Background(), cred.GetAsOAuth2TokenSource())
+
+	u, err := url.Parse("https://quay.io/api/v1/repository")
```
evankanderson, line 61: Does Quay serve the OCI v2 repository API at this path?

evankanderson (commented):
Thanks for the documentation!  This looks pretty good, but there's some old code we should kill off here.

internal/providers/quay/auth.go
shreyaabaranwal: Good catch ,  removed it. DockerHub's manager doesn't implement this either, so Quay didn't need it.

internal/providers/quay/auth.go
shreyaabaranwal: Removed , same reasoning, this branch wasn't needed since we're not doing an OAuth token flow here.

internal/providers/quay/manager.go
shreyaabaranwal: Removed the EncryptedToken branch and its TODO. This is brand-new code with no existing secrets to migrate, so it only needs the EncryptedAccessToken path. Left DockerHub's manager.go untouched, since retiring that path there needs a DB check first, per your comment.

internal/providers/quay/quay.go
shreyaabaranwal: - Quay uses two separate APIs for different purposes: its REST API (quay.io/api/v1/repository) for ListImages/repository listing, and the standard OCI /v2/... API for tags, manifests, and digests.

- I verified both paths directly: the REST endpoint returns the expected repository response, while oci.New(cred, "quay.io", ...) correctly uses Quay’s OCI Distribution API through go-containerregistry.

- So this separation is intentional and follows Quay’s API design , REST for repository discovery, OCI for image operations.

evankanderson (approved):
It looks like this clone of the DockerHub code is lightly tested (coverage went down).  If you're going to build out more of these, we should find a way to bring coverage up again.

PR #6755, opened 2026-09-08
Add Starlark JSON module

evankanderson:
# Summary

Adds JSON encoding and decoding tools to the `mindev test` Starlark runtime.  This can be handy for varying JSON API output during tests; building a `json.encode` method in Starlark is not fully feasible, as Starlark prevents recursion and other Turing-complete features to ensure predictable runtime.

# Testing

Added a test which performs JSON encoding and decoding.  Tested that `assert.eq("", json_str)` in the test fails, then replaced with the correct value.  Starlark seems to preserve dict key order for iteration.

### DIFF
--- a/pkg/ruletest/runner.go
+++ b/pkg/ruletest/runner.go
@@ -16,6 +16,7 @@ import (
 	"sort"
 	"strings"
 
+	"go.starlark.net/lib/json"
 	"go.starlark.net/starlark"
 	"go.starlark.net/starlarktest"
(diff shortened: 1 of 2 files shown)


krrish175-byte (approved):
The test is correct and deterministic, but the comment reads like it's relying on unspecified behavior. Consider changing it to "Starlark dicts preserve insertion order (per spec)." or just dropping the comment since the test is self-explanatory. LGTM

PR #6756, opened 2026-09-08
Fix DockerHub provider-class manager missing OAuthManager

shreyaabaranwal:
## Summary

DockerHub's provider-class manager didn't implement OAuthManager, which StoreProviderToken (the token-enrollment RPC) requires -- `minder provider enroll --class dockerhub -t <token>` failed on main as a result.

Found while building a new Quay.io provider that mirrors this same manager pattern (see #6753).

## Testing

`go test ./internal/providers/dockerhub/... -v`, `golangci-lint` , all clean, no regressions.


internal/providers/dockerhub/auth.go
```diff
@@ -0,0 +1,45 @@
...
+	"golang.org/x/oauth2"
+
+	"github.com/mindersec/minder/internal/db"
+	"github.com/mindersec/minder/internal/providers/manager"
+	provv1 "github.com/mindersec/minder/pkg/providers/v1"
+)
+
+// NewOAuthConfig implements the providerClassOAuthManager interface.
+// DockerHub only supports the user-input (token) authorization flow, so
+// there is no OAuth2 authorization code flow to configure.
```
evankanderson: Maybe add a comment that this is needed by the provider interface to store _any_ kind of credential?

evankanderson (dismissed):
I'm willing to approve this, but it would be great to have some sort of automated test coverage (probably in controlplane, sadly) that shows we can actually create some form of Dockerhub provider with some parameters.

I'm not going to block this fix on that testing; let me know whether you'd prefer to add that test or get this merged and file a subsequent issue for testing (which might or might not fall on you).

internal/providers/dockerhub/auth.go
shreyaabaranwal: Let's merge this now , I'll file a follow-up issue for the controlplane test coverage.

 Thanks for the flexibility on that!

evankanderson (approved):

PR #6769, opened 2026-09-09
fix: apply repository filter in ListArtifacts

th3Gli4ch:
## Summary

- `ListArtifacts` with `from="repository=<name>"` silently ignored the filter and returned artifacts from all repositories in the project
- The `repoSlubList` was parsed correctly but the loop only checked `OriginatedFrom.Valid` - no name comparison was ever made against the slug list
- Fix resolves each slug to a repo entity UUID upfront via `GetEntityByName`, then checks `OriginatedFrom.UUID` membership per artifact in O(1)

## Root cause

`internal/controlplane/handlers_artifacts.go` lines 274–281 had a TODO comment where the actual filtering logic should have been:

```go
if len(filter.repoSlubList) > 0 {
    if !ent.OriginatedFrom.Valid {
        continue
    }
    // TODO: Implement efficient repo filtering  ← no name check happened
}

Changes

- internal/controlplane/handlers_artifacts.go: resolve slugs to repo entity UUIDs before the artifact loop, then filter by set membership
- internal/controlplane/handlers_artifacts_test.go: table-driven tests covering no-filter, single repo, unknown repo, and multi-repo cases

Test plan

- [ ] No filter → all artifacts returned
- [ ] repository=repo-a → only artifacts whose parent is repo-a are returned
- [ ] repository=repo-unknown → empty list (unknown slug gracefully skipped)
- [ ] repository=repo-a,repo-b → artifacts from both repos returned, others excluded

### DIFF
--- a/internal/controlplane/handlers_artifacts.go
+++ b/internal/controlplane/handlers_artifacts.go
@@ -292,7 +292,7 @@ func (filter *artifactListFilter) listArtifacts(
 	// Filter by repository if needed and convert to protobuf
 	results := []*pb.Artifact{}
 	for _, ent := range artifactEnts {
-		if len(allowedRepoIDs) > 0 {
+		if len(filter.repoSlugList) > 0 {
 			if !ent.OriginatedFrom.Valid {
 				continue
 			}
--- a/internal/controlplane/handlers_artifacts.go
+++ b/internal/controlplane/handlers_artifacts.go
@@ -267,18 +267,34 @@ func (filter *artifactListFilter) listArtifacts(
 		return nil, fmt.Errorf("failed to get artifact entities: %w", err)
 	}
 
+	// Resolve repo slugs to entity IDs once, before the loop.
+	allowedRepoIDs := make(map[uuid.UUID]struct{}, len(filter.repoSlubList))
+	for _, slug := range filter.repoSlubList {
+		repoEnt, err := filter.store.GetEntityByName(ctx, db.GetEntityByNameParams{
+			ProjectID:  projectID,
+			EntityType: db.EntitiesRepository,
+			Name:       slug,
+			ProviderID: provider.ID,
+		})
+		if err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				continue
+			}
+			return nil, fmt.Errorf("failed to look up repository %q: %w", slug, err)
+		}
+		allowedRepoIDs[repoEnt.ID] = struct{}{}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/controlplane/handlers_artifacts.go [hunk 1]
```diff
@@ -267,18 +267,34 @@ func (filter *artifactListFilter) listArtifacts(
 		return nil, fmt.Errorf("failed to get artifact entities: %w", err)
 	}
 
+	// Resolve repo slugs to entity IDs once, before the loop.
+	allowedRepoIDs := make(map[uuid.UUID]struct{}, len(filter.repoSlubList))
+	for _, slug := range filter.repoSlubList {
+		repoEnt, err := filter.store.GetEntityByName(ctx, db.GetEntityByNameParams{
+			ProjectID:  projectID,
+			EntityType: db.EntitiesRepository,
+			Name:       slug,
```
evankanderson, line 280: It looks like these "slubs" (typo?) are for the encompassing GitHub repository, not for the individual entity.  That's a sort of weird github-ism, so this is probably worth a comment.  (The entire API seems like it doesn't really take into account the DockerHub or new Quay.io providers.)

internal/controlplane/handlers_artifacts.go
```diff
@@ -267,18 +267,34 @@ func (filter *artifactListFilter) listArtifacts(
...
+			return nil, fmt.Errorf("failed to look up repository %q: %w", slug, err)
+		}
+		allowedRepoIDs[repoEnt.ID] = struct{}{}
+	}
+
 	// Filter by repository if needed and convert to protobuf
 	results := []*pb.Artifact{}
 	for _, ent := range artifactEnts {
-		// Apply repository filter if specified
 		if len(filter.repoSlubList) > 0 {
```
evankanderson: This feels like it should be checking `len(allowedRepoIDs)` now.

evankanderson (dismissed):
This seems to have been removed in https://github.com/mindersec/minder/pull/5913, alongside a large number of other changes in the refactor.  I'll admit that I didn't manually test this endpoint at that time, thanks for the fix!

I'm approving this, but I'll give it a day for you to either say "merge as-is" or patch the two small comments I had.  (Don't feel the need to build a big filtering mechanism; my observation is just that the current filtering mechanism is kinda not-right.)

evankanderson (dismissed):

evankanderson (approved):

PR #6773, opened 2026-09-12
Fix GetArtifactByName ignoring owner and repo in name lookup

th3Gli4ch:
# Summary

- `GetArtifactByName` accepts `repoOwner/repoName/artifactName` but queried `properties.PropertyName` using only the trailing `artifactName` segment
- `properties.PropertyName` for artifacts is stored as `owner/artifactName`, so the lookup never matched correctly and the owner prefix was silently dropped
- `garbage/garbage/myartifact` returned the same result as `real-owner/real-repo/myartifact`

Fixes 

- Construct the correct lookup key as `repoOwner/artifactName` to match the stored format
- Add a post-fetch repo validation: if `ArtifactPropertyRepo` is cached, reject the request when it doesn't match the requested `repoOwner/repoName`
- Add unit tests covering: invalid name format, provider not found, artifact not found, repo mismatch rejection, and happy path

### DIFF
--- a/internal/controlplane/handlers_artifacts.go
+++ b/internal/controlplane/handlers_artifacts.go
@@ -18,6 +18,7 @@ import (
 	"github.com/mindersec/minder/internal/db"
 	"github.com/mindersec/minder/internal/engine/engcontext"
 	"github.com/mindersec/minder/internal/logger"
+	ghprops "github.com/mindersec/minder/internal/providers/github/properties"
 	"github.com/mindersec/minder/internal/util"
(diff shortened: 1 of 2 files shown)


evankanderson (approved):
I think this is correct for the current GitHub model, where `ghcr.io` creates a namespace for each organization, but packages can have a different name than the repo name.  We'll need to make this a bit more rigorous for the other OCI providers in the future that don't connect source repository organization with registry namespaces.

Issue #6774, opened 2026-09-13
Expose OCI image identity and preserve raw manifest/index in artifact ingeste


shreyaabaranwal:
## Description

The artifact ingester currently exposes Sigstore verification data through
`input.ingested`, but does not expose the identity of the image being
evaluated (registry, repository, tag, or digest).

This makes it difficult for rules to use a DataSource to inspect the
underlying OCI image. In particular, this is needed for planned rules around
multi-arch support, SBOM presence, and layer-size limits.

While exploring a multi-arch policy for epic #6436, I also found a second
limitation in `GetManifest()` (`internal/providers/oci/oci.go`). It currently
uses `remote.Image()`, which resolves a multi-platform image to a single
platform-specific manifest. As a result, the other platform entries in the
original image index are lost.

Together, these mean that exposing the current manifest output as-is would
not be sufficient for rules that need to inspect the full OCI image.

## Proposed solution

Expose the information needed by these rules at the artifact level:

1. Include the image reference in the artifact ingester output:
   - registry
   - repository
   - tag
   - digest

2. Preserve access to the original OCI manifest or image index, along with
   its media type, so that multi-platform information is not lost.

The exact implementation (for example, whether to use `remote.Get()` or
`remote.Index()`) can be determined during the design/implementation phase.

## Alternatives considered

I considered using the existing `GetManifest()` / `remote.Image()` path as-is,
but it resolves multi-platform images to a single platform and therefore
cannot provide the information needed for a multi-arch policy.

## Why this is needed

This was identified while whiteboarding a policy to verify that a container
image supports both `linux/amd64` and `linux/arm64`.

The same underlying information would also be useful for other planned
artifact-property rules, including:

- multi-architecture support
- SBOM presence
- layer-size limits

So this appears to be a shared prerequisite rather than a blocker specific
to a single rule.

## Acceptance criteria

- `input.ingested` for artifact entities exposes the image registry,
  repository, tag, and digest.
- Rules can access the original OCI manifest or image index without losing
  platform entries.
- The manifest/index media type is preserved so consumers can distinguish
  between a single-platform manifest and a multi-platform image index.

shreyaabaranwal:
Both halves of this issue are now merged:
- Image identity (registry/repository/tags/digest): #6775
- Raw manifest/index preservation: #6776

Closing this out. Follow-ups tracked separately: #6778 (GitHub-as-OCI) and
#6779 (cross-provider consistency).

PR #6775, opened 2026-09-13
Expose OCI image identity in artifact ingester

shreyaabaranwal:
## Summary

Part of #6774.

Preserves artifact version tags (previously discarded) and surfaces an `imageIdentity` (registry, repository, tags, digest) alongside each verification result, so Rego rules can key off exact image references instead of only sigstore verification data.

- `getAndFilterArtifactVersions` now returns the full `[]*pb.ArtifactVersion` objects instead of bare digest strings, since returning just `Sha` discarded `Tags`.
- `Repository` is built as `owner/name` when `artifact.Owner` is populated (GHCR, where owner is a distinct registry path segment per `container.BuildImageRef`), or just `name` otherwise (e.g. DockerHub, where the owner/namespace is baked into the provider's configured base URL rather than tracked per artifact , see `internal/providers/dockerhub.New`). This assumes only GitHub-backed artifacts populate `Owner` today (verified: no DockerHub/Quay properties package sets it); a future OCI provider that starts populating `Owner` would need this revisited.
- `Registry` is populated only when the provider implements the `OCI` interface (`interfaces.As[provifv1.OCI]`); the GitHub/GHCR path is left empty rather than hardcoding `ghcr.io`, since that's not guaranteed to be correct for every GitHub-backed registry.
- 
## Note on output shape

This changes the shape of what rules receive per artifact version: entries
go from `{"Verification": ...}` to `{"Identity": ..., "Verification": ...}`.
This is additive — existing Rego keying off `.Verification` by name is
unaffected. Checked minder-rules-and-profiles: both rule types that
reference `.Verification` (`artifact_signature`, `artifact_attestation_slsa`)
access it by key, not positionally, so nothing there breaks.

## Known gap / scope note

Raw manifest/index preservation (the other half of #6774 , adding `GetRawManifest` to the `OCI` provider interface so a multi-platform index isn't resolved down to a single platform) is **not included in this PR**. It's held pending scope confirmation from @evankanderson.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/engine/ingester/artifact/...`
- [ ] Maintainer review of `Repository`/`Registry` conventions across providers


internal/engine/ingester/artifact/artifact.go
```diff
@@ -234,11 +245,48 @@ func (i *Ingest) getVerificationResult(
...
+
+// getRegistryForProvider returns the registry hostname for providers that
+// expose one generically via the OCI interface (e.g. DockerHub). GitHub/GHCR
+// artifacts don't go through this interface today and ghcr.io isn't tracked
+// anywhere generic enough to surface here, so we leave it empty rather than
+// hardcode a value that may not hold for every GitHub-backed registry.
+func getRegistryForProvider(prov interfaces.Provider) string {
+	ocicli, err := interfaces.As[provifv1.OCI](prov)
+	if err != nil {
+		return ""
```
intelligent-ears: This is the part I'd want to think hardest about. For GHCR — probably the most common case — `Identity.Registry` ends up empty while `Repository` is `owner/name`, so a rule that wants to match a full image reference can't assemble one, and can't tell "unknown registry" apart from "empty string".

Given the goal is letting rules key off exact image references, a blank registry on the main path undercuts that a bit. I agree hardcoding `ghcr.io` here is wrong, but is there somewhere upstream that already knows the host? `container.BuildImageRef` takes a registry argument, so the concept exists somewhere in the GitHub path — might be worth following that back.

If it genuinely isn't available, no objection to leaving it empty, but then it's worth saying so in the field's doc comment so rule authors know to expect it rather than treating it as a bug.

Worth a one-line note that the error is deliberately discarded — right now this returns `""` both for "not an OCI provider" (expected) and for any other assertion failure, and a reader can't tell which cases are intended.

internal/engine/ingester/artifact/artifact_test.go
```diff
@@ -58,6 +58,7 @@ func TestArtifactIngestMatching(t *testing.T) {
 		mockSetup     func(*mockghclient.MockGitHub, *mockverify.MockArtifactVerifier)
 		artifact      *pb.Artifact
 		params        map[string]interface{}
+		wantIdentity  *imageIdentity
```
intelligent-ears, line 62: All three new cases set `Owner: "stacklok"`, so `buildRepository` only gets exercised on the owner-populated branch. The empty-Owner path is the one your comment above flags as fragile if a future OCI provider starts populating Owner, so it's the one most worth pinning — a case with `Owner: ""` asserting `Repository: "name"` would lock in the documented behaviour.

Same for `Registry`: it's empty in every expected identity, so `getRegistryForProvider`'s OCI branch isn't covered.

intelligent-ears (commented):
Nice work — the `buildRepository` comment explaining the Owner assumption is exactly the kind of thing that saves the next person an hour, and returning the full `ArtifactVersion` is clearly the right call over threading tags through separately.

A few things below. The registry one is the only substantive question; the rest are small.

One thing worth adding to the description: this changes the shape of what rules receive. Entries go from `{"Verification": ...}` to `{"Identity": ..., "Verification": ...}`. That's additive so existing rego keying off `.Verification` keeps working, but it's worth stating outright since it's rule-visible, and worth a quick grep through minder-rules-and-profiles for anything indexing into that structure positionally rather than by key.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Fixed in c6ea7f1 , turns out the sigstore verifier already defaults to
ghcr.io for GitHub-backed artifacts (see newContainerAuth in
internal/verifier/sigstore/container/container.go, which defaults to
"ghcr.io" and the GitHub path never overrides it via WithRegistry). So
GitHub now resolves to that same default instead of leaving it blank.
Also split out the "not OCI, not GitHub" case explicitly so it's clear
that's the only genuinely-unknown case, rather than silently swallowing
any assertion error.

internal/engine/ingester/artifact/artifact_test.go
shreyaabaranwal: Added both , an empty-Owner test case (asserting Repository falls back to
just the artifact name), and a focused TestGetRegistryForProvider that
exercises all three branches (OCI, GitHub, neither) directly using the
existing mocks.

intelligent-ears (dismissed):

internal/engine/ingester/artifact/artifact.go
```diff
@@ -234,11 +245,56 @@ func (i *Ingest) getVerificationResult(
...
 	}
-	return versionResults, nil
+	return results, nil
+}
+
+// ghcrRegistry is the registry GitHub-backed artifacts are resolved against.
+// This mirrors the default used by the sigstore verifier itself: newContainerAuth
+// in internal/verifier/sigstore/container/container.go defaults to "ghcr.io"
+// unless a provider overrides it via WithRegistry (the OCI-provider case below).
+const ghcrRegistry = "ghcr.io"
```
evankanderson, line 262: We should extract this into a constant if we expect the value to be the same in both places.  It looks like `internal/verifier` would be the place to put the constant, since this depends on that file.

internal/engine/ingester/artifact/artifact.go
```diff
@@ -234,11 +245,56 @@ func (i *Ingest) getVerificationResult(
...
+// (e.g. DockerHub) report their own; GitHub is handled as a special case
+// since it authenticates against GHCR without implementing the OCI
+// interface. Any other provider type leaves this empty rather than guess.
+func getRegistryForProvider(prov interfaces.Provider) string {
+	if ocicli, err := interfaces.As[provifv1.OCI](prov); err == nil {
+		return ocicli.GetRegistry()
+	}
+	if _, err := interfaces.As[provifv1.GitHub](prov); err == nil {
+		return ghcrRegistry
+	}
```
evankanderson, line 275: Can we make the `GitHub` provider implement `provifv1.OCI`?  (It looks like we somehow generate a mock for it anyway, but the actual implementation doesn't implement the `OCI` trait.)

This might be an "open an issue to implement"; I haven't looked at how much work it is.

internal/engine/ingester/artifact/artifact.go
```diff
@@ -234,11 +245,56 @@ func (i *Ingest) getVerificationResult(
...
+
+// buildRepository returns the artifact's path within its registry.
+//
+// For GitHub/GHCR, the owner is a distinct path segment (ghcr.io/<owner>/<name>)
+// and is tracked per-artifact via artifact.Owner, so we reconstruct that path here.
+//
+// For generic OCI providers (e.g. DockerHub), the owner/namespace is baked into
+// the provider's configured base URL at setup time rather than tracked per
+// artifact (see internal/providers/dockerhub.New), so artifact.Owner is empty
+// and the repository is just the artifact name.
```
evankanderson, line 287: It seems like we're tracking `cfg.Namespace()`, which would be the equivalent of `artifact.Owner` in GHCR, no?

You should also feel free to refactor existing code (including adding new functions to interfaces) to make these more consistent.  The differences here are mostly historical accident, rather than intent -- GHCR was added because we already had GitHub auth and credentials, and wanted to show the ability to do _something_ with sigstore.  DockerHub was added to show that we could create entities from someplace other than GitHub, but was mostly a demo of "how to add another provider" rather than driven by specific feature requests.

Ideally, it should be possible to write common rules that should work across most OCI providers, just like we can write rules about Git contents that work across GitHub and GitLab (even though their APIs for e.g. branch protection are quite different).

internal/engine/ingester/artifact/artifact.go
```diff
@@ -180,31 +181,41 @@ func (i *Ingest) getVerificationResult(
 	ctx context.Context,
 	cfg *ingesterConfig,
 	artifact *pb.Artifact,
-	checksums []string,
-) ([]verification, error) {
-	var versionResults []verification
+	versions []*pb.ArtifactVersion,
+) ([]map[string]any, error) {
```
evankanderson, line 185: Can we narrow the type here?  I'm not a huge fan of untyped interfaces unless we're passing things across language boundaries (e.g. JSON, Rego, Starlark).

evankanderson: In this case, you might define an `imageInfo` struct that has a `Verification` and `Identity` fields, which I *think* will pass through Rego marshalling, though we should test it.


```suggestion
) ([]imageInfo, error) {
```

internal/engine/ingester/artifact/artifact.go
```diff
@@ -180,31 +181,41 @@ func (i *Ingest) getVerificationResult(
...
 	if err != nil {
 		return nil, fmt.Errorf("error getting verifier: %w", err)
 	}
 
+	registry := getRegistryForProvider(i.prov)
+	repository := buildRepository(artifact)
+
 	// Loop through all artifact versions that apply to this rule and get the provenance info for each
-	for _, artifactChecksum := range checksums {
+	for _, version := range versions {
```
evankanderson, line 197: It shouldn't happen, but since we have a `*pb.ArtifactVersion`, you may want to add a nil-check here.

internal/engine/ingester/artifact/artifact.go
```diff
@@ -180,31 +181,41 @@ func (i *Ingest) getVerificationResult(
...
+		identity := imageIdentity{
```
evankanderson, line 198: I think I know why you called this "identity", but it feels like it might be slightly confusing given that there's also a "signerIdentity" in the verification.  What do you think about calling this "Image" or "ImageInfo"?

evankanderson (commented):
One interesting behavior that I don't think we need to change (but should probably document), is that we have a separate ingestion result for **each tag** even if there are multiple tags that point to the same SHA hash.  (That is, while ArtifactVersion supports a list, we only even create a 1-length list.)

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Good call, will extract it , moving to internal/verifier so there's one
source of truth.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: That would genuinely solve the ghcr.io-default problem more cleanly than
the fallback we have now. I'd rather scope it as a separate issue though ,
it touches the GitHub provider's auth/credential path in ways that feel
bigger than this PR. Happy to open it.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: You're right about cfg.Namespace() , I described it as "not tracked" when
what I meant was "not exposed through the OCI interface today." Will fix
the comment's wording.

On the bigger consistency point , completely agree that's the right
long-term direction. I'd like to keep this PR scoped to identity + raw
manifest though, and open a separate tracking issue under #6436 for
making the OCI provider interface consistent (namespace/owner handling,
GitHub-as-OCI, etc.) as its own piece of work. Does that split make sense?

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Agreed, will define imageInfo with Verification and Identity fields and
update the function body accordingly. Will also add a quick test to
confirm it marshals through Rego correctly before considering this done ,
good call flagging that as unverified.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Going with imageInfo for the wrapper struct (Verification + Identity
fields), and renaming imageIdentity to imageRef for this one , avoids
colliding with imageInfo and steers clear of confusion with
go-containerregistry's own v1.Image type.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Good catch , added this.

internal/engine/ingester/artifact/artifact.go
evankanderson: That would be great to open as an issue, as you're probably more familiar with this code than I am at the moment, and you're probably aware of any "gotchas" that would need to be considered.

internal/engine/ingester/artifact/artifact.go
evankanderson: Yep, that's a perfectly fine resolution -- "not now" is _often_ an acceptable resolution for these sorts of warts.

internal/engine/ingester/artifact/artifact.go
evankanderson: Mostly, the types help me (as a human) understand what's going on when I come back and read it in 6 months.  Obviously, the original code was only partially written this way, but we should aim to do a little better with each revision.  😁

evankanderson:
I agree with all your comments, I'll be happy to re-review and approve when you've pushed the branch.

internal/engine/ingester/artifact/artifact.go
```diff
@@ -61,6 +61,24 @@ type verifiedAttestation struct {
...
+	Registry   string   `json:"registry,omitempty"`
+	Repository string   `json:"repository"`
+	Tags       []string `json:"tags,omitempty"`
+	Digest     string   `json:"digest"`
+}
+
+// imageInfo is the typed result returned per artifact version. Field
+// names are capitalized (Go-exported) rather than JSON-tagged, matching
+// how this is currently keyed when accessed from Rego ("Identity",
+// "Verification") -- this preserves that shape exactly.
```
evankanderson, line 76:
```suggestion
// "Verification").
```

(The "exactly" claim reads as LLM-generated, without adding value)

internal/engine/ingester/artifact/rego_marshal_test.go
```diff
@@ -0,0 +1,70 @@
...
+// conversion the rego evaluator applies to interfaces.Ingested.Object before
+// handing it to a policy as input.ingested. internal/engine/eval/rego/eval.go's
+// Eval sets Input.Ingested = res.Object and passes the Input to
+// rego.EvalInput; OPA's rego package then round-trips that value through
+// encoding/json (util.RoundTrip) before converting it to an ast.Value via
+// ast.InterfaceToValue -- so the conversion is JSON-tag-aware, not raw
+// reflection. Since imageInfo's own Identity/Verification fields carry no
+// json tags, the round trip is expected to preserve "Identity" and
+// "Verification" as capitalized top-level keys, matching the shape
+// previously produced by the map[string]any this struct replaced.
```
evankanderson, line 24: This test is fine for now, at some point I want to explore if we can have this run through `pkg/ruletest`, which effectively hooks up the whole engine (though Ingest is partially mocked).

evankanderson (approved):
Two small comments, but nothing blocking, so I'm merging as-is.  I'd tend to go over the LLM-generated comments when I'm using it and walk back the claims and excitement about two levels; I find shorter comments generally easier to read, and LLMs tend to over-explain basics.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Fair point, will keep an eye on that going forward , good reminder to trim
the "over-explaining" tendency before pushing, not just check the logic.

internal/engine/ingester/artifact/rego_marshal_test.go
shreyaabaranwal: Sounds good ! happy to explore this.

PR #6776, opened 2026-09-14
Expose raw OCI manifest/index via GetRawManifest

shreyaabaranwal:
**Stacked on #6775** , this branch is based on `feature/6774-artifact-identity`,
which is not merged yet. The diff therefore currently includes the changes
from #6775 as well. Once #6775 merges, this PR's diff will be much cleaner.

**Please do not merge this PR before #6775.**

## Summary

This is the second part of #6774, focused on preserving the raw OCI
manifest/index.

Adds `GetRawManifest` to the OCI provider interface so rules can inspect an
artifact's manifest or image index exactly as returned by the registry,
without it being silently resolved to a single platform.

## Why

The existing `GetManifest()` implementation
(`internal/providers/oci/oci.go`) uses `remote.Image()`, which resolves a
multi-platform image index to a single platform-specific manifest. This
discards the other platform entries.

As a result, rules cannot reliably answer questions such as:

- Does this image support `linux/amd64`?
- Does this image support `linux/arm64`?
- Does this image support both platforms?

This was the original use case identified while whiteboarding the
multi-architecture policy that motivated #6774.

`GetRawManifest` uses `remote.Get()` instead, which preserves the manifest
or image index as returned by the registry, along with its media type. This
allows rules to distinguish between a single-platform manifest and a
multi-platform image index and inspect all available platform entries.

## Scope

For now, this is limited to providers that implement the OCI interface:

- DockerHub
- Quay

GitHub/GHCR does not currently implement the OCI interface. It only
implements `ArtifactProvider` for version listing, so GitHub-backed
artifacts will not have a `Manifest` entry yet.

This is an intentional scope boundary and a known follow-up rather than
something this PR attempts to address. Extending GHCR to support the OCI
interface can be handled separately.

## Design decisions

- **Fetch by digest rather than tag:** the digest is always available on an
  artifact version, while a tag may not be. This also ensures we fetch the
  exact artifact being evaluated.

- **Manifest fetch failures are non-fatal:** failures are logged and the
  manifest entry is omitted rather than failing the entire rule evaluation.
  Sigstore verification is the primary purpose of the artifact ingester, so
  a secondary manifest-fetch failure should not block the rest of the
  evaluation.

## Test plan

- `go build ./...`
- `go test ./internal/engine/ingester/artifact/... ./internal/providers/oci/...`


internal/engine/ingester/artifact/artifact.go
```diff
@@ -180,31 +181,59 @@ func (i *Ingest) getVerificationResult(
...
 		return nil, fmt.Errorf("error getting verifier: %w", err)
 	}
 
+	registry := getRegistryForProvider(i.prov)
+	repository := buildRepository(artifact)
+	// Raw-manifest support is scoped to providers implementing the OCI
+	// interface (DockerHub, Quay) for now. GitHub/GHCR doesn't implement
+	// this interface today; that's a known, documented gap pending a
+	// decision on whether/how to extend GHCR.
+	ocicli, ociErr := interfaces.As[provifv1.OCI](i.prov)
```
evankanderson: This seems like it's probably another reason to extend the GHCR provider.

One quirk (as of a year ago) is that GHCR auth doesn't work with app tokens, only with PATs / shared secrets.  At the time, that meant that the GHCR provider couldn't handle private images.  We don't need to fix that while adding the OCI support, but it's something to keep in mind based on a GitHub-side limitation.

internal/engine/ingester/artifact/artifact.go
```diff
@@ -180,31 +181,59 @@ func (i *Ingest) getVerificationResult(
...
+	for _, version := range versions {
+		identity := imageIdentity{
+			Registry:   registry,
+			Repository: repository,
+			Tags:       version.Tags,
+			Digest:     version.Sha,
+		}
+
+		var rawManifest *provifv1.RawManifest
+		if ociErr == nil {
```
evankanderson: I think `ocicli` will be nil here as well, and I'd be inclined to simply check that (and maybe not even store `ociErr` in this case, since we have a fallback).

internal/engine/ingester/artifact/artifact.go
```diff
@@ -180,31 +181,59 @@ func (i *Ingest) getVerificationResult(
...
+			rawManifest, err = ocicli.GetRawManifest(ctx, artifact.GetName(), version.GetSha())
+			if err != nil {
+				// Deliberately non-fatal: a transient manifest-fetch
+				// failure shouldn't block sigstore verification results,
+				// which is this ingester's primary purpose.
+				zerolog.Ctx(ctx).Debug().Err(err).Str("name", artifact.GetName()).
```
evankanderson: Maybe log this at `Info`, since it seems unusual enough that an operator might want to look into it.

internal/engine/ingester/artifact/artifact.go
```diff
@@ -234,11 +263,63 @@ func (i *Ingest) getVerificationResult(
...
-			versionResults = append(versionResults, *verResult)
+			// Append the identity, verification result, and (when
+			// available) raw manifest to the list
+			entry := map[string]any{
+				"Identity":     identity,
+				"Verification": *verResult,
+			}
+			if rawManifest != nil {
+				entry["Manifest"] = *rawManifest
+			}
```
evankanderson: If you use a struct here (as per #6775) and allocate it at the beginning of the block, you can just assign to the `Manifest` pointer up on line 212.

internal/providers/github/mock/github.go
evankanderson, line 1: I haven't looked into this, but I'm wondering why the `OCI` mock is being generated here.

internal/providers/oci/oci.go
```diff
@@ -157,6 +157,28 @@ func (o *OCI) GetManifest(ctx context.Context, contname, tag string) (*v1.Manife
 	return man, nil
 }
 
+// GetRawManifest returns the manifest or image index exactly as returned by
+// the registry, fetched by digest (not tag) since digests are canonical and
+// always available on an artifact version, unlike tags. Unlike GetManifest,
+// this does not resolve a multi-platform index down to a single platform.
+func (o *OCI) GetRawManifest(ctx context.Context, contname, digest string) (*provifv1.RawManifest, error) {
```
evankanderson, line 164: Why not return `remote.Descriptor` here?  That seems likely to be a stable interface, and we won't need to plumb more fields through later.

evankanderson (commented):
This looks promising -- I think the big thing I'd focus on is getting the return types correct -- I think re-using go-containerregistry types for the OCI interface makes sense, and then we talked about the other in #6775.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Good context, thanks , will add this to #6778 so it's not lost. Agreed we
don't need to solve it here.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Good call, will drop ociErr and just check ocicli != nil directly.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Agreed, bumping to Info.

internal/engine/ingester/artifact/artifact.go
shreyaabaranwal: Will do , adding a Manifest *provifv1.RawManifest field to imageInfo (now
that it exists per #6775) and assigning directly instead of the map-entry
approach.

internal/providers/github/mock/github.go
shreyaabaranwal: That's pre-existing and unrelated to this PR , .mk/gen.mk's mockgen
invocation for github/mock/github.go uses -source mode against
providers.go, which generates mocks for every interface in that file
(GitHub, OCI, GitLab, etc.) into the one destination, not just
GitHub-specific ones. Didn't touch that generation setup here.

internal/providers/oci/oci.go
shreyaabaranwal: Yeah, that makes sense , I hadn't considered the context/fetcher being carried along with remote.Descriptor. v1.Descriptor seems like a much cleaner fit here. I'll switch to that and keep the raw manifest bytes in Content.

krrish175-byte:
Tests are failing. Please fix those

internal/engine/ingester/artifact/artifact.go
```diff
@@ -413,4 +416,4 @@ func signerIdentityFromCertificate(c *certificate.Summary) (string, error) {
 	builderURL = strings.TrimPrefix(builderURL, c.SourceRepositoryURI)
 
 	return builderURL, nil
-}
+}
```
evankanderson: It looks like you may need to run `go fmt` on this file, and the test.

internal/engine/ingester/artifact/rego_marshal_test.go
```diff
@@ -67,4 +68,10 @@ func TestImageInfoRegoInputShape(t *testing.T) {
...
 	require.Equal(t, true, verificationOut["is_signed"])
 	require.Equal(t, true, verificationOut["is_verified"])
-}
+
+	manifestOut, ok := entry["Manifest"].(map[string]any)
+	require.True(t, ok, "expected Manifest key to survive the rego round trip")
+	require.Equal(t, "application/vnd.oci.image.manifest.v1+json", manifestOut["mediaType"])
+	require.Equal(t, "sha256:1234", manifestOut["digest"])
+	require.NotEmpty(t, manifestOut["content"])
+}
```
evankanderson: (And this one, too)

I suspect `go fmt` is somewhat pedantic about file termination with/without newlines, but we try to keep that formatting so as not to introduce spurious diffs through the codebase.

evankanderson (commented):
The linter is angry, which should be a small fix.  (Yes, I wish it would suggest the fix automatically...)

internal/engine/ingester/artifact/rego_marshal_test.go
shreyaabaranwal: Ah, that explains it , thanks for catching it. Fixed in 39f533a.

evankanderson (approved):

PR #6777, opened 2026-09-15
Fix nil schema validation, OCI auth, and webhook dead code

th3Gli4ch:
- REST datasource: ValidateArgs now skips validation when no input schema is defined, matching the constructor's documented nil-schema allowance
- OCI provider: pass credentials to all registry calls (ListTags, getImage, GetDigest, GetReferrer, getDigestFromRef) via new authOptions helper; private registries were silently getting anonymous requests despite valid stored credentials
- GitHub webhook handler: remove dead URL path parsing that split and immediately discarded the result on every request

# Summary

This PR fixes three independent bugs across the REST datasource, OCI provider, and GitHub webhook handler.

REST datasource nil schema crash. ValidateArgs in internal/datasources/rest/handler.go returned an error when inputSchema was nil, despite the constructor explicitly documenting that nil schemas are allowed (// schema may be nil). Any datasource defined without an inputSchema field was silently broken at evaluation time. The fix makes ValidateArgs return nil when no schema is present, skipping validation rather than rejecting the call.

OCI provider ignoring credentials. GetAuthenticator() in internal/providers/oci/oci.go was fully implemented but never called. Every registry operation — ListTags, getImage, GetDigest, GetReferrer, and getDigestFromRef — used the go-containerregistry remote API with no auth option, meaning all calls were anonymous. Private OCI registries would return 401 despite valid credentials being stored. Two TODO comments in getDigestFromRef even acknowledged the missing auth. The fix adds an authOptions(ctx) helper that bundles context, user-agent, and credentials, wires it into all four call sites, and converts getDigestFromRef from a package-level function to an *OCI method so it has access to the authenticator.

GitHub webhook dead code. On every incoming GitHub webhook, hook.go split the request URL path and extracted the last segment into segments[len(segments)-1], then immediately discarded it with _ =. The comment above it even noted the value was not being used. The strings import existed solely for this. Both lines and the import are removed.
Fixes #(related issue)

# Testing

All three packages build and pass their existing test suites. The Test_restHandler_ValidateArgs/"No schema" case was updated to reflect the corrected behavior (wantErr: false). OCI auth has no unit tests since the registry calls require a live endpoint — manual verification is to enroll a private OCI registry and confirm artifacts resolve instead of returning 401.


internal/providers/oci/oci_test.go
```diff
@@ -4,23 +4,40 @@
...
+	return mockTokenSource{token: m.token}
+}
+
 func TestResolveCreatedAt(t *testing.T) {
 	t.Parallel()
 
 	buildTime := time.Date(2024, time.January, 2, 3, 4, 5, 0, time.UTC)
-	// poison is returned by the config getter for cases where the annotation is
-	// present: if the implementation wrongly consulted the config, the assertion
-	// against buildTime (or the expected error) would fail.
```
evankanderson, line 23: I see we're deleting this comment, but it still seems useful explaining why the variable is named "poison".

internal/providers/oci/oci_test.go
```diff
@@ -96,3 +112,55 @@ func TestResolveCreatedAt(t *testing.T) {
...
+
+	o := New(nil, "invalid.registry.local", "invalid.registry.local/myrepo")
+	ctx := context.Background()
+
+	assert.True(t, o.CanImplement(minderv1.ProviderType_PROVIDER_TYPE_OCI))
+	assert.False(t, o.CanImplement(minderv1.ProviderType_PROVIDER_TYPE_GITHUB))
+	assert.Equal(t, "invalid.registry.local", o.GetRegistry())
+
+	// These should all return errors because we have no real registry,
+	// but they will hit the lines in oci.go.
```
evankanderson, line 127: I'd rather have tests that illustrate the "good" path, rather than tests which simply run error cases to get high coverage numbers.  Coverage is a helpful indicator of "testedness", but one has to be careful of [Goodhart's Law](https://en.wikipedia.org/wiki/Goodhart%27s_law).

evankanderson (commented):
This PR could reasonably be broken into 3:

1. For the empty inputSchema, I don't see any documentation _either way_ about whether a REST methods can simply point to a fixed endpoint.  I **think** I can think of examples where that's useful, so I'm inclined to accept the behavior change.

2. For the GitHub webhook change, I think it's a reasonable cleanup, though it might be worth documenting that the current webhook includes a per-repository (I think) random component that could be checked to see if it aligns.  i.e. we should be able to look up the repo that the event is for, and validate that the UUID in the URL corresponds to the UUID for that repo in the database.

3. For the OCI changes, I'd probably want to at least have the assurance that the code has been tested in an end-to-end environment.  If I need to do that, I'll need some time to figure out what that e2e environment looks like, but I'm willing to give it a go if needed.  If you already have done some e2e testing, can you document the rough setup steps for someone to be able to find later and turn into documentation?

Thanks for these fixes, by the way!

th3Gli4ch:
I don't have access to a private Quay.io or Docker Hub namespace to test against , and spinnin up a local registry:2 instance with Bearer token auth isn't straightforward since the standard Docker registry uses htpasswd (Basic auth) rather than OAuth2 Bearer tokens - which is what GetAuthenticator() returns. The in-process test registry in `TestOCI_WithRegistry` exercises the complete call path  including `ListTags`, `GetDigest`, and `GetManifest` against a real HTTP server. `TestOCI_Auth` also covers the credential construction paths.

I’ve additionally added an end-to-end test file, `oci_e2e_test.go`, gated behind the `e2e` build tag. It can be run against a live private registry using:

For the live validation, I’m happy to defer the registry-specific testing to you if you already have a test namespace available. Otherwise, I can set one up if you can point me toward a suitable registry.

I have sent the other fixes individually as mentioned in the comments Kindly review it for any further changes thank you

PR #6780, opened 2026-09-16
Fix permit empty schemas for endpoints in REST datasource handler

th3Gli4ch:
ValidateArgs rejected calls with return errors.New("input schema cannot be nil") whenever no schema was configured  but the constructor explicitly accepts nil for inputSchema. A nil schema is valid: it means the REST endpoint takes no arguments (e.g. a fixed polling URL). Rejecting it made schema-less datasource methods unusable.

Fix: return nil from ValidateArgs when inputSchema is nil, skipping validation rather than erroring.

Test: the existing "No schema" test case is updated from wantErr: true to wantErr: false to match the corrected behaviour.

Why: REST datasource methods that point to a fixed endpoint (no parameterization needed) should be able to omit the input schema entirely. The constructor already allows nil for inputSchema, but ValidateArgs was rejecting it with an error, making schema-less handlers unusable at runtime. Skip validation instead.


evankanderson:
(Adjusted the title to make the behavior change clear when producing release notes.)

evankanderson (approved):
It looks like jsonschema doesn't provide warnings or errors for additional fields, so passing arguments for an empty schema (`"{}"`) would provide the same result as skipping the validation entirely.

Issue #6437, opened 2026-05-04
Roadmap: Improve Rule Identity


evankanderson:
## This is a Roadmap Epic, and needs further design and breakdown into work items

Minder datasources support
[using the `providerAuth` field](../understand/providers.md#defining-a-data-source)
to authenticate rules to the API which manages the entity, but do not support
any form of other authentication. Rule identity would be useful for calling
third-party or custom API endpoints to collect security data (either
company-specific like approved security exceptions, or paid-access such as
vulnerability feeds). Minder should consider defining identities for executing
rule types (based on the enclosing project), and using
[OIDC](https://openid.net/developers/how-connect-works/) or
[SPIFFE](https://spiffe.io/) to provide a unique identity for each Minder
project or rule executing in the project. For APIs which require static secrets
like API keys, secret stores such as Vault, OpenBao, or cloud provider APIs can
avoid the need for building secret storage directly into Minder.

GitHub Actions may
[provide an example](https://docs.github.com/en/actions/reference/security/oidc#oidc-token-claims)
of the data which might be presented alongside a Minder rule identity.

### Solution Proposal

* Define a standard mechanism for adding a Minder-attested identity to outbound HTTP requests from rules (in either datasources or via `http.send`)
   * TODO: should this authentication should be explicitly requested by the rule?
* Build a rule which uses attested identity to fetch and use a secret from an existing secrets manager (OpenBao / Vault or a cloud provider)
* Document usage of the Minder-attested identity on the documentation site

dakshhhhh16:
[...]

PR #6782, opened 2026-09-16
Fix/oci auth

th3Gli4ch:
GetAuthenticator() was implemented but never actually called. ListTags, GetDigest, GetReferrer, and getDigestFromRef were all making unauthenticated requests, regardless of the configured credentials. This meant authenticated registries could silently fail or return 401s.

Fix: added an authOptions() helper that bundles the context, user-agent, and credentials (when available) into []remote.Option, then passed it through all four call sites.

Testing:

TestOCI_Basic verifies CanImplement and GetRegistry

TestOCI_Auth covers GetAuthenticator() for three cases: nil credential (anonymous), valid OAuth2 token, and a non-OAuth2 credential type (expects an error)

TestOCI_WithRegistry starts an in-process OCI registry, pushes a real random image, then calls our client's ListTags (tag is returned), GetDigest (valid sha256: digest returned), and GetManifest (non-nil manifest returned). This exercises the actual auth plumbing end-to-end without needing a remote registry.

oci_e2e_test.go (gated behind //go:build e2e) is for live validation against a real private registry when credentials are available


evankanderson (changes requested):
Unfortunately, I don't think this works in an end-to-end manner.  (You've added an integration test, but I think this actually needs to be tested with a local Minder instance from setup through entity registration and policy evaluation.)


I tried configuring this in a local Minder server with a DockerHub provider using a PAT (`minder provider enroll -c dockerhub --provider-config <(echo '{"dockerhub": {"namespace": "evancustcodian"}}') --token $TOKEN`), but it looks like the PAT method requires a username as well as a password (`authn.Basic`), and we don't collect the username in the provider.  For Docker Teams / Enterprise customers, it would be nice if we could use an [OIDC authenticator](https://docs.docker.com/security/authentication/oidc-connections/create-manage/) based on a Minder-provided identity, which ties in with #6437, but is quite a bit more work.

For [Quay.io], it looks like the token is [used with the username `$token` and basic auth](https://docs.quay.io/glossary/access-token.html), but the recommended method is a [Robot Account](https://docs.quay.io/glossary/robot-accounts.html), which _also_ needs both username and password data, rather than an OAuth authenticator.

Issue #6779, opened 2026-09-15
Make OCI provider interface consistent across GitHub/DockerHub/Quay


shreyaabaranwal:
### Please describe the enhancement

Artifact identity is currently represented differently across providers.
GHCR tracks the owner/namespace through `artifact.Owner`, while DockerHub
and Quay derive the namespace from `cfg.Namespace()`.

There is currently no consistent way for OCI-related code and rules to
access the registry, repository, and namespace/owner across providers.

### Solution Proposal

Define a consistent way to expose registry, repository, and
namespace/owner information across OCI-capable providers.

The goal is to let artifact rules work with a common representation
regardless of whether the artifact comes from GitHub/GHCR, DockerHub,
or Quay, instead of having provider-specific handling in the artifact
ingester.

The exact interface or data-model change can be determined as part of
the design.

### Describe alternatives you've considered

Keep the current provider-specific handling, with the artifact ingester
dealing with the differences for each provider.

This works in the short term, but would make each future OCI-related
rule more complicated and require additional provider-specific logic.

### Additional context

This came up during review of #6775. The current differences appear to
be largely historical rather than intentional: GHCR was added around
existing GitHub authentication, while DockerHub was initially added
primarily as a demo provider.

This is part of the broader OCI provider work tracked in #6436.

### Acceptance Criteria

- There is a documented, consistent representation of registry,
  repository, and namespace/owner for OCI-capable providers.
- OCI-related code can use this representation without provider-specific
  branching.
- Existing artifact and rule behavior remains unchanged.
- The provider-specific logic in `buildRepository` and
  `getRegistryForProvider` can be simplified where applicable.
- The chosen representation is documented for future OCI-related rules
  and providers.

evankanderson:
A few more specific policy examples (because I like to derive functionality from actual use cases, rather than just from abstract reasoning):

* Kubernetes supports the last year of minor releases with monthly patch releases.  This means that they will be actively patching to produce e.g. `1.37.0`, `1.36.4`, and `1.35.8` (and sometimes something like a `1.34.12`, depending on the month).  They probably want to be able to express something like "each supported release is _signed by identity X_, _has an SBOM_, etc.
* Minder produces binary releases for clients (`v0.3.2`, etc) and container / helm releases on a rolling basis (`v0.0.20260918.5412+ref.d33db3f`).  The two distribution tracks have different version numbers and release schedules (and probably different checks for platform binaries and OCI artifacts).
[...]

Issue #6778, opened 2026-09-15
Make GitHub provider implement provifv1.OCI


shreyaabaranwal:
## Summary

While working on #6774/#6775, I found that GitHub-backed artifacts currently
don't have access to the OCI provider interface.

The GitHub provider currently implements `ArtifactProvider` for version
listing, but does not implement the OCI interface (`GetManifest`,
`GetRawManifest`, `GetRegistry`, `GetAuthenticator`, etc.).

Because of this, the artifact ingester currently falls back to a hardcoded
`ghcr.io` registry for GitHub-backed artifacts. It also means GitHub/GHCR
artifacts cannot use the OCI-based functionality being added in #6776.

## Proposed solution

Add OCI interface support to the GitHub provider so GitHub-backed artifacts
can use the same OCI functionality as the DockerHub and Quay providers.

This would allow the artifact ingester to get the registry and other OCI
information from the provider itself instead of relying on a GitHub-specific
`ghcr.io` fallback.

It would also allow GitHub/GHCR artifacts to use raw-manifest support from
#6776.

The exact implementation details, particularly authentication and
credential handling when accessing GHCR, can be worked out as part of this
issue.

## Alternatives considered

Keep the existing GitHub-specific `ghcr.io` fallback and treat GHCR OCI
support as a separate limitation.

This works for the current artifact verification flow, but it means
GitHub-backed artifacts won't have the same OCI capabilities as
DockerHub/Quay providers as more OCI-based rules are added.

## Additional context

This came up during review of #6775.

A `MockOCI` implementation already exists alongside the GitHub provider
mocks, but the real GitHub provider does not currently implement the OCI
interface.

The main open question is how the GitHub provider should handle
authentication and credentials when accessing GHCR's registry API.

## Acceptance criteria

- The GitHub provider implements the required OCI interface methods:
  - `GetDigest`
  - `GetReferrer`
  - `GetManifest`
  - `GetRawManifest`
  - `GetRegistry`
  - `GetAuthenticator`
- Existing GitHub-provider-based artifact verification continues to work
  without changes in behavior.
- The artifact ingester no longer needs a GitHub-specific hardcoded
  `ghcr.io` fallback once the OCI support is available.
- GitHub/GHCR-backed artifacts can use the raw-manifest functionality added
  in #6776.
- Authentication and credential handling for GHCR are covered by tests.

evankanderson:
[...]

PR #6791, opened 2026-09-17
Implement provifv1.OCI on GitHub provider

shreyaabaranwal:
Part of #6778.

## Summary

Implements the OCI provider interface (GetDigest, GetReferrer, GetManifest,
GetRawManifest, GetRegistry, GetAuthenticator) on the GitHub provider, so
GitHub-backed artifacts get the same OCI capabilities as DockerHub and Quay.

## Why

GitHub-backed artifacts currently fall back to a hardcoded ghcr.io registry
default in the artifact ingester, since the GitHub provider only
implemented ArtifactProvider (version listing), not the OCI interface. This
also meant GitHub/GHCR artifacts couldn't use the raw-manifest support
added in #6776.

## Implementation

- `GetRegistry`/`GetAuthenticator` delegate to `GitHubCredential`'s existing
  `GetAsContainerAuthenticator` (already used in production by the sigstore
  verifier).
- `GetDigest`/`GetReferrer`/`GetManifest`/`GetRawManifest` are implemented
  by embedding an `*oci.OCI` client configured with baseURL
  `ghcr.io/<namespace>`, reusing the existing OCI provider code.

### 1. GHCR currently assumes a single namespace

The OCI methods added for GHCR (`GetAuthenticator`, `GetDigest`,
`GetManifest`, `GetRawManifest`) resolve against the provider-configured
namespace (`ghcrwrap.Namespace()`).

This is different from the existing Sigstore verification path, which
resolves the owner per artifact using `artifact.Owner`.

This matches the current DockerHub/Quay model, where one provider instance
is configured for a single namespace. However, it may not work for GitHub
App installations that can access multiple organizations.

This provider/namespace inconsistency is already tracked in #6779, so this
PR keeps the existing behavior rather than expanding its scope.

### 2. Raw manifest fetching for private GHCR images

With the GitHub provider now implementing the OCI interface,
`GetRawManifest` will also be called for GHCR-backed artifacts.

The current OCI implementation does not pass an authenticator to the
underlying `remote` calls. As a result:

- Public GHCR images can be fetched successfully.
- Private GHCR images may fail to fetch their manifest.
- On failure, the manifest is logged and omitted, while the rest of the
  artifact verification continues.

This is a consequence of adding OCI support to the GitHub provider and does
not change the existing Sigstore verification behavior.

Authentication support for these OCI calls is not yet tracked as a
separate issue , raised for discussion in the review thread on #6778.

## Test plan

- `go build ./...`
- `go test ./internal/providers/github/... ./internal/engine/ingester/artifact/...`


shreyaabaranwal:
The "Unit testing" check failure looks like a flake , verified
TestAll/TestHandleGitHubWebHook/repository_ruleset_created passes cleanly
on a fresh main checkout (no changes from this PR touch webhook handling).
Might need a maintainer to re-trigger the check.

internal/providers/github/ghcr/ghcr.go
```diff
@@ -83,6 +83,11 @@ func (g *ImageLister) getNamespace() string {
 	return g.cfg.GetNamespace()
 }
 
+// Namespace returns the configured GHCR namespace.
+func (g *ImageLister) Namespace() string {
+	return g.getNamespace()
```
krrish175-byte: The new Namespace() method just calls getNamespace(), but getNamespace() still has two other direct callers inside ghcr.go (the GetNamespaceURL return and the ListPackages call). One option worth considering: rename getNamespace() to Namespace() and update those two internal call sites to use g.Namespace() instead. That removes the double-wrapping entirely and keeps only one method to look at. Not a blocker, but it would be cleaner.

internal/providers/github/common.go
```diff
@@ -900,6 +914,40 @@ func (c *GitHub) GetNamespaceURL() string {
...
+	return container.GHCRRegistry
+}
+
+// GetAuthenticator returns the authenticator for accessing GHCR.
+//
+// Uses GetAsContainerAuthenticator rather than ghcrOCI's own GetAuthenticator,
+// since that one assumes an OAuth2 token and returns a Bearer, not the Basic
+// auth GHCR expects.
+func (c *GitHub) GetAuthenticator() (authn.Authenticator, error) {
+	return c.GetCredential().GetAsContainerAuthenticator(c.ghcrwrap.Namespace()), nil
```
krrish175-byte, line 951: The credential passed to oci.New() is stored in ghcrOCI.cred but never actually used: GetDigest/GetReferrer/GetManifest/GetRawManifest delegate to ghcrOCI which makes unauthenticated remote calls, and GetAuthenticator bypasses ghcrOCI.GetAuthenticator() entirely (for the good reason explained in the comment above). So the cred field in ghcrOCI is essentially dead right now. This is consistent with the known limitation documented in the PR, but it might be worth a TODO comment so someone fixing private image support knows exactly where to look.

internal/providers/github/common_test.go
```diff
@@ -241,6 +244,14 @@ func (m *mockCredential) GetCacheKey() string {
 	return "mock-cache-key"
 }
 
+func (m *mockCredential) GetAsContainerAuthenticator(owner string) authn.Authenticator {
+	_ = m
```
krrish175-byte: The receiver m is always considered used by the Go compiler in methods, so this _ = m assignment is not needed. Same issue on line 243.

internal/providers/github/common_test.go
```diff
@@ -1661,6 +1672,25 @@ func TestCreateSecurityAdvisory(t *testing.T) {
...
+func TestGetAuthenticator(t *testing.T) {
+	t.Parallel()
+
+	th := setupTest(t)
+	th.gh.ghcrwrap = ghcr.FromGitHubClient(th.client, "test-owner")
+	th.delegate.EXPECT().GetCredential().Return(&mockCredential{})
+
+	authenticator, err := th.gh.GetAuthenticator()
+	require.NoError(t, err)
+	assert.Equal(t, &authn.Basic{Username: "test-owner", Password: "mock-token"}, authenticator)
```
krrish175-byte, line 1690: The tests for GetRegistry and GetAuthenticator are solid. It would also be worth adding tests for the four delegation methods: GetDigest, GetReferrer, GetManifest, and GetRawManifest. They are simple wrappers around ghcrOCI, so even a table-driven test that passes a bad image name and checks for an error would be enough to confirm the wiring is correct and catch a regression if the delegation ever breaks.

krrish175-byte (commented):
Good implementation overall. The PR description is thorough and clearly documents the known limitations. A few inline notes below.

internal/providers/github/common_test.go
shreyaabaranwal: Fair point, adding tests for those four.

internal/providers/github/common_test.go
shreyaabaranwal: You're right, that's not needed , removing it.

internal/providers/github/ghcr/ghcr.go
shreyaabaranwal: Good catch, will collapse the two into one exported method.

internal/providers/github/common.go
shreyaabaranwal: Makes sense, adding a TODO pointing at this.

shreyaabaranwal:
The "Unit testing" failure looks like another flake , this time in
internal/eea/eea_test.go (TestFlushAll/TestAggregator), a race-detector
test that's entirely unrelated to this PR's changes (only touches
internal/providers/github). 
@krrish175-byte 
Ran it against clean main with -race and it
passed cleanly. 

Might need a re-trigger.

shreyaabaranwal:
Opened #6794 to track this separately, so it doesn't get lost as "just a
flake" next time someone hits it.

shreyaabaranwal:
@krrish175-byte all checks are green now, and I've addressed the review
comments .


 let me know if there's anything else you'd like changed!

krrish175-byte (approved):
looks good from my side. @evankanderson PTAL if I missed anything.

internal/providers/github/common_test.go
evankanderson: I'm not going to object to the extra tests, but (IMO) there's not a lot of extra value in testing one-line delegations like this (though it does make coverage numbers look a bit better).  Generally, this is a place where I might consider either hoisting the OCI tests (e.g. make them exportable to the various implementations), or building a central table-driven test in the OCI implementation which iterates through the different providers.

Given the size of the current implementation, I'm not going to block this PR on additional testing.

evankanderson (approved):
This looks good to me, though I think we'll need to start thinking about a manual testing strategy for the OCI rules pretty soon.  (Once we've got manual completed, we can think about automation, but it's good to start with a human end-to-end test that things works using a live Minder instance.)

evankanderson:
I'm currently trying to figure out why the fuzz test is consistently failing.

shreyaabaranwal:
@evankanderson Yeah, that makes sense. I agree it’d be useful to have a manual end-to-end test for the OCI rules first, especially with a live Minder instance, before we think about automating it.

Also, this might save you some digging , I hit the same ClusterFuzzLite failure on #6796, and it looks unrelated to the PR.

This turned out to be a pre-existing nil-pointer panic in `internal/providers/github/webhook/handlers_packages.go` `(processPackageEvent).` A webhook payload of `null` unmarshals without an error, leaving `event` == nil, and the subsequent `event.Action` access panics.

I also checked the git history and confirmed that this file hasn’t changed since v0.0.88, so it predates both this PR and #6796 by quite a bit.

I haven’t opened a tracking issue yet since I wasn’t sure what the right scope/naming would be. Happy to open one if you think it’d be useful.

PR #6796, opened 2026-09-20
Fix race condition in TestAggregator/TestFlushAll

### LINKED ISSUE #6794: Race condition in internal/eea tests (TestFlushAll/TestAggregator)
## Summary

While investigating a CI flake on #6791, I found that
`internal/eea/eea_test.go`'s `TestFlushAll` and `TestAggregator` can
occasionally fail under the `-race` detector.

The tests access `testing.T` methods (`t.Log`, `require.NoError`,
`assert.NoError`, etc.) from background goroutines while the test/subtest
may already be finishing. The goroutines are not synchronized with the
test lifecycle before those calls are made.

## Solution Proposal

Synchronize the background goroutines with the main test goroutine using
buffered channels.

Instead of calling `t.Log`/`assert`/`require` from worker goroutines,
have them return their results through the channels and perform the test
assertions from the main goroutine after receiving the results.

This applies to:
- the eventer/`FlushAll` goroutines in `TestFlushAll`
- the worker goroutines in `TestAggregator`

## Alternatives Considered

**Re-run CI when the test flakes.**

This may unblock an occasional failure, but it leaves a flaky test in the
suite and makes a red `Unit testing` check harder to distinguish from a
real regression.

## Additional Context

I reproduced the failures with:

`go test ./internal/eea/... -race -run "TestFlushAll|TestAggregator"`

using a clean `main` checkout, confirming that the race is independent of
#6791.
I have not yet implemented the fix.

## Acceptance Criteria

- `TestFlushAll` and `TestAggregator` no longer call `testing.T` methods
  from background goroutines.
- `go test ./internal/eea/... -race` passes reliably across repeated runs.
- No changes to `internal/eea` production behavior.

shreyaabaranwal:
Fixes #6794.

## Summary

Synchronizes the background goroutines in TestAggregator and TestFlushAll
with the main test goroutine using buffered channels, instead of calling
t.Log/assert/require from within the goroutines themselves.

## Why

The -race detector occasionally fails these tests because testing.T
methods were being called from a background goroutine concurrently with
(or after) the test/subtest finishing , testing.T isn't safe for
concurrent use. See #6794 for the full investigation.

## Testing

- go build ./...
- go test ./internal/eea/... -race -run "TestFlushAll|TestAggregator" -v
- Ran the -race suite 5 times in a row (-count=1 each time) to confirm the
  fix, not just once , race conditions are probabilistic, a single pass
  doesn't prove anything.

No production code changed , this is test-only.


internal/eea/eea_test.go
```diff
@@ -90,12 +90,15 @@ func TestAggregator(t *testing.T) {
 	// This tests that flushing sends messages to the executor engine
 	evt.Register(constants.TopicQueueEntityEvaluate, flushedMessages.Add, aggr.AggregateMiddleware)
 
+	t.Log("Running eventer")
```
evankanderson: Ah - you can't use `t.Log` inside the goroutine?  That's too bad, as it's nice to be able to tell when the eventer actually started vs things sending events into the eventer.

evankanderson: On line 366, we do `<-evt.Running()` to block until the eventer is running.  Maybe we want to do that here and then log right afterwards?

internal/eea/eea_test.go
```diff
@@ -372,15 +374,19 @@ func TestFlushAll(t *testing.T) {
...
+				flushErr <- aggr.FlushAll(ctx)
 			}()
 
 			t.Log("Waiting for flush")
 			flushedMessages.Wait()
+			require.NoError(t, <-flushErr)
 
 			assert.Equal(t, int32(1), flushedMessages.count.Load(), "expected one message")
+
+			cancel()
```
evankanderson, line 390: We deferred `cancel()` earlier -- is this needed to stop the eventer?  If so, a comment seems worthwhile.

evankanderson (dismissed):
Thanks for tracking this down!  I'd seen some of these races from time to time, but wasn't able to reproduce them locally (possibly due to having too much concurrency on my main machine).

Since this is a subtle race condition, I've got a couple comments that you _might_ want to apply.  I'm willing to accept this as-is to fix the races, though -- just let me know whether you want to push another commit or have me merge as-is.  (If I don't hear in 3-4 days, I will probably merge as-is.)

shreyaabaranwal:
Applied both , moved the log after evt.Running() and added a comment on
the explicit cancel(). Thanks for the thorough look!

krrish175-byte (approved):

